PR #59, opened 2014-08-21
Make the header sexier

marksomnian:
This PR makes the header of the index page sexier by increasing the height of the header image from `432px` to `569px` (the height of the image) and adding `background-attachment: fixed;`. This makes the header image stay in place when scrolling, making it look nicer.

### DIFF
--- a/css/com/layout/header.css
+++ b/css/com/layout/header.css
@@ -172,8 +172,9 @@
 
 .page_index .global-header {
   padding: 64px 0;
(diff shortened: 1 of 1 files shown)


yole:
Sorry for not addressing it in a timely manner. I believe this is already obsolete, and it will most definitely become obsolete with the new Web site design that will be rolled out with Kotlin 1.0 release.

PR #87, opened 2014-09-23
Mention the "use" extension method in exceptions.md

mikehearn:
Compare it to Java's try-with-resources statement.


yole:
I've added this example to the 'idioms' section.

PR #89, opened 2014-09-25
Clarified description of super-type list in traits

evacchi:
Added a clearer explanation of the super-type list of a trait,
and the implication of declaring a class type in this list.


yole:
Super types for traits have been dropped from the language, so I'm closing this as no longer relevant.

PR #94, opened 2014-10-07
Mention how to create no-args c'tors

mikehearn:


dittert:
It looks like this is already integrated into classes.md.

abreslav:
@dittert, thanks, yes, it's already there. Thanks @mikehearn

PR #110, opened 2014-11-18
Document constructor references

cypressious:
I didn't really know which class to choose as an example. If you have a better alternative to Object, please let me know.


abreslav:
Merged manually. Thanks

PR #118, opened 2014-12-31
Example error

miltador:
Tried to run an example, related to "boxing of numbers does not preserve identity" and got the results:
falsefalse
But the example says it must be truefalse
I have no idea why that's happening, using latest Kotlin. Just learning language and noticed that example is wrong. Can someone explain me what's going on? :)


abreslav:
Thanks for your PR, we'll fix the corresponding bug

PR #127, opened 2015-01-31
Updated Readme with JDK prerequisite steps.

Zoltu:
Someone more familiar with IntelliJ than I still needs to fill in the steps sfor how to add a new SDK.  I managed to cobble it together but I don't remember what I did and it took me far longer than it should.  Note: I am coming from C# development so I am entering this "Getting Started" guide with no knowledge of Java, JDKs, JREs, IntelliJ, etc.  I understand that this isn't the primary audience of Kotlin, but I think there is value in the getting started instructions being verbose enough for a non-Java developer to follow them.  This includes steps on how to acquire and install the appropriate dependencies.


hhariri:
Thanks. I've added one reference of JDK you mention and I've also added a new paragraph with links to some more detailed instructions for people new to JVM/IDEA. Let me know if that helps.

max-kammerer:
I'm not sure but there was some problems with jdk8 streams during compilation under this jdk.

hhariri:
Which?

max-kammerer:
Clashing java streams with kotlin ones

PR #128, opened 2015-02-01
Point the navbar at the actual API docs

mikehearn:
Currently it goes to a useless page that simply redirects users to an ancient 0.1-SNAPSHOT view of the docs.


hhariri:
Thanks. We still are not pointing to it because there are some unfinished aspects. Once it's done we can update the link. I'll close this for now.

PR #130, opened 2015-02-02
Command-line tutorial: Add path parameter to list_folders script command

ubuntudroid:
Trying to run the list_folders.kts script without supplying a path results in a ArrayIndexOutOfBoundsException. Therefore I've added a parameter placeholder to the example command.


hhariri:
Thanks.

PR #134, opened 2015-02-11
Adds note about unimplemented feature.

Zoltu:
The feature being used in this code block is not yet implemented in the latest build.  Either it should be removed from the documentation or annotated as such.  I have chosen to annotate it so it will be easy to add the documentation back in later, but removal also works if that is preferred.


abreslav:
Thanks for the report, I removed the whole section in #135, closing this one now

PR #136, opened 2015-02-12
Update Application.kt and Gradle to use package-level main()

evacchi:
I have updated the tutorial to use a simpler approach for the static main() definition as described here:
https://devnet.jetbrains.com/thread/460557

It does introduce the requirement for introducing the directive 

```java
springBoot {
    mainClass = '_DefaultPackage'
}
```

which in turn may require to be clearer on how top-level functions are transformed into methods of a <package-name>Package class, but I feel like it might be better than introducing the `platformStatic` annotation and all the `class object` boilerplate.

Feel free to disregard this pull request if you feel that this may introduce more problems than benefits, though.

### DIFF
--- a/code/tutorials/spring-boot-restful/build.gradle
+++ b/code/tutorials/spring-boot-restful/build.gradle
@@ -42,11 +42,6 @@ sourceSets {
     }
 }
 
-springBoot {
-    mainClass = '_DefaultPackage'
-}
-
-
 dependencies {
     compile("org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version") // New
     compile("org.springframework.boot:spring-boot-starter-web")
--- a/code/tutorials/spring-boot-restful/build.gradle
+++ b/code/tutorials/spring-boot-restful/build.gradle
@@ -42,6 +42,11 @@ sourceSets {
     }
 }
 
+springBoot {
+    mainClass = '_DefaultPackage'
+}
+
+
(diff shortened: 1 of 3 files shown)


hhariri:
Eduardo, thanks for the PR. I love how you're really focusing on minimizing the number of lines for Kotlin :). In this tutorial though I was trying to remain as "faithful" as possible to the original Springboot stuff so that people appreciate that everything in Java can be used and consumed from Kotlin. That's why I just really did a one-to-one conversion.

How about though, what we could do is add a section to the tutorial describing your alternative? So we keep the original one and the alternative approach?

evacchi:
I'm perfectly fine with that. Shall I update this pull request, or would you rather cherry pick the changes yourself?

hhariri:
If you would update it, even better, then I can simply merge.

evacchi:
There you go, I have updated the tutorial text; I have also reverted the changes in the actual source code. Feel free to modify everything as you wish.

hhariri:
Brilliant. Thank you!

hhariri:
Moved running to end and added you as author

evacchi:
whoops, my bad: there is one broken link: [top-level functions] points to http://kotlinlang.org/docs/functions.html instead of http://kotlinlang.org/docs/reference/functions.html

hhariri:
Can you fix it and PR?

PR #164, opened 2015-03-20
Typo: Matadata -> Metadata

MariusVolkhart:
Correct a typo


yole:
Thanks for noticing and fixing Marius!

PR #177, opened 2015-04-07
Adding Kotlin2JS

bbarker:


bashor:
Thanks for your efforts we updated this reference according to M11 in separate PR.

PR #179, opened 2015-04-09
remove mention of .kts files from the command-line tutorial

yole:


yole:
We want to keep this info.

PR #181, opened 2015-04-12
Reconfigure maven-compiler-plugin

DarthAndroid:
Instead of running the kotlin compiler in the `process-sources` phase, we should run it in the `compile` phase and configure the `maven-compiler-plugin` to run after the kotlin plugin. This can be achieved by redefining the `maven-compiler-plugin`'s executions after the `kotlin-maven-plugin`'s executions with the same execution `id` as the original definition.


abreslav:
Thanks for your contribution! I integrated your changes manually by adding them next to the old version, because it is shorter and satisfies many users.

PR #184, opened 2015-04-16
fixing typo

rciovati:


abreslav:
This change is incorrect: as the previous paragraph says, the `when` expression is not supposed to have an argument.

PR #185, opened 2015-04-17
Changes in 'Learn -> Videos -> Conferences and Events'

svtk:
Added an Android talk, removed one talk in russian (it was added to the events list before).

I'd like to add two more talks, but now I can't (https://youtrack.jetbrains.com/issue/KT-7499).


hhariri:
Any reason the other one was removed?

PR #187, opened 2015-04-21
index page fixes: fix standalone compiler link, promote Android, remove ...

yole:
...Expressive block, clarify text in Safe block.

Screenshot: https://www.evernote.com/shard/s23/sh/2bcdad40-a3ba-4bf6-9003-4af8da25f17d/8575fb8cbb0bda7982c9c3c713323196

### DIFF
--- a/_includes/pages/index/get-kotlin.html
+++ b/_includes/pages/index/get-kotlin.html
@@ -15,7 +15,7 @@
                 </p>
 
                 <div class="option-link-wrap">
-                    <a href="{{ site.baseurl }}/docs/tutorials/getting-started.html"
+                    <a href="{{ site.baseurl }}/docs/tutorials/command-line.html"
                        class="option-link">Download Compiler</a>
                 </div>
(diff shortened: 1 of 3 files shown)


_includes/pages/index/get-kotlin.html
```diff
@@ -15,7 +15,7 @@ <h3 class="option-title">Compiler</h3>
                 </p>
 
                 <div class="option-link-wrap">
-                    <a href="{{ site.baseurl }}/docs/tutorials/getting-started.html"
+                    <a href="{{ site.baseurl }}/docs/tutorials/command-line.html"
                        class="option-link">Download Compiler</a>
```
hhariri: Was it decided that this points to a different tutorial in the end or to the compiler?

yole: I think we wanted to point to the tutorial. The link to the compiler itself is the first thing in the tutorial.

PR #191, opened 2015-04-29
Translations in Korean

minsoopark:
I'm not sure if it is good for you :)

It's not completed yet, but I'm keeping going.

and it doesn't have to be merged immediately.


yole:
Thanks for putting the effort into translating the documentation! Unfortunately, since we don't have the resources to update the Korean translation of the documentation as the language changes, we won't be able to publish it on kotlinlang.org at this time. You're welcome to publish the translation on your own Web site.

PR #193, opened 2015-05-03
Command line Url is broken

jmfayard:
should be http://kotlinlang.org/docs/tutorials/command-line.html


jmfayard:
I have not jenkins installed so I could not test

yole:
Thanks for reporting the problem! I've pushed the correct fix to it.

PR #194, opened 2015-05-03
Fix NPE in example code

phase:
I added a constructor to the `Customer` class so calling `println(customer.getName())` wouldn't throw a NPE.


hhariri:
Thanks.

PR #202, opened 2015-06-01
Clarified abstract classes / functions not needing open

bdavisx:
I think the new text is clearer around not needing open on abstract classes and functions.


yole:
Thanks for your contribution!

PR #206, opened 2015-06-09
Added Annotation Mark in Classes and Inheritance chapter.

kyungkoo:
added missed annotation mark @ in 86 line.


yole:
It's not missed. The @ character is not required in this context.

PR #216, opened 2015-07-23
Fixed invalid Example.

kyungkoo:
1. Array's size property is deprecated. so It changed to size() method.
2. Character changed to Char because of compile error.


yole:
Thanks for your fix!

kyungkoo:
Thanks :+1:

PR #223, opened 2015-07-28
Add docs on tail recursion to functions.md and reorganise

mikehearn:
The tailRecursive annotation was undocumented, this fixes it.

Also the document had become a bit disorganised, with "how to call functions" sitting at the bottom of the document instead of the top, and nesting levels not reflecting the actual content. Cleans up a bit.

### DIFF
--- a/docs/reference/functions.md
+++ b/docs/reference/functions.md
@@ -16,6 +16,42 @@ fun double(x: Int): Int {
 }
 ```
 
+## Function Usage
+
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
@mikehearn also it would be great to notice that tail recursion is not allowed with in try/catch/finally blocks

mikehearn:
Done, thanks.

abreslav:
Merged manually, thanks

PR #229, opened 2015-08-17
Add Fuel into OSS Kotlin list

kittinunf:
I would love to add [Fuel](https://github.com/kittinunf/Fuel) into the list.

:sparkles:


hhariri:
Cool, thanks!

PR #233, opened 2015-08-24
Make folder structure slightly clearer

keirlawson:
I got confused by this when trying to set up my first Kotlin Gradle project


yole:
Thanks for your fix!

PR #235, opened 2015-08-31
Update kotlin-doc.md

K0zka:


yole:
Sorry, this isn't a correct change. The document describes the KDoc language, which is not being replaced by anything. The document generation tool called KDoc has indeed been removed from M13, but it was never documented, and it's not what this document is talking about.

PR #237, opened 2015-09-04
Update ranges.md

RodrigoQuesadaDev:
According to current Progression's code, the start and end elements can be of any type (should the rest of the page be updated to reflect this fact?). Also, according to the current documentation characters are not numbers any more: http://kotlinlang.org/docs/reference/basic-types.html#numbers


yole:
Thanks for the fix!

PR #239, opened 2015-09-09
Update extensions.md

arturh:


yole:
Thanks for your fix!

udalov:
I don't think the fix is correct. `List#get` takes an `Int` parameter, not a `T`.

PR #241, opened 2015-09-10
docs updated for M13

yole:


docs/reference/basic-types.md
```diff
@@ -191,8 +191,8 @@ of each array element given its index:
 
 ``` kotlin
 // Creates an Array<String> with values ["0", "1", "4", "9", "16"]
-val asc = Array(5, {i -> (i * i).toString()})
-```
+val asc = Array(5, { i -> (i * i).toString() })
+```1
```
udalov: bad character

docs/reference/classes.md
```diff
@@ -309,3 +309,37 @@ inside that class.
...
+    class Const(val number: Double) : Expr()
+    class Sum(e1: Expr, e2: Expr) : Expr()
+    object NotANumber : Expr()
+}
+```
+
+Note that multiple levels of inheritance are also allowed, as long as the entire hierarchy is nested within the
+enclosing sealed class.
+
+The key benefit of using sealed classes comes into play when you use them in a `when` expression. If it's possible
```
udalov: Would make sense to add a link to the `when`-expression section

docs/reference/coding-conventions.md
```diff
@@ -18,10 +18,17 @@ If in doubt default to the Java Coding Conventions such as:
 * use 4 space indentation
 * public functions should have documentation such that it appears in Kotlin Doc
 
-Kotlin does not have fields as a primary concept in the language -- it only has properties.
-Avoid the use of prefixes on properties, such as _ or m_ or other kinds of notation.
-If you need access to a backing field of a property, use the $ prefix: `$foo` to refer to a field behind property `foo`;
-never create a private property and call it `_foo`
+In some cases, it's necessary to define a [backing property](properties.html#backing-properties) for a property.
+(For example, it can be necessary when the public view of a property is a read-only collection, and inside the class
+the collection is mutable). In that case, the name of the backing property should use the _ prefix:
```
udalov: Personally I don't think this should be in the official style guide. I'm not sure we even use this exact scheme in our project. Why prefix and not postfix, for example? Also it's located way too high for its relevance

docs/reference/control-flow.md
```diff
@@ -143,9 +143,9 @@ for (item: Int in ints) {
 
 As mentioned before, *for*{: .keyword } iterates through anything that provides an iterator, i.e.
 
-* has an instance- or extension-function `iterator()`, whose return type
-  * has an instance- or extension-function `next()`, and
-  * has an instance- or extension-function `hasNext()` that returns `Boolean`.
+* has an member- or extension-function `iterator()`, whose return type
+  * has an member- or extension-function `next()`, and
+  * has an member- or extension-function `hasNext()` that returns `Boolean`.
```
udalov: `an` -> `a`

docs/reference/delegated-properties.md
```diff
@@ -85,41 +85,41 @@ For a **mutable** property (a *var*{:.keyword}), a delegate has to _additionally
  
 ## Standard Delegates
 
-The `kotlin.properties.Delegates` object from the standard library provides factory methods for several useful kinds of delegates.
+The Kotln standard library provides factory methods for several useful kinds of delegates.
```
udalov: typo

docs/reference/functions.md
```diff
@@ -161,18 +164,14 @@ fun double(x: Int) = x * 2
...
-* Functions with block body must always specify return types explicitly, unless it's intended for them to return `Unit`, [in which case it is optional](#unit-returning-functions). 
+Functions with block body must always specify return types explicitly, unless it's intended for them to return `Unit`, [in which case it is optional](#unit-returning-functions).
 Kotlin does not infer return types for functions with block bodies because such functions may have complex control flow in the body, and the return
 type will be non-obvious to the reader (and sometimes even for the compiler). 
 
 
 ### Variable number of arguments (Varargs)
 
-The last parameter of a function may be marked with `vararg` annotation
+The last parameter of a function may be marked with `vararg` modiifer:
```
udalov: typo

docs/reference/functions.md
```diff
@@ -191,8 +190,9 @@ allowing a variable number of arguments to be passed to the function:
 
 Inside a function a `vararg`-parameter of type `T` is visible as an array of `T`, i.e. the `ts` variable in the example above has type `Array<out T>`.
 
-Only one parameter may be annotated as `vararg`. It may be the last parameter or the one before last,
-if the last parameter has a function type (allowing a lambda to be passed outside parentheses).
+Only one parameter may be annotated as `vararg`. If a `vararg` parameter is not the last one in the list, values for the
```
udalov: annotated -> marked?

docs/reference/inline-functions.md
```diff
@@ -158,17 +158,17 @@ inline fun <reified T> TreeNode.findParentOfType(): T? {
...
-Though reflection may not be needed in many cases, we can still use it with a reified type parameter: `javaClass()` gives us access to it:
+Though reflection may not be needed in many cases, we can still use it with a reified type parameter:
 
 ``` kotlin
-inline fun methodsOf<reified T>() = javaClass<T>().getMethods()
+inline fun methodsOf<reified T>() = T::class.members
 
 fun main(s: Array<String>) {
-  println(methodsOf<String>().joinToString("\n"))
+  println(methodsOf<StringBuilder>().joinToString("\n"))
```
udalov: I'd rename `methodsOf` to `membersOf`

docs/reference/java-interop.md
```diff
@@ -91,6 +107,12 @@ mnemonic notation for them:
 * `(Mutable)Collection<T>!` means "Java collection of `T` may be mutable or not, may be nullable or not",
 * `Array<(out) T>!` means "Java array of `T` (or a subtype of `T`), nullable or not"
 
+#### Nullability annotations
+
+Java types which have nullability annotations are represented not as platform types, but as actual nullable or non-null
+Kotlin types. Currently, the compiler supports the JetBrains flavor of the nullability annotations (`@Nullable` and `@NotNull`
+from the `org.jetbrains.annotations` package.
```
udalov: Missing '(', also I guess would make sense to link to https://www.jetbrains.com/idea/help/nullable-and-notnull-annotations.html

docs/reference/java-interop.md
```diff
@@ -282,11 +304,11 @@ To retrieve the type information from an object, we use the javaClass extension
 val fooClass = foo.javaClass
 ```
 
-Instead of Java's `Foo.class` use javaClass<Foo>().
+Instead of Java's `Foo.class` use Foo::class.java().
```
udalov: Unneeded parentheses

docs/reference/java-interop.md
```diff
@@ -333,15 +356,14 @@ if (Character.isLetter(a)) {
 ### Java Reflection
 
 Java reflection works on Kotlin classes and vice versa. As mentioned above, you can use `instance.javaClass` or 
-`javaClass<ClassName>()` to enter Java reflection through `java.lang.Class`. You can then "convert" to Kotlin reflection
+`ClassName::class.java` to enter Java reflection through `java.lang.Class`. You can then "convert" to Kotlin reflection
 by calling `.kotlin`:
```
udalov: This suggestion no longer makes sense :) To get to Kotlin reflection I'd just use `ClassName::class`, no need to do `.java` and then `.kotlin`

docs/reference/java-interop.md
```diff
@@ -333,15 +356,14 @@ if (Character.isLetter(a)) {
...
  
 ``` kotlin 
 val kClass = x.javaClass.kotlin  
 ```
  
-In much the same way you can convert from Kotlin reflection to Java: `ClassName::class.java` is the same as `javaClass<ClassName>()`.
-Other supported cases include acquiring a Java getter/setter method or a backing field for a Kotlin property, 
+Other supported cases include acquiring a Java getter/setter method or a backing field for a Kotlin property,
 getting a containing `KPackage` instance for a Java class, and getting a `KProperty` for a Java field.
```
udalov: Not related to your changes, but I don't understand the part about `KPackage` and I think it should be removed...

docs/reference/java-interop.md
```diff
@@ -376,9 +398,11 @@ Kotlin code can be called from Java easily.
 
 ### Package-Level Functions
 
-All the functions and properties declared inside a package `org.foo.bar` are put into a Java class named `org.foo.bar.BarPackage`.
+All the functions and properties declared in a file `example.kt' inside a package `org.foo.bar` are put into a Java
```
udalov: Bad formatting, quote instead of backtick

docs/reference/java-interop.md
```diff
@@ -376,9 +398,11 @@ Kotlin code can be called from Java easily.
...
+class named `org.foo.bar.Example`.
```
udalov: `ExampleKt` in fact

docs/reference/java-interop.md
```diff
@@ -391,20 +415,71 @@ fun bar() {
...
+demo.Example.bar();
 ```
 
-For the root package (the one that's called a "default package" in Java), a class named `_DefaultPackage` is created.
+The name of the generated Java class can be changed using the `@JvmName` annotation:
+
+``` kotlin
+package demo
+
+@file:JvmName("DemoUtils")
```
udalov: This should come before the `package` directive

docs/reference/java-interop.md
```diff
@@ -531,7 +608,7 @@ And we want to call it from Java and catch the exception:
 ``` java
 // Java
 try {
-  demo.DemoPackage.foo();
+  demo.Example.foo();
```
udalov: `ExampleKt`

docs/reference/kotlin-doc.md
```diff
@@ -97,7 +97,7 @@ Specifies the version of the software in which the element being documented was
 Excludes the element from the generated documentation. Can be used for elements which are not part of the official
 API of a module but still have to be visible externally.
 
-> KDoc does not support the `@deprecated` tag. Instead, please use the `@deprecated` annotation.
+> KDoc does not support the `@Deprecated` tag. Instead, please use the `@deprecated` annotation.
```
udalov: `@Deprecated` annotation?

docs/reference/lambdas.md
```diff
@@ -70,20 +70,14 @@ fun <T, R> List<T>.map(transform: (T) -> R): List<R> {
...
 ```
 
 One other helpful convention is that if a function literal has only one parameter,
 its declaration may be omitted (along with the `->`), and its name will be `it`:
 
 ``` kotlin
-ints map {it * 2} // Infix call + Implicit 'it'
-```
-
-These conventions allow to write [LINQ-style](http://msdn.microsoft.com/en-us/library/bb308959.aspx) code:
```
udalov: Probably not very relevant, but I think it would be nice if something from the official documentation would be google-able with queries like "kotlin linq"

docs/reference/lambdas.md
```diff
@@ -96,7 +90,7 @@ A function literal or a function expression is an "anonymous function", i.e. a f
 but passed immediately as an expression. Consider the following example:
 
 ``` kotlin
-max(strings, {a, b -> a.length < b.length})
+max(strings, { a, b -> a.length < b.length })
```
udalov: Also `length()` is now a function

docs/reference/packages.md
```diff
@@ -48,11 +48,6 @@ import foo.Bar // Bar is accessible
 import bar.Bar as bBar // bBar stands for 'bar.Bar'
 ```
 
-## Visibility and Package Nesting
+## Visibility of Toplevel Declarations
```
udalov: Top-level?

docs/reference/visibility-modifiers.md
```diff
@@ -13,18 +13,8 @@ There are four visibility modifiers in Kotlin:
 
 * `private` --- visible only in the declaring scope and its subscopes (inside the same module);
```
udalov: I think this part about "its subscopes" no longer makes sense for package members, so probably we should just list the four modifiers here without explanation, given that they're thoroughly explained later

docs/reference/properties.md
```diff
@@ -137,6 +130,37 @@ public val table: Map<String, Int>
...
 
+
+## Late-Initialized Properties
+
+Normally, properties declared as having a non-null type must be initialized in the constructor.
+However, fairly often this is not convenient. For example, properties can be initialized through dependency injection,
+or in the setup method of a unit test. In this case, you cannot supply a non-null initializer in the constructor,
+but you still want to avoid null checks when referencing the property inside the body of a class.
+
+To handle this case, you can annotate the property with the `lateinit` modifier:
```
udalov: annotate -> mark?

docs/reference/visibility-modifiers.md
```diff
@@ -78,8 +67,8 @@ Examples:
 open class Outer {
     private val a = 1
     protected val b = 2
-    val c = 3 // internal by default
-    public val d: Int = 4 // return type required
+    internal val c = 3
+    public val d = 4  // public by default
```
udalov: The comment here doesn't help I think

docs/reference/annotations.md
```diff
@@ -76,30 +76,73 @@ annotation class Suspendable
...
+```
+
+If you don't specify a use-site target, the target is chosen according to the `@Target` annotation of the annotation
+being used. If there are multiple applicable targets, the first applicable target from the following list is used:
+
+  * `param`
+  * `field`
+  * `property`
+  * `get`
+  * `set`
```
udalov: I think `set` has a higher priority than `get`

docs/reference/delegated-properties.md
```diff
@@ -85,41 +85,41 @@ For a **mutable** property (a *var*{:.keyword}), a delegate has to _additionally
...
-    println(lazy)
+    println(lazyValue)
+    println(lazyValue)
 }
 ```
 
 
-If you want **thread safety**, use `blockingLazy()`: it works the same way, but guarantees that the values will be 
-computed only in one thread, and that all threads will see the same value.
+If you want **thread safety**, use the overload of `lazy()` that takes a lock as a parameter or pass `LazyThreadSafetyMode.SYNCHRONIZED`.
```
ilya-g: SYNCHRONIZED is a default value. So you have to opt-out from thread safety, rather than opt-in for it.

docs/reference/visibility-modifiers.md
```diff
@@ -9,24 +9,10 @@ title: "Visibility Modifiers"
...
-altering the implementation.
-
-``` kotlin
-public val foo: Int = 5    // explicit return type required
-public fun bar(): Int = 5  // explicit return type required
-public fun bar() {}        // block body: return type is Unit and can't be changed accidentally, so not required
-```
-
-Below please find explanations of these for different type of declaring scopes.  
+Below please find explanations of these for different type of declaring scopes.
```
udalov: for -> four

docs/reference/java-interop.md
```diff
@@ -28,6 +28,22 @@ fun demo(source: List<Int>) {
...
+Methods that follow the Java conventions for getters and setters (no-argument methods with names starting with `get`
+and single-argument methods with names starting with `set`) are represented as properties in Kotlin. For example:
+
+``` kotlin
+import java.util.Calendar
+
+fun calendarDemo() {
+    val calendar = Calendar.getInstance()
+    if (calendar.firstDayOfWeek == Calendar.SUNDAY) {  // call getFirstDayOfWeek()
+        calendar.firstDayOfWee = Calendar.MONDAY       // call setFirstDayOfWeek()
```
ilya-g: typo

docs/reference/lambdas.md
```diff
@@ -70,20 +70,14 @@ fun <T, R> List<T>.map(transform: (T) -> R): List<R> {
...
 
 ``` kotlin
-ints map {it * 2} // Infix call + Implicit 'it'
-```
-
-These conventions allow to write [LINQ-style](http://msdn.microsoft.com/en-us/library/bb308959.aspx) code:
-
-``` kotlin
-strings filter {it.length == 5} sortBy {it} map {it.toUpperCase()}
+ints.map { it * 2 } // Infix call + Implicit 'it'
```
ilya-g: Infix call mentioned in comment, but in fact it's not

docs/reference/annotations.md
```diff
@@ -8,29 +8,29 @@ title: "Annotations"
...
+This is very similar to how this works in Java.
 
 ``` kotlin
-@fancy class Foo {
-  @fancy fun baz(@fancy foo: Int): Int {
-    return (@fancy 1)
-  }
-}
+@Target(FIELD)
+@Retention(AnnotationRetention.SOURCE)
```
abreslav: Why qualified name here and a simple name above?

docs/reference/annotations.md
```diff
@@ -8,29 +8,29 @@ title: "Annotations"
...
+### Usage
 
 ``` kotlin
-fancy class Foo {
-  fancy fun baz(fancy foo: Int): Int {
-    @fancy fun bar() { ... }
-    return (@fancy 1)
+@Fancy class Foo {
+  @Fancy fun baz(@Fancy foo: Int): Int {
+    return (@Fancy 1)
```
abreslav: `@Fancy` is only applicable to fields in this example

docs/reference/annotations.md
```diff
@@ -76,30 +76,73 @@ annotation class Suspendable
...
+```
+
+The same syntax can be used to annotate the entire file. To do this, put an annotation with the target `file` at
+the top level of a file:
 
 ``` kotlin
-import org.junit.Test
-import org.junit.Assert.*
+@file:JvmName("Foo")
+```
```
abreslav: I'd clarify: before the package directive, or all import if the package directive is not there

docs/reference/annotations.md
```diff
@@ -8,29 +8,29 @@ title: "Annotations"
...
 
 ``` kotlin
-annotation class fancy
+annotation class Fancy
 ```
 
-### Usage
+To specify the possible targets, retention policy and other attributes of the annotation, you can annotate the
+annotation class itself with the corresponding annotations (@Target, @Retention, @Repeatable, @MustBeDocumented).
+This is very similar to how this works in Java.
```
abreslav: Needs to be explained nevertheless. Our docs shouldn't be too Java-dependent.

docs/reference/classes.md
```diff
@@ -309,3 +309,37 @@ inside that class.
 Even more specifically, if you declare a [companion object](object-declarations.html#companion-objects) inside your class,
 you'll be able to call its members with the same syntax as calling static methods in Java/C#, using only the class name
 as a qualifier.
+
+
+## Sealed Classes
+
+Sealed classes are used for representing restricted class hierarchies, when a value can have one of the types from a
+limited set, but cannot have any other type. They are, in a sense, an extension of enum classes: the set of values
```
abreslav: Comparing to enum classes is a very long shot, to my mind. Referring to algebraic data types explicitly would be better.

docs/reference/classes.md
```diff
@@ -309,3 +309,37 @@ inside that class.
...
+``` kotlin
+sealed class Expr {
+    class Const(val number: Double) : Expr()
+    class Sum(e1: Expr, e2: Expr) : Expr()
+    object NotANumber : Expr()
+}
+```
+
+Note that multiple levels of inheritance are also allowed, as long as the entire hierarchy is nested within the
+enclosing sealed class.
```
abreslav: I doubt this is correct. Doesn't `sealed` only restrict direct subclasses, and has no idea about indirect ones? // cc @mglukhikh

docs/reference/equality.md
```diff
@@ -17,13 +17,10 @@ In Kotlin there are two types of equality:
 Referential equality is checked by the `===` operation (and its negated counterpart `!==`). `a === b` evaluates to
 true if and only if `a` and `b` point to the same object.
 
-Alternatively, you can check for reference equality using a built-in inline function `identityEquals()`
-that can be called in the following way
+Alternatively, you can check for reference equality using a built-in inline function `identityEquals()`:
```
abreslav: I believe we should get rid of `identityEquals` in favor of `===`, but this is subject to discussion

docs/reference/idioms.md
```diff
@@ -170,7 +165,7 @@ fun transform(color: String): Int {
 }
 ```
 
-### Return on try catch block
+### 'try/catch' as an expression
```
abreslav: I'd leave out "as an": try/catch expression

docs/reference/idioms.md
```diff
@@ -184,7 +179,7 @@ fun test() {
 }
 ```
 
-### Return on if statement
+### 'if' as an expression
```
abreslav: same here

docs/reference/lambdas.md
abreslav: +1 // cc @belovrv

docs/reference/java-interop.md
```diff
@@ -391,20 +410,71 @@ fun bar() {
...
+
+Having multiple files which have the same generated Java class name (the same package and the same name or the same
+@JvmName annotation) is normally an error. However, the compiler has the ability to generate a single Java facade
+class which has the specified name and contains all the declarations from all the files which have that name.
+To enable the generation of such a facade, use the @JvmMultifileClass annotation in all of the files.
+
+``` kotlin
+// oldutils.kt
+@file:JvmName("Utils")
+@file:JvmMultifileClass
```
dnpetrov: Do we want to promote `@[...]`  syntax? 
E.g., here you can write `@file:[JvmName("Utils") JvmMultifileClass]`.

docs/reference/visibility-modifiers.md
yole: No, actually 'for' is correct here. We have four modifiers for two different types of enclosing scopes. :)

udalov: Ok :)

docs/reference/java-interop.md
```diff
@@ -391,20 +410,69 @@ fun bar() {
...
+// Java
+new demo.Foo();
+demo.DemoUtils.bar();
+```
+
+Having multiple files which have the same generated Java class name (the same package and the same name or the same
+@JvmName annotation) is normally an error. However, the compiler has the ability to generate a single Java facade
+class which has the specified name and contains all the declarations from all the files which have that name.
+To enable the generation of such a facade, use the @JvmMultifileClass annotation in all of the files.
+
```
dnpetrov: Do we want to advertise it in M13?

PR #253, opened 2015-09-22
Improve "this expressions", remove incorrect code

mr-max:
Running the current code sample of this-expressions.md results in errors.


yole:
Pushed a different fix for this page.

PR #256, opened 2015-09-28
docs update for M14

yole:


docs/reference/data-classes.md
```diff
@@ -23,10 +23,14 @@ the primary constructor_:
   * `copy()` function (see below).
   
 If any of these functions is explicitly defined in the class body or inherited from the base types, it will not be generated.
-  
-*NOTE* that if a constructor parameter does not have a `val` or `var` in front of it, it will not be included in computation
-of all these functions; nor will be properties declared in the class body or inherited from the superclass.
 
+To ensure consistency and meaningful behavior of the generated code, data classes have to fulfil the following requirements:
+
+  * The primary constructor needs to have at least one parameter;
```
cy6erGn0m: this is not exact correct as one parameter is not enough: it should be `val` or `var`

abreslav:
Looks good to me

PR #257, opened 2015-09-29
Start working on translate-to-cn

Jween:
A startup pull for translate kotlin docs from English to Chinese-Simplified.
also contains three translated pages: Base Syntax, idioms and Coding-Conventions inside Getting started section.
More pages will be translated maybe 1 to 3 pages per workday-in-China


yole:
Thanks for your effort! However, I have to request you to host any translations of the Web site in your own projects. No one on our team speaks Chinese, and therefore we are unable to maintain an official Chinese translation of the documentation and to update it when the language changes.

PR #270, opened 2015-10-21
docs updates for 1.0 beta

yole:


docs/reference/annotations.md
```diff
@@ -74,6 +74,27 @@ annotation class Special(val why: String)
 @Special("example") class Foo {}
 ```
 
+Allowed parameter types are:
+
+ * types that correspond to Java primitive types (Int, Long etc.);
+ * strings;
+ * classes (`Foo::class`);
+ * enums;
+ * other annotations;
```
abreslav: Example would make sense: there were questions about having to write `@` before these annotations

yole: There is one below.

PR #271, opened 2015-10-21
Added return types in example, and added operator modifier to plus()

nhaarman:


yole:
Thanks for the fixes!

PR #273, opened 2015-10-22
precedence of "?:" over "===" ?!?!

stangls:
On the page http://try.kotlinlang.org/#/Examples/Hello, world!/Simplest version/Simplest version.kt
the following:

```java
class X(){
    fun equals(x:X) = x===this
}

    var x=X()
    println ( x ) 
    println ( x.equals(x) )
    println ( x == x )
    println ( if (x==null) x===null else x==x )
    println ( x?.equals(x) ?: x === null ) // this yields "false"
    println ( x?.equals(x) ?: (x === null) )
}
```
prints:
```java
X@40bde56d
true
true
true
false
true
```


yole:
Thanks for the correction!

PR #275, opened 2015-10-25
'with' use case

beyondeye:
A very useful idioms that is hard to learn about from current documentation  is using the 'with' function for calling multiple methods on an object instance


yole:
Thanks for your contribution! Please correct the issues noted above and rebase your changes into a single commit, and I'll merge it.

beyondeye:
Done!

PR #278, opened 2015-10-25
added `General Overviews` section

voddan:
General Overviews is a section for introduction sessions longer than `Get Started`


voddan:
@belovrv what's the status of the PR?

PR #282, opened 2015-10-25
Sort oss-projects.yml alphabetically

cbeust:


yole:
I think that a topical sorting would be more appropriate than an alphabetic one, to be honest.

cbeust:
Yup, that's fine too I think anything is better than the current random order :)

PR #284, opened 2015-10-27
Specify bash shell in INSTALL.sh shebang

homburg:
On Ubuntu, the install script produces the following syntax error with default shebang `#!/bin/sh`
``` 
./INSTALL.sh: 10: ./INSTALL.sh: [[: not found
```
Default `/bin/sh` on ubuntu is a link to `/bin/dash` and it seems the script is intended for bash shell.

Specifying the shebang explicitly as `/bin/bash` should work on linux and osx


yole:
Thanks for your contribution!

PR #286, opened 2015-10-28
Update ranges.md

mzhua:


yole:
The original code is actually correct.

PR #289, opened 2015-10-31
Add operator keyword to rangeTo methods

hannesstruss:
As per [Operator Overloading](https://kotlinlang.org/docs/reference/operator-overloading.html):

> Functions that overload operators need to be marked with the `operator` modifier.


belovrv:
Thanks

PR #300, opened 2015-11-03
Update idioms.md

tomtanti:
IntArray is expected, not Unit.


yole:
The original heading was correct: `apply()` allows to use methods that return `Unit`, such as `fill` in this example. I've corrected the return type of the method.

PR #301, opened 2015-11-04
Update oss-project by adding Result

kittinunf:
Adding [Result](https://github.com/kittinunf/Result) into the list

Thank you for making this awesome language.


yole:
Merged manually. Thanks!

PR #308, opened 2015-11-09
replace .size():Int method to .size property

java1cprog:
'size():Int' is deprecated. Use property 'size" instead


yole:
Thanks!

PR #316, opened 2015-11-12
Add Blaze tool

jjlauer:
Execute your Kotlin or Kotlin scripts with the Blaze scripting stack for the JVM


jjlauer:
Curious about the status of this PR?

yole:
The "Resources" page is intended to show libraries written in Kotlin or tools created specifically for Kotlin developers. While we do appeciate that Blaze supports Kotlin, it's not a Kotlin-specific tool, so we'd rather not show it on this page.

PR #323, opened 2015-11-21

clemp6r:
Add Wizbii in the list of companies using Kotlin


yole:
Thanks!

clemp6r:
Thanks. 

PS: it would be nice that the list be randomized.

PR #325, opened 2015-11-22
Fixed named companion example

Endran:
The example of the named companion was incorrect. It omitted the mandatory name of the companion object. This has now been added.


yole:
The example is actually correct. The name of the companion object is not mandatory when accessing its members.

PR #335, opened 2015-12-05
Fix typo in control-flow docs

PhilGlass:
Consequently doesn't make sense in this context. I would guess the intention was either "sequentially" or "consecutively" (both work, but I'd say the former is slightly clearer).


yole:
Thanks!

PR #338, opened 2015-12-12
Fix minor grammar error

segunfamisa:
This is just a very minor fix in the grammar I found while reading the doc


yole:
The original text is correct. For more information on the term "implementation inheritance", please check out https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming)

PR #339, opened 2015-12-22
Update docs for Beta 4

abreslav:


abreslav:
@yole When you have time, give it a scan please

docs/reference/reflection.md
```diff
@@ -97,7 +97,14 @@ value using `get()` or retrieve the property name using the `name` property. For
 the [docs on the `KProperty` class](/api/latest/jvm/stdlib/kotlin.reflect/-k-property/index.html).
 
 For a mutable property, e.g. `var y = 1`, `::y` returns a value of type [`KMutableProperty<Int>`](/api/latest/jvm/stdlib/kotlin.reflect/-k-mutable-property/index.html),
-which has a `set()` method.
+which has a `set()` method.                     
+
+A property reference can be used where a function with no parameters is expected:
```
udalov: Not really correct because of KProperty0 / KProperty1 and Function0 / Function1 distinction, I'd say something like the number of receivers of the property should be equals to the number of parameters of the function

abreslav: Maybe I'm overly cautious here, but it seemed like too many details to me

PR #345, opened 2015-12-28
remove example which is no longer valid (KT-10519)

yole:


docs/reference/classes.md
```diff
@@ -267,22 +267,6 @@ and provide our own implementation that eliminates the ambiguity.
...
-
-interface B {
-  fun f() { print("B") }
-}
-
-class C() : A(), B {
-  // We are not required to override f()
-}
-```
-
```
abreslav: I believe this should still be valid:

```
interface A {
    fun f()
}

open class B {
    open fun f() { ... }
}

class C : B(), A {
    // We are not required to override f()
}
```

PR #347, opened 2015-12-30
Fixing Kotlin Example that shows Java Code

desertjim:
Catch block was in Java format instead of Kotlin


yole:
This is in fact an example showing Java code, so it's correct that it uses Java syntax.

desertjim:
Ah sorry.  I thought it was supposed to show Kotlin since the mark down indicated it was Kotlin code.

PR #350, opened 2016-01-09
Fix parameter types

doubleo2:
Extension function `swap` parameters should be of type `T` to match the `MutableList` type


yole:
The parameters are named "index1" and "index2". The type of index in a list is `Int`, not the type of the list element.

PR #351, opened 2016-01-09
Update inline-functions.md

BiaoWu:
line 168 : Type parameters must be placed before the name of the function.


yole:
Thanks for fixing this!

PR #352, opened 2016-01-10
Make description when to use @JvmStatic easier to understand

stefreak:


yole:
I've reworded the text slightly differently. Thanks anyway!

stefreak:
Nice, that's even better! :)

PR #356, opened 2016-01-12
unnecessary phase configuration

sebastiankirsch:
Since you defined a default phase for the mojos, explicitly configuring a phase is unnecessary.


abreslav:
Thanks!

PR #360, opened 2016-01-17
Update type-safe-builders.md

foliot:
Fixed the Element interface and reformatted.

Great job with Kotlin by the way! The language is extremely well-designed and a huge improvement over Java.


foliot:
Come to think of it, it may be better to make the Element interface an abstract class and implement toString there. That makes more sense, though it is different than what is on the Try Kotlin site.

yole:
Thanks for your contribution!

PR #369, opened 2016-02-07
Added awesome kotlin list.

IRus:


voddan:
not sure it should be just another link on the page. It provides an alternative to the list itself (or derived from it), so it may be put at the top with a proper explanation of what it is (in html)

IRus:
Close this because of #383

PR #378, opened 2016-02-11
Array.size changed to Array.size()

zikil:
Do not use deprecated API.


yole:
As of Kotlin 1.0 RC, the Array.size property is not deprecated, and the Array.size() method does not exist. The documentation text is correct.

PR #381, opened 2016-02-12
increased the number of OSS kotlin projects

voddan:
increased the number of OSS kotlin projects from 400 to 2000


yole:
Thanks!

PR #390, opened 2016-02-15

dgallego:
Keeping in mind that we have recently started to use Kotlin in our Android projects, we think it is important to support it by joining the community :)


yole:
Merged manually.

PR #403, opened 2016-02-16
added `data classes` and `read-only collections` as kotlin features

voddan:
added `data classes` and `read-only collections` as kotlin features over Java


yole:
Thanks!

PR #407, opened 2016-02-17
Fixed gradle scope reference

robfletcher:
`compileTest` should be `testCompile`.


cy6erGn0m:
agree, see proof https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations

yole:
Thanks, and sorry for the confusion!

PR #412, opened 2016-02-21
Add class member references in Kotlin but not in Scala

jtdev:
For the Scala vs Kotlin comparison: https://kotlinlang.org/docs/reference/comparison-to-scala.html
A feature in Kotlin that Scala doesn't have.


yole:
Thanks!

PR #418, opened 2016-02-25
add !! operator to the list of sources for NPEs

Boris-de:
As this is an explicit list of all possible sources of NPEs I think it should also contain the `!!`-operator that is described at the end of this document ("this will [...] throw an NPE if b is null")

### DIFF
--- a/docs/reference/null-safety.md
+++ b/docs/reference/null-safety.md
@@ -17,6 +17,7 @@ would be the equivalent of a `NullPointerException` or NPE for short.
 Kotlin's type system is aimed to eliminate `NullPointerException`'s from our code. The only possible causes of NPE's may be
 
 * An explicit call to `throw NullPointerException()`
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
Doesn't it throw KotlinNullPointer instead of NullPointerException?

Boris-de:
You are right, it does throw a `KotlinNullPointerException`, but this extends from `java.lang.NullPointerException`. So basically it is a normal NPE and it would seem overly specific to say that it doesn't throw it because it only throws a subclass.

Perhaps it should be rephrased to be exact on what this list includes. Either its only for plain java-NPEs in which my addition to the list would be obsolete, or its for all kinds of NPEs in which case the operator should be added.
If the list is only about the plain java-NPEs perhaps the section about the operator could be change to state precisely that it does generate a KotlinNullPointerException to emphasize the difference to the list of possible sources of NPEs.

PR #421, opened 2016-03-02
I would like to fix unfair comparison and mistakes

kmizu:
I found many mistakes in this page, I would like to fix mistake.

> In Scala, sometimes it’s very hard to tell what’s happening in your code without using a debugger, because too many implicits get into the picture

Because implicit conversions and parameters are resolved in compile time, the debugger is useless.  It's wrong.

> Yield operator

Scala has no 'yield' operator(`yield` is just a keyword used in for-expression.  Did you researched about Scala in a serious manner really?

> We plan to support [Project Valhalla](http://openjdk.java.net/projects/valhalla/) once it is released as part of the JDK

In serious comparison to actual language, a word such as `plan` should not be used.  If Scala creator say 'Built-in XML will be planned to be removed in future', do you write it in language comparison really?

> Member references

There are member references in Scala as the followings:

```scala
def  isOdd(x: Int) = x % 2 != 0
val numbers = List(1, 2, 3)
println(numbers.filter(isOdd _)) // prints List(1, 3)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Working with site
 
 ## Rake tasks
 
-Site uses [rake](https://github.com/jimweirich/rake) for build management.
+Site uses [rake](https://github.com/ruby/rake) for build management.
 To list available commands, run `rake` from project folder with no arguments.
 
 - `rake build` generates site in the `_site` folder. Use `dest` parameter to specify another location: `rake build dest=/path`.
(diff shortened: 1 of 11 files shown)


abreslav:
Looks like something has gone wrong with git on your side. The PR contains commits other than yours. Could you fix this?

kmizu:
Hmmm.  This PR contained only my commits several days ago.  my commits may be rebased after that.  I will create PR again at all.

kmizu:
@abreslav I sent a new PR https://github.com/JetBrains/kotlin-web-site/pull/429.

PR #423, opened 2016-03-04
added "Kotlin tour" section

voddan:
As was rightfully mentioned here https://medium.com/@elviraw/kotlin-and-ceylon-3ee011125b7d#.7ega2fcy9

 > Ceylon has a very nice Tour to get you started. Kotlin, not so much.

I personally think that the kotlin documentation in its current state is very good at "touring" a beginner through the kotlin language. So the problem here IMHO is it that many people do not read the reference through. I assume this happens because they think it is boring and takes a lot of time. The section I wrote is meant to solve the problem by attracting their attention and point the beginners to reading the reference.


yole:
Merged manually with some rewording.

voddan:
Thanks! Your version sounds nicely than mine.

PR #425, opened 2016-03-05
Fixed typo with arrayOf

vkotovv:
Incorrect init method: array() instead of arrayOf()


vkotovv:
Also, I found array() call [here](https://kotlinlang.org/docs/reference/dynamic-type.html):
```kotlin
dyn.whatever(1, "foo", dyn) // 'whatever' is not defined anywhere
dyn.whatever(*array(1, 2, 3))
```
Should I fix that as well?

yole:
Thanks! I've fixed the usage in dynamic-type.md as well.

PR #427, opened 2016-03-07
"API" button on the front page

voddan:
because this is the place on the site where I (as a Kotlin user) go most often


yole:
We decided not to add an extra item to the top-level navigation, but instead we made a more easy-to-find link to the standard library docs on the "Learn" page.

voddan:
Ok, sounds good

PR #428, opened 2016-03-08
Collections and Pattern Matching

kolotyluk:
Additional useful information for developers accustom to Scala collections and pattern matching. I recommend putting this section in alphabetical order for easier reading.

### DIFF
--- a/docs/reference/comparison-to-scala.md
+++ b/docs/reference/comparison-to-scala.md
@@ -32,8 +32,14 @@ Taking this into account, if you are happy with Scala, you most likely do not ne
 * Yield operator
 * Actors
     * Kotlin supports [Quasar](http://www.paralleluniverse.co/quasar/), a third-party framework for actor support on the JVM
-* Parallel collections
-    * Kotlin supports Java 8 streams, which provide similar functionality
+* Collections
+    * Scala has its own unique collections library, while Koltin relies on Java collections with some extensions
+        * Scala's basic list type is an immutable linked-list similar to LISP, while Kotlin is based on Java ArrayList
+        * Scala has Traversable and Iterable, while Kotlin just supports Iterable.
+    * Parallel collections
+        * Kotlin supports Java 8 streams, which provide similar functionality
+* Pattern Matching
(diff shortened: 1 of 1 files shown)


docs/reference/comparison-to-scala.md
```diff
@@ -32,8 +32,14 @@ Taking this into account, if you are happy with Scala, you most likely do not ne
 * Yield operator
 * Actors
     * Kotlin supports [Quasar](http://www.paralleluniverse.co/quasar/), a third-party framework for actor support on the JVM
-* Parallel collections
-    * Kotlin supports Java 8 streams, which provide similar functionality
+* Collections
+    * Scala has its own unique collections library, while Koltin relies on Java collections with some extensions
```
ilya-g: I think it should be formulated like "Scala has its own comprehensive hierarchy of collection types, incompatible with java.util.* collections, while Kotlin just utilizes java.util.* collections with some enhancements and extensions" (possible link to Collections page).

I'm sure there's a ton of differences could be derived from that fact, why do you think the particular two of them (namely about lists and traversable) deserve emphasis?

docs/reference/comparison-to-scala.md
```diff
@@ -32,8 +32,14 @@ Taking this into account, if you are happy with Scala, you most likely do not ne
...
+        * Scala's basic list type is an immutable linked-list similar to LISP, while Kotlin is based on Java ArrayList
+        * Scala has Traversable and Iterable, while Kotlin just supports Iterable.
+    * Parallel collections
+        * Kotlin supports Java 8 streams, which provide similar functionality
+* Pattern Matching
+    * Scala's match is (currently) more powerful than Kotlin's when, but this may change in the future.
```
ilya-g: We shouldn't mention in comparison those Kotlin features that are under consideration  without clear plans when and how they are going to be implemented.

docs/reference/comparison-to-scala.md
soc: I think https://github.com/JetBrains/kotlin-web-site/pull/500 is more accurate.

yole:
The "Comparison to Scala" page has been removed from the Web site.

PR #433, opened 2016-03-11
added benchmarking idiom

voddan:
initial SO question: http://stackoverflow.com/questions/35930456/how-to-benchmark-a-kotlin-program

### DIFF
--- a/docs/reference/idioms.md
+++ b/docs/reference/idioms.md
@@ -256,3 +256,11 @@ stream.buffered().reader().use { reader ->
     println(reader.readText())
 }
 ```
+
+### Benchmarking
+
+``` kotlin
+val benchmark = measureNanoTime {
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
Not sure about title "Benchmark" as in fact it is not..
Also we need some disclaimer here about actual benchmarking

voddan:
How would you call it? 
Wiki says : "Benchmarking is usually associated with assessing performance characteristics of computer hardware [....] Software benchmarks are, for example, run against compilers or database management systems."
https://en.wikipedia.org/wiki/Benchmark_(computing)

yole:
I think the correct advice is to use JMH. measureNanoTime() is not a good benchmarking tool, and we shouldn't promote it as such.

PR #437, opened 2016-03-22
Substituting type `E` for `T` as, `T` is used throughout the page

aganezov:
Second fixed entry, where `? extends E` is substituted with `? extends T` is not just for overall conciseness, but also due to the explanation after corresponding code example, where `? extends T` is used.

### DIFF
--- a/docs/reference/generics.md
+++ b/docs/reference/generics.md
@@ -49,8 +49,8 @@ interface. What's the signature of this method? Intuitively, we'd put it this wa
 
 ``` java
 // Java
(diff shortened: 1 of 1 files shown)


yole:
The examples show the `Collection` interface from Java, which uses `E` as the type parameter name: https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html

The text below the examples talks about a generic "? extends" clause, not related to any specific type, so it's not necessary for it to match the name of the type parameter used in the examples.

PR #442, opened 2016-03-30
add-design-patterns-section

voddan:
an extension to the `idioms` charapter with some Java design patterns converted to Kotlin

### DIFF
--- a/docs/reference/design-patterns.md
+++ b/docs/reference/design-patterns.md
@@ -19,7 +19,7 @@ Patterns that have dedicated Kotlin syntax for their support.
 
 ``` java
 public class Singleton {
-    String info = "any usefull shared information";
+    String data = "some usefull shared data";
 
     private Singleton() {}
 
@@ -33,7 +33,7 @@ public class Singleton {
 }
 
 // Usage:
-String info = Singleton.getInstance().info;
+String data = Singleton.getInstance().data;
 ```
 
 Java code above provides you with a lazy-initialised thread-safe [singleton instance](https://en.wikipedia.org/wiki/Singleton_pattern).
@@ -41,11 +41,11 @@ Kotlin [`object`](object-declarations.html#object-declarations) has the same sem
 
 ``` kotlin
 public object Singleton {
-    val info = "any usefull shared information"
+    val data = "any usefull shared information"
 }
 
 // Usage:
-val info = Singleton.info
+val data = Singleton.data
 ```
 
 
--- a/docs/reference/design-patterns.md
+++ b/docs/reference/design-patterns.md
@@ -48,6 +48,60 @@ public object Singleton {
 val data = Singleton.data
 ```
 
+### Delegate (Decorator)
+
+``` java
+interface Printer {
+    void print(String msg);
+}
+
+final class HP3000 implements Printer {
+    void print(String msg) {
+        System.out.println(msg);
+    }
+}
+
+class MyPrinter implements Printer {
+    Printer p = new HP3000();
+
+    void print(String msg) {
+        p.print(msg);
+    }
+}
+
+// Usage:
+MyPrinter mp = new MyPrinter();
+mp.print("Hello!");
+```
+
+[Delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern) allows inheriting behaviour of a class without extending it.
+This is invaluable when the class, implementing the behaviour, is final.
+Also this technique has no limitation on the number of delegated interfaces, thus providing a mean for [multiple inheritance](https://en.wikipedia.org/wiki/Multiple_inheritance).
+Kotlin cuts the boilerplate here with its [class delegation](https://kotlinlang.org/docs/reference/delegation.html#class-delegation) syntax:
+
+``` kotlin
+// --- interface `Printer` and class `HP3000` are defined in the Java snipet above ---
+
+class MyPrinter : Printer by HP3000()
+
+// Usage:
+val mp = MyPrinter()
+mp.print("Hello!")
+```
+
+The delegate object can be a constructor parameter, allowing runtime behaviour configuration, or acting as *dependency injection*:
+
+``` kotlin
+class ConfigurablePrinter(printer: Printer = HP3000()) : Printer by printer
+
+// Usage:
+val workPrinter = ConfigurablePrinter()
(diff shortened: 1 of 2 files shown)


docs/reference/design-patterns.md
```diff
@@ -0,0 +1,363 @@
...
+
+[Value objects](https://en.wikipedia.org/wiki/Value_object) are used to transfer data between parts of a system.
+This Kotlin class is *identical* to the Java class above:
+
+``` kotlin
+class StudentVO(var name: String?,
+                var rollNo: Int)
+```
+
+The pattern might be improoved by requaring field immutability and prohibiting nullable values.
```
cypressious: requiring

cypressious:
Saying that the strategy pattern is obsolete is a very strong opinion and I don't agree. In fact, it can even be improved using Kotlin language features like extension methods to make it more useful:

```kotlin
abstract class Foo(val bar: Bar) {
    fun doStuff() = bar.doStuffInternal()

    abstract fun Bar.doStuffInternal()
}
```

voddan:
@cypressious I agree that "obsolete" is out of place here. For example I came up with a variation of Builder pattern in Kotlin. 

I am thinking about removing those groups all together. They look cool (like "kool"), but don't communicate the idea right. Opinions?

docs/reference/design-patterns.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Natively supported patterns
+
+Patterns that have dedicated Kotlin syntax for their support.
+
+### Singleton
+
+``` java
+public class Singleton {
+    String data = "some usefull shared data";
```
MarcinMoskala: add constant

docs/reference/design-patterns.md
```diff
@@ -0,0 +1,363 @@
...
+    private Singleton() {}
+
+    private static class LazyHolder {
+        private static final Singleton INSTANCE = new Singleton();
+    }
+
+    public static Singleton getInstance() {
+        return LazyHolder.INSTANCE;
+    }
+}
```
MarcinMoskala: some override?

docs/reference/design-patterns.md
voddan: @MarcinMoskala What do you mean? You referred to line 22, right?

docs/reference/design-patterns.md
```diff
@@ -0,0 +1,363 @@
...
+
+class HolidayStrategy implements PriceStrategy {
+    double convertPrice(double price) {
+        return price * 1.1;
+    }
+}
+
+// Usage:
+Item tvset = Item(100.0);
+double price = tvset.getTodaysPrice(HolidayStrategy);
```
jcomo: I believe this should be `new HolidayStrategy()`

cy6erGn0m: or `object HolidayStrategy`

docs/reference/design-patterns.md
MarcinMoskala: In kotlin there is const
In java "static final"
It should be added to static string to make it work quicker

docs/reference/design-patterns.md
voddan: @cy6erGn0m Java doesn't have `object` syntax, or does it?

docs/reference/design-patterns.md
voddan: the whole point of this field was to be mutable and shared between threads

dstd: but in [Kotlin-version](https://github.com/JetBrains/kotlin-web-site/pull/442/files#diff-17261ec8f99c1daa36b1928b4b9b11f3R44) this field is immutable

voddan: fair enough!

yole:
This text is good, but it's not really reference material, and we don't have an "articles" section or anything like that where such text could be published. Maybe you could publish it on Medium, and we would then post a link?

voddan:
@yole in my understanding it is not that different from "idioms".
I sow a bunch of SO question on how to implement known java patterns, so this section is a mix of FAQ and "idioms"

Can you explain/compare this to idioms? They are a part of the docs after all

BTW I tried to structure the text as a list of independent sections/elements, so that could be updated over time. Sure there are lots of interesting solutions that I don't know and that are pretty hard to google

voddan:
https://medium.com/@adinugroho/singleton-in-kotlin-502f80fd8a63#.4ph5q7o0u

docs/reference/design-patterns.md
```diff
@@ -0,0 +1,372 @@
...
+Car car = new Car(cb);
+```
+
+The [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) is used to create objects with `final` fields without providing a constructor with a massive list of arguments.
+Also it allows having default values for `final` fields that are optional to specify.
+All of this is easily achieved in Kotlin with the [primary constructor](classes.html#constructors) syntax.
+Note how using [named arguments](functions.html#named-arguments) makes the syntax well-readable even for longer parameter lists:
+
+``` kotlin
+class Car(val year: Int,
```
neworld: I am just curious, why such line breaking in the declaration. I mean, why class name and properties are in the same line. I prefer to break these lines like JetBrain already do ([random picked class](https://github.com/JetBrains/kotlin/blob/master/idea/idea-core/src/org/jetbrains/kotlin/idea/core/KotlinIndicesHelper.kt#L55-L62))

That code style has an advantage. You don't need to re-align all properties each time then you rename class:

```diff
-class User(val id: Int,
-           val name: String,
-           val surname: String) : Entity {
+data class User(val id: Int,
+                val name: String,
+                val surname: String) : Entity {
```

voddan: Hm, I didn't put much thinking into it. The style I use (here and in other projects) does not cause much trouble and IJ auto-formatter supports it. I don't think the docs use one style or another as for now. I believe this is an interesting and very important discussion, and it should be followed by creating of an extensive style guid in the docs.

For now I hope the team merges my patch ASAP, so that others could patch the text easily. I an going to raise the issue after Kotlin 1.0.3 is out

neworld: You believe or not, but our project has entities with +30 properties, and I think this is quite important.

But I agree with you. We can polish it afterward

PR #446, opened 2016-04-06
add trimMargin() example to basic-types.md

helmbold:
Explain trimMargin with an example. This is an important function to make multiline strings useful.


yole:
Thanks for your contribution!

PR #447, opened 2016-04-06
added property-over-function section

voddan:
a summary of the conventions used in stdlib


voddan:
Thank you @yole! 
Is now a good time to discuss merging #442 ?

PR #450, opened 2016-04-13
Idioms: using of `reified` for convenient overloads of generic functions

dstd:
Example for using of `reified`. It's quite frequently asked at https://kotlinlang.slack.com/


yole:
Thanks for your contribution!

PR #452, opened 2016-04-21
Update basic-types.md

andu1603:
In suggested example all lines will produce the same output. Escaped strings have the same behavior (looks like '$' character doesn't have to be escaped). I got this result in the online IDE (http://try.kotlinlang.org/).

### DIFF
--- a/docs/reference/basic-types.md
+++ b/docs/reference/basic-types.md
@@ -276,10 +276,14 @@ val str = "$s.length is ${s.length}" // evaluates to "abc.length is 3"
 ```
 
 Templates are supported both inside raw strings and inside escaped strings.
-If you need to represent a literal `$` character in a raw string (which doesn't support backslash escaping), you can use the following syntax:
+If you need to represent a literal `$` character in a raw string (which doesn't support backslash escaping), you can use one of the following syntax:
 
 ``` kotlin
+val s = 9.99
 val price = """
+$9.99
 ${'$'}9.99
+${'$'}$s
+$$s
 """
 ```


yole:
Sorry, I don't see the point of your change. The original example shows the syntax for escaping the $ character, and produces the expected result. Your suggested example is much more confusing.

andu1603:
Your example show one of the way to escape $ character, but it's pointless
for numbers, because for numbers we could use $ without escaping. Also you
don't mention that it's possible to escape it with doubling $, that looks
easier.

суббота, 23 апреля 2016 г. пользователь Dmitry Jemerov написал:

> Closed #452 <https://github.com/JetBrains/kotlin-web-site/pull/452>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/JetBrains/kotlin-web-site/pull/452#event-638528038>
>


-- 
С Уважением, Анастасия Дудукина.

yole:
The syntax shown in the original example is the one that works in all cases. It's true that other syntactic options work in some other cases, and I don't see much value in explaining all of the possibilities in detail here. And showing the examples without explaining the cases when they are applicable is also not too useful.

andu1603:
Thank you for your response.

суббота, 23 апреля 2016 г. пользователь Dmitry Jemerov написал:

> The syntax shown in the original example is the one that works in all
> cases. It's true that other syntactic options work in some other cases, and
> I don't see much value in explaining all of the possibilities in detail
> here. And showing the examples without explaining the cases when they are
> applicable is also not too useful.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/JetBrains/kotlin-web-site/pull/452#issuecomment-213722767>
>


-- 
С Уважением, Анастасия Дудукина.

PR #453, opened 2016-04-22
Fix wrong name in Collections/map example

MatthewJWalls:
Small typographic error in Map usage sample: 'map' should be 'readWriteMap'


yole:
Thanks for the fix!

PR #454, opened 2016-05-03
add example for filtering non-null values

helmbold:


yole:
Thanks for your contribution!

PR #457, opened 2016-05-08
Update extensions.md to illustrate that extension functions can overload

sreich:
add example showing that extension functions really can overload other functions, through differing arguments. including java functions. i didn't know they could, until i asked in slack.


yole:
Documented in b74f0862ebb51c34f98898e72cfe1b5cc77fb78a

PR #458, opened 2016-05-12
Add maven example without using process-sources

ansell:
Adds a maven example that doesn't require the process-sources hack by switching off the default-compile and default-testCompile targets that are treated specially by maven.

Also add examples of maven configurations that will generate the two types of Jar file, a minimal and a self-contained Jar file.

### DIFF
--- a/docs/reference/using-maven.md
+++ b/docs/reference/using-maven.md
@@ -84,48 +84,46 @@ In maven terms that means kotlin-maven-plugin should be run before maven-compile
 
 ``` xml
 <plugin>
-	<artifactId>kotlin-maven-plugin</artifactId>
-	<groupId>org.jetbrains.kotlin</groupId>
-	<version>${kotlin.version}</version>
-	<executions>
-		<execution>
-			<id>compile</id>
-			<goals>	<goal>compile</goal> </goals>
-		</execution>
-		<execution>
-			<id>test-compile</id>
-			<goals>
-				<goal>test-compile</goal>
-			</goals>
-		</execution>
-	</executions>
+    <artifactId>kotlin-maven-plugin</artifactId>
+    <groupId>org.jetbrains.kotlin</groupId>
+    <version>${kotlin.version}</version>
+    <executions>
+        <execution>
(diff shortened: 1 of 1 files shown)


ansell:
This pull request has a matching pull request to kotlin-examples to synchronize the pom.xml files there with the suggestions here https://github.com/JetBrains/kotlin-examples/pull/34

cy6erGn0m:
@ansell Why do you think that the executions are guaranteed to be executed in the declarations order?

yole:
Thanks for your contribution, and sorry for the delay with handling it!

ilya-g:
The new method of mixed kotlin-java compilation seems kinda cumbersome to me, compared to one with employing `process-sources` phase. What specifically is wrong with the old one?

ansell:
The process-sources method does actually complete the compilation, but it doesn't do it in the phases that Maven recognises in its default lifecycle. I am not sure if the kotlin plugin supports incremental compilation. However, it wouldn't have been possible to support this during process-sources phase in some corner cases, because there may be other source-processing going on (per the convention for the process-sources phase) that could interfere with the compilation.

The underlying issue that makes the new method complex is that Maven directly references the "default-compile" and "default-testCompile" execution ID's into its default lifecycle as the first target for the compile and testCompile phases, respectively. That works as expected in the majority of cases (all of the cases where Java is the only language being used) but doesn't work for Kotlin/Java shared modules where Kotlin has to compile its sources first. If Maven didn't do that, the new method may not seem as cumbersome, as it would contain about 10 less lines.

Hence, feel free to continue using the process-sources method, as it is unlikely that you will run into the corner cases. However, in general, the solution is to make sure that default-compile and default-testCompile are disabled and make sure that the kotlin plugin is above the maven-compiler-plugin in your pom.xml file.

You can put the goal/phase definitions into pluginManagement in a parent pom.xml file if you don't want to redeclare them each time, but you would still need to reference the kotlin plugin in each module pom.xml to ensure it gets called before the maven-compiler-plugin.

PR #461, opened 2016-05-20
add example with let

helmbold:


charlesdurham:
This exists under idioms, https://kotlinlang.org/docs/reference/idioms.html#execute-if-not-null

yole:
Thanks for your contribution!

charlesdurham:
"with" is misspelled

yole:
fixed

PR #462, opened 2016-05-25
Update gradle page

cypressious:
- Add note about mixes Java/Kotlin code under Android
- Add kotlin-test-junit
- Add section on kapt


yole:
That's really useful, thanks!

PR #463, opened 2016-05-27
Update to list of companies using Kotlin

jmarkovic:
Added Infinum to list of companies using Kotlin in production.


yole:
Thanks!

PR #468, opened 2016-06-02
added a link to the logos resourse

voddan:
"Where I can get a logo" is a very frequently asked question. This info should be somewhere on the site and searchable.


yole:
Thanks!

PR #470, opened 2016-06-05
Adds if/in range example for idioms.md.

IonoclastBrigham:


yole:
I find the use of `if` together with a sequence (range with a step different from 1) to be quite confusing, inefficient and definitely not idiomatic. I've added a few more natural-looking examples to this section instead.

In any case, thanks for your contribution!

PR #473, opened 2016-06-08
added about overriding default parameters

voddan:
see http://stackoverflow.com/questions/37700828/why-is-it-possible-to-omit-default-values-in-overridden-member-functions-of-sub/37701185?noredirect=1#comment62876675_37701185


yole:
Added a more detailed note. Thanks!

PR #474, opened 2016-06-12
clarify the use of method references with overloaded functions (https…

yole:
…://youtrack.jetbrains.com/issue/KT-12461)


udalov:
👍

PR #476, opened 2016-06-14
document how anonymous inner class instances are created in the nested classes section

jnizet:
It's hard to find that information in the doc because it's currently in the chapter about objects, which is not where a Java developer looks for it.
I also added a note on how to create instances of Java functional interfaces, and added a link to the nested classes chapter in the section discussing how to instantiate classes.


yole:
Thanks for your contribution!

PR #479, opened 2016-06-26
Overriding protected methods

karavayeu:
Adding missing information about modifier for overridden protected methods.


yole:
Documented with a slightly different wording. Thanks!

PR #481, opened 2016-07-03
Clarify backing field generation condition

maxblumental:
I made the point with backing fields in default accessors a bit more verbose. I am not sure if got it right, but if I am I think my variant is easier to understand.


yole:
Reworded slightly differently. Thanks for your contribution, anyway!

PR #483, opened 2016-07-05
Object expressions example code does not compile

LarsEckart:
I'm using Intellij 2016.1.2, kotlin plugin 1.0.3-release-IJ2016.1-103
The example code on this site did not compile. I got "Error:(11, 1) Kotlin: Right-hand side has anonymous type. Please specify type explicitly".

The type of "val ab" could not be deferred automatically, so a type for it had to be declared.


yole:
Thanks!

PR #484, opened 2016-07-06
added a context to xml and fixed broken link

oshai:
- pom.xml is more understanable when in context
- the link to examples zip was old so I replaced it with just a link to the repo assuming developers will know what to do there

### DIFF
--- a/docs/reference/using-maven.md
+++ b/docs/reference/using-maven.md
@@ -121,4 +121,4 @@ For OSGi support see the [Kotlin OSGi page](kotlin-osgi.html).
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


yole:
What's wrong with the link to the examples zip? This is an archive of the current state of the 'master' branch in the 'kotlin-examples' project, automatically generated by GitHub.

oshai:
This is my mistake, I looked at the wrong file after downloading (the parent pom), anyway I think it can be better to have another link to the repo.

oshai:
I reverted the link

yole:
Thanks for your contribution!

PR #485, opened 2016-07-06
Add idiom for constructors of generic classes in idioms.md

rocoty:
Added a new idiom for constructors of classes that require generic type information.

### DIFF
--- a/docs/reference/idioms.md
+++ b/docs/reference/idioms.md
@@ -271,6 +271,27 @@ stream.buffered().reader().use { reader ->
 inline fun <reified T: Any> Gson.fromJson(json): T = this.fromJson(json, T::class.java)
 ```
 
+### Convenient form for a generic "constructor" that requires the generic type information
+
+``` kotlin
+class Foo<T : Any>(val clazz: Class<T>) {
+
+    companion object Factory {
+
+        inline operator fun <reified T : Any> invoke(): Foo<T> {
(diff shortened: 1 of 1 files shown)


yole:
Sorry, but I don't think that overriding `operator invoke()` in a companion object is an idiom that should be endorsed in the documentation. At the very least, the same trick can be accomplished with a top-level function that has the same name as the class, which requires much less magic.

rocoty:
I see. That makes sense. Would it be possible to add your suggested variant to the idioms section then? I've found such generic factory functions to be very convenient.

rocoty:
Alternatively, could it somehow be made into a language feature to have reified type parameters in constructors? I take it the biggest issue you had with my pull request was the "hacky" use of operator overloading, and so the idea itself isn't necessarily a bad one.

yole:
Using a factory function with a reified type parameter and forwarding it to a class constructor is definitely a good idea. However, in my personal opinion, it's better to use factory functions that have descriptive names, not necessarily the same as the name of the class.

It might be possible to support inline constructors with reified type parameters, but that would be a major new language feature, best discussed elsewhere (for example, as a feature request in https://youtrack.jetbrains.com/issues/KT).

rocoty:
Alright. Will do.

rocoty:
Opened an issue on this here:
https://youtrack.jetbrains.com/issue/KT-13127

PR #488, opened 2016-07-08
added github-edit-link to docs-page

voddan:
This should add a link to **Edit** on github to pages like https://kotlinlang.org/docs/events.html>
I suppose it will go to https://github.com/JetBrains/kotlin-web-site/blob/master/_data/events.yml

I need this link every time I want to add a video to the page :(


yole:
No, unfortunately it won't; it will edit the template of the page, which is not very useful.

voddan:
Is there a way to do the magic?

yole:
I'm pretty sure there is, but it requires a bit more work.

PR #490, opened 2016-07-11
clarified the semantics for companion object initialisation

voddan:
the initial SO question: http://stackoverflow.com/questions/37262468/static-initialisation-block-in-kotlin

maybe it is worth it to describe the "static init" use case in more detail in the corresponding section and/or the idiom section


yole:
Thanks!

PR #493, opened 2016-07-23
added "Modern Web Development with Kotlin" book

voddan:
the references use two different styles - html and markdown. I used html, but feel free to clean it up

the description for the book is copied from leanpub


yole:
Thanks!

PR #495, opened 2016-07-27
corrected the use of the wrong superlative form of "easy"

Maviza101:
Corrected the use of the superlative form of the word "easy". The appropriate form is "easiest", but I found "most easiest" in the page.


yole:
Thanks!

Maviza101:
You're welcome!

PR #496, opened 2016-07-27
corrected some sentences; removed ref to obsolete plugin

Maviza101:
I corrected the grammar of a couple of sentences, improved the clarity of many others, and at the article's end, I removed a reference to the "Kotlin Extensions for Android" plugin because, as of this moment (27/07/2016), it is indicated as being obsolete in both Android Studio and IntelliJ IDEA.

### DIFF
--- a/_data/companies-using-kotlin.yml
+++ b/_data/companies-using-kotlin.yml
@@ -185,3 +185,8 @@
 - company: Hootsuite
   url: https://hootsuite.com/products/mobile-apps
   description: Hootsuite's mobile apps schedule, publish, and monitor conversations on Twitter, Facebook, Instagram, YouTube, and more. Using Kotlin in our Android app.
+
+- company: Lunar Labs
+  url: https://joinlunar.com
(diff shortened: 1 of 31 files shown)


yole:
The IDE plugin for the Android extensions is indeed obsolete, because its functionality is now part of the main Kotlin plugin. However, the functionality described in https://kotlinlang.org/docs/tutorials/android-plugin.html and the corresponding Gradle plugin are not by any means obsolete.

Maviza101:
Yeah. I now see that I shouldn't have removed the link to https://kotlinlang.org/docs/tutorials/android-plugin.html, I should only have changed its text from "Kotlin Android Extensions plugin" to "Kotlin plugin". So, what needs to be done then is to replace all references to the IDE plugin with refs to the main Kotlin plugin, while keeping the info on how to use the functionality. Let me see if I can create another pull request then...

Maviza101:
There! I have written something more appropriate.

p7nov:
Unfortunately, this is currently outdated.

PR #497, opened 2016-07-30
Remove member references

soc:
Analog to https://kotlinlang.org/docs/reference/reflection.html#function-references:

```
def isOdd(x: Int) = x % 2 != 0
// isOdd: (x: Int)Boolean

val numbers = List(1, 2, 3)
// numbers: List[Int] = List(1, 2, 3)

println(numbers.filter(isOdd))
// List(1, 3)
```


yole:
Thanks!

PR #498, opened 2016-07-30
Remove built-in XML

soc:
XML support has been removed from the core and was made optional, pending deprecation and replacement with string interpolation.


yole:
Thanks!

PR #499, opened 2016-08-01
Remove actors

soc:
In both languages actors are not built-into the language, they are regular libraries.


yole:

PR #500, opened 2016-08-01
Adapt collections

soc:
"supports Java 8 streams, which provide similar functionality" is not a difference between the languages. Both languages allow the use of Java 8 streams. I added a comment on immutable collections instead.

### DIFF
--- a/_data/companies-using-kotlin.yml
+++ b/_data/companies-using-kotlin.yml
@@ -185,3 +185,8 @@
 - company: Hootsuite
   url: https://hootsuite.com/products/mobile-apps
   description: Hootsuite's mobile apps schedule, publish, and monitor conversations on Twitter, Facebook, Instagram, YouTube, and more. Using Kotlin in our Android app.
+
+- company: Lunar Labs
+  url: https://joinlunar.com
+  description: All of Lunar OS's pre installed mobile apps are written in Kotlin.
+    
--- a/_data/companies-using-kotlin.yml
+++ b/_data/companies-using-kotlin.yml
@@ -181,4 +181,7 @@
 - company: Infinum
   url: https://infinum.co/
   description: Using Kotlin to build expressive, less verbose Android apps for startups and businesses.
-
+  
+- company: Hootsuite
+  url: https://hootsuite.com/products/mobile-apps
+  description: Hootsuite's mobile apps schedule, publish, and monitor conversations on Twitter, Facebook, Instagram, YouTube, and more. Using Kotlin in our Android app.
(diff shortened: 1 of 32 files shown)


ilya-g:
Maybe we should rephrase it like "the similar functionality can be achieved in Kotlin with the help of Java 8 parallel streams"?

Definitely, both languages support Java 8 streams, but does anyone use them in Scala?

ilya-g:
Regarding to collections, what do you think about this comparison #428?

soc:
Hi @ilya-g!

- I think it's not correct to make a huge distinction between Scala collections vs. Java collections when Kotlin is considering going into the same direction. I think it's more accurate to describe the situation as is without editorializing.

- The thing about list is irrelevant imho.

- Ditto regarding traversable/iterable.

ilya-g:
> I think it's more accurate to describe the situation as is

I agree. And the situation is that in Kotlin we use Java collection implementations for now. 
Scala has its own hierarchy of collections designed from scratch. It results in better design decisions but comes at a price of being incompatible with java collections. 
And I think this distinction is worth mentioning.

soc:
If Kotlin adds its own immutable collections, it is in exactly the same position.

ilya-g:
Even if Kotlin goes in the same direction, it doesn't mean it'll come to the same end point. 

We plan to add immutable collections and avoid disrupting java collections compatibility at the same time. You can refer to the details about the proposed immutable collections [here](https://github.com/Kotlin/kotlinx.collections.immutable/blob/master/proposal.md).

soc:
Well, all I'm saying is that the editorializing sounds a tad bit desperate (the heading is called "What Scala has that Kotlin does not").

ilya-g:
> the heading is called "What Scala has that Kotlin does not"

I'd agree that comparison of collection specifics is not suitable to be in that list and deserves its own subsection.

_data/companies-using-kotlin.yml
```diff
@@ -174,3 +174,22 @@
...
+  url: https://play.google.com/store/apps/developer?id=Maxwell+Forest+Pty+Ltd
+  description: Safedome (since 1.1.3) and other internal Android apps written in Kotlin.
+  
+- company: Infinum
+  url: https://infinum.co/
+  description: Using Kotlin to build expressive, less verbose Android apps for startups and businesses.
+  
+- company: Hootsuite
+  url: https://hootsuite.com/products/mobile-apps
+  description: Hootsuite's mobile apps schedule, publish, and monitor conversations on Twitter, Facebook, Instagram, YouTube, and more. Using Kotlin in our Android app.
```
Groostav: Hootsuite's using kotlin?

I'm going to have so much fun lambasting the scala guys at [our next meetup](https://www.meetup.com/vancouver-scala/)...

yole:

PR #501, opened 2016-08-01
Remove comment about Option ...

soc:
... as it could cause readers to think that Option was related to null handling.


yole:

PR #502, opened 2016-08-11
Switch wildcard section to use `E` rather than `T`

mkobit:
The `Collection<? extends E> items` uses `E`, and then all the documentation under it refers to `T`.

This change aligns the documentation section with the code section above it.

### DIFF
--- a/docs/reference/generics.md
+++ b/docs/reference/generics.md
@@ -76,9 +76,9 @@ interface Collection<E> ... {
 }
 ```
 
-The **wildcard type argument** `? extends T` indicates that this method accepts a collection of objects of *some subtype of* `T`, not `T` itself. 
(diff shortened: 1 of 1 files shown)


yole:
Different classes use different names for type parameters; that's just as fine as using different names for local variables that refer to the same thing. The type parameter of `Collection` is called `E` because it stands for "element". The type parameter of a generic class is usually called `T` for "type".

mkobit:
I agree with you in general that _type_ `T` make the most sense on this page. My thoughts when going thhrough this page was that most of this section refers to the `Collection<? extends E> items` shown above, so that using `E` might be clearer to refer back up that method signature. 

I always find myself rereading and referencing Java documentation about generics, so I thought it might be helpful.

I'll defer to your judgement though!

PR #503, opened 2016-08-14
Correction: you can assign an object to a variable

BlackHC:
See http://try.kotlinlang.org/#/UserProjects/74rm3tekp1tls37je8c11lchs9/e5p5cvptptv9b84f1u1d5jjd1k

```
object HelloObject {
    fun print() {
        println("Hello I'm Mr Meeseeks!")
    }
}

    val x = HelloObject
    x.print()
}
```
seems to work.

Maybe I misunderstood the statement. Also I'm not sure whether removing the whole sentence is the best idea. Please provide feedback :)

### DIFF
--- a/docs/reference/object-declarations.md
+++ b/docs/reference/object-declarations.md
@@ -89,8 +89,7 @@ object DataProviderManager {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


yole:
Rephrased to make it clear what the docs actually intend to say.

PR #506, opened 2016-08-26
Fix code highlighting

pilgr:
Code block to config plugin don't have the background on https://kotlinlang.org/docs/tutorials/android-plugin.html. But it has background on github preview. So I hope that specifying language can solve the issue.


yole:
Thanks!

PR #510, opened 2016-09-09
Typo in FAQ

Harmitage:
If it is some kind of a wordplay, I don't get it, but "Objected-Oriented" is an accepted term.


yole:
Thanks! No, it wasn't any kind of wordplay.

PR #520, opened 2016-10-09
Remove redundant visibility modifier from annotations.md

nonylene:
Remove `public` modifier from `annotations.md`.


yole:
Thanks!

PR #522, opened 2016-10-14
typo

sauerbraten:
head and body are members of the HTML class, not the html function, obviously... (see here: https://kotlinlang.org/docs/reference/type-safe-builders.html#declarations)


yole:
Thanks!

PR #528, opened 2016-10-20
Fix If/Loop formatting and indentation

cypressious:


yole:
Thanks for taking care of this! Really appreciated!

PR #534, opened 2016-10-29
Update Android Studio tutorial to mention plugin

nanodeath:
Android Studio doesn't come with the Kotlin plugin. It must be manually installed, which the docs will now call out.


yole:
Thanks for your contribution! I've added this note to the beginning of the tutorial, because it's more logical that way.

PR #553, opened 2016-11-14
changed the name of argument to the let block.

Maviza101:
I changed the name of the argument to the let block from "it" to "item". The previous code was actually correct, but this one demystifies the origin of the said variable.


yole:
The original code actually demonstrates the correct way to use a `let` function. The `it` implicit lambda parameter is extremely common in Kotlin code, so it should not need any demystifying. Moreover, `it` can be used when `let` is applied to a longer expression, not just a single local variable.

Maviza101:
I agree. My only concern is for Kotlin noobs (like myself) who might not have seen a LOT of Kotlin code. Besides, while using Kotlin for Android, most of the lambda parameters I've seen (especially for listeners), seem to be clearer than this e.g
`foo.setOnClickListener { view -> bar(view) }`

PR #554, opened 2016-11-14
Add compiler installation instructions for MacPorts users

breun:
Kotlin is also available from MacPorts for Mac OS X/macOS users.


yole:
Thanks!

PR #558, opened 2016-11-21
Update ranges.md

elect86:
Added `until` sample

comments made cleaner

### DIFF
--- a/pages/docs/reference/ranges.md
+++ b/pages/docs/reference/ranges.md
@@ -11,12 +11,12 @@ Range expressions are formed with `rangeTo` functions that have the operator for
 Range is defined for any comparable type, but for integral primitive types it has an optimized implementation. Here are some examples of using ranges
 
 ``` kotlin
-if (i in 1..10)  // equivalent of i in [1, 10], 10 is included
+if (i in 1..10) {  // equivalent of i in [1, 10], 10 is included
     println(i)
-    
-if (i in 1 until 10)  // equivalent of i in [1, 10), 10 is excluded
+}
+if (i in 1 until 10) { // equivalent of i in [1, 10), 10 is excluded
     println(i)
-
+}
 ```
 
 Integral type ranges (`IntRange`, `LongRange`, `CharRange`) have an extra feature: they can be iterated over.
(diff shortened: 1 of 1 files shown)


pages/docs/reference/ranges.md
```diff
@@ -11,9 +11,12 @@ Range expressions are formed with `rangeTo` functions that have the operator for
 Range is defined for any comparable type, but for integral primitive types it has an optimized implementation. Here are some examples of using ranges
 
 ``` kotlin
-if (i in 1..10) { // equivalent of 1 <= i && i <= 10
+if (i in 1..10)  // equivalent of i in [1, 10], 10 is included
     println(i)
-}
```
cypressious: Please don't remove the curly braces.

pages/docs/reference/ranges.md
```diff
@@ -11,9 +11,12 @@ Range expressions are formed with `rangeTo` functions that have the operator for
 Range is defined for any comparable type, but for integral primitive types it has an optimized implementation. Here are some examples of using ranges
 
 ``` kotlin
-if (i in 1..10) { // equivalent of 1 <= i && i <= 10
+if (i in 1..10)  // equivalent of i in [1, 10], 10 is included
```
cypressious: The word "equivalent" implies this is real Kotlin syntax which it isn't. Maybe plain english would be more appropriate.

pages/docs/reference/ranges.md
```diff
@@ -11,9 +11,12 @@ Range expressions are formed with `rangeTo` functions that have the operator for
...
     println(i)
-}
+    
+if (i in 1 until 10)  // equivalent of i in [1, 10), 10 is excluded
```
cypressious: Same here with the word "equivalent".

pages/docs/reference/ranges.md
cypressious: Previously, it was valid Kotlin syntax, too.

elect86: Ok, I removed `equivalent`

yole:
Added a paragraph mentioning this function in a more appropriate place.

PR #562, opened 2016-11-27
Add required parens for == translation.

swankjesse:
Otherwise this fails.
```
    fun eq(a: Any?, b: Any?): Boolean {
        return a?.equals(b) ?: b === null
    }

    @Test fun testEquals() {
        assertTrue(eq("a", "a"))
    }
```


yole:
Thanks!

PR #566, opened 2016-12-13
Minor grammar correction

RabieJradi:
Swapped "is" with "are" in sentence.


yole:
The original grammar is correct: http://ell.stackexchange.com/questions/34322/how-to-use-any-of

PR #568, opened 2016-12-14

HugoMillwood:
The Android development team at Expressen is in love with the language :heart:.


yole:
Thanks!

PR #578, opened 2016-12-23
Additional explaination of overriding properties

Groostav:
I feel that the properties section could use expansion. I understand its syntactic sugar to us but to many kotlin users it is likely to be an important feature.

My use-case was in trying to explain to a colleague why creating a kotlin interface with `fun getThingies(): Thingies` was a bad idea, and that he should instead put `val thingies: Thingies` on his interface. I wanted to refer him to the kotlin lang docs, but was a little disappointed with what is here right now, so I figured I would suggest these changes.

My main goal was to explain how properties can be overriden, and some of the specifics around syntax for it. I think this syntax is fairly intuitive, but was not intuitive enough for my afforementioned collegue. In this sense, a few extra words elaborating on the subject I think are warrented. 

My phrase `compatible type` is in reference to covariance on getters. I'm not sure this is a topic worth even mentioning here, but I thought the phrase `same type` would simply incorrect. 

please note: several links will probably need updating,
- this breaks all links to `overriding members`. I think this is worth it because trying to differentiate between methods and properties is likely a disservice. If backwards compatibility is hotly desired it would not be hard to create a 3-# section `overriding members` with two 4-# sections `overriding methods` and `overriding properties`
- https://kotlinlang.org/docs/reference/properties.html#overriding-properties should point to this new section.

### DIFF
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -195,7 +195,7 @@ to inherit from this class. By default, all classes in Kotlin are final, which
 corresponds to [Effective Java](http://www.oracle.com/technetwork/java/effectivejava-136174.html),
 Item 17: *Design and document for inheritance or else prohibit it*.
 
-### Overriding Members
+### Overriding Methods
 
 As we mentioned before, we stick to making things explicit in Kotlin. And unlike Java, Kotlin requires explicit
 annotations for overridable members (we call them *open*) and for overrides:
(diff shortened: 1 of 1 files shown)


Groostav:
hmm, [this search for `overriding-members`](https://github.com/JetBrains/kotlin-web-site/search?q=overriding-members) turns up nothing. Am I missing something?

yole:
Merged with some additional corrections. Thanks!

PR #582, opened 2016-12-26
Mapped typed: broken anchor link fix

Miha-x64:
broken 'platform-types' anchor link replaced with 'notation-for-platform-types'


yole:
The correct anchor is #null-safety-and-platform-types . I'll fix this.

PR #583, opened 2016-12-30
Fix syntactic error in properties.md

Lukle:


yole:
The original grammar is actually correct.

PR #587, opened 2017-01-05
fix heading levels of generics.md

hltj:
Fix issue as the followed snapshot that "Generic functions" & "Generic constraints" should be sub-headings of "Generic" in the TOC of the PDF. While currently they are in same level.

headinglevels-generic


yole:
Thanks!

PR #588, opened 2017-01-05
All book headers described in same style

Turansky:
Now anchors works for all books


Turansky:
I understand, that can be different reasons to declare headers in different style.
If some important reason exists, please write about it.

yole:
The reason is that the page should look nice, which it doesn't if you use the default Markdown formatting.

PR #595, opened 2017-01-13
Clarifies intro of method dispatching for extensions

reitzig:
In the current [docs](https://kotlinlang.org/docs/reference/extensions.html#extensions-are-resolved-statically), it says:

> By defining an extension, you do not insert new members into a class, but merely make new functions callable with the dot-notation on instances of this class.

As the next paragraph explains, the current version is misleading: we won't actually *call* the method defined at the instance type, but at the variable type. Proposal:

> [...] on variables of this type.


yole:
Thanks! The new wording is indeed more correct.

PR #598, opened 2017-01-15
Added missing `is`

nkiesel:


belovrv:
the original syntax is actually correct

PR #621, opened 2017-02-08
Update Chicago Kotlin Users Group URL

dajulia3:
We had to recreate the Chicago Kotlin Users Group, since, while I was out on vacation, the meetup group expired. Unfortunately, Meetup support has not been responsive, so I created a new Meetup and will be re-inviting all the old members. So, I had to update the url for the meetup group.


belovrv:
Thanks!

PR #623, opened 2017-02-09
Drop the table of Kotlin versions

abreslav:


abreslav:
In hindsight, I've hard-coded 1.0.6. as a sample version, but it would be better to have the latest release there. Do we have such a macro somewhere? // cc @belovrv

yole:
``{{ site.data.releases.latest.version }}`` should work as such a macro

PR #624, opened 2017-02-11
Fix Eclipse install button

moosichu:
The URL of the button image was out-of-date and pointed to a stale image. Therefore you couldn't even drag and drop the link to your Eclipse workspace!

Used the external button given on the eclipse website which works as of writing:
http://marketplace.eclipse.org/content/kotlin-plugin-eclipse#group-external-install-button


yole:
Thanks a lot for the fix!

PR #630, opened 2017-02-15
Update getting-started-with-gradle.md

h0tk3y:


yole:
Does this apply to 1.1 only?

h0tk3y:
@yole Yes, it does: the 1.0.6 Gradle plugin requires explicit `outputFile`, and the JS stdlib has only been renamed for 1.1.

yole:
Merged manually.

PR #634, opened 2017-02-20
Fix hash references in all articles

hltj:
I have written a tool to find all missed hash references in all markdown files and then I fixed them after manually verified.
Here are all the fixes.

I will publish the tool as a opensource project a few days later (and post a comment here too).


yole:
Thanks a lot! Please do let us know when you publish the tool - we'll then integrate it in our Web site build process to make sure that such problems are caught before deployment.

hltj:
OK. I will.

hltj:
[Here](https://github.com/hltj/markdown-hash-checker) is the tool, python scripts actually,  https://github.com/hltj/markdown-hash-checker .

It only support  check for markdown files under a single directory. I will implement recursive sub-directories support soon.

PR #639, opened 2017-02-26
Added a (missing) paragraph on "Default Imports"

voddan:
The original SO question: http://stackoverflow.com/a/40076096/3144601


yole:
Unfortunately this list doesn't match what the Kotlin compiler actually does, so I've had to correct it. But thanks for the contribution, anyway!

PR #649, opened 2017-03-01
More details on static methods

Miha-x64:


yole:
Merged with some additional edits. Thanks for your contribution!

PR #652, opened 2017-03-05
Fix hash references in recently added contents

hltj:
These invalid hash references were detected by the tool I mentioned in #634 and then fixed manually.

I have published the tool here https://github.com/hltj/markdown-hash-checker .


yole:
Thanks! I've filed an internal issue to integrate the links checker into the site build process.

PR #653, opened 2017-03-05
translation

springAppl:
translate part of the article

### DIFF
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -2,35 +2,27 @@
 type: doc
 layout: reference
 category: "Syntax"
-title: "Classes and Inheritance"
+title: "类和继承"
 related:
     - functions.md
     - nested-classes.md
     - interfaces.md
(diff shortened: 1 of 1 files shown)


hltj:
Maybe your should not submit pull request here.
And it had already been translated completely in https://github.com/hltj/kotlin-web-site-cn . see https://github.com/hltj/kotlin-web-site-cn/blob/master/pages/docs/reference/classes.md .
Here is the Chinese site of Kotlin Programming Language: https://kotlin-zhcn.github.io/ , most pages have been translated, and welcome to fork and PR [here](https://github.com/hltj/kotlin-web-site-cn).

yole:
First, accepting this pull request would replace part of the English documentation with the Chinese one, which is definitely not what we want.

Second, we aren't publishing translated documentation on our official Web site at this time, because we don't have a possibility to ensure that the information is up to date. Please see the comment above for an unofficial Chinese translation.

PR #658, opened 2017-03-07
KT-16141: Spread markers around the city location 

voddan:
I took the liberty to add a small modification to the code of the site so that problems described at KT-16141 would be (partially) resolved.

Please check the code carefully since it was my first time writing JS. I did my best to emulate the code style of the project.

### DIFF
--- a/static/js/page/events/map/Map.js
+++ b/static/js/page/events/map/Map.js
@@ -117,14 +117,56 @@ Map.prototype._createMarkers = function (events) {
   var map = this;
   var markers = [];
 
+  var cities = [];
+  var eventsByCities = {};
+
   events.forEach(function (event) {
+    let city = event.city;
+
     if (!event.city) {
-      return;
+        return;
     }
-    markers.push(new Marker(event, map));
+    
+    if(city in cities) {
+      eventsByCities[city] = [event] + eventsByCities[city];
+    } else {
+      cities = [city] + cities;
+      eventsByCities[city] = [event];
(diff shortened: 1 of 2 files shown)


voddan:
I was able to run the site when I replaced "0.0.0.0" with "localhost" in `kotlin-website.py`. That's where the second commit is coming from.

kisenka:
@voddan thanks! In some cases dots placed on the water :)

image

@belovrv If it's ok I'll accept this PR.

kisenka:
@voddan ping

voddan:
@kisenka Yes? I thought this PR was dead since it has never been merged for some reason.

kisenka:
My question was "In some cases dots placed on the water. If it's ok I'll accept this PR."

voddan:
@kisenka Oh, I didn't get that it was a question. 

Dots on water are ok since there is really no other solution - I mean, if I place dots only on land someone might think that those are all JetBrains offices 0_o.

Sorry I didn't react to your previous comment properly - I was sure you were in the process of merging it or rejected the PR for some internal reason. 

Frankly I would prefer a real solution with a drop-down list than this hack, but it is still better than no solution at all

kisenka:
@voddan no problem. I'm check your additions again and merge them.

belovrv:
I don't think its a good solution for the problem. @kisenka, let's make a dot with a number of events and clicking on it will show all of them in the left menu.

PR #661, opened 2017-03-10
Link to coroutines in Yield/Actors comparison with Scala

elizarov:


yole:
Looks like your pull request is the wrong way around.

elizarov:
Ops. Yes.

PR #665, opened 2017-03-13
Correct ranges.md

wojteko22:
Add necessary casting in functions and change -1.0 to -1L


yole:
Thanks for your contribution! I've made some additional changes to ensure that the docs match the standard library APIs more closely.

PR #666, opened 2017-03-13
Update idioms.md

nazerke:


yole:
I don't see what you intend to accomplish with this change. The current idioms page shows the correct way to use `if` with a range; your suggested change will not work.

PR #672, opened 2017-03-19
Addition of the word interfaces.

bitluke:
To make it read well it should be "Interfaces in Kotlin are very similar to Java 8 Interfaces" . leaving it as it is before makes it not clear what Java 8 it is similar to . Is it Classes ? Generic classes ?


yole:
I don't really think that this makes the text read better. It's clear from the context that we're talking about interfaces.

PR #674, opened 2017-03-22
Update getting-started-with-gradle.md

kengorab:
Minor issues, but on the website right now, the `compileKotlin2Js` block is not formatted well, and the $ is missing from the `outputFile` path. When I preview the change to the `compileKotlin2Js` block on github, it seems to render fine (before I made my change removing the space before the triple-backtick on line 85) but maybe whatever rendering engine you use doesn't detect it properly, because it looks like this:

screen shot 2017-03-21 at 8 37 06 pm

At any rate, some friends and I were trying this out at work today, and noticed it so I figured I'd fix it.


yole:
Thanks!

PR #677, opened 2017-03-23
Extended books in index.md

milos85vasic:
Added Fundamental Kotlin book in books section of index.md.


yole:
Added this information to the new Books page: http://kotlinlang.org/docs/books.html

PR #684, opened 2017-03-30
(for 1.1.2) Info about omitting dependencies versions

h0tk3y:
Added info about the default versions for `org.jetbrains.kotlin` dependencies.
Formatted indents uniformly.


yole:
Merged manually.

PR #688, opened 2017-04-11
Fix colon spacing on "Constructor References" example

rafaelzlisboa:
So it corresponds to the rest of this file and https://kotlinlang.org/docs/reference/coding-conventions.html#colon


yole:
Thanks!

PR #689, opened 2017-04-11
Small change to the wording of map example

StuStirling:
Very small change but I believe it makes more sense now.


yole:
Thanks!

PR #698, opened 2017-04-21
Add Sydney to user groups

mgouline:
Submitted the KUG form to @meilalina, this is to add Sydney KUG to the list of user groups with updated URL.


belovrv:
Thanks

PR #699, opened 2017-04-21
Fixes type declarations in generic swap example

jeffkole:


yole:
The example is correct. The parameters are indices into the list. The type of an index parameter is Int.

jeffkole:
Sorry about the distraction.

PR #703, opened 2017-04-30
Code typo fixed.

sunnykgupta:
Looks like someone typed `it` instead of `item`


yole:
This is no typo, and indeed the correct way to use "let". See http://kotlinlang.org/docs/reference/lambdas.html#it-implicit-name-of-a-single-parameter for more information.

PR #711, opened 2017-05-08
Clarify phrasing of intro to type projections

smichel17:
The live version's phrasing is a little weird and it was difficult for me to parse. Here's a few tweaks that make it easier for me to understand.


yole:
Thanks! The old wording was quite confusing indeed.

PR #713, opened 2017-05-10
Code Example: replace call to nonexistent Array.size() with Array.size

damianw:
`Array.size()` has not existed for some time and makes this example confusing.


yole:
Thanks, but we're about to drop this example entirely and replace it with a better one.

PR #715, opened 2017-05-12
Correct the type to T, not E

PaulWoitaschek:
The following text speaks about `T` the whole time, not `E`. So I changed the collection type to `T`

### DIFF
--- a/pages/docs/reference/generics.md
+++ b/pages/docs/reference/generics.md
@@ -71,8 +71,8 @@ That's why the actual signature of `addAll()` is the following:
 
 ``` java
 // Java
-interface Collection<E> ... {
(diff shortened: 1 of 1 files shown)


yole:
I don't think this is a helpful change. T is the name that is used for a type parameter with no specific meaning. E is "element type" of a collection. This is also the name used in the JDK: https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html

PaulWoitaschek:
It really makes no sense currently. There is no wildcard type argument ? extens T

wildcard

yole:
OK, this makes sense. I've changed T to E in the subsequent paragraph.

PR #716, opened 2017-05-12
doc rewording and make them more concise

king6cong:


yole:
You're removing the description of a very important mnenonic rule: "return" returns from the nearest function declared with the "fun" keyword. Removing the rule would make the description more confusing.

king6cong:
I think the removed sentence

> A *return*{: .keyword }  statement without a label
always returns from the function declared with the *fun*{: .keyword } keyword.

has similar meaning to the following

> whereas a *return*{: .keyword } inside
an anonymous function will return from the anonymous function itself

yole:
The rule is more general than that. A `return` in a regular function (declared with the `fun` keyword) returns from that function, a `return` in an anonymous function returns from the anonymous function, and a `return` in a lambda returns from the enclosing `fun` function.

PR #730, opened 2017-05-20
Correct example code

Khang-NT:


yole:
The example code is correct. The example shows that we're accessing a list of Strings (`strs`), expecting to get a `String` out of it, but actually it contains an object of a different type.

PR #740, opened 2017-05-24
Update coding-conventions.md

jeffknupp:
Fix grammar.


yole:
This document is being rewritten, so I won't be able to accept this as such, but I'll make sure that this part is correct in the new version of the document. Thanks anyway!

PR #743, opened 2017-05-26
Give an example of how to overload an operator

spartanhooah:


yole:
This isn't a good example. The + operator is not supposed to modify the object on its left hand side.

PR #747, opened 2017-05-27
Replaced java.lang.Char to java.lang.Character

WinterGuardian:
Replaced the incorrect java.lang.Char to java.lang.Character


yole:
Thanks!

PR #749, opened 2017-05-27
Provide an example of overloading an operator

spartanhooah:


yole:
You've submitted the same example in https://github.com/JetBrains/kotlin-web-site/pull/743/, and I commented on why it's not a good example. It's still not good. I'll add a different example myself.

spartanhooah:
How is it the same? My new example returns a new object rather than modifying the LHS. Oh well, I guess as long as an example gets added, I don't really care.

yole:
Oh, sorry, I somehow missed that. My bad! I'll add your example now too.

yole:
https://github.com/JetBrains/kotlin-web-site/commit/bce7abc414ce3717d77ca7179e4a6757160c5794

PR #751, opened 2017-05-30
Fix minor typo

perryprog:


yole:
There is no typo here. A length of a String is an Int, not a String.

perryprog:
@yole Oh, I feel stupid now! Sorry for wasting your time!

PR #752, opened 2017-05-30
Update command-line.md

Akshar-Patel:
"sdk" command would not work without executing " $ source "$HOME/.sdkman/bin/sdkman-init.sh" "


yole:
The instructions say "Next, open a new terminal window". Running `source` is not necessary if you do that.

PR #753, opened 2017-05-31
small typo in example

merlosy:
in example of class Person, "var" instead of "val"


merlosy:
oups sorry didnt read the next paragraph 
:/

PR #754, opened 2017-05-31
Missed update for 1.1, and remove repeated example

willhains:
The function `eval(expr)` was repeated in both example code blocks, and the second one still referred to the nested `Const` and `Sum`.


yole:
Thanks!

PR #757, opened 2017-05-31
#beginner -> #getting-started

Miha-x64:
#beginner seems to be unused, real beginner questions are in #getting-started.


hhariri:
Thanks.

PR #759, opened 2017-05-31
Added the negations to the `not to do` chapter

voddan:
This is a minor detail, but I don't think it is a good idea for any instruction to read `Perform threats of violence`, ever. I added the repeating negations so that even without the contexts the phrases could not be misunderstood.


hhariri:
Thanks. Good idea.

PR #765, opened 2017-06-06
Update link to a new version of Yested framework.

jean79:
Original Yested framework is no longer in the development. It has been replaced by a new version which is being actively developed.


yole:
Thanks!

PR #766, opened 2017-06-08
added return value to fun double(x: Int)

alexvas:


yole:
You can't return a value from a function like that; you need to use the `return` keyword.

PR #767, opened 2017-06-08
make return value of the fun double(...) explicit

alexvas:
Either go that way or add three dots to the body of the function see `fun powerOf(...)` below.


yole:
Thanks!

PR #769, opened 2017-06-09
show how to map nullable values if not null

silmeth:
An example how to do equivalent of Java’s 
``` java
Mapped mapped = Optional.of(data)
    .map(d -> transformData(d)))
    .orElse(defaultValue);
```
with Kotlin’s nullable reference for chaining.

I had hard time finding how to do this. I believe more examples of what can be achieved using the `let` method would be helpful.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -155,6 +155,14 @@ data?.let {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


yole:
Thanks! This is a useful example indeed.

PR #771, opened 2017-06-12
added license note

edgeone89:
added license note of the reference


yole:
This is not a good place to show this information. The license for the documentation is definitely not the first thing that a user expects to see when they start reading the documenation for a language.

edgeone89:
> The license for the documentation is definitely not the first thing that a user expects to see when they start reading the documenation for a language.

why?
https://www.postgresql.org/docs/9.6/static/index.html

PR #772, opened 2017-06-12
added license note

edgeone89:
added license note of the tutorials


yole:
Rejecting for the same reason as the other pull request.

PR #782, opened 2017-06-25
Added fix for kotlin tests in mixed builds.

DALDEI:
Current version of Kotlin plugin and Gradle may have problems with the kotlin compile classpath for kotlin test source.  Added directions on how to fix the test classpath.

### DIFF
--- a/pages/docs/reference/using-gradle.md
+++ b/pages/docs/reference/using-gradle.md
@@ -66,6 +66,43 @@ sourceSets {
 }
 ```
 
+### Mixed Java and Kotlin Test configuration
+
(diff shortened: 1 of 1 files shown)


h0tk3y:
@DALDEI, the behavior you described is more likely a bug. Could you please report it at [kotl.in/issue](kotl.in/issue)  with an example of how it can be reproduced?

h0tk3y:
This change is actually not necessary for mixed Java & Kotlin projects, the `test` Kotlin sources should be able to access classes from `main` without adding the `main` sources to the `test` source set. Applying the change, on contrary, will lead to `main` classes being compiled twice.

If you observe the behavior that makes the change necessary, please report it at [kotl.in/issue](https://kotl.in/issue), since it would be by no means the designed behavior.

I'm closing the PR since the configuration it suggests is not necessary when the plugin functions normally.

PR #787, opened 2017-07-01
Update generics.md

vseregin:


yole:
This example is actually wrong, and your change doesn't make it work right; we'll replace it with a different example altogether.

PR #788, opened 2017-07-03
Changes in wildcard type argument paragraph

chandilsachin:
"The **wildcard type argument** `? extends E` indicates that this method accepts a collection of objects of *some subtype of* `E`, not `E` itself" : this line is confusing. As far as i con understand at first read, it says `addAll` can not accept `Foo` list having it declared like `void addAllCollection<? extends Foo>`. Check this link http://www.angelikalanger.com/GenericsFAQ/FAQSections/TypeArguments.html#What is a wildcard?

### DIFF
--- a/pages/docs/reference/generics.md
+++ b/pages/docs/reference/generics.md
@@ -76,7 +76,7 @@ interface Collection<E> ... {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


yole:
Reworded this differently to be closer to the original intent. Thanks anyway!

PR #791, opened 2017-07-04
Added table of Kotlin visibility modifiers

edgeone89:
Added table of Kotlin visibility modifiers like in java tutorials

### DIFF
--- a/pages/docs/reference/visibility-modifiers.md
+++ b/pages/docs/reference/visibility-modifiers.md
@@ -12,6 +12,16 @@ Classes, objects, interfaces, constructors, functions, properties and their sett
 There are four visibility modifiers in Kotlin: `private`, `protected`, `internal` and `public`.
 The default visibility, used if there is no explicit modifier, is `public`.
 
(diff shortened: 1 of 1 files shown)


yole:
This table is somewhat misleading. Unlike Java, in Kotlin you can apply visibility modifiers to top-level declarations, and the table doesn't reflect that. Also, I'm not sure whether it really adds much value here.

edgeone89:
top-level declarations can be added in the table; and the table adds clarity while preserving compactness-conciseness of info

DipanshKhandelwal:
Even it doesn't even explains all of those attributes, like Kotlin also have a concept of open classes and variables.Check [this](https://kotlinlang.org/docs/reference/classes.html#inheritance) out.

edgeone89:
@DipanshKhandelwal 
could you, please, give your variant?

p7nov:
Closing this since it doesn't provide the complete picture of the visibility modifiers usage and isn't updated for a long time.

PR #797, opened 2017-07-13
Add String as Basic Type

borrom-dev:


yole:
I've reworded this differently, because String isn't really a built-in type with an optimized implementation; it's a regular class. Thanks for your contribution, anyway!

PR #804, opened 2017-07-20
Added output as a comment on "print" and "println" functions for easy…

diamondobama:
… comprehension


yole:
There's a Run button next to the samples, which shows the exact output of each line. I don't see how duplicating this information in comment would aid comprehension.

PR #805, opened 2017-07-26
Update null-safety.md

spitfire55:
Minor typo in safe call example combined with let.


spitfire55:
TIL the "it" keyword exists in Kotlin! Awesome! Turns out this is not a typo...

PR #806, opened 2017-07-27
Added caveat about null safety with Java interop

alexreidy:


h0tk3y:
There are runtime null checks generated for usages of platform types. On null value, these checks throw `java.lang.IllegalStateException` or `kotlin.TypeCastException`, but not NPE, so this should not be listed as a possible cause of NPE.

alexreidy:
This throws a `java.lang.NullPointerException`:
```kotlin
val q = ConcurrentLinkedQueue<String>()
q.poll().length
```
Unfortunately this is exactly how I learned about this caveat which is why I think it's worth pointing this out and referring people to the Java interop null safety section so they don't skim over `null-safety.html` and proceed without reading the interop docs thinking "great, I will never see NPEs in normal code" which is very incorrect if they're calling Java unaware that the little exclamation point means null pointers can fly. Also I'm not sure there's a huge difference between a NPE and an IllegalStateException - odds are they will both crash a program if a developer is unaware that they can be thrown by certain code, right? That's where my concern comes from: people skimming the docs and being absent minded when it comes to Java interop and nulls. It's easy especially for beginners to get used to the null safety of pure Kotlin and forget they're playing in a minefield when calling Java. Kotlin's interop is so seamless that if their code isn't NPE-prone they can be writing production Kotlin before learning the hard way to respect that little exclamation point.

PR #810, opened 2017-08-01
clarify expressions vs statements for checking

helmbold:
see https://discuss.kotlinlang.org/t/sealed-classes-and-when-expressions/3980


yole:
Thanks!

PR #813, opened 2017-08-06
Mention syntax for describing optional lambda

Flamenco:


yole:
This note should go in the "Function Types" section. It's not related to the syntax of lambda expressions.

PR #826, opened 2017-08-15
Add public modifier and final keyword for clarity

shaunkawano:
For the sample Java source codes written on `Instance Fields` section to indicate how generated source code from `@JvmField` annotation looks like, I have added `public` visibility modifier and `final` keyword. This is to explicitly indicate that

> The field will have the same visibility as the underlying property.

and also how visibility modifiers in Kotlin work.
https://kotlinlang.org/docs/reference/visibility-modifiers.html#visibility-modifiers


yole:
This example shows a Java class that accesses the field, not a Java class that is generated from the Kotlin example.

PR #827, opened 2017-08-17
Update idioms.md

SuperBeagleDog:
Add more comments:
Filtering a list(note that: list can not contains null elements, you can call list.filterNotNull() to filter null elements first)


yole:
First, the note is incorrect: you can use `filter` on a list containing null elements. Second, this type of note does not belong in a caption. Finally, `filterNotNull` is already [documented](http://kotlinlang.org/docs/reference/null-safety.html#collections-of-nullable-type).

SuperBeagleDog:
Thanks for your answer.

PR #832, opened 2017-08-21
Update ranges.md

qwertukg:


yole:
There's no advantage in using `forEach` to iterate over a range in this manner, so I don't think there's any value in showing this example.

PR #835, opened 2017-08-23
Follow convention of using it in lambdas

softarn:
The convention is to use `it` for lambdas that are short and not nested. The Kotlin docs aren't following this convention though and explicitly declaring them. I propose that we change them according to the convention.

This is only one example, but if you agree I'll update with the others I find when reading through the docs.


yole:
The convention is to use `it` for lambdas that are short and not nested _when it makes code easier to read_. In this particular example, I definitely find the original code to be easier to read, and I would be against a wholesale change that updates all of the examples.

softarn:
True maybe the original is easier to read. Thanks for fast response! 👍

PR #836, opened 2017-08-23
Explain restrictions for public API inline functions

h0tk3y:
Add section on public API inline function bodies and `@PublishedApi`.


yole:
See also https://youtrack.jetbrains.com/issue/KT-19881

PR #838, opened 2017-08-25
Add Kotlin @JvmOverloads for constructor example

jottinger:
#fixes KT-19917


h0tk3y:
I think that the existing text covers the case of non-default parameters preceding the default ones: the function example shows that. Also, I think this statement makes it quite clear that `@JvmOverloads` can be used for constructors:

> The annotation also works for constructors, static methods etc. It can't be used on abstract methods, including methods defined in interfaces.

One can expect that it works the same way (except for the no-arg constructor, which is mentioned).

jottinger:
h0tk3y, I know. But I was looking explicitly for it, and the lack of example actually served as a barrier; feel free to reject the PR, but it would have saved me some work by being more explicit.

h0tk3y:
I've added a constructor to the existing example. 
The commit: https://github.com/JetBrains/kotlin-web-site/commit/4297c43c655613575cc7161d8f77f7fa172cee11

PR #839, opened 2017-08-27
Moved Parameters to the beginning, Infix notation — later

Miha-x64:
Moved Parameters up (it is the most important about functions :), Infix notation — down, because is not as popular as return types, named & default arguments, etc


h0tk3y:
Right, thanks!

PR #844, opened 2017-09-03
Small fixes of using-maven.md

dtonhofer:
1) Added hint regarding the Maven Dependencies Diagram Tool
2) The XML under "Compiling Kotlin only source code" and "Compiling Kotlin and Java sources" has been updated to look exactly like the one produced by IntellijIDEA. If haven been able to cross-check the "maven-compiler-plugin" part under "Compiling Kotlin and Java sources" because my IDEA has a major problem loading archetypes.


ilya-g:
Is the explicit phase specification required for these goals?

p7nov:
Unfortunately, we haven't received any answer from the author. Thank you, we'll merge similar changes in another commit.

PR #845, opened 2017-09-04
Metioned property delegation operators in Operator overloading

Miha-x64:
Mentioned property delegation operator functions in Operator overloading.


h0tk3y:
Thanks!

PR #849, opened 2017-09-10
Added Conditional single-expression functions

DipanshKhandelwal:
Added Conditional single-expression functions which returns true or false according to the condition checked.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -238,6 +238,13 @@ fun transform(color: String): Int = when (color) {
 }
 ```
 
+### Conditional single-expression functions
+
+```kotlin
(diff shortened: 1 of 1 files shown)


yole:
I'm not sure why a single-expression function that returns true or false is substantially different from a single-expression function that returns a value of any other type, and why it's worth highlighting as a separate idiom.

DipanshKhandelwal:
I think its a bit different, because its not just returning a value but its computing before returning the value.

yole:
Yes, almost all functions do that, in all languages. There is not anything specific to Kotlin in this.

DipanshKhandelwal:
OK, thanks . I got it now , it's nothing to be written exclusively.

PR #850, opened 2017-09-10
Fixed Gradle terminology

zsmb13:
Updated terminology around which `build.gradle` file to apply the Extensions plugin in to match the usual parlance.


h0tk3y:
Thanks!

PR #852, opened 2017-09-10
Update idioms.md

DipanshKhandelwal:
I think it would be be better to write the variable in the braces, while using its value.
This would be easier to understand for everyone new to the language.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -46,7 +46,7 @@ val positives = list.filter { it > 0 }
 ### String Interpolation
 
 ``` kotlin
-println("Name $name")
+println("Name ${name}")
 ```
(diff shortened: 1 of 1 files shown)


yole:
Thanks for your contribution, but this really isn't idiomatic Kotlin code. The idiomatic way to write Kotlin is to avoid using redundant elements in the source code, and curly braces are redundant in this context.

DipanshKhandelwal:
OK, i got it but then at some places it is used with braces while at others it is not.
Even when it contains a single word variable.

yole:
The correct thing to do is to remove braces in those places where they are used with a single variable.

DipanshKhandelwal:
Yes, i got it. Thanks.

PR #853, opened 2017-09-10
Using similar convention for referencing variables

DipanshKhandelwal:
In some places the value is referenced as it as  $variable while at some places as  ${variable} .
Like in this snippet, in the first part of the code it is referenced as it is without any braces, while below, in the same code it is used with braces. I think it must follow a single convention.


yole:
Thanks!

PR #856, opened 2017-09-11
Update basic-syntax.md

DipanshKhandelwal:
Creating basic class and object.
In basic-syntax.html, none of the snippet described any way for creation of classes and its object.

### DIFF
--- a/pages/docs/reference/basic-syntax.md
+++ b/pages/docs/reference/basic-syntax.md
@@ -627,7 +627,7 @@ class Triangle (var sideA: Double, var sideB: Double, var sideC: Double): Shape(
 
     override fun getArea(): Double {
         var s = (sideA + sideB + sideC)/2
-        return sqrt(s*(s - sideA)*(s - sideB)*(s - sideC))
+        return Math.sqrt(s*(s - sideA)*(s - sideB)*(s - sideC))
     }
     override fun getPerimeter(): Double = (sideA + sideB + sideC)
 }
--- a/pages/docs/reference/basic-syntax.md
+++ b/pages/docs/reference/basic-syntax.md
@@ -610,24 +610,24 @@ See [Higher-order functions and Lambdas](lambdas.html).
 ``` kotlin
 //sampleStart
 fun main(args: Array<String>) {
-    val rectangle: Rectangle = Rectangle (5.0, 2.0) //no 'new' keyword required in Kotlin while initializing
+    val rectangle: Rectangle = Rectangle(5.0, 2.0) //no 'new' keyword required in Kotlin while initializing
     val triangle: Triangle = Triangle (3.0, 4.0, 5.0)
-    print( "Area of rectangle is - ${rectangle.getArea()} and it's perimeter is - ${rectangle.getPerimeter()}" )
-    print( "Area of triangle is - ${triangle.getArea()} and it's perimeter is - ${triangle.getPerimeter()}" )
+    println("Area of rectangle is - ${rectangle.getArea()} and it's perimeter is - ${rectangle.getPerimeter()}")
+    println("Area of triangle is - ${triangle.getArea()} and it's perimeter is - ${triangle.getPerimeter()}")
 }
 //sampleEnd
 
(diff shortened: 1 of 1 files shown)


yole:
Thanks for your contribution! There are a couple of things that need to be corrected before we can merge that. First, the "object" term has a specific meaning in Kotlin, which doesn't apply here. The correct term to use is "instance". Second, the code example needs to be valid code: the constructor arguments need to match the class body, and the methods need to have bodies defined.

DipanshKhandelwal:
OK, ill do it ASAP.

pages/docs/reference/basic-syntax.md
```diff
@@ -602,3 +602,45 @@ fun main(args: Array<String>) {
...
+    override fun isSquare(): Boolean = length == height
+    override fun getArea(): Double = height*length
+    override fun getPerimeter(): Double = 2*(height + length)
+}
+
+class Triangle (var sideA: Double, var sideB: Double, var sideC: Double): Shape(3) {
+
+    override fun getArea(): Double {
+        var s = (sideA + sideB + sideC)/2
+        return sqrt(s*(s - sideA)*(s - sideB)*(s - sideC))
```
h0tk3y: This call cannot be resolved when the code is run at try.kotlinlang.org. Perhaps you meant `Math.sqrt(...)`

DipanshKhandelwal: Yes, thanks.

DipanshKhandelwal:
@yole , I have changed and tested code. Please give it a look.

pages/docs/reference/basic-syntax.md
```diff
@@ -602,3 +602,45 @@ fun main(args: Array<String>) {
...
 See [Higher-order functions and Lambdas](lambdas.html).
+
+## Creating basic classes and its instances:
+
+<div class="sample" markdown="1">
+
+``` kotlin
+//sampleStart
+fun main(args: Array<String>) {
+    val rectangle: Rectangle = Rectangle (5.0, 2.0) //no 'new' keyword required in Kotlin while initializing
```
h0tk3y: It would be nice of you to re-format the code block, so that it follows the default code style that is used for Kotlin in IntelliJ IDEA and suggested by the [Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html). The readability would slightly benefit from it, and no one will stumble upon surprising code style.
You can do it automatically with IntelliJ IDEA.

Here are the points that might surprise the reader:
* The type and function names should not be separated from the opening parenthesis by a space (e.g. `Rectangle (5.0, 2.0)` → `Rectangle(5.0, 2.0)`), and the arguments are not (e.g. `print( "..." )` → `print("...")`).
* Also, please surround all binary operators including `*` and `/` with whitespaces (e.g. `s*(s - sideA)*(s - sideB)*(s - sideC)` → `s * (s - sideA) * (s - sideB) * (s - sideC)`).
* The Kotlin naming conventions also suggest using *UpperCamelCase* for type names (`interface rectangleProperties` → `interface RectangleProperties`).
* And the last, the colon `:` should be prepended by whitespace if and only if it means subtyping relationship (`class Triangle(...): Shape(3)` → `class Triangle(...) : Shape(3)`.

pages/docs/reference/basic-syntax.md
```diff
@@ -602,3 +602,45 @@ fun main(args: Array<String>) {
...
+    val triangle: Triangle = Triangle (3.0, 4.0, 5.0)
+    print( "Area of rectangle is - ${rectangle.getArea()} and it's perimeter is - ${rectangle.getPerimeter()}" )
```
h0tk3y: Note: the two strings will be printed on the same line. Maybe `println` is better here.

DipanshKhandelwal: I'll remove all syntax errors.
Thanks.

DipanshKhandelwal:
@h0tk3y, thanks for the review.
I hope its alright now.

PR #881, opened 2017-09-11
Update compatibility.md

DipanshKhandelwal:


pages/docs/reference/compatibility.md
```diff
@@ -125,8 +125,8 @@ available APIs to the runtime of version 1.0.
 ```
 
 This means that you're using the Kotlin compiler 1.1 against the standard or reflection library of version 1.0. This can be handled in different ways:
-* If you intend to use the APIs from the 1.1 Standard Library, or language features that depend on those APIs, you should upgrade the dependency to the version 1.1.
```
h0tk3y: (see [another comment](https://github.com/JetBrains/kotlin-web-site/pull/885/files#r139399030))

PR #885, opened 2017-09-11
Update javascript-dce.md

DipanshKhandelwal:


pages/docs/reference/javascript-dce.md
```diff
@@ -11,11 +11,11 @@ Since version 1.1.4, Kotlin/JS includes a dead code elimination (DCE) tool.
 This tool allows to strip out unused properties, functions and classes from the generated JS.
 There are several ways you get unused declarations:
 
-* Functions can be inlined and never get called directly (which happens always except for few situations).
```
h0tk3y: Several guides on formatting vertical lists in English I've found (like [*Handling Vertical Lists*](http://www.getitwriteonline.com/archive/101406VerticalLists.htm)) suggest using periods when the list items are (or contain) complete sentences:

> If the items in the list are complete sentences, put a period or a question mark at the end of each of them-and capitalize the first word-just as you would do with any sentence. Do not use "and" or "or" after the next-to-last item.

> If even one of the items in the list contains a complete sentence, put a period at the end of every item in the list-and capitalize the first word-just as you would do if they were all complete sentences.

I believe the original formatting did fit better. There are more cases in the other pull requests that I haven't accepted yet where periods were replaced with semicolons after complete sentences. Though formatting is rather a matter of style, I don't think we should make changes in it that are not recommended by style guides in written English.

pages/docs/reference/javascript-dce.md
```diff
@@ -75,10 +75,10 @@ can be found [here](https://github.com/JetBrains/kotlin-examples/tree/master/gra
 
 * As for 1.1.x versions, DCE tool is an *experimental* feature.
  This does not mean we are going to remove it, or that it's unusable for production.
-  This means that we can change names of configuration parameters, default settings, etc.
```
h0tk3y: (see [another comment](https://github.com/JetBrains/kotlin-web-site/pull/885/files#r139399030))

DipanshKhandelwal:
Thanks for the review.

h0tk3y:
Thank you for fixing it and for your contribution overall!

PR #895, opened 2017-09-13
Update classes.md

DipanshKhandelwal:


pages/docs/reference/classes.md
```diff
@@ -149,7 +149,7 @@ Creating instances of nested, inner and anonymous inner classes is described in
 
 Classes can contain
 
-* Constructors and initializer blocks
```
yole: Please format the link correctly.

DipanshKhandelwal:
The link is correct, the part after # is the internal link for the page.

yole:
The reference is correct, but this is not a link. See the [Markdown syntax documentation](https://daringfireball.net/projects/markdown/syntax#link) for the right way to format links.

DipanshKhandelwal:
Oh sorry, I got it now.

PR #896, opened 2017-09-14
(for 1.1.60) Add a section about JSR-305 support to java-interop.md

h0tk3y:
[KEEP-79](https://github.com/Kotlin/KEEP/blob/41091f1cc7045142181d8c89645059f4a15cc91a/proposals/jsr-305-custom-nullability-qualifiers.md)
[KT-20225](https://youtrack.jetbrains.com/issue/KT-20225)


sdeleuze:
@h0tk3y Does compiler flag and possible values changed since Kotlin 1.1.4? It seems we are currently using `-Xjsr305-annotations=enable` while `-Xjsr305=strict` seems to be documented in that PR.

pages/docs/reference/java-interop.md
```diff
@@ -131,6 +131,135 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
+
+```java
+// FILE: test/package-info.java
+@NonNullApi // declaring all types in package 'test' as non-nullable by default
+package test;
+```
+
+##### `@UnderMigration` annotation (since Kotlin 1.1.60)
+
+The `@UnderMigration` annotation (from `kotlin.annotations.jvm`) can be used to define a migration status
```
dzharkov: Note, that these annotations belong to a separate library artifact called `kotlin-annotations-jvm` (in the end it should available in maven/gradle along with other artifacts)

dzharkov: Also I'd explain/give an example on how it can be used by a library maintainer

pages/docs/reference/java-interop.md
```diff
@@ -131,6 +131,135 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
+}
+
+@Nonnull(when = When.MAYBE)
+@TypeQualifierDefault({ElementType.METHOD, ElementType.PARAMETER})
+public @interface NullableApi {
+}
+
+@NullableApi
+interface A {
+    String foo(String x); // fun foo(x: String?): String?
```
dzharkov: Probably it's worth adding an example with type arguments (that should rename platform) in case of using `NullableApi`

sdeleuze: And with generic type arguments, varargs and array elements when they will be supported since it is not obvious to know what types should be annotated.

h0tk3y:
`TODO` When the support for arrays and varargs becomes clear, document it as well.

PR #897, opened 2017-09-15
Attempt to show that data-classes are not just DTOs

Miha-x64:
Many people think data-classes are just DTOs, stupid bags with data.
Attempting to say that `data` modifier brings some functionality.

I'd be glad to see some suggestions how to make this more clear.


h0tk3y:
Well, almost every class holds data. I'd rather say it like
> We frequently create classes whose main purpose is to hold data.

PR #904, opened 2017-09-25
Update inline-functions.md

RamV13:
- Updating sentence structure for explanation of tradeoff between code size and runtime
- Adding annotation for suppressing the inline function warning for nothing to inline


yole:
Thanks!

PR #905, opened 2017-09-25
Update coroutines.md

jsaund:
Description of real laziness of a sequence was a bit confusing -- simplifying terms to make it clearer


yole:
Thanks!

PR #906, opened 2017-09-26
Add JSR-305 support docs for 1.1.50

h0tk3y:


h0tk3y:
(this is the [1.1.60 reference](https://github.com/JetBrains/kotlin-web-site/pull/896) stripped down to what's working in 1.1.50)

PR #911, opened 2017-10-08
Replace variable named 'data' with 'value'

rjoncontract:
Though Kotlin allows usage of keywords like `data` and `open` for variable names. The syntax highlighter on the rendered HTML page highlights them as keywords. This could be confusing to developers who are just being introduced to the `data` keyword.


yole:
Thanks!

PR #912, opened 2017-10-09
Update oss-projects.yml

laviua:


yole:
We're no longer updating the OSS projects page. Please submit the project to https://kotlin.link/ instead.

PR #913, opened 2017-10-09
Added 'Restrictions' in 'Experimental status'

Miha-x64:
This note may help people not to do stupid things like I do, including these:
```
val func = suspend fun(call: ApplicationCall, query: String) {
    call.respondText(...)
}

class SomeRouteHandler : suspend (PipelineContext<Unit>, Unit) -> Unit {
    override suspend fun invoke(...) { ... }
}
```


h0tk3y:
Thanks! I think these restrictions should rather be mentioned in the main coroutines page, because they are not strictly bounded to the experimental state (may be lifted earlier or later). 

Here's the commit: https://github.com/JetBrains/kotlin-web-site/commit/c081f0c9388be205d1e0a824bc6dff44deb5b867

PR #914, opened 2017-10-15
Modified sample in "Providing a delegate"

AnsgarLin:
Make the first example to be easier to modify into a compilable code after copy-paste


h0tk3y:
Thanks!

PR #915, opened 2017-10-16
Fix property references example in the reflection docs

ltongues:
Description of the first code example implies that `x` is immutable, so I changed its declaration to use `val` and added a separate example for a mutable variable.


h0tk3y:
Thanks!

PR #916, opened 2017-10-16
Remove trailing whitespace

rillig:
Replaced "[ \t]+$" with "" in all files.


h0tk3y:
Sorry, we've got a big diff since the PR which now conflicts with the changes.
In my opinion, removing the whitespaces in the Markdown files does not make much sense: the Markdown processor that generates the actual site pages removes all redundant whitespaces anyway, including the trailing ones.

PR #917, opened 2017-10-19
Mentioned top-level variables

Miha-x64:
I was unable to find explicit mention of top-level variables.
This is the simplest fix.


h0tk3y:
Thanks!

PR #924, opened 2017-10-30
Public Review: Procedures around Kotlin Language Evolution

abreslav:
This PR contains the initial version of the documents describing how Kotlin's evolution is governed. These documents describe the general guidelines on how decisions about language and core library changes are made.

To understand the context, please read the [langauge-committee.md](https://github.com/JetBrains/kotlin-web-site/blob/5688f9cd1ee14283423417b0b6f3a7bb1b5d5582/pages/docs/evolution/language-committee.md) file first.

These documents have been prepared by the Kotlin Language Committee, and are now offered here to gather feedback from the community. Your comments are very welcome

### DIFF
--- a/pages/docs/evolution/breaking-changes.md
+++ b/pages/docs/evolution/breaking-changes.md
@@ -207,7 +207,7 @@ we defer to the existing
 ## `7` The scope of this policy
 
 
-## `7.1` In scope
+### `7.1` In scope
 
 `7.1.1` Language: syntax, static checks, execution
     semantics of language constructs.
@@ -232,7 +232,7 @@ we defer to the existing
 `7.1.7` KDoc syntax.
 
 
-## `7.2` Out of scope
+### `7.2` Out of scope
 
 `7.2.1` Build tools and plugins for them (e.g. Gradle
     support).
@@ -264,16 +264,16 @@ we defer to the existing
 
 ## `8` Different requirements for different platforms
 
-## `8.1` Kotlin/JVM
+### `8.1` Kotlin/JVM
 
 No additional requirements on deprecation/breaking changes.
 
-## `8.2` Kotlin/JS
+### `8.2` Kotlin/JS
 
 `8.2.1` Binary compatibility is not guaranteed for the
 JS world, because everything is usually built from sources.
 
-## `8.3` Kotlin/Native
+### `8.3` Kotlin/Native
 
 `8.3.1` Until Kotlin/Native reaches an official
     release, no compatibility guarantees are given for Kotlin/Native.
@@ -289,7 +289,7 @@ JS world, because everything is usually built from sources.
 
 Notation: NC — new compiler, OC — old compiler.
 
-## `9.1` Binary Change, issues
+### `9.1` Binary Change, issues
 
 `9.1.1` A binary produced from unchanged code with NC
     fails where the one produced by OC worked.
@@ -297,7 +297,7 @@ Notation: NC — new compiler, OC — old compiler.
 *Examples of failures*: linkage error, runtime exceptions not
         present in previous versions, deadlocks, race conditions
 
-## `9.2` Binary Change, non-issues
+### `9.2` Binary Change, non-issues
 
 `9.2.1` A binary compiled with NC may be rejected by
     OC, when it relies on new language features unsupported in OC.
@@ -315,7 +315,7 @@ Notation: NC — new compiler, OC — old compiler.
 `9.2.5` Adding supertypes to existing library
 classes/interfaces.
 
-## `9.3` Source Change, issues
+### `9.3` Source Change, issues
 
 `9.3.1` Sources that compiled with OC don't compile
     with NC against the same classpath.
@@ -324,7 +324,7 @@ classes/interfaces.
     behavior changes in a way significant for contract-abiding use
     cases.
 
-## `9.4` Source Change, non-issues
+### `9.4` Source Change, non-issues
 
 `9.4.1` Code compilable with NC fails to compile with
     OC (e.g. due to new language features supported in NC).
@@ -333,12 +333,12 @@ classes/interfaces.
     build configuration or compiler settings explicitly (i.e. in
     addition to advancing the compiler version).
 
-## `9.5` Library Change, issues
+### `9.5` Library Change, issues
 
 `9.5.1` Making a contract on existing API more strict
     than it used to be in a previous version.
 
-## `9.5` Library Change, non-issues
+### `9.5` Library Change, non-issues
 
 `9.5.2` Relaxing a contract on existing APIs.
 
@@ -354,7 +354,7 @@ classes/interfaces.
     stdlib at runtime.
 
 
-## `9.6` Performance changes
+### `9.6` Performance changes
 
 We recognize that runtime performance and bytecode size are important
 metrics, and will make reasonable effort to keep them in a good shape,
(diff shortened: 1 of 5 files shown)


jvgogh:
Looks good to me!

pages/docs/evolution/breaking-changes.md
```diff
@@ -0,0 +1,483 @@
...
+* `4.2.1` Provide and describe a replacement for
+    deprecated constructs as part of the warning.
+
+* `4.2.2` Allow to disable the warnings in settings,
+    or promote them to errors.
+
+* `4.2.3` Along with this version, provide automated
+    migration aids (e.g. in the IDE).
+
+`4.3` Allow a long enough period of time for
```
voddan: IMHO it is important to specify in this document the exact minimum migration period.

Many teams wait a certain time before updating to a new version of a software to allow for the vendor to release "hot bug fix" updates. I assume those teams use the same approach to the Kotlin compiler, IDE and IDE plugin. 

The problem arises when a team's migration period is significantly bigger than Kotlin's migrating period. In that case it may happen that a team skips several Kotlin compiler version, thus not following the intended migration path.

By specifying the minimum migration period for Kotlin components, for example 45 days, the Kotlin team guaranties that anybody can update their toolchain once per 45 days and not miss a version.

To clarify, this would still allow the length of the migration period to vary "based on the impact and complexity of the change".

pages/docs/evolution/breaking-changes.md
```diff
@@ -0,0 +1,483 @@
...
+    discontinued.
+
+* `4.5.2` If possible, keep a compatibility mode that
+    does not support new features, but allows the deprecated one as in
+    version N.M+1 or N+1.X.
+
+* `4.5.3` Migration aids can be removed from the
+    compiler at this point.
+
+`4.6` A version that supports automated migration
```
voddan: It is unclear which tools are covered by this paragraph. Since it mentions "automated migration", I assume that it includes not only the compiler, but also the IDE and the IDE plugin. In this case, it is unclear to me how it will work. Will JetBrains support installing older versions of the Kotlin plugin? Will the release notes contain "upgrade with x - y plugin version" notices? 

As a solution I can propose extracting the language migration functionality into separate plugins. Here is an example of how it would work: suppose I want to migrate my code base from Kotlin 1.6 to Kotlin 1.8; then I install the latest version of the Kotlin plugin, plus two smaller plugins: "Kotlin Migrator 1.6 -> 1.7" and "Kotlin Migrator 1.7 -> 1.8", where the additional plugins provide inspections and quick fixes needed to migrate from 1.6 to 1.7, and from 1.7 to 1.8 respectively. This configuration allows for different update cycles for the main plugin and for the migration assisting plugins. The downside is that it would probably require a somewhat stable internal compiler API to allow the plugins to communicate.

pages/docs/evolution/breaking-changes.md
abreslav: Language versions don't come out every 45 days, it takes closer to a year to release a new version. These documents are intended to set gudielines, not provide guaratees, so there's no need to write out exact numbers.

pages/docs/evolution/breaking-changes.md
abreslav: That's a good option

pages/docs/evolution/breaking-changes.md
voddan: My comment related to minor language versions rather than to major versions. The minor versions come more often than once a year, but I agree with you that their cooking period (~ 6 months) is probably more than enough.

pages/docs/evolution/breaking-changes.md
```diff
@@ -0,0 +1,483 @@
...
+
+* `4.4.1` In the release notes, repeat the information
+    about the deprecation plan.
+
+* `4.4.2` Allow to demote the errors to warnings through
+    settings.
+
+* `4.4.3` Keep migration aids available.
+
+`4.5` In the version N.M+2 or N+1, report errors for
```
udalov: ... or N+2?

pages/docs/evolution/breaking-changes.md
pdvrieze: From my perspective (university, teaching), it is important that I can continue using the same language version throughout the entire academic year while still having bug fixes for "significant" bugs (students are steered well away from edge cases and optimization is not an issue either). Changes in language versions throughout a course just increase opportunities for confusion as well as create problems with lab setup (which is often done once per year)

abreslav: Release versions are kept updated with important big fixes, and you don't have to migrate to a next release even if it's come out already, so I don't think there should be any problems in your case

artem-zinnatullin:
Can you please explicitly cover how new major version of Kotlin will work with previous one used in the same classpath/binary (and as a build system dependency)?

Will users have to update everything to new major version (project itself and all dependencies that use Kotlin) or new version will have own package ie `kotlin2` and users will be able to use both at runtime? See http://jakewharton.com/java-interoperability-policy-for-major-version-updates/

As the community grows, we have more and more libraries in Kotlin and migration from one major version to another could be very painful (as it is in Swift for instance).

Thanks!

abreslav:
I think we are not ready to set a policy for such things yet, but we realize that binary compatibility issues are important and have to be accounted for. Why not now: nobody knows what happens with Jigsaw adoption and development, maybe the world of the classpath goes away by itself :)

pages/docs/evolution/change-review-process.md
```diff
@@ -47,3 +37,42 @@ Reviews that need a decision before the next committee meeting (e.g.
 because of a release deadline) can be done by the committee in
 email/chat. These type of reviews need to be requested by the Lead
 Kotlin Language Designer (currently Andrey Breslav).
+
+# Issue Template
+
+``` java
```
LouisCAD: Why ```java?

abreslav: Because the website doesn't support markdown as a language yet

pages/docs/evolution/breaking-changes.md
ilya-g: or N+1.x?

pages/docs/evolution/change-review-process.md
```diff
@@ -0,0 +1,78 @@
...
+# Proposed behavior/pattern
+> 1. How it fixes the issue?
+> 2. Describe automated migration, if possible
+
+# Impact of this change
+> If possible: 
+> * number of impacted users
+> * how common this pattern is
+> * if we believe it's a rare case, why
+> * is automated migration possible
```
ilya-g: "is automated migration possible" — is it the same as "Describe automated migration, if possible" in Proposed behavior?

dnpetrov:
I'd suggest rewording
> 3.2.3 Bad code compiled due to a bug, and worked reasonably, must follow the deprecation policy
as
> 3.2.3 Bad code compiled and worked reasonably, must follow the deprecation policy

The actual reason why the code is considered "bad" might be more complex than just a "bug in the compiler". The point here is that some code that worked _reasonably_ (due to some coincidence) would behave differently after the change.

dnpetrov:
Case study: https://youtrack.jetbrains.com/issue/KT-20830
Technically, there's no "bad" Kotlin code: here a Java method annotated as `NotNull` returns null.
However, in the end the result of this method is not used. No exception is thrown.

1.) Is this code "good" or "bad"?

2.) Does this code "work reasonably"?

We've discussed it face-to-face with @erokhins , and are kinda stuck.

erokhins:
It is good to mention in `change-review-process.md` that for bugs that was introduced in the latest version language committee discussion can be established via e-mail.
It was shortly discussed at Mar 23rd, 2018: [doc](https://docs.google.com/document/d/1ReH84Cw_ZhGOUM_MdMQbLjzB0edXIeaFuBUF5molsuI/edit#heading=h.xmrj33x4ns7i)

abreslav:
This was merged back in 2018, just not closed properly. Thanks everyone for your comments!

PR #928, opened 2017-11-03
Minor typo: Should be 2 spaces!

jaredsburrows:
I figured this was a minor typo in the documentation.


yole:
We should totally compromise on 3.

PR #931, opened 2017-11-10
Update using-ant.md

vadimgurov:


udalov:
This is not entirely correct. Kotlin tasks should work fine with Ant 1.8.2 or later. They also should work with earlier versions of Ant, printing a warning that 1.8.2+ is recommended. Have you met a specific problem on Ant < 1.9?

vadimgurov:
Had problems with ant 1.7 that was fixed after upgrade to 1.9. Did not test 1.8. There is no right ant version mentioned in doc.

yole:
Thanks!

PR #935, opened 2017-11-17
Remove Kotlin formatting from print statements

kjperry:
Both of these "this prints:" examples had Kotlin keywords in them, so they were displayed in multicolor.


h0tk3y:
Thanks!

PR #937, opened 2017-11-20
Fix tailrec example

ArtsiomCh:
Tailrec part of `if (x == Math.cos(x)) x` is equivalent of `if (x == y) return x` not `if (x == y) return y`, if I get it right...


h0tk3y:
Thanks!

PR #939, opened 2017-11-21
remove contradict suggestion for factory name

oshai:
suggestion to name factory method as the class name contradicts suggestion in another part of the document:
"Exception: factory functions used to create instances of classes can have the same name as the class being created"
contradicts:
"If you declare a factory function for a class, avoid giving it the same name as the class itself. Prefer using a distinct name making it clear why the behaviour of the factory function is special. Example:"

which is here: https://github.com/JetBrains/kotlin-web-site/blob/yole/styleguide/pages/docs/reference/coding-conventions.md#factory-functions

### DIFF
--- a/pages/docs/reference/coding-conventions.md
+++ b/pages/docs/reference/coding-conventions.md
@@ -65,16 +65,6 @@ fun processDeclarations() { ... }
 var declarationCount = ...
 ``` 
 
-Exception: factory functions used to create instances of classes can have the same name as the class being created:
-
-``` kotlin
-abstract class Foo { ... }
-
-class FooImpl : Foo { ... }
-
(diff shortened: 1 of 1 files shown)


yole:
This pattern is used by the standard library, and we want to allow using it in user code too. The rule for factory functions says "avoid"; it does not forbid doing that.

oshai:
I understand, I just think that the guideline should be more clear and say in what cases to use one and in what to use other.in the current phrasing it is not clear if it is a good practice or not and in what cases exactly.

Of course one can ignore the guidelines anyway, I just think it is better to be more strict here or not say anything at all.

yole:
I've added some additional clarification text.

PR #940, opened 2017-11-22
Update java-interop.md to clarify JSR-305 support

wild-lynx:


h0tk3y:
Thanks! I've placed it with a more noticeable note block and reworded it a little.

wild-lynx:
Thank you!

PR #945, opened 2017-11-26
Syntax Highlighting

parthdeveloper007:
you guyz forgot to add a syntax heighlight there in `build.gradle`


yole:
`gradle` is not a supported language of Pygments; you need to use `groovy`

yole:
The current version of the tutorial already has correct syntax highlighting.

PR #946, opened 2017-11-26
Shortcuts

parthdeveloper007:
There is only need of special key combination to make easy for a programmer to simply convert his/ her file from java to kotlin by `Ctrl+Shift+Alt+K` andits all done. That's it....


yole:
Shortcuts are platform-dependent. The instruction provided works in all cases.

PR #957, opened 2017-11-29
Fix typo in exceptions

StefMa:
I'm not a nativ english speaker.
But for me `And this is *not* good` sounds way better than `And this is *no* good`...

But yeah. I'm not a nativ english speaker.. So maybe it will not be merged. I'm fine with it 😄

### DIFF
--- a/pages/docs/reference/exceptions.md
+++ b/pages/docs/reference/exceptions.md
@@ -70,7 +70,7 @@ catch (IOException e) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


h0tk3y:
Several guides on written English style suggest using *not good* to disqualify something from being *good*, for example: http://www.dictionary.com/e/not-good-vs-no-good/

I think we can leave the sentence as it is.

PR #960, opened 2017-11-29
Fixing the name of organizer user group for one event.

bajicdusko:
Organizer UG name fix for the event in Banja Luka form 5th October.


belovrv:
Thanks!

PR #965, opened 2017-12-04
Add two sections about type checks with generics and unchecked casts

h0tk3y:


pages/docs/reference/typecasts.md
```diff
@@ -102,3 +102,71 @@ val x: String? = y as? String
...
+        // `list` is smart-cast to `ArrayList<String>`
+    }
+}
+```
+
+The same syntax with omitted type arguments can used for casts that do not take type arguments into account: `list as ArrayList`. 
+
+Inline functions with [reified type parameters](inline-functions.html#reified-type-parameters) have their actual type arguments
+ inlined at each call site, which enables `arg is T` checks for the type parameters, but if `arg` is an instance of a 
+generic type itself, *its* type arguments are still erased.
```
yole: Show an example for this case?

pages/docs/reference/typecasts.md
```diff
@@ -102,3 +102,71 @@ val x: String? = y as? String
...
+
+val intsFile = File("ints.dictionary")
+
+// We saved a map with `Int`s into that file, so we want 
+// this map to have `Int` for value type:
+val intsDictionary: Map<String, Int> = readDictionary(intsFile) as Map<String, Int>
+// Warning: Unchecked cast: `Map<String, *>` to `Map<String, Int>`
+```
+
+The compiler produces a warning for the cast in the last line since it cannot be fully checked at runtime, it provides
```
yole: Rephrase this sentence? Two "it"'s referring to different things is confusing.

pages/docs/reference/typecasts.md
```diff
@@ -102,3 +102,71 @@ val x: String? = y as? String
...
+// this map to have `Int` for value type:
+val intsDictionary: Map<String, Int> = readDictionary(intsFile) as Map<String, Int>
+// Warning: Unchecked cast: `Map<String, *>` to `Map<String, Int>`
+```
+
+The compiler produces a warning for the cast in the last line since it cannot be fully checked at runtime, it provides 
+no guarantees that the map holds `Int` actual values.  
+
+In general, you should avoid unchecked casts by doing proper design: in the example above, there could be interfaces
+ `DictionaryReader<T>` and `DictionaryWriter<T>` with type-safe implementations.
```
yole: I don't like the "doing proper design" wording. In cases like this, you may design the API so that clients are not required to do any unsafe casts, but the implementation will still use unsafe casts, and it's unavoidable.

PR #973, opened 2017-12-25
Fix sample code

AnsgarLin:
'Any?' is not compilable with an error message "Type argument is not within its bounds"


h0tk3y:
Thanks!

PR #974, opened 2017-12-25
Generic functions parameter types can be omitted at call site if they can be inferred from context

wild-lynx:


pages/docs/reference/generics.md
```diff
@@ -261,6 +261,10 @@ To call a generic function, specify the type arguments at the call site **after*
 ``` kotlin
 val l = singletonList<Int>(1)
 ```
+Parameter types can be omitted if they can be inferred from context, so the following example works as well:
```
h0tk3y: Did you mean 'Type arguments can be omitted...'?

wild-lynx:
Yes, I did. Thank you!

PR #982, opened 2018-01-03
change abstract classes to interfaces in generics

silmeth:
To make them more closely corresponding to Java examples (`interface Source<out T>`) and to Kotlin’s own standard library (`interface Comparable<in T>`).

It’s been confusing for me – why Kotlin examples used `abstract class` instead of `interface` (I believe that’s because of historical reasons, not relevant today).


yole:
Thanks!

PR #984, opened 2018-01-07
object declarations clarifications

Miha-x64:
1. Speaking not so loud about singleton usefulness. Singletons should be used with caution, if ever used.
2. Note about thread-safety. As I can see, the question about thread-safety is asked frequently.


h0tk3y:
Thanks!

PR #988, opened 2018-01-11
Update command-line.md

wusp:
.kts file refers to Kotlin script file.


yole:
After your edit, the text effectively says "a script is a script", which doesn't make too much sense.

PR #990, opened 2018-01-11
Add PHP to the list of generator languages

bradynpoulsen:
PHP is fairly popular in the web application community and can be a cool reference to mention when introducing Kotlin Coroutines


yole:
Unlike all the other languages currently listed in the documentation, we are not familiar with PHP and did not look at its design when designing coroutines, so we'd rather not mention it.

PR #991, opened 2018-01-12
New Tutorial: C Interop

hhariri:


pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+layout: tutorial
+title:  "Interop with C Libraries"
+description: "A look at how to interop with C libraries"
+authors: Hadi Hariri 
+date: 2018-01-09
+showAuthorInfo: false
+---
+
+
+When writing native applications, often times we need to access certain functionality that is not included in the Kotlin standard library,
```
yole: "oftentimes" is spelled as a single word

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+title:  "Interop with C Libraries"
+description: "A look at how to interop with C libraries"
+authors: Hadi Hariri 
+date: 2018-01-09
+showAuthorInfo: false
+---
+
+
+When writing native applications, often times we need to access certain functionality that is not included in the Kotlin standard library, 
+such as for instance making HTTP calls, reading and writing from disk, etc.
```
yole: "such as for instance" is redundant

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+* [Create Kotlin Bindings](#generating-bindings)
+* [Consume a generated Kotlin API](#consuming-the-kotlin-api)
+* [Link the library into the application](#compiling-and-linking-the-library)
+
+
+## Generating Bindings
+
+An ideal scenario for interop is to call C functions as if we were calling Kotlin functions, that is, following the same signature and conventions. And that is precisely what the 
+`cinterop` tool provides us with. It takes a series of C libraries and generates the corresponding Kotlin bindings for it, which then allow us
```
yole: Normally it takes a single C library

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+to make some HTTP calls, so we'll create a file named `libcurl.def` with the following contents
+
+```
+headers = curl/curl.h
+headerFilter = curl/*
+linkerOpts.osx = -L/opt/local/lib -L/usr/local/opt/curl/lib -lcurl
+linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl```
+```
+
+A few things going on in that file so let's see them one by one. The first entry is `headers` which is the list of header files that we want to generate
```
yole: are going on

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+headers = curl/curl.h
+headerFilter = curl/*
+linkerOpts.osx = -L/opt/local/lib -L/usr/local/opt/curl/lib -lcurl
+linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl```
+```
+
+A few things going on in that file so let's see them one by one. The first entry is `headers` which is the list of header files that we want to generate 
+Kotlin stubs for. We can multiple files to this entry, separating each one with `\` on a new line. In our case we only want `curl.h`. The files we reference
```
yole: We can add

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+need to relative to the folder where the definition file is, or by a available on the system path (in our case it would be `/usr/bin/curl`).
```
yole: "a available" - a word is missing

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+The second line is the `headerFilter`. This is used to denote what exactly we want included. In C, when one file references another file with the `#include` directive, 
+all the headers are also included. Sometimes this may not be needed, and we can use this parameter, [using glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)), to fine tune things. 
+
+The third and forth line are about providing linker options, which can vary based on different target platforms. In our cause we are defining it for macOS and Linux.
```
yole: lines

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+```kotlin
+import libcurl.*
+
+fun main(args: Array<String>) {
+    val curl = curl_easy_init()
+    if (curl != null) {
+        curl_easy_setopt(curl, CURLOPT_URL, "http://example.com")
+        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L)
+    }
```
yole: if `curl != null` is false, you need to skip the entire body, not just the setopt calls.

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+The next step is to compile our application. We already covered the basics of compiling a Kotlin/Native application from the command line in the [A Basic Kotlin/Native application](basic-kotlin-native-app.html) tutorial.
+The only difference in this case is that we have to include the library that `cinterop` generated for us. 
+
+
+    konanc Main.kt -library build/c_interop/libcurl
+
+We can see that we're passing in as `library` parameter the outut path of `cinterop`. 
+
+If there are no errors during compilation, we should seen the output as a file named `program.kexe`, which on execution should output
```
yole: should see

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Compiling and Linking the library
+
+The next step is to compile our application. We already covered the basics of compiling a Kotlin/Native application from the command line in the [A Basic Kotlin/Native application](basic-kotlin-native-app.html) tutorial.
+The only difference in this case is that we have to include the library that `cinterop` generated for us. 
+
+
+    konanc Main.kt -library build/c_interop/libcurl
+
+We can see that we're passing in as `library` parameter the outut path of `cinterop`.
```
yole: output

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+date: 2018-01-09
+showAuthorInfo: false
+---
+
+
+When writing native applications, oftentimes we need to access certain functionality that is not included in the Kotlin standard library, 
+such as making HTTP calls, reading and writing from disk, etc.
+
+Kotlin/Native provides us with the ability to consume standard C libraries, opening up an entire ecosystem of functionality that exists 
+for pretty much anything we could need.
```
olonho: Maybe add something like
"Also note, that platform libraries allows one to use native for the platform functionality without need of creating any interoperability definitions."

See smth like https://github.com/JetBrains/kotlin-native/blob/master/PLATFORM_LIBS.md and https://github.com/JetBrains/kotlin-native/blob/master/samples/csvparser/src/main/kotlin/CsvParser.kt as an example.

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+## Generating Bindings
+
+An ideal scenario for interop is to call C functions as if we were calling Kotlin functions, that is, following the same signature and conventions. And that is precisely what the 
+`cinterop` tool provides us with. It takes a C library and generates the corresponding Kotlin bindings for it, which then allow us
+to use the library as if it were Kotlin code. 
+
+In order to generate these bindings, we need to create a library definition file (`def`) that contains some information about the headers we need to generate. In our case we want to use the famous `libcurl` library
+to make some HTTP calls, so we'll create a file named `libcurl.def` with the following contents
```
olonho: calls -> requests

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+linkerOpts.osx = -L/opt/local/lib -L/usr/local/opt/curl/lib -lcurl
+linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl```
+```
+
+A few things are going on in that file so let's see them one by one. The first entry is `headers` which is the list of header files that we want to generate 
+Kotlin stubs for. We can add multiple files to this entry, separating each one with `\` on a new line. In our case we only want `curl.h`. The files we reference
+need to be relative to the folder where the definition file is, or be available on the system path (in our case it would be `/usr/bin/curl`).
+
+The second line is the `headerFilter`. This is used to denote what exactly we want included. In C, when one file references another file with the `#include` directive, 
+all the headers are also included. Sometimes this may not be needed, and we can use this parameter, [using glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)), to fine tune things.
```
olonho: Maybe add something like:
"Note, that headerFilter is an optional argument, required if library you're using is being installed as a system library and you do not want to fetch external dependencies (such as system stdint.h header) into your interoperability library. It may be important for both library size optimizations and fixing potential conflicts between system and Kotlin/Native provided compilation environment."

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+An ideal scenario for interop is to call C functions as if we were calling Kotlin functions, that is, following the same signature and conventions. And that is precisely what the 
+`cinterop` tool provides us with. It takes a C library and generates the corresponding Kotlin bindings for it, which then allow us
+to use the library as if it were Kotlin code. 
+
+In order to generate these bindings, we need to create a library definition file (`def`) that contains some information about the headers we need to generate. In our case we want to use the famous `libcurl` library
+to make some HTTP calls, so we'll create a file named `libcurl.def` with the following contents
+
+```
+headers = curl/curl.h
```
olonho: Please mention `package` argument, allowing to tell which package definition produced from def file go, like 

package = curl

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+Compiler options can also be specific, albeit in this case it is not necessary. 
+
+The convention that is followed is that each library gets its own definition file, usually named the same as the library. For more information on all
+the options available to `cinterop`, see [README available on GitHub](https://github.com/JetBrains/kotlin-native/blob/master/INTEROP.md)
+
+Once we have the definition file ready, we can invoke `cinterop` and have it generate the bindings for us
+
+    cinterop -def libcurl.def -o build/c_interop/libcurl
+    
+The output should be a Kotlin compiled library (extension `.klib`) and the Kotlin source header files. We're also instructing `cinterop`
```
olonho: Producing Kotlin source header files is not smth we want to advertise. Also it will likely go away soon (will generate descriptors instead). I'd suggest to rewrite it without mentioning source code. `klib` utility (klib contents) is the recommended way of inspecting contents of the klib.

pages/docs/tutorials/native/interop-with-c.md
hhariri: As in renaming the default package? I don't get this working. I add package = curl for instance and it still produces package libcurl.

PR #992, opened 2018-01-13
Removed broken link to Kotlin-X umbrella. All projects are in https:/…

abudzinskyy:
…/github.com/Kotlin.


abudzinskyy:
https://github.com/kotlinx/kotlinx is not valid link. I haven't found an alternative link for Kotlin-X umbrella.

PR #995, opened 2018-01-20
Updated data class documentation to explain excluding properties.

AdamMc331:
https://youtrack.jetbrains.com/issue/KT-22377

### DIFF
--- a/pages/docs/reference/data-classes.md
+++ b/pages/docs/reference/data-classes.md
@@ -66,6 +66,33 @@ val jack = User(name = "Jack", age = 1)
 val olderJack = jack.copy(age = 2)
 ```
 
+## Excluding Properties
+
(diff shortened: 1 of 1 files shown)


pages/docs/reference/data-classes.md
```diff
@@ -66,6 +66,33 @@ val jack = User(name = "Jack", age = 1)
 val olderJack = jack.copy(age = 2)
 ```
 
+## Excluding Properties
```
ilya-g: The title and the section placement do not look consistent to me. 

The first section also mentions this behavior as "The compiler automatically derives the following members from all properties declared in the primary constructor". 

Probably this should be titled as "Properties declared in the class body" and placed the second.

h0tk3y:
Thanks! I've accepted @ilya-g's suggestion about the title and placement and added a few changes from myself (runnable code sample, some rephrasing).

AdamMc331:
I like the improvements. Thank you both. :)

PR #996, opened 2018-01-22
New Tutorial: KLib

hhariri:


pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+
+* [What is a Kotlin/Native Libary](#what-is-kotlin/native-library)
+* [How to create a Kotlin/Native Library](#creating-a-kotlin/native-libraries)
+* [How to consume Kotlin/Native libraries](#consuming-a-kotlin/native-library)
+* [How to use and create library repositories](#using-and-creating-library-repositories)
+
+
+## What is Kotlin/Native Library
+
+The concept of a library, a collection of resources to be used by applications, is available in Kotlin/Native and it is represented
```
yole: "Resources" is usually used to mean "not code". A library contains code. Better to change this to "functions" or "APIs" or something like that.

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+(as we would for instance with DLL files on the Windows platform).
+
+A `klib` file is a compressed archive in zip format with the following directory structure:
+
+File `utils.klib`
+
+```
+    targets/                    /* One or more target platforms */
+    
+        macbook/                /* When targeting macbook, this entry would appear */
```
yole: Seriously, "macbook"? Do I need to compile programs differently for my iMac?

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+            
+    linkdata/                   /* A set of [ProtoBuf](https://github.com/google/protobuf) files with serialised linkage metadata */ 
+    
+    resources/                  /* Resource files such as images, etc. */
+
+    manifest                    /* A file in Java Property Format describing the library */ 
+```
+
+**Note**: It's important to note that until Kotlin/Native stabilises, the internal format of a `klib` file may change somewhat. Usually when working with `klib` files however,  
+knowing the internals is not required.
```
yole: If it's not required to know, why is it the first thing you tell your readers about?

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+To create a library from this file, we can use the compiler with the parameter `-produce library` or `-p library` for short:
+
+    konanc utils.kt -p library
+    
+By default, the output of the filename is `library.klib`. We can override it using the `-output` or `-o` parameter:
+
+    konanc utils.kt -p library -o utils
+    
+
+The second way to create a library is using the `cinterop` tool which allows us to create a Kotlin/Native library of an existing
```
yole: "from an existing"

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+import utils.*
+
+fun main(args: Array<String>) {
+    printWarning("App is about to shut down!")
+}
+```
+
+Notice how we need to import the library on the first line using the `import` statement. 
+
+In order for the compiler to correctly link in the library, we need to pass the library name using the `-library` or `-l` paramter
```
yole: Typo in "parameter"

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+
+```kotlin
+import utils.*
+
+fun main(args: Array<String>) {
+    printWarning("App is about to shut down!")
+}
+```
+
+Notice how we need to import the library on the first line using the `import` statement.
```
yole: I think it's important to use precise wording here. The "import" statement does not import a library; it imports a package, and the declarations in this package are part of that library.

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+Notice how we need to import the library on the first line using the `import` statement. 
+
+In order for the compiler to correctly link in the library, we need to pass the library name using the `-library` or `-l` paramter
+
+    konanc sample.kt -l utils
+    
+this would then produce a single executable file with no runtime dependencies.
```
yole: Capitalize this?

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+
+In order for the compiler to correctly link in the library, we need to pass the library name using the `-library` or `-l` paramter
+
+    konanc sample.kt -l utils
+    
+this would then produce a single executable file with no runtime dependencies.    
+    
+## Using and creating library repositories
+
+Often it would be useful to use the same library across multiple applications. To avoid having various copies of the same library, the Kotlin
```
yole: "would be" -> "is"

pages/docs/tutorials/native/working-with-klib.md
```diff
@@ -0,0 +1,140 @@
...
+    
+    
+### Installing libraries to custom repositories
+
+In addition to using the default repository, we can also have libraries stored in custom repositories, which can be useful for instance
+if we want to share libraries amongst a group of projects.
+
+In order to do this, we can once again use the command `klib`: 
+
+    klib install utils -repository jetbrains
```
yole: I think it would be more clear to show an example with a full path to the repository.

pages/docs/tutorials/native/working-with-klib.md
hhariri: Well technically resources can mean code also, and the reason I said resources is because the library can contain images and other things. But I'll rephrase it somewhat.

pages/docs/tutorials/native/working-with-klib.md
hhariri: This is the compiler option. If they change it, we can update it.

pages/docs/tutorials/native/working-with-klib.md
hhariri: Cause it's a warning and if someone comes across this tutorial and sees it's not this they might now why I've not updated it yet :). But point taken.

pages/docs/tutorials/native/working-with-klib.md
hhariri: I've re-worded this to make it less cryptic and more straightforward

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+to use the library as if it were Kotlin code. 
+
+In order to generate these bindings, we need to create a library definition file (`def`) that contains some information about the headers we need to generate. In our case we want to use the famous `libcurl` library
+to make some HTTP calls, so we'll create a file named `libcurl.def` with the following contents
+
+```
+headers = curl/curl.h
+headerFilter = curl/*
+linkerOpts.osx = -L/opt/local/lib -L/usr/local/opt/curl/lib -lcurl
+linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl```
```
kikitux: have extra backticks

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -0,0 +1,140 @@
...
+
+fun main(args: Array<String>) {
+    val curl = curl_easy_init()
+    if (curl != null) {
+        curl_easy_setopt(curl, CURLOPT_URL, "http://example.com")
+        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L)
+    }
+    val res = curl_easy_perform(curl)
+    if (res != CURLE_OK) {
+        println("curl_easy_perform() failed ${curl_easy_strerror(res)}")
```
kikitux: this need to be:

```
println("curl_easy_perform() failed ${curl_easy_strerror(res)?.toKString()}")
```

PR #998, opened 2018-01-24
Add a three-letter acronym

guenhter:
It is stated in the description, that only two letter acronyms are capitalized. 
But when scanning over the text, this can be overlooked. 
Therefore a  three-letter-acronym is added to make it clear with the examples, that 
two letters are capitalized, and three ore more are just with capitalized first letter.


h0tk3y:
Thanks!

PR #1000, opened 2018-01-25
interop c // fix markdown and convert cpointer to KString

kikitux:
extra backticks removed

added `?.toKString()` to the curl error, otherwise a pointer message is print


kikitux:
this is how the code looks with those extra ``` 

screenshot 2018-01-25 23 58 13

h0tk3y:
Thanks!

PR #1002, opened 2018-02-01
Update comparison-to-java.md Ternary-Operator

Wasabi375:
There is this endless discussion about whether or not to add the ternary operator to Kotlin https://discuss.kotlinlang.org/t/ternary-operator/2116/122 and someone pointed out, that the docs don't list this as a difference to java. 
It seems like the ternary operator wont be added so I think it is a good idea to state this here, as it makes changing from java easier.


h0tk3y:
Thanks! I've changed the line to be uniform with the other points in the list, which all just mention a Java feature and provide a link to the docs page that describes the Kotlin alternative.

PR #1003, opened 2018-02-01
Improve code examples in manual section "Return at Labels" (#KT-22646)

ltongues:
https://youtrack.jetbrains.com/issue/KT-22646


h0tk3y:
Thanks! I've also described the way to simulate `break` with a nesting lambda as described in the StackOverflow answer, it might be useful. And I've converted the code samples to make them runnable.

PR #1006, opened 2018-02-06
JS-7113: Make Kotlin version more notable on the main page

AlexandrPrendota:


templates/inc/header.html
```diff
@@ -27,6 +27,7 @@
                     read more</a>
             </span>
             </div>
+            <div class="current-version">Version 1.2.21</div>
```
satamas: Use data from releases.yml

PR #1009, opened 2018-02-07
Add documentation for sam-with-receiver

yanex:


pages/docs/reference/compiler-plugins.md
```diff
@@ -275,3 +275,98 @@ As with all-open, add the plugin JAR file to the compiler plugin classpath and s
...
+}
+
+apply plugin: "kotlin-sam-with-receiver"
+```
+
+Or using the Gradle plugins DSL:
+
+```groovy
+plugins {
+  id "org.jetbrains.kotlin.plugin.sam.with.receiver" version "{{ site.data.releases.latest.version }}"
```
h0tk3y: Note that we haven't started publishing this plugin to the plugins portal yet.

yanex: Oh, I didn't know that it has to be explicitly published. I'll remove that part then, thanks.

PR #1010, opened 2018-02-08
Clarification on inheritance of `expect` declarations

wild-lynx:


udalov:
IMHO this page shouldn't go into the details, but only present an overview of necessary concepts. Maybe there should be another page or even section about the current rules for `expect`/`actual` declarations?

PR #1011, opened 2018-02-08
Update delegation.md

SerryJohns:
I've corrected grammar in the first sentence introducing delegation.
`implementation` to `implement`


yole:
The grammar is correct. See for example http://wiki.c2.com/?ImplementationInheritance for an explanation of the term "implementation inheritance".

PR #1014, opened 2018-02-12
Document that 'this' in infix calls cannot be omitted

yole:


h0tk3y:
Thanks! I will also consider rephrasing the requirements part, because the first two points are now required for the third one, which, in turn, may not be noticeable enough in this list. Looks like this part had been originally written before the `infix` modifier was introduced.

PR #1022, opened 2018-02-23
Fix confusing wrapping on backing properties example.

nachtien:
The last word, field, is wrapped to the next line and can be a bit confusing.  This prevents wrapping on standard zoom.


h0tk3y:
With an up-to-date version of Chrome and default font size the line is not wrapped:

image

Anyway, I've rephrased the line to make it more compact. Thanks!

PR #1025, opened 2018-03-08
correct kiwi -> kiwifruit

ke5stl:
'kiwi' is a bird, or a person that normally does fly, but for some reason does not (eg: grounded wing commander), or someone from New Zealand, or, sometimes, the nzd/usd currency pair.  just as grapefruit are not called grape (a grape is something else entirely), neither should the kiwifruit be called a kiwi.  yeah, I understand that in many people's minds, kiwi is also the fruit, but, it's just not right.  Have a look at this: https://medium.com/new-zealand-thoughts/kiwi-kiwi-or-kiwifruit-8ac0b30ac447


yole:
TIL

ke5stl:
much appreciated!  thank you :)

PR #1035, opened 2018-03-16
Added meaning of Pojo and Poco

wtoalabi:
To clear things up for those seeing the terms for the first time!


h0tk3y:
All of the terms *DTO*, *POJO*, and *POCO* appear to be well-known and easy to find.
This page should not explain them, it's better to keep the abbreviations as they are, since they are more noticeable this way.

PR #1036, opened 2018-03-27
Update Effective Java citation in classes.md

Javaru:
The page lists the *Design and document for inheritance or else prohibit it* Effective Java entry as being item 17 in "Effective Java".  This is technically an incorrect reference since listing the book as "Effective Java" implies the first edition, in which the mentioned item is #15. The citation should therefore be "Effective Java, 2nd Edition, Item #17". But given that the third edition was released in this past December, I think it, as the latest edition. should be cited. In the third edition it's item #19. Thus I recommend chaining the cite to  "Effective Java, 3rd Edition, Item #19".


h0tk3y:
Thanks!

h0tk3y:
I've also update the other places where *Effective Java* is mentioned [in this commit](https://github.com/JetBrains/kotlin-web-site/commit/5e78185c69a4a30eaa5e89538948ec935341f949)

PR #1042, opened 2018-04-13
Fix typo: @Parcelize, instead of @Parcelable

dextorer:
There is no such thing as the `@Parcelable` annotation, so it's rather a typo. This PR fixes the typo.


h0tk3y:
Thanks!

PR #1043, opened 2018-04-15
Clarify multiplatform project setup

SUPERCILEX:


pages/docs/reference/multiplatform.md
```diff
@@ -103,6 +103,7 @@ buildscript {
     }
 }
 
+apply plugin: 'java'
```
h0tk3y: The `java` plugin is not really needed in a Kotlin MPP platform module for JVM.

pages/docs/reference/multiplatform.md
```diff
@@ -111,7 +112,7 @@ repositories {
 
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    expectedBy project(":")
+    expectedBy project(":common")
```
h0tk3y: This sample shows a project layout where the root project is the common one, with platform modules being its subprojects. Why should we change this?

PR #1045, opened 2018-04-18
Fixed typo in properties.md

mangefox:
Comment claimed statement was in error but it was perfectly valid. Edited line to what I assume was the intended meaning


yole:
This statement (as well as the statement you want to replace it with) are only valid as declarations of an unused local variable. If you're declaring a top-level property or the property of a class, an initializer is required.

In other words, there is no typo here.

PR #1048, opened 2018-04-25
Fix incorrect property declaration

YakovBurtsev:
prop declaration in code is generated by compiler should use var keyword instead of val keyword, as it was used var keyword in original Also I add set method.


h0tk3y:
Thanks!

PR #1053, opened 2018-05-16
Call compile method for Gradle Kotlin DSL

StefMa:
In Gradle Kotlin DSL all configurations are methods.
Means the current documentation is just wrong and lead to issues.
This will fix it 👍


h0tk3y:
Thanks!

PR #1054, opened 2018-05-20
Fix for Chrome not respecting negative z-index

Krizzu:
Hey,

So recently I started to wonder why my mouse scroll (or even touch pad scroll) stopped working on the website. I dig bit more and found out that chrome has problem with negative z-index.

Search popup (when hidden) uses `z-index: -1` to hide its body behind other elements, and remove that property once search icon is clicked.

My Fix is just a proposal, to prevent using negative z-index. I replaced it with `top: 100%` property, which makes search body shop up sliding from the bottom.

Thanks

### DIFF
--- a/static/js/com/search/search.scss
+++ b/static/js/com/search/search.scss
@@ -42,7 +42,7 @@ body._no-scroll {
 
   &._hidden {
     opacity: 0;
-    z-index: -1;
+    top: 100%;
   }
 
   &:focus {


Krizzu:
Hey,

Any updates on this?

AlexanderPrendota:
@Krizzu Hello! I guess it's some of our bug but I can't reproduced it in my side

Krizzu:
Hey @AlexanderPrendota,

I'm using Chrome Version 67.0.3396.99 (Official Build) (64-bit).
macOS Sierra 10.12.6

I think it might be related to my Chrome/PC, because I tried Kotlin website on different hardware and it worked with no issue.

So the question is if You guys can do anything about? I might be a special snowflake here, haven't seen anyone with the same issue

AlexanderPrendota:
So, well. What am I doing wrong?
1) Open kotlinlang.org
2) Click on search icon
3) Insert some symbols 
4) Close search panel

and mouse scroll doesn't work, right?

Krizzu:
For me it's simply going to website. Scrolling with mouse/keys/tab doesn't move the screen. It's like a invisible full size element is blocking scroll ability. Although, I can scroll through content using scroll indicator on the right.

AlexanderPrendota:
@Krizzu I can merged your PR but I want to reproduce this problem first

Krizzu:
@AlexanderPrendota Hey,
I'd love to help with giving repro steps, but I simply can't. I can scroll normally on other browsers, but my Chrome is somehow broker. I tried reinstalling/disabling extensions and still nothing.

If I remove negative `z-index` style from `search-popup` I can scroll without issues.
Don't know what to say, IMHO I wouldn't recommend merging anything just because one guy has an issue.

That's why I opened this PR - to discuss a possible solutions. If there is none, we can just skip this.

nikpachoo:
fixed with #2477

PR #1057, opened 2018-05-26
fix example usage of trimIndent

jnizet:
the previous example didn't trim anything, as noted in the documentation:
"In case if there are non-blank lines with no leading whitespace characters (no indent at all) then the common indent is 0, and therefore this function doesn't change the indentation."

The new snippet formatting is the one applied by IntelliJ when choosing formatting based on the Kotlin style guide


h0tk3y:
Thanks!

PR #1064, opened 2018-06-17
Add note on visibility for extensions

Esenin:
I don't have prior experience of Java, C#, or Gosu, I'm also not familiar with a concept of extensions and it was **not** obvious to me that an extension declared outside of a class' scope does not have an access to the class' private (and protected?) members.

I was able to deduce an answer to my question about visibility from the following paragraph, though:

> Extensions do not actually modify classes they extend. By defining an extension, you do not insert new members into a class, but merely make new functions callable with the dot-notation on variables of this type.

However, I believe that the documentation would benefit by adding this information explicitly. 
Not sure how to formulate it better -- I'm open to any suggestions.

### DIFF
--- a/pages/docs/reference/extensions.md
+++ b/pages/docs/reference/extensions.md
@@ -260,9 +260,10 @@ C().caller(D1())  // prints "D.foo in C" - extension receiver is resolved static
 
 ## Note on visibility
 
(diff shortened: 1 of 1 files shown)


h0tk3y:
Thanks!

PR #1069, opened 2018-06-27
Update ranges.md

jacksonweekes:


h0tk3y:
The original explanation is true, it's `1 <= i`, `i` is greater or equals `1`.

PR #1079, opened 2018-07-18
Update operator-overloading.md

Arneball:
The example is updated so that Point.unaryMinus() prints the right string.


AlexanderPrendota:
Thanks. Merged as [`149e38d`](https://github.com/JetBrains/kotlin-web-site/commit/149e38d51c08a5fb9d85f4cfdf783cde53270d01)

PR #1081, opened 2018-07-21
Refactor name to avoid misunderstanding

1step2hell:
I'm a beginner with Kotlin, and I didn't know the method "repeat()" of CharSequence in Kotlin. First time I saw code like this:
val repeat: String.(Int) -> String = { times -> repeat(times) }
consider it as a recursive function type certainly, and I was confused: how does it recurse? There even wasn't a recursive condition.
This mistake last a while till I found out that method "repeat()" belongs to CharSequence in Kotlin. It might happen to some other new guys like me.
This is not a good sample. If you want to explain something, focus on that point only, try to avoid things which can bring misunderstanding.


h0tk3y:
Thanks!

PR #1084, opened 2018-08-02
Tutorial kn dynamic libraries

jonnyzzz:
https://youtrack.jetbrains.com/issues?q=EVAN-5371
The text is ready for the factual check. Language & Grammar will be re-checked afterwards

### DIFF
--- a/pages/docs/tutorials/native/dynamic-libraries.md
+++ b/pages/docs/tutorials/native/dynamic-libraries.md
@@ -64,7 +64,7 @@ The `kotlinc` generates the following files, depending on the OS,
 where you run the compiler:
 - macOS: `demo_api.h` and `libdemo.dylib`
 - Linux: `demo_api.h` and `libdemo.so`
-- Windows: `demo_api.h` and `demo.dll`
+- Windows: `demo_api.h`, `demo_symbols.def` and `demo.dll`
 
 Let's check the C API for our Kotlin code in the `demo_api.h` 
 
@@ -192,7 +192,8 @@ demo_KBoolean (*IsInstance)(demo_KNativePtr ref, const demo_KType* type);
 Those functions deals with Kotlin/Native objects. One calls 
 `DisposeStablePointer` to release a Kotlin object and `DisposeString` to release Kotlin String, 
 which has the `char*` type in C. One may use the `IsInstance` function to check, if a
-Kotlin type, a `demo_KNativePtr` is an instance of another type. 
+Kotlin type, a `demo_KNativePtr` is an instance of another type. The actual set of
+operations generated depend on the actual usage from your code.
  
 Kotlin/Native has garbage collection, but it does not help to deal
 with Kotlin objects from C language. Kotlin/Native has an interop with Objective-C and 
@@ -200,9 +201,12 @@ Swift, and integrates with their reference counters. You may want to find more d
 in the [documentation](https://github.com/JetBrains/kotlin-native/blob/master/OBJC_INTEROP.md)
 or to check the related [TUTORIAL_KOTLIN_macOS_FRAMEWORK_LINK].
 
+NOTE. It is possible to have several Kotlin/Native dynamic libraries in one application, 
+but it is NOT possible to share objects between them.
+
 ### Our Library Functions
 
-Let's take a look on the nests structure with path `kotlin.root.demo`. The path 
+Let's take a look on the `kotlin.root.demo` field, it
 mimics the package structure of our Kotlin code with `kotlin.root.` prefix.
 
 There is `kotlin.root.demo.DemoClazz` field, that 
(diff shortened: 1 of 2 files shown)


pages/docs/tutorials/native/dynamic-libraries.md
```diff
@@ -0,0 +1,339 @@
...
+ 
+Kotlin/Native has garbage collection, but it does not help to deal
+with Kotlin objects from C language. Kotlin/Native has an interop with Objective-C and 
+Swift, and integrates with their reference counters. You may want to find more details
+in the [documentation](https://github.com/JetBrains/kotlin-native/blob/master/OBJC_INTEROP.md)
+or to check the related [TUTORIAL_KOTLIN_macOS_FRAMEWORK_LINK].
+
+### Our Library Functions
+
+Let's take a look on the nests structure with path `kotlin.root.demo`. The path
```
olonho: What is "on the nests structure"?

pages/docs/tutorials/native/dynamic-libraries.md
```diff
@@ -0,0 +1,339 @@
...
+gcc main.c libdemo.so
+```
+
+The compiler generates the executable called `a.out`. We run it to see Kotlin code
+being executed from C library in action. On Linux we'll need to include `.` into the `LD_LIBRARY_PATH`
+to let the application to load the `libdemo.so` library from the current folder.
+
+## Compiling and Running the Example on Windows
+
+To start with, you'll need Microsoft Visual C++ compiler installed and support AMD x64
```
olonho: AMD x64  -> x86_64

pages/docs/tutorials/native/dynamic-libraries.md
```diff
@@ -0,0 +1,339 @@
...
+## Using Generated Headers from C
+
+The usage from C not complicated and strait forward. We create a `main.c` file with the following 
+code: 
+```c
+#include "demo_api.h"
+#include "stdio.h"
+
+int main(int argc, char** argv) {
+  //initialize Kotlin/Native library
```
olonho: This comment is incorrect for the end user. Smth like "Obtain reference for calling Kotlin/Native functions".

pages/docs/tutorials/native/dynamic-libraries.md
```diff
@@ -0,0 +1,339 @@
...
+  lib->DisposeString(response);
+
+  //create Kotlin object instance
+  demo_kref_demo_DemoClazz newInstance 
+          = lib->kotlin.root.demo.DemoClazz.DemoClazz();
+  long result = lib->kotlin.root.demo.DemoClazz.foo(newInstance, 42);
+  printf("DemoClazz returned %ld\n", result);
+  lib->DisposeStablePointer(newInstance.pinned);
+
+  return 0;
```
olonho: Following things need to be mentioned:
 - Only operations explicitly mentioned in the source code are exposed to the C world
 - If you're using multiple Kotlin/Native libraries in the same C program object handles of those are generally incompatible, and shall not be passed each to another.

pages/docs/tutorials/native/dynamic-libraries.md
```diff
@@ -0,0 +1,339 @@
...
+target. The easiest way is to have a version of Microsoft Visual Studio installed on 
+your Windows machine. 
+
+We will be using `x64 Native Tools Command Prompt <VERSION>` console. You'll see the 
+shortcut to open the console in the start menu. It comes with a Microsoft Visual Studio
+package.  
+
+On Windows Dynamic libraries are included either via a special static library wrappers,
+or via a manual code with the [LoadLibrary](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684175.aspx)
+function call. We will follow the first option and generate the static wrapper library
```
olonho: Please fix all that: we generate .def file, and we could also generate both static and dynamic libraries by K/N.

pages/docs/tutorials/native/dynamic-libraries.md
jonnyzzz: I'll use `field` instead to refer to a chain of fields. It should be easier to understand, I guess

pages/docs/tutorials/native/dynamic-libraries.md
jonnyzzz: fixed

pages/docs/tutorials/native/dynamic-libraries.md
jonnyzzz: Does it mean we cache initialized state inside and it will always return the same reference even if called multiple times? from multiple threads?

pages/docs/tutorials/native/dynamic-libraries.md
jonnyzzz: Could you please elaborate on the first. What else, beyond `DisposeStablePointer`, `DisposeString` and `IsInstance` it may generate?

pages/docs/tutorials/native/dynamic-libraries.md
jonnyzzz: I found the .def file generation with K/N 0.8.2. Cool!. On Windows one needs to have a static library wrapper that uses the DLL. Are we able to generate one? Right now I see no such library generated.

jonnyzzz:
merged with review #1091

PR #1085, opened 2018-08-03
Remove the example due to empty range.

Nestarneal:
4..1 is an empty range. If we want to iterate from 4 to 1, we should use 4 downTo 1 instead of 4..1


AlexanderPrendota:
It's just example. So, All is file =)

PR #1094, opened 2018-08-14
Update compiler-plugins.md

s1monw1:
Changed wording. I recommend to not specifically say "data class" in that case, because actually using data classes for JPA entities might be a bad practice (equals/hashCode dilemma).


h0tk3y:
Thsnks!

PR #1096, opened 2018-08-15
Add clarifying explanation for computed properties

runningcode:
When using a computed property with a `get()` the behavior will be different than without using `get()` so add some explanation for this.

Feel free to help me out with the wording and syntax. :)

### DIFF
--- a/pages/docs/reference/properties.md
+++ b/pages/docs/reference/properties.md
@@ -115,6 +115,28 @@ var setterWithAnnotation: Any? = null
 ```
 </div>
 
+
+### Computed Properties
+
+When using a custom getter for a property (`get()`) the value is computed once and stored. However, when omitting the `get()`, the value is computed on every access.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+``` kotlin
+val isEmpty get() = this.size == 0 // Computed once and stored.
+val isEmpty = this.size == 0 // Computed on every access.
+```
+</div>
+
+When using a `var` the behavior is the same in both cases. The value is computed on every access.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+``` kotlin
+var isEmpty get() = this.size == 0  // Computed on every access.
+var isEmpty = this.size == 0 // Computed on every access.
+```
+</div>
+
+
 ### Backing Fields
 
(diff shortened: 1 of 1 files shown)


pages/docs/reference/properties.md
```diff
@@ -115,6 +115,28 @@ var setterWithAnnotation: Any? = null
...
+val isEmpty get() = this.size == 0 // Computed on every access.
+val isEmpty = this.size == 0 // Computed once and stored.
+```
+</div>
+
+When using a `var` the behavior is almost the same except that the computed value is stored until it changes or is overwritten.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+``` kotlin
+var isEmpty get() = this.size == 0  // Computed on every access.
```
jonnyzzz: Setter or an initializer is missing here. It will not compile.

runningcode:
@jonnyzzz Thanks for pointing that out. How does it look now?

pages/docs/reference/properties.md
```diff
@@ -115,6 +115,29 @@ var setterWithAnnotation: Any? = null
 ```
 </div>
 
+
+### Computed Properties
+
+When using a computed property, the value is computed once and stored. This behavior changes when using adding `get()`. The `get()` causes the value to be computed on every access.
```
jonnyzzz: when using addin`. It seems one word is an extra here

pages/docs/reference/properties.md
```diff
@@ -115,6 +115,29 @@ var setterWithAnnotation: Any? = null
...
+val isEmpty get() = this.size == 0 // Computed on every access.
+val isEmpty = this.size == 0 // Computed once and stored.
+```
+</div>
+
+When using a `var` the behavior is almost the same except that the computed value is stored until it changes or is overwritten.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+``` kotlin
+var isEmpty = false
```
jonnyzzz: could you please avoid using backing fields here, which are discussed below. It looks like you may have a setter declared too

pages/docs/reference/properties.md
jonnyzzz: What is the computed property? Is it the one with `get` and `set` declared?

yole:
Thanks for your contribution! Unfortunately the text in this PR does not exactly match the terminology we're using (a property with an initializer and no getter isn't a computed property), so I've pushed a different change that makes the role of custom accessors clearer.

PR #1098, opened 2018-08-19
Stylistic changes for cos() fixpoint computation

dtonhofer:
Let's keep the example stylistically good: It seems unwise to do exact floating point comparison in an infinite loop as stated. Let's compare against a target precision of 10^-10 instead (machine epsilon of Double is 2^-52 ~ 1.11e-16, and the result is expected to be < 1.0, so 10^-15 sounds safe, but let's just go for 10^-10). 

Additionally, the loop code optimizes by only calling Math.cos(x) once, but the compiler hopefully won't do that, as the language is not purely functional. So I added the second Math.cos(x) code into the loop, for clarity.

Also fixed something that looks like a russianism in the text (apologies)

### DIFF
--- a/pages/docs/reference/functions.md
+++ b/pages/docs/reference/functions.md
@@ -310,7 +310,7 @@ class MyStringCollection {
 
 ## Function Scope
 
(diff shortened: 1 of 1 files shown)


h0tk3y:
Thanks!

PR #1100, opened 2018-08-22
Update oss-projects.yml

tuttza:
I thought TornadoFX GUI framework would be worthy of adding to the list.


AlexanderPrendota:
okay! Thanks!

PR #1103, opened 2018-08-28
Build-style usage of methods that return

hamzamuric:


h0tk3y:
Even though the example does not create an array filled with a certain value in the idiomatic way, it shows an idiomatic usage of `apply` (as said in the title) and becomes meaningless if `apply` is removed. A better usage of `apply` not connotating with arrays would improve the sample.

PR #1111, opened 2018-09-10
Add synthetic properties information for Fragments

sjaramillo10:


p7nov:
We don't publish Android tutorials on the Kotlin website anymore, so this is obsolete. Thanks.

PR #1114, opened 2018-09-13
Refine compiler's error message

xamgore:
It's not clear for the first time, where is the problem in the original code.
Now it's more intuitive.


AlexanderPrendota:
Thanks!

PR #1122, opened 2018-09-27
Update and comment Kotlin/Native tutorials (WIP)

SvyatoslavScherbina:


pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -41,7 +41,7 @@ linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl
 
 A few things are going on in that file so let's see them one by one. The first entry is `headers` which is the list of header files that we want to generate 
 Kotlin stubs for. We can add multiple files to this entry, separating each one with `\` on a new line. In our case we only want `curl.h`. The files we reference
-need to be relative to the folder where the definition file is, or be available on the system path (in our case it would be `/usr/bin/curl`).
+need to be relative to the folder where the definition file is, or be available on the system path (in our case it would be `/usr/include/curl`).
```
SvyatoslavScherbina: > The files we reference need to be relative to the folder where the definition file is

I'm not sure it works as described.

> or be available on the system path 

Or on the user-specified path, e.g. with `-I` compiler options.

pages/docs/tutorials/native/interop-with-c.md
```diff
@@ -41,7 +41,7 @@ linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl
...
 
 The second line is the `headerFilter`. This is used to denote what exactly we want included. In C, when one file references another file with the `#include` directive, 
 all the headers are also included. Sometimes this may not be needed, and we can use this parameter, [using glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)), to fine tune things.
```
SvyatoslavScherbina: Btw, extended Java globs are supported, so that `**` can be used.

SvyatoslavScherbina: (below)
> is an optional argument and mostly only used when the library we're using is being installed as a system library and we do not want to fetch external dependencies 

How does it differ when the same library isn't installed as a system one?

pages/docs/tutorials/native/interop-with-c.md
SvyatoslavScherbina: Moreover, `cinterop` doesn't use headers from system paths like `/usr/include` by default.
On macOS it uses Xcode-provided SDKs, on other platforms Kotlin/Native has its own "system roots" (aka sysroots), located somewhere in `~/.konan/dependencies/`.

SvyatoslavScherbina:
[Mapping Strings from C](https://github.com/JetBrains/kotlin-web-site/blob/SvyatoslavScherbina-patch-1/pages/docs/tutorials/native/mapping-strings-from-c.md) tutorial doesn't mention that `const char*` parameters are mapped directly to `String`.

jonnyzzz:
merged with #1131

PR #1126, opened 2018-09-28
Updated usage of sealed classes with when.

aschatten:
You don't need an `else` clause even if `when` is used as a statement. A somewhat ugly and artificial example from [Kotlin Koans](https://try.kotlinlang.org/#/Kotlin%20Koans/Introduction/Smart%20casts/Task.kt)
```
fun eval(expr: Expr): Int {
    when (expr) {
        is Num -> {
            return expr.value
        }
        is Sum -> {
            return eval(expr.left) + eval(expr.right)
        }
    }
}
```


yole:
The wording is somewhat confusing but correct. The idea is that, if you use `when` as an expression, the compiler verifies that all inheritors of a `sealed` class are accounted for, and you can omit the `else` clause when that's the case. If `when` is a statement, there are no exhaustiveness checks performed.

PR #1151, opened 2018-10-05
Enabling experimental Gradle metadata incorrect

jstuyts:
A single quote was missing at the end. This has been added.


h0tk3y:
Thanks!

PR #1152, opened 2018-10-05
Fix dead link in server overview docs

miknie:
Replace dead article link for Corda with link to their homepage.


AlexanderPrendota:
Thanks!

PR #1158, opened 2018-10-08
Update extensions.md

tjancic:


yole:
The parameters to the `swap` function are indices in the list. The type of a list index is `Int`. The example is correct as written.

djulieani:
fun <T> MutableList <T> .swap (index1: T , index2: T ) {

PR #1159, opened 2018-10-09
Added Link to Free Bootcamp by Google on edx

amitsk:


p7nov:
Thanks for the suggestion.
However, at the moment we list only the courses provided by the JetBrains team.

PR #1162, opened 2018-10-12
docs(android): Link to Android Studio stable

mreichelt:
- before the main weg page and the Android sub page were
  linking to Android Studio Preview. Since Android Studio 3.0 stable
  is out for some time now we really should link to the stable version
  instead!


AlexanderPrendota:
Thanks

PR #1182, opened 2018-10-30
Add short rationale to Java features that are not present in Kotlin

MarcinMoskala:
Non-private fields should be removed because we have them using @JvmField

### DIFF
--- a/pages/docs/reference/comparison-to-java.md
+++ b/pages/docs/reference/comparison-to-java.md
@@ -21,11 +21,10 @@ Kotlin fixes a series of issues that Java suffers from:
 ## What Java has that Kotlin does not
 
 * [Checked exceptions](exceptions.html)
-* [Primitive types](basic-types.html) that are not classes
-* [Static members](classes.html)
-* [Non-private fields](properties.html)
-* [Wildcard-types](generics.html)
-* [Ternary-operator `a ? b : c`](control-flow.html#if-expression)
+* [Primitive types](basic-types.html) that are not classes - are used under the hood, but cannot be used explicitely. 
+* [Static members](classes.html) - replaced with [companion objects](object-declarations.html#companion-objects), and @JvmStatic.
+* [Wildcard-types](generics.html) - replaced with declaration-site variance and type projections.
+* [Ternary-operator `a ? b : c`](control-flow.html#if-expression) - replaced with [if expression](control-flow.html#if-expression). 
 
 ## What Kotlin has that Java does not
 
--- a/pages/docs/reference/comparison-to-java.md
+++ b/pages/docs/reference/comparison-to-java.md
@@ -23,7 +23,7 @@ Kotlin fixes a series of issues that Java suffers from:
 * [Checked exceptions](exceptions.html)
 * [Primitive types](basic-types.html) that are not classes - The byte-code uses primitives where possible, but they are not explicitly available.
 * [Static members](classes.html) - replaced with [companion objects](object-declarations.html#companion-objects), [top-level functions](functions.html), [extension functions](extensions.html#extension-functions), or [@JvmStatic](java-to-kotlin-interop.html#static-methods).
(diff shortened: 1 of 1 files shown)


MarcinMoskala:
The reason is that when you read this page currently it looks like this are advantages of Java, not as a list of things that were intentionally removed. I often see people referencing to this list as "Kotlin has less developed generics because it does not supports Wildcard-types" or "You cannot use Primitive types in Kotlin". This is why I think there should be a short explanation what is used instead.

MarcinMoskala:
I know that above there are already reasons why these features are not in Kotlin. But people do not read it and they just search for some cheap arguments why Java is better to later give it with info that it comes from Kotlin documentation. Adding such explanation there should help.

pages/docs/reference/comparison-to-java.md
```diff
@@ -21,11 +21,10 @@ Kotlin fixes a series of issues that Java suffers from:
 ## What Java has that Kotlin does not
 
 * [Checked exceptions](exceptions.html)
-* [Primitive types](basic-types.html) that are not classes
-* [Static members](classes.html)
-* [Non-private fields](properties.html)
-* [Wildcard-types](generics.html)
-* [Ternary-operator `a ? b : c`](control-flow.html#if-expression)
+* [Primitive types](basic-types.html) that are not classes - are used under the hood, but cannot be used explicitely.
```
yole: explicitely -> explicitly

pages/docs/reference/comparison-to-java.md
```diff
@@ -21,11 +21,10 @@ Kotlin fixes a series of issues that Java suffers from:
 ## What Java has that Kotlin does not
 
 * [Checked exceptions](exceptions.html)
-* [Primitive types](basic-types.html) that are not classes
-* [Static members](classes.html)
-* [Non-private fields](properties.html)
-* [Wildcard-types](generics.html)
-* [Ternary-operator `a ? b : c`](control-flow.html#if-expression)
+* [Primitive types](basic-types.html) that are not classes - are used under the hood, but cannot be used explicitely. 
+* [Static members](classes.html) - replaced with [companion objects](object-declarations.html#companion-objects), and @JvmStatic.
```
yole: Link to docs for `@JvmStatic?`

pages/docs/reference/comparison-to-java.md
```diff
@@ -21,11 +21,10 @@ Kotlin fixes a series of issues that Java suffers from:
...
+* [Wildcard-types](generics.html) - replaced with declaration-site variance and type projections.
```
yole: Link to docs too?

pages/docs/reference/comparison-to-java.md
mikezx6r: [Primitive types](basic-types.html) that are not classes. The byte-code uses primitives where possible, but they are not explicitly available.

pages/docs/reference/comparison-to-java.md
mikezx6r: Replaced with companion objects, top-level functions, extension functions, or @JvmStatic.

MarcinMoskala:
I responded to all comments. Any more feedback?

mikezx6r:
In Slack https://kotlinlang.slack.com/archives/C0922A726/p1540905520498700?thread_ts=1540898720.491000&cid=C0922A726 
there was a suggestion (I agree with it) that all of the 'Things Java has that Kotlin does not' items are actually issues or confusions in Java that Kotlin addresses. They are not things Java has that Kotlin doesn't, but rather issues Java has that Kotlin addresses.

That's a larger change, and requires more discussion, so I'm not proposing changes to your current PR, but perhaps a new one that moves these items up to the top level.

For example primitives. Java has them for performance enhancements. Kotlin doesn't need them because the compiler mostly takes care of that, thereby reducing the confusion of 'should I use primitives or classes'.

PR #1183, opened 2018-10-30
Fix a "bug" in code snippet

mouse4d:
Java String.format uses vararg, so we should pass arguments as vararg. Otherwise it compiles ok but throws a runtime exception |java.util.IllegalFormatConversionException: x != [Ljava.lang.Object;"

I banged my head for half a day to understand what's wrong and in the end Andrey Mischenko showed me the answer.


yole:
Thanks for your contribution!

PR #1184, opened 2018-10-30
Fix wrong task name in using-gradle.md

Bluexin:


yole:
Thanks for your contribution!

PR #1185, opened 2018-10-30
docs(ranges) : remove incorrect example

NinoDLC:
The range 4..1 is empty, thus incorrect. Should use downTo instead


NinoDLC:
How come you let a faulty example in the doc ? I spent 30 minutes wondering why the code wouldn't work, trying stuff on the compiler. 

This is extremely misleading to me.

hhariri:
Yes, sorry. It was my bad.

PR #1186, opened 2018-10-30
fix small typo ("a Apple" -> "an Apple")

jasonjmcghee:
Issue does not exist in header of https://kotlinlang.org/docs/tutorials/native/apple-framework.html, but link to the page has the typo.


h0tk3y:
Thanks!

PR #1193, opened 2018-11-04
Fix broken libcurl.def in tutorial

emanguy:
I downloaded the Kotlin/Native compiler using the linux artifact on version 0.9.3. I'm running Ubuntu 18.04 with libcurl4 and libcurl4-openssl-dev installed. The original .def file did not work, with `cinterop` saying "Could not find curl/curl.h". I then added `-I/usr/include/x86_64-linux-gnu`, and curl.h was found but it was complaining about a missing definition for the macro `__glibc_clang_prereq`. After adding the second compiler include (`-I/usr/include`) creation of the .klib worked.

### DIFF
--- a/pages/docs/tutorials/native/interop-with-c.md
+++ b/pages/docs/tutorials/native/interop-with-c.md
@@ -35,6 +35,7 @@ to make some HTTP calls, so we'll create a file named `libcurl.def` with the fol
 ```
 headers = curl/curl.h
 headerFilter = curl/*
+compilerOpts.linux = -I/usr/include -I/usr/include/x86_64-linux-gnu
 linkerOpts.osx = -L/opt/local/lib -L/usr/local/opt/curl/lib -lcurl
 linkerOpts.linux = -L/usr/lib/x86_64-linux-gnu -lcurl
 ```


jonnyzzz:
Thank you for the pull request. It does not seem to work for me on Ubuntu 18.04 (Docker) and Kotlin/Native 1.0.1 (which is the Beta release included into Kotlin 1.3.0)
```
root@757711954ed4:/kotlin-native-linux-1.0.1/bin# ./cinterop -def curl.def
Exception in thread "main" java.lang.Error: /tmp/tmp1393423407243474707.c:1:10: fatal error: 'curl/curl.h' file not found
	at org.jetbrains.kotlin.native.interop.indexer.UtilsKt.ensureNoCompileErrors(Utils.kt:145)
	at org.jetbrains.kotlin.native.interop.indexer.IndexerKt.indexDeclarations(Indexer.kt:916)
	at org.jetbrains.kotlin.native.interop.indexer.IndexerKt.buildNativeIndexImpl(Indexer.kt:906)
	at org.jetbrains.kotlin.native.interop.indexer.NativeIndexKt.buildNativeIndex(NativeIndex.kt:56)
	at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.processCLib(main.kt:242)
	at org.jetbrains.kotlin.native.interop.gen.jvm.MainKt.interop(main.kt:36)
	at org.jetbrains.kotlin.cli.utilities.InteropCompilerKt.invokeInterop(InteropCompiler.kt:88)
	at org.jetbrains.kotlin.cli.utilities.MainKt.main(main.kt:18)
```

Installing the `libcurl4-openssl-dev` package made it work. Thanks for the fix!

PR #1197, opened 2018-11-07
Fixed a typo in sam-with-receiver samples

jmatsu:
TaskRunner is defined as  the sample of SAM with receiver stuff but TaskHandler class, which is not defined, is used.


yole:
Thanks for your contribution!

PR #1199, opened 2018-11-08
Lambda argument should be moved out of parentheses

shuhuaxie:
I learned this from Android Studio code hint


yole:
This syntax feature is explained later in the document: http://kotlinlang.org/docs/reference/lambdas.html#passing-a-lambda-to-the-last-parameter

To make the initial example more clear, it does not use any syntax sugar.

PR #1200, opened 2018-11-09
Escape keyword used as a parameter name.

pchmielowski:
As `expect` is a keyword in Kotlin, the parameter name should be escaped by backtics.


yole:
`expect` is a soft keyword. It has a special meaning only when it is used as part of a modifier list. When it's used as a name, it does not require escaping.

PR #1208, opened 2018-11-14
Update android-frameworks.md

ThunderVader:
without it annotation @AutoValue can't be found


yole:
We actually don't want to promote AutoParcel because we now have our own solution which is much better (the `@Parcelize` annotation). I've removed this section from the tutorial.

PR #1223, opened 2018-11-26
Update javascript-dce.md for Kotlin version 1.3

dkotrada:
Because Kotlin 1.3 is released.


yole:
The section needs updating, but this doesn't automatically mean that the feature is no longer experimental.

dkotrada:
because it was explicitly version 1.1. .... just wanted to improve it.

PR #1225, opened 2018-11-28
Missing spring dependency-management gradle plugin

gavvvr:
You either need to use the plugin or implicitly specify the version of `spring-boot-starter-web`


p7nov:
Thanks! Sorry for long waiting.

PR #1228, opened 2018-11-29
Sync to coroutine's README.md.

jgoamakf:
To be consistent with https://github.com/Kotlin/kotlinx.coroutines/blob/master/README.md.

- Versions of Kotlin and coroutine are set to latest
- Avoid deprecated gradle configuration.


yole:
Thanks!

PR #1234, opened 2018-12-07
KT-28714 Fix plugin statement

KenjiOhtsuka:
https://youtrack.jetbrains.com/issue/KT-28714

According to the code in https://github.com/JetBrains/kotlin-examples/blob/master/tutorials/servlet-web-applications/build.gradle , `war` is surrounded by `'`.


yole:
Thanks!

PR #1236, opened 2018-12-10
Corrected by replacing `if` with `for`

karanatwal:


yole:
The original example is correct. See the comment above it for an explanation of what the code does.

PR #1237, opened 2018-12-10
Update on Using nullable values and checking for null part

MrKwon:
On ## Using nullable values and checking for *null*{: .keyword }
define function `` fun parseInt(str: String): Int? { } ``


yole:
If you want to show how the function is implemented, it's better to use `.toIntOrNull()` as in the following example, and not to duplicate its logic in this example code.

MrKwon:
Thanks for feedback !

PR #1238, opened 2018-12-13
Add note about lack of thread-safety of immutable collection ops

kszilagyi:
I think it's important to note this, as a lot of people (like myself) might come from languages where the standard library has persistent collections (like Scala) and operations on them therefore are completely thread-safe. Given the ones in Kotlin look very similar one might immediately assume they can be used the same way while a little more care needed in practice.

### DIFF
--- a/pages/docs/reference/collections.md
+++ b/pages/docs/reference/collections.md
@@ -45,7 +45,7 @@ val items = listOf(1, 2, 3)
 
 </div>
 
-Currently, the `listOf` method is implemented using an array list, but in future more memory-efficient fully immutable collection types could be returned that exploit the fact that they know they can't change. 
+Currently, the `listOf` method is implemented using an array list, but in future more memory-efficient fully immutable collection types could be returned that exploit the fact that they know they can't change.
 
 Note that the read-only types are [covariant](generics.html#variance). That means, you can take a `List<Rectangle>` and assign it to `List<Shape>` assuming `Rectangle` inherits from `Shape` (the collection types have the same inheritance relationship as the element types). This wouldn't be allowed with the mutable collection types because it would allow for failures at runtime: you might add a `Circle` into the `List<Shape>`, creating a `List<Rectangle>` with a `Circle` in it somewhere else in the program.
 
--- a/pages/docs/reference/collections.md
+++ b/pages/docs/reference/collections.md
@@ -45,7 +45,7 @@ val items = listOf(1, 2, 3)
 
 </div>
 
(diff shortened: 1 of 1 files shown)


yole:
Thanks for submitting! Could you please correct a couple of issues before we can merge this? The note is misplaced here (it needs to be a separate paragraph after "... as well as all the utilities you would expect such as sort, zip, fold, reduce and so on"), and "synchronization" is spelled differently.

kszilagyi:
Thanks for your comments! I have fixed them, let me know if you spot anything else.

yole:
Thanks!

pages/docs/reference/collections.md
```diff
@@ -84,6 +84,8 @@ val item = rwList.firstOrNull()
 
 ... as well as all the utilities you would expect such as sort, zip, fold, reduce and so on.
 
+It's important to note, that the operations on readonly collections which return a modified collection (like `+`, `filter`, `drop`, etc.) do not create their results atomically, therefore the results are not safe to use from a different thread without proper synchronization.
```
ilya-g: I'm late to the party, but I don't think this note adds clarity: 
- First it only applies to those read-only collections that are backed by mutable implementations, which can be referenced as mutable and modified through that mutable reference by a different thread. 
-  Second, "the results are not safe to use from a different thread without proper synchronization" makes an impression that if you synchronize access to the _results_ in a different thread you'll be safe, however what should actually be synchronized is the collection modification in the other thread and collection query in the thread that makes that query.

kszilagyi: Hey, thanks for your comment, I think you are right that it's not clear enough but not for the reason you think. I will create a new MR to try to make it clear and possibly have a link to this comment. 

There are two kinds of thread safety issues with the collections. The first is the one which your points are referring to. Given that's fairly straightforward and mutable backing is mentioned in the documentation I am not worried about that too much. But there is a much more subtle issue here what I was trying to point out: even if the backing collection is immutable and doesn't change, the results of the operations are not safe to read from a different thread. 

Here is the implementation of the `+` method from the standard library:
```
public operator fun <T> Collection<T>.plus(element: T): List<T> {
    val result = ArrayList<T>(size + 1)
    result.addAll(this)
    result.add(element)
    return result
}
```
As you can see, the result is constructed in multiple steps. Given there are no locks, the compiler is allowed to inline the whole block and reorder the statements. If I pass the result to a new thread after I called `+`, it is not guaranteed that it will see the fully constructed object. The new thread might see an empty ArrayList, an ArrayList with the original elements or the right one or basically anything in between, depending of what kind of reordering/optimisation the compiler has done.

This would be safe if `+` returned a properly implemented persistent collection as they can't be in a half state. But given it doesn't we have to be careful when we share the results between threads.

kszilagyi: New pull request for clarification: https://github.com/JetBrains/kotlin-web-site/pull/1243

PR #1240, opened 2018-12-14
docs: added multi-language examples, moved kapt part to kapt.md

p7nov:
- added multi-language samples with Kotlin DSL
- moved the kapt-gradle part from using-gradle.md to kapt.md
- revised the text.

### DIFF
--- a/pages/docs/reference/kapt.md
+++ b/pages/docs/reference/kapt.md
@@ -26,7 +26,8 @@ apply plugin: 'kotlin-kapt'
 
 Or you can apply it using the plugins DSL:
 
-<div class="sample" markdown="1" mode="groovy" theme="idea">
+<div class="multi-language-sample" data-lang="groovy">
+<div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
 plugins {
@@ -35,10 +36,24 @@ plugins {
 ```
 
 </div>
+</div>
+
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+```kotlin
+plugins {
+    kotlin("kapt") version "{{ site.data.releases.latest.version }}"
+}
+```
+
+</div>
+</div>
 
 Then add the respective dependencies using the `kapt` configuration in your `dependencies` block:
 
-<div class="sample" markdown="1" mode="groovy" theme="idea">
+<div class="multi-language-sample" data-lang="groovy">
+<div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
 dependencies {
@@ -46,6 +61,19 @@ dependencies {
 }
 ```
 
+</div>
+</div>
+
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+```kotlin
+dependencies {
+    kapt("groupId:artifactId:version")
+}
+```
+
+</div>
 </div>
 
 If you previously used the [Android support](https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#annotationProcessor_config) for annotation processors, replace usages of the `annotationProcessor` configuration with `kapt`. If your project contains Java classes, `kapt` will also take care of them.
@@ -56,7 +84,8 @@ If you use annotation processors for your `androidTest` or `test` sources, the r
 
 Use `arguments {}` block to pass arguments to annotation processors:
 
-<div class="sample" markdown="1" mode="groovy" theme="idea">
+<div class="multi-language-sample" data-lang="groovy">
+<div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
 kapt {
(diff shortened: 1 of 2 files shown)


pages/docs/reference/kapt.md
```diff
@@ -66,14 +95,59 @@ kapt {
...
+
+</div>
+</div>
+
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+```kotlin
+kapt {
+    useBuildCache = true
```
h0tk3y: When the code is the same (or may be written in a way that is valid in both Groovy and Kotlin DSL), a multi-language sample may be not necessary.

pages/docs/reference/using-gradle.md
```diff
@@ -58,6 +58,27 @@ apply plugin: 'org.jetbrains.kotlin.<...>'
 </div>
 </div>
 
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+
+```kotlin
```
h0tk3y: As [the `plugins { ... }` DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) and the Gradle plugin portal become more popular, we may want to restructure the plugin usage instructions to recommend using just this in the first place:
```
plugins {
    kotlin("jvm") version "..." // it's `id 'org.jetbrains.kotlin.jvm' version '...'`  in Groovy
}
```
We should still mention this way to apply the plugin as a fallback option.

pages/docs/reference/using-gradle.md
```diff
@@ -154,49 +175,97 @@ With Gradle Kotlin DSL, configure source sets with `java.sourceSets { ... }` ins
...
 When targeting JavaScript, a different plugin should be applied:
 
-<div class="sample" markdown="1" mode="groovy" theme="idea">
+<div class="multi-language-sample" data-lang="groovy">
+<div class="sample" markdown="1" mode="groovy" theme="idea" data-highlight-only>
 
 ``` groovy
-apply plugin: "kotlin2js"
+plugins {
+    id 'kotlin2js' version '{{ site.data.releases.latest.version }}'
```
h0tk3y: The Kotlin/JS plugin is somewhat special because it's not yet properly published to be applied with the `plugins { ... }` DSL, so this like won't work without the workaround code in `settings.gradle` that is added by our New Project wizard:
```
pluginManagement {
    resolutionStrategy {
        eachPlugin {
            if (requested.id.id == "kotlin2js") {
                useModule("org.jetbrains.kotlin:kotlin-gradle-plugin:${requested.version}")
            }
        }
    }
}
```
I think it should be mentioned here as a part of the instruction.

pages/docs/reference/using-gradle.md
```diff
@@ -154,49 +175,97 @@ With Gradle Kotlin DSL, configure source sets with `java.sourceSets { ... }` ins
...
 
-This plugin only works for Kotlin files so it is recommended to keep Kotlin and Java files separate (if it's the case that the same project contains Java files). As with
-targeting the JVM, if not using the default convention, we need to specify the source folder using *sourceSets*:
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
 
-<div class="sample" markdown="1" mode="groovy" theme="idea">
+```kotlin
+plugins {
+    kotlin("js") version "{{ site.data.releases.latest.version }}"
```
h0tk3y: This won't work either, for the same reason as the Groovy sample above. As far as I can see, our New Project wizard adds the workaround to the `settings.gradle` file and applies the plugin as `id("kotlin2js")` in Kotlin DSL, too.

pages/docs/reference/using-gradle.md
```diff
@@ -252,72 +357,128 @@ dependencies {
...
+}
+```
 
-If you target JavaScript, use `compile "org.jetbrains.kotlin:kotlin-stdlib-js"` instead.
+</div>
+</div>
 
-If you're targeting JDK 7 or JDK 8, you can use extended versions of the Kotlin standard library which contain
-additional extension functions for APIs added in new JDK versions. Instead of `kotlin-stdlib`, use one of the
+To run applications that use `kotlin-stdlib`, JDK 6 is required. On JDK 7, 8, or later, you can use extended versions of the Kotlin standard library which contain
```
h0tk3y: In fact, running those application requires a JRE, not JDK (the latter is used for building but is not required for running a Java application). Also, formally, the Android runtime is nether a JRE nor a JDK.
Using `kotlin-stdlib` with Java 7, 8 and above is fine, so saying that JDK 6 is required may be confusing.
Consider rewording this to something like:
> The Kotlin standard library `kotlin-stdlib` targets Java 6 and above. There are extended versions of the standard library that add support for some of the features of JDK 7 and JDK 8.

pages/docs/reference/using-gradle.md
```diff
@@ -252,72 +357,128 @@ dependencies {
...
+ ```
+ 
+</div>
+</div>
 
-## Annotation processing
+## Annotation Processing
 
-See the description of [Kotlin annotation processing tool](kapt.html) (`kapt`).
+Kotlin supports annonation processing via the _Kotlin annotation processing tool_(`kapt`). Usage of kapt with Gradle is described on the [kapt page](kapt.html).
```
h0tk3y: This link could contain an anchor to the section, like `[kapt page](kapt.html#gradle-build-cache-support-since-1220)` (not sure I correctly translated the title to an anchor, please check before adding it)

pages/docs/reference/using-gradle.md
```diff
@@ -210,6 +279,25 @@ apply plugin: 'com.android.application'
...
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+```kotlin
+buildscript {
+    dependencies {
+        classpath(kotlin("gradle-plugin", version = "{{ site.data.releases.latest.version }}'"))
+    }
+}
+plugins {
+    id("com.android.application")
```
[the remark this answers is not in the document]
h0tk3y: See the Gradle docs about the mechanism used for these workarounds: https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_resolution_rules

pages/docs/reference/using-gradle.md
h0tk3y: Here's a relevant issue: https://youtrack.jetbrains.com/issue/KT-20156

pages/docs/reference/kapt.md
p7nov: Strongly agree.

pages/docs/reference/using-gradle.md
p7nov: +

pages/docs/reference/using-gradle.md
p7nov: +

pages/docs/reference/using-gradle.md
p7nov: +

pages/docs/reference/using-gradle.md
p7nov: +

pages/docs/reference/using-gradle.md
p7nov: +

PR #1243, opened 2018-12-19
Clarify thread safety issue with collections

kszilagyi:
Clarify thread safety issue with collection operations and add link to comment.


kszilagyi:
I am happy to write up the explanation onto a separate page if that is preferred over the link to the comment.

yole:
At this point I'm not sure if this is worth it. I don't see any evidence that a significant number of users were confused by the current behavior of the functions, I don't like seeing links to github comments in the docs, and it's totally overkill to have a separate page explaining this gotcha (and not any other gotchas that people might run into at some point).

p7nov:
Thanks for the addition.
However, the new detailed documentation on collections is already published.

PR #1245, opened 2018-12-27
Clean up language for building multiplatform projects

zmarkan:
Unify the language explaining the same thing when describing what host OS-es are supported. 
It's "drier" but may help non-native English speakers as it reduces complexity.


h0tk3y:
Thanks!

PR #1246, opened 2019-01-02
Fix README to point to correct data folder.

Pugzy:
Currently the README has links to `_data` however this is actually the `data` folder.


yole:
Thanks!

PR #1254, opened 2019-01-18
Update object-declarations.md

bloodlee:
Remove the useless parentheses


bloodlee:
@yole Thanks for the explanation. It's clear now.

PR #1269, opened 2019-01-25
Update coroutines-basic-jvm.md

utahwithak:
You can't use `sumBy` because it deals purely with ints and results in overflow! With this big of numbers you need to work with longs.  The initial example while slow got it correct. Subsequent examples were all computationally incorrect.


ilya-g:
Fixed by PR #1296.

PR #1273, opened 2019-02-01
docs(functions): clarify implicit `this.` dispatch

softmoth:
Perhaps a full-blown description of function dispatch is called for, to be explicit about how various candidates are ordered (I assume it involves the number of parameters and how specific their types are, generics, etc.). But for now, this seems a decent place to call out this particular detail.


p7nov:
No answer from the author, so I'm adding my version in https://github.com/JetBrains/kotlin-web-site/pull/1748

PR #1278, opened 2019-02-07
Correct Groovy syntax for all tasks Compiler Options

pal-tgilbert:
"It is also possible to configure all Kotlin compilation tasks in the project" 
Specified Groovy syntax doesn't work  in Android Studio 3.3.1, while corrected version does

### DIFF
--- a/pages/docs/reference/using-gradle.md
+++ b/pages/docs/reference/using-gradle.md
@@ -583,7 +583,7 @@ It is also possible to configure all Kotlin compilation tasks in the project:
 <div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile::class.java).all {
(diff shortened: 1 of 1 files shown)


pages/docs/reference/using-gradle.md
```diff
@@ -583,7 +583,7 @@ It is also possible to configure all Kotlin compilation tasks in the project:
 <div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile::class.java).all {
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile.class).all {
```
pal-tgilbert: FYI: you can even remove the .class
However the one currently listed in the docs definitely doesn't work in Android Studio.

h0tk3y:
I've removed `.class`, too.
Thanks @pal-tgilbert!

PR #1292, opened 2019-02-12
Clarify 'Null Safety' examples

jinbeomhong:
- In the 'Checking for null in conditions' example, type of `b` is `String` by default. So `b != null` is always `true`. So, it will be better to specify the variable's type to `String?`:
- In the 'Safe Calls' example, it will be better to mention that a safe call on non-null receiver of type `String` is unnecessary.


yole:
Thanks!

PR #1295, opened 2019-02-13
small fix in comment

MarcoIeni:


yole:
The comment is correct: i is greater or equal than 1 and less or equal than 10.

PR #1296, opened 2019-02-17
Avoid sum overflow calculation.

sherifnosseir:
The current code does not return the correct answer when using coroutines.
The sumBy function returns an `Int`, however the answer does not fit in an Int resulting in a wrong value.
answer: 500000500000
current: 1784293664

https://www.wolframalpha.com/input/?i=sum+1..1000000


ilya-g:
Indeed, `toLong()` conversion after `await()` is required.

ilya-g:
I've added `toLong()` after `await()` and pushed it as be5365d92c4f6805da57eb7f556b4647b36fbbb6

PR #1299, opened 2019-02-19
Adds the Arrow library to OSS projects

MaureenElsberry:
Arrow is a Functional companion to the standard library - adds to OSS list of resources https://arrow-kt.io/. 

Thanks!


yole:
We're no longer maintaining this list of resources; it's going to be removed eventually.

PR #1306, opened 2019-02-26
Removed the space before the colon

tsayukov:
In according to [coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#colon) there are no cases where a space before a colon is needed.


h0tk3y:
Thanks!

PR #1310, opened 2019-03-01
Add proper terminology for PascalCase

rsheasby:


yole:
Could you please edit the text so that it does not have two clauses in parentheses one after the other?

rsheasby:
Done! I tend to think it reads better without the explanation of what PascalCase is so I removed it entirely. Most developers should already know what it is or be able to figure it out from the example.

yole:
I'd leave the explanation in. The Kotlin documentation is accessed by many novice developers who are many years younger than Pascal, so the term would be unfamiliar to them.

rsheasby:
Sure I'll rephrase and add it back. I'm sure they would be able to figure it out from the example but whichever you prefer.

rsheasby:
Done.

rsheasby:
Just wanted to follow up on this PR? Are there more changes that need to happen before it can be merged?

p7nov:
Thanks for your suggestion. 
However, we decided to use the term `camel case` throughout the docs to avoid introducing terms that may be unfamiliar.

Regards, Pavel
Kotlin Technical Writer.

PR #1314, opened 2019-03-05
Update return type of greeting()

akhilbojedla:
returning ResponseEntity being more relevant than retuning an object


yole:
How exactly is this more relevant? Your suggested code is longer and has exactly the same end result as the original one.

PR #1317, opened 2019-03-06
docs: scope functions

p7nov:
added documentation on scope functions.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -322,8 +322,6 @@ with(myTurtle) { //draw a 100 pix square
 ```
 </div>
 
-For more details about `with` and other similar functions, see [Scope Functions](scope-functions.html). 
-
 
 ### Java 7's try with resources
 
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -299,7 +299,7 @@ fun transform(color: String): Int = when (color) {
 ```
 </div>
 
-### Calling multiple methods on an object instance ('with')
+### Calling multiple methods on an object instance (`with`)
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
@@ -321,6 +321,9 @@ with(myTurtle) { //draw a 100 pix square
 }
 ```
 </div>
+
+For more details about `with` and other similar functions, see [Scope Functions](scope-functions.html). 
+
 
 ### Java 7's try with resources
 
--- a/pages/docs/reference/scope-functions.md
+++ b/pages/docs/reference/scope-functions.md
@@ -24,7 +24,7 @@ data class Person(var name: String, var age: Int, var city: String) {
 fun main() {
 //sampleStart
     Person("Alice", 20, "Amsterdam").let {
-        println("Calling run()")
+        println(it)
         it.moveTo("London")
         it.incrementAge()
         println(it)
@@ -46,7 +46,7 @@ data class Person(var name: String, var age: Int, var city: String) {
 fun main() {
 //sampleStart
     val alice = Person("Alice", 20, "Amsterdam")
-    println("Without run()")
+    println(alice)
     alice.moveTo("London")
     alice.incrementAge()
     println(alice)
@@ -67,7 +67,7 @@ Because the scope functions are all quite similar in nature, it's important to u
 
 ### Context object: `this` or `it`
 
-Each scope function uses one of two ways to access the context object in the passed lambda function: as a lambda receiver (`this`) or as a lambda argument (`it`). Both provide the same technical capabilities, so we'll describe the pros and cons of each for different cases and provide recommendations on their use.
+Inside the lambda of a scope function, the context object is available by a short reference instead of its actual name. Each scope function uses one of two ways to access the context object: as a lambda [receiver](lambdas.html#function-literals-with-receiver) (`this`) or as a lambda argument (`it`). Both provide the same capabilities, so we'll describe the pros and cons of each for different cases and provide recommendations on their use.
 
 <div class="sample" markdown="1" theme="idea">
 ```kotlin
@@ -89,7 +89,7 @@ fun main() {
 
 #### this
 
-`run`, `with`, and `apply` refer to the context object as a lambda receiver - by keyword `this`. Hence, in their lambdas, the object is available as it would be in ordinary class functions. In most cases, you can omit `this` when accessing the members of the receiver object, making the code shorter. On the other hand, if `this` is omitted, it can be hard to distinguish between the receiver members and external objects or functions. So, having the context object as a receiver (`this`) is recommended for lambdas that mainly operate the object members: call its functions or assign properties.
+`run`, `with`, and `apply` refer to the context object as a lambda receiver - by keyword `this`. Hence, in their lambdas, the object is available as it would be in ordinary class functions. In most cases, you can omit `this` when accessing the members of the receiver object, making the code shorter. On the other hand, if `this` is omitted, it can be hard to distinguish between the receiver members and external objects or functions. So, having the context object as a receiver (`this`) is recommended for lambdas that mainly operate on the object members: call its functions or assign properties.
 
 <div class="sample" markdown="1" theme="idea">
 ```kotlin
@@ -108,20 +108,23 @@ fun main() {
 
 #### it
 
-In turn, `let` and `also` have the context object as a lambda argument. If the argument name is not specified, the object is accessed by the implicit default name `it`. `it` is shorter than `this` and expressions with `it` are usually easier for reading. However, you can't omit the argument name, so there is no way to shorten calls to the object functions or properties. Hence, having the context object as `it` is better when the object is mostly used as an argument in function calls. `it` is also better if you operate multiple variables in the code block.
+In turn, `let` and `also` have the context object as a lambda argument. If the argument name is not specified, the object is accessed by the implicit default name `it`. `it` is shorter than `this` and expressions with `it` are usually easier for reading. However, when calling the object functions or properties you don't have the object available implicitly like `this`. Hence, having the context object as `it` is better when the object is mostly used as an argument in function calls. `it` is also better if you use multiple variables in the code block.
 
 <div class="sample" markdown="1" theme="idea">
 ```kotlin
-fun writeToLog(str: String) {}
-fun writeToFile(str: String) {}
+fun writeToLog(message: String) {
+    println("INFO: $message")
+}
 
 fun main() {
 //sampleStart
-    "Hello".also {
-        println("Also called on the string $it")
-        writeToLog(it)
-        writeToFile(it)
+    fun getRandomInt(): Int {
+        return Random.nextInt(100).also {
+            writeToLog("getRandomInt() generated value $it")
+        }
     }
+    
+    val i = getRandomInt()
 //sampleEnd
 }
 ```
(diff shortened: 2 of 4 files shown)


pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Return value
+
+Scope functions return different values:
+* `apply` and `also` return the context object.
+* `let`, `run`, and `with` return the lambda result.
+
+These two options let you choose the proper function depending on what you do next in your code.
+
+Having the object itself as a return value is useful for chaining other operations on it.
```
yole: Not really. Chained operations can be simply put inside `apply`. Object as the return value is useful for returning the object from the function in which a scoped function has been used.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+fun main() {
+//sampleStart
+    val numbers = mutableListOf("one", "two", "three", "four", "five")
+    numbers.map { it.length }.filter { it > 3 }.let(::println)
+//sampleEnd
+}
+```
+</div>
+
+`let` is often used for executing a code block only with non-null values. To perform actions on a non-null object, use the safe call operator `?` on it and call `let` with the actions in its lambda.
```
yole: This is actually the most common use of `let`, so it should be on top.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+        println("let() called on $it")        
+        processNonNullString(it)      // OK: the function is called only on non-nulls
+        it.length
+    }
+//sampleEnd
+}
+```
+</div>
+
+Another case for using `let` is introducing local variables for improving code readability. To define a variable, provide its name as the lambda argument so that it can be used instead of the default `it`.
```
yole: I believe this is an artificial example. To introduce a variable, you can simply declare it.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+```kotlin
+fun processNonNullString(str: String) {}
+
+fun main() {
+//sampleStart
+    val str: String? = "Hello"   
+    //processNonNullString(str)       // compilation error: str can be null
+    val length = str?.let { 
+        println("let() called on $it")        
+        processNonNullString(it)      // OK: the function is called only on non-nulls
```
yole: It's important to emphasize here that inside the lambda you should use `it`, not the original variable. Many people write code like `str?.let { processNonNullString(str) }`, and while this works, we need to discourage it.

pages/docs/reference/scope-functions.md
ilya-g: >  Chained operations can be simply put inside apply.

This isn't idiomatic. If there's a chain of operations, both `apply` and `also` should be inside that chain, not outside.

pages/docs/reference/scope-functions.md
ilya-g: I believe it became the most common even without us writing this page, so it won't harm to leave the narrative order going from basic usages to advanced. Otherwise `let` might get a reputation of a fancy `if (it != null)` replacement.

pages/docs/reference/scope-functions.md
ilya-g: But then that variable could get broader scope that required, and again break the chain of calls. I agree though that the example below doesn't show it in full.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+|Function|Object reference|Return value|Is extension function|
+|---|---|---|---|
+|`let`|`it`|Lambda result|Yes|
+|`run`|`this`|Lambda result|Yes|
+|`run`|-|Lambda result|No: called without the context object|
+|`with`|`this`|Lambda result|No: takes the context object as an argument.|
+|`apply`|`this`|Context object|Yes|
+|`also`|`it`|Context object|Yes|
+
+Here is a short guide for choosing scope functions depending on the intended purpose:
```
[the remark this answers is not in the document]
ilya-g: That section in coding conventions could already have external references to it, so we could remove its content and leave only a link to here.

pages/docs/reference/scope-functions.md
p7nov: Added more clarification to the start of the section **Context object: this or it**. I think it's a better place than the description of a particular function.

pages/docs/reference/scope-functions.md
p7nov: Broadened the description, replaced the example.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,530 @@
...
+In turn, `let` and `also` have the context object as a lambda argument. If the argument name is not specified, the object is accessed by the implicit default name `it`. `it` is shorter than `this` and expressions with `it` are usually easier for reading. However, you can't omit the argument name, so there is no way to shorten calls to the object functions or properties. Hence, having the context object as `it` is better when the object is mostly used as an argument in function calls. `it` is also better if you operate multiple variables in the code block.
+
+<div class="sample" markdown="1" theme="idea">
+```kotlin
+fun writeToLog(str: String) {}
+fun writeToFile(str: String) {}
+
+fun main() {
+//sampleStart
+    "Hello".also {
```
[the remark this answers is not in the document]
p7nov: Replaced the example

pages/docs/reference/scope-functions.md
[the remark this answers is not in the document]
p7nov: I'd remove the guidance from **Coding conventions** in this case. The guidance in the new page is shorter (which is better if you know the functions) and there are detailed descriptions on the same page (which is good if you don't know the details or need to refresh).
In **Coding conventions**, I'd leave the heading about scope functions and a short introductory word with the link to **Scope functions**. What do you say?

pages/docs/reference/idioms.md
```diff
@@ -322,6 +322,9 @@ with(myTurtle) { //draw a 100 pix square
 ```
 </div>
 
+For more details about `with` and other similar functions, see [Scope Functions](scope-functions.html).
```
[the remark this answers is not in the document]
p7nov: Agree. Adding links for all idioms is out of this PR's scope, so I'd remove this link for now.

pages/docs/reference/scope-functions.md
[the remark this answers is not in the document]
ilya-g: Seems fine.

pages/docs/reference/scope-functions.md
p7nov: Added another example on using also when returning the context object from a function.

pages/docs/reference/scope-functions.md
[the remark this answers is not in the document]
p7nov: Replaced the old guidance in **Coding conventions** with the link to the new doc.

p7nov:
@yole @ilya-g guys, please have another look at the updated PR.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,576 @@
...
+data class Person(var name: String, var age: Int = 0, var city: String = "")
+
+fun main() {
+//sampleStart
+    val isTeenager = Person("Adam").run {
+        age = 32
+        city = "London"
+        age > 12 && age < 20
+    }
+    println("Is Adam a teenager? $isTeenager")
```
ilya-g: I can't help but feel that this `run` sample is quite contrived compared to its original use case from [KT-5235](https://youtrack.jetbrains.com/issue/KT-5235). I'm tempted to rewrite it with `apply` for setting the properties + `run` for evaluating the expression, but `apply` has not been introduced at this moment.

Perhaps this one could illustrate it better:
```
    class MultiportService(var url: String, var port: Int) {
        fun prepareRequest(): String = "Default request"
        fun query(request: String): String = "Result for query '$request'"
    }
    
// start
    val service = MultiportService("https://example.kotlinlang.org", 80)
      
    val result = service.run {
        port = 8080
        query(prepareRequest() + ", and some extra")
    }
    // an alternative with 'let' could be shown
// end
    println(result)
```
The idea here is the necessity to call several methods of a context object that depend on each other's result or the state of object they mutate.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,576 @@
...
+
+The scope functions differ by the result they return:
+* `apply` and `also` return the context object.
+* `let`, `run`, and `with` return the lambda result.
+
+These two options let you choose the proper function depending on what you do next in your code.
+
+#### Context object 
+
+The return value of `apply` and `also` is the context object itself. Hence, they can be used as steps in chains of function calls on the object.
```
ilya-g: It is not clear, what's different from `let` and `run` that can also be used in the middle of call chains. Perhaps, we can call them "side steps" (not sure, maybe "tee steps", or "branch steps"), because they return the same context object on which we can continue chaining calls.

p7nov: Updated the description

pages/docs/reference/scope-functions.md
p7nov: Thanks, I've replaced the example with this (with small modifications and the alternative with `let`).

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,587 @@
...
+        add("five")
+        filter { it.endsWith("e") }.size        
+    }
+    println("There are $resultSize elements that end with e.")
+//sampleEnd
+}
+```
+</div>
+
+Additionally, you can ignore the return value and use a scope function for creating a temporary scope.
```
ilya-g: In the sample below the result is not ignored.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,587 @@
...
+```kotlin
+fun processNonNullString(str: String) {}
+
+fun main() {
+//sampleStart
+    val str: String? = "Hello"   
+    //processNonNullString(str)       // compilation error: str can be null
+    val length = str?.let { 
+        println("let() called on $it")        
+        processNonNullString(it)      // OK: the function is called only on non-nulls
```
ilya-g:
```suggestion
        processNonNullString(it)      // OK: 'it' is not null inside '?.let { }'
```
Otherwise it's unclear which function was meant: `processNonNullString` or `let`.

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,586 @@
...
+
+The Kotlin standard library contains several functions whose sole purpose is to execute a block of code within the context of an object. When you call such a function on an object with a [lambda expression](lambdas.html) provided, it forms a temporary scope. In this scope, you can access the object without its name. Such functions are called _scope functions_. There are five of them: `let`, `run`, `with`, `apply`, and `also`.
+
+Basically, these functions do the same: execute a block of code on an object. What's different is how this object becomes available inside the block and what is the result of the whole expression.
+
+Here's a typical usage of a scope function:
+
+
+<div class="sample" markdown="1" theme="idea">
+```kotlin
```
AlexanderPrendota: \n

pages/docs/reference/scope-functions.md
```diff
@@ -0,0 +1,586 @@
...
+//sampleStart
+    Person("Alice", 20, "Amsterdam").let {
+        println(it)
+        it.moveTo("London")
+        it.incrementAge()
+        println(it)
+    }
+//sampleEnd
+}
+```
```
AlexanderPrendota: \n

PR #1319, opened 2019-03-07
docs: Doc on @Experimental

p7nov:
Added documentation on @Experimental and @useExperimental annotations.

### DIFF
--- a/pages/docs/reference/experimental.md
+++ b/pages/docs/reference/experimental.md
@@ -100,7 +100,7 @@ To use an experimental API in all functions and classes in a file, add the file-
 ### Module-wide use
 
 If you don't want to annotate every usage of experimental APIs in your code, you can accept the experimental status for your whole module. Module-wide use of experimental APIs can be propagating and non-propagating as well:
-* To accept the experimental status without propagation, compile the module with the argument `Xuse-experimental`, specifying the fully qualified name of the experimental API marker you use: `-Xuse-experimental=org.mylibrary.ExperimentalMarker`. Compiling with this argument has the same effect as if every declaration in the module having the annotation`@UseExperimental(ExperimentalMarker::class)`.
+* To accept the experimental status without propagation, compile the module with the argument `Xuse-experimental`, specifying the fully qualified name of the experimental API marker you use: `-Xuse-experimental=org.mylibrary.ExperimentalMarker`. Compiling with this argument has the same effect as if every declaration in the module had the annotation`@UseExperimental(ExperimentalMarker::class)`.
 * To accept and propagate the experimental status to your whole module, compile the module with the argument `-Xexperimental=org.mylibrary.ExperimentalMarker`. In this case, _every declaration_ in the module becomes experimental. The use of the module requires the acceptance of its experimental status as well.
 
 If you build your module with Gradle, you can add arguments like this:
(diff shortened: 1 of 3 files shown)


pages/docs/reference/experimental.md
```diff
@@ -0,0 +1,232 @@
...
+ ```kotlin
+ //client code
+ @file:UseExperimental(ExperimentalDateTime::class)
+ ```
+ </div>
+
+### Module-wide use
+
+If you don't want to annotate every usage of experimental APIs in your code, you can accept the experimental status for your whole module. Module-wide use of experimental APIs can be propagating and non-propagating as well:
+* To accept the experimental status without propagation, compile the module with the argument `Xuse-experimental`, specifying the fully qualified name of the experimental API marker you use: `-Xuse-experimental=org.mylibrary.ExperimentalMarker`. Compiling with this argument has the same effect as if every declaration in the module having the annotation`@UseExperimental(ExperimentalMarker::class)`.
```
udalov: > as if every declaration in the module having the annotation

having -> had

pages/docs/reference/experimental.md
```diff
@@ -0,0 +1,232 @@
...
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.3</version>
+        <configuration>
+            <source>1.8</source>
+            <target>1.8</target>
+            <forceJavacCompilerUse>true</forceJavacCompilerUse>
+            <fork>true</fork>
+            <compilerArgs>
+                <arg>-Xuse-experimental=org.mylibrary.ExperimentalMarker</arg>
+            </compilerArgs>
```
ilya-g: It looks strange that our compiler options are specified for Java compiler plugin. Does it really work?

pages/docs/reference/experimental.md
```diff
@@ -0,0 +1,232 @@
...
+<div class="multi-language-sample" data-lang="kotlin">
+<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+
+```kotlin
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+// ...
+
+val compileKotlin: KotlinCompile by tasks
+
+compileKotlin.kotlinOptions.freeCompilerArgs += "-Xuse-experimental=org.mylibrary.ExperimentalMarker"
```
ilya-g: I'm not sure that this is an idiomatic way to specify compiler arguments in Gradle Kotlin script. 
@h0tk3y Do you know a better way?

pages/docs/reference/experimental.md
```diff
@@ -0,0 +1,232 @@
...
+</div>
+
+Experimental marker annotations must meet several requirements:
+* `BINARY` [retention](/api/latest/jvm/stdlib/kotlin.annotation/-annotation-retention/index.html)
+* No `EXPRESSION` and `FILE` among [targets](/api/latest/jvm/stdlib/kotlin.annotation/-annotation-target/index.html)
+* No parameters.
+
+A marker annotation can have one of two severity [levels](/api/latest/jvm/stdlib/kotlin/-experimental/-level/index.html) of informing about experimental API usage:
+* `Experimental.Level.ERROR`. Acceptance is required. Otherwise, the code that uses marked API won't compile. This level is used by default.
+* `Experimental.Level.WARNING`. Acceptance is not required. Without it, the compiler raises a warning.
```
ilya-g: "is not required, but advisable", otherwise it looks that it's ok to omit opt-in, while warnings show it's not.
Also "mandatory" could be used instead of "required".

pages/docs/reference/experimental.md
```diff
@@ -0,0 +1,232 @@
...
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+```kotlin
+@Experimental(level = Experimental.Level.WARNING)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+annotation class ExperimentalDateTime
+```
+</div>
+
+If you publish several features in the experimental state, declare a marker for each. Separate markers make the use of experimental features safer for your clients: they'll be able to use only the features that they explicitly accept.
```
ilya-g: Also it makes it easier for you to graduate these features to stable independently.

pages/docs/reference/experimental.md
ilya-g: It'd be ideal to provide a guidance for experimental API developers how to choose between `Level.ERROR` and `Level.WARNING` for their annotations. Do we know some rationale for this?

pages/docs/reference/experimental.md
p7nov: My bad: wrong plugin. Fixed.

pages/docs/reference/experimental.md
p7nov: As discussed with @udalov, there are no general rules or good real-life examples of using `Level.WARNING`. I think it's OK to leave the choice to readers.

pages/docs/reference/experimental.md
h0tk3y: Well, this is more or less the idiomatic way. An alternative that might looks arguably better as it doesn't require declaring the variable is:
```
tasks.getByName("compileKotlin") {
    kotlinOptions { /* ... */ }
}
```
We'll provide a way to access `kotlinOptions` through the *target – compilation* model soon.

p7nov: Updated the example slightly differently (also approved by @h0tk3y :)

PR #1320, opened 2019-03-07
Added "Swapping two variables" idiom

elizarov:


pages/docs/reference/idioms.md
```diff
@@ -358,3 +358,13 @@ if (b == true) {
 }
 ```
 </div>
+
+### Swapping two variables
+
+```kotlin
```
AlexanderPrendota: add \n after div

elizarov: After which div? I've added one more to the end of file. I don't see line breaks being 
used anywhere else in this file.

ilya-g:
This looks like a clever hack rather than idiom 🤓 

I'd rather prefer to have something like `swap(::a, :b)` inline function in stdlib.

elizarov:
I'd prefere it, too, but it would not really inline, as far as I can tell, and it will not work with array elements.

elizarov:
On a side note, all idioms look like clever hacks until you start using them and learn to recognize them :)

PR #1321, opened 2019-03-07
Concretize finally block occurrence in Exceptions

okylymnyk:
As `finally` can have zero or one block, it is recommended to specify the same here.


p7nov:
Thanks!

PR #1324, opened 2019-03-07
Extension Function Page: Correct the method param types.

birajpatel:
The types of input params should be `T` and not `Int`


yole:
The parameter of `swap` is the index in the list. The type of an index parameter in a list is `Int`.

PR #1326, opened 2019-03-10
Update mpp-ios-android.md

queiroz:
Fixed case sensitive "iOS" to "ios" as stated at "We create a similar file for the iOS target in the `SharedCode/src/iosMain/kotlin/actual.kt`" and to avoid a common mistake when adding dependencies to sourceSets as Groovy is case sensitive. `sourceSets { iosMain.dependencies { } }`.


p7nov:
Thanks!

PR #1327, opened 2019-03-11
Fix usage of unary and binary in operator overloading intro

blaugold:


yole:
The usage is correct. Unary operations do not have a left-hand-side type, they have just one type. Binary operators have two operand types, and it's necessary to specify for which one the extension function is defined.

PR #1346, opened 2019-03-28
Update object-declarations.md to reflect relaxed restriction on final keyword since Java 8

benelog:
Since Java 8,  'final' keyword is not mandatory for variables accessed from anonymous inner classes when variables are effectively final.


yole:
Thanks!

PR #1360, opened 2019-04-17
Add note to example for backing fields

webleaf:


yole:
The difference between `val` and `var` is one of the key concepts in Kotlin, and it's not clear why it makes sense to remind about it in this specific place.

webleaf:
@yole This documentation is mostly read by those who are still not familiar with the language. Keywords `val` and `var` are very similar. So, just not paying attention to it, I spent a lot of time trying to understand why in this example no "default implementation of at least one of the accessors".
Moreover, this is a key concept...

yole:
So do you actually expect every documentation page to describe all the key concepts of the language?

webleaf:
@yole of course no. I suggested this note, because this note (in this place) would save me time and I decide that it might be useful to someone else.
Yes, this is exactly the page "Properties and Fields", which begins with a description of the difference between `var` and `val`.
But despite this, I still stuck in place, which I propose to comment. Especially, this note on the main topic of the page.

PR #1361, opened 2019-04-22
Update mpp-ios-android.md with consistent source paths

projektir:
Making the source folders consistent


p7nov:
Thanks for the suggestion.
The tutorial is already replaced with a Hands-On lab.

PR #1367, opened 2019-04-25
Idiom to check if collection contains an element

arekolek:
Using `in` instead of `contains`

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -58,12 +58,8 @@ val positives = list.filter { it > 0 }
 val emails = ... // can be a list, set, map, sequence
 
 if ("john@example.com" in emails) { ... }
-    ...
-}
 
 if ("jane@example.com" !in emails) { ... }
-    ...
-}
 ```
 </div>
 
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -51,17 +51,17 @@ val positives = list.filter { it > 0 }
 ```
 </div>
 
-### Check if collection contains an element
+### Checking element presence in a collection.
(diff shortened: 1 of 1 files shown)


pages/docs/reference/idioms.md
```diff
@@ -51,6 +51,22 @@ val positives = list.filter { it > 0 }
 ```
 </div>
 
+### Check if collection contains an element
```
p7nov:
```suggestion
### Checking element presence in a collection.
```

pages/docs/reference/idioms.md
```diff
@@ -51,6 +51,22 @@ val positives = list.filter { it > 0 }
...
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+```kotlin
+val emails = ... // can be a list, set, map, sequence
```
p7nov: No need to declare the list in this example, just use a self-explaining name in the lines below, like _emailsList_

pages/docs/reference/idioms.md
```diff
@@ -51,6 +51,22 @@ val positives = list.filter { it > 0 }
...
+
+if ("john@example.com" in emails) {
```
p7nov:
```suggestion
if ("john@example.com" in emails) { ... }
```

pages/docs/reference/idioms.md
```diff
@@ -51,6 +51,22 @@ val positives = list.filter { it > 0 }
...
+    ...
+}
+
+if ("jane@example.com" !in emails) {
```
p7nov:
```suggestion
if ("jane@example.com" !in emails) { ... }
```

PR #1370, opened 2019-04-26
Update apple-framework.md - grammatic error

MoeHamoud:
In "There are some small differences, though. In Kotlin *any object has the only one instance.*"


yole:
Thanks!

PR #1372, opened 2019-04-30
remove "disable bitcode" instructions from Kotlin/Native iOS+Android tutorial

gimlids:
The purpose of this PR is twofold:

1. Find out for myself whether Kotlin/Native supports LLVM bitcode output for Apple store submission.

2.
    a. If bitcode is supported, remove this scary instruction that suggests bitcode is not supported, rendering Kotlin/Native very very risky to ever use in an iOS project. Apple may one day require bitcode for iOS apps (they already do for tvOS and watchOS apps).

    b. Perhaps this instruction can be augmented or replaced with https://github.com/JetBrains/kotlin-native/blob/master/FAQ.md#q-how-do-i-enable-bitcode-for-my-kotlin-framework, if there is a problem with removing this section outright.


jonnyzzz:
Thank you for the fixes!

PR #1378, opened 2019-05-05
Update Android Section of using-gradle.md

AndrewReitz:
The current docs will not work since you cannot use buildscript block and plugins block together. 
Updated to show how it can be setup for both Gradle and Kotlin build script files.

### DIFF
--- a/pages/docs/reference/using-gradle.md
+++ b/pages/docs/reference/using-gradle.md
@@ -300,10 +300,8 @@ buildscript {
     }
 }
 
-plugins {
-    id 'com.android.application'
-    id 'kotlin-android'
-}
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
 ```
 
 </div>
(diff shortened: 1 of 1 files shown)


pages/docs/reference/using-gradle.md
```diff
@@ -313,12 +311,27 @@ plugins {
...
+        google()
+    }
+
+    resolutionStrategy {
+        eachPlugin {
+            if (requested.id.id in setOf("kotlin-android", "kotlin-kapt", "kotlin-android-extensions")) {
+                useModule("org.jetbrains.kotlin:kotlin-gradle-plugin:${requested.version}")
+            }
+
+            if (requested.id.id in setOf("com.android.application", "com.android.feature")) {
```
p7nov, line 327: Why not to use `if(requested.id.namespace == "com.android") {` as in the Gradle example here: https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:plugins_resolution_strategy ?
And similar for Kotlin plugins.

AndrewReitz: That would also work. I did this back when this wasn't documented and had to dig through source code, so there is no reason other than this is what I saw gradle doing at that time.

tprochazka: Why such a complex solution with custom resolutionStrategy?

In Kotlin normally works just
```
    kotlin("android") version("1.3.72")
    kotlin("android.extensions") version("1.3.72")
```

And in Groovy
```
    id "org.jetbrains.kotlin.android" version('1.3.72')
    id "org.jetbrains.kotlin.android.extensions" version('1.3.72')

```

AndrewReitz: This is for loading the Kotlin extensions it's for loading the android plugin which isn't available in the Gradle plugins portal, and I don't believe the kotlin ones will load the way you suggested either due to how the ids and plugins are named.

tprochazka: You have right, this works for me
```
plugins {
    id 'com.android.application'
    id "org.jetbrains.kotlin.android"
}

```must because I have this in the root gradle file

```
buildscript {
    ext.kotlin_version = "1.3.72"
    repositories {
        google()
        jcenter()
    }
    dependencies {
        classpath "com.android.tools.build:gradle:4.1.0-beta03"
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
    }
}

```

p7nov:
Folks, sorry for the long waiting.
Now we consider [Google docs](https://developer.android.com/) the source of truth for all Android-related topics and don't duplicate their content on the Kotlin website. Thus, we don't publish instructions on setting up build scripts for Android projects (see [here](https://kotlinlang.org/docs/reference/using-gradle.html#targeting-android)).

PR #1379, opened 2019-05-06
Typo fix in description

muthuraj57:


p7nov:
Thanks for the suggestion.
However, the original text is more accurate: we use frameworks in Kotlin projects, not vice versa.

PR #1384, opened 2019-05-13
Remove footnote about tailrec functions in JVM backend

serebit:
The footnote claims that tailrec functions are only available in the JVM backend, but this is no longer the case.


p7nov:
Thanks for noticing. However, tailrec is still not supported by all platforms, so we'll just add Native support to this sentence.

PR #1386, opened 2019-05-15
Update sam-with-receiver section

nglorioso:
Matching the groovy configuration to the example and Maven configuration


p7nov:
Thanks!

PR #1392, opened 2019-05-23
Corrected comment

happyherp:
In the comment it says that the range will go to 10. But the code says it should go to 4. I changed the comment.


AlexanderPrendota:
thanks!

PR #1398, opened 2019-06-02
Specify the referred compiler and clarify -d option.

nuno-andre:


p7nov:
Thanks for the addition!

PR #1402, opened 2019-06-05
Using dukat instead of ts2kt in examples for converting typescript de…

Schahen:
Today we've released dukat version good enough to substitute ts2kt and we'll introduce further improvements only to dukat. That said, docs should be updated.

### DIFF
--- a/pages/docs/tutorials/javascript/working-with-javascript.md
+++ b/pages/docs/tutorials/javascript/working-with-javascript.md
@@ -104,7 +104,7 @@ The standard library provides us with a series of wrappers around DOM as well as
 when we want to use a library such as jQuery? Kotlin does not have its own "header" files for all the different libraries available on the JavaScript ecosystem
 however, TypeScript does. The [Definitely Typed repository](https://github.com/DefinitelyTyped/DefinitelyTyped/)  provides us with a very large selection of header files. 
 
-Using the tool `dukat` (TypeScript to Kotlin) we can convert any `d.ts` files to Kotlin. To install the tool we can use `npm`
+Using the `dukat` tool we can convert any TypeScript declaration files to Kotlin. To install the tool we can use `npm`
 
 ```bash
 npm -g install dukat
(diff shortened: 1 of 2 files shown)


pages/docs/tutorials/javascript/working-with-javascript.md
```diff
@@ -98,23 +98,23 @@ fun printHello() {
...
 
-## Using ts2kt to generate header files for Kotlin
+## Using dukat to generate header files for Kotlin
 
 The standard library provides us with a series of wrappers around DOM as well as functions to work with JavaScript, using static typing. What happens however
 when we want to use a library such as jQuery? Kotlin does not have its own "header" files for all the different libraries available on the JavaScript ecosystem
 however, TypeScript does. The [Definitely Typed repository](https://github.com/DefinitelyTyped/DefinitelyTyped/)  provides us with a very large selection of header files. 
 
-Using the tool `ts2kt` (TypeScript to Kotlin) we can convert any `d.ts` files to Kotlin. To install the tool we can use `npm`
+Using the tool `dukat` (TypeScript to Kotlin) we can convert any `d.ts` files to Kotlin. To install the tool we can use `npm`
```
bashor: Maybe rephrase it somehow like "Using the tool `dukat` we can convert any `d.ts` files with TypeScript definitions to Kotlin. ..."?

Schahen: Well, to be stricter  they are called (in typescript code / API I mean) - TypeScript declaration files - which are precisely one that end with "*.d.ts" - so, may be:

> Using `dukat` tool we can convert any TypeScript declaration files to Kotlin

It's even shorter )

bashor: LGTM

epabst: For correct grammer, I think it should be "Using the dukat tool we can convert any TypeScript declaration files to Kotlin."

PR #1403, opened 2019-06-10
adding a tip for those using `implementation`

sinadarvi:
if you use `implementation` in `dependencies` section the code above will not generate any `.js` file from `.jar` files, so we can just modify our task so it will consider them for the task job


p7nov:
The tutorials on Kotlin/JS were completely rewritten, so this isn't relevant now.
Thank you

PR #1411, opened 2019-06-14
Fix maven-assembly-plugin config error

XQDD:


p7nov:
Hey, thanks for the suggestion.
However, the current configuration is correct as well, so we'll leave it as is.

PR #1417, opened 2019-06-30
Fix typo

cdlans:
Change
> `List<? super String>` a **supertype** of `List<Object>`
to
> `List<? super String>` a **supertype** of `List<String>`


p7nov:
Thank for your suggestion.
However, the current doc describes the PECS (Producer Extends Consumer Super) principle correctly.
You can find a good illustration of the principle here: https://stackoverflow.com/questions/2723397/what-is-pecs-producer-extends-consumer-super

PR #1418, opened 2019-07-01
fix: added info about stateless lambdas optimization (KT-32067)

p7nov:
Added a note about creating stateless lambdas as singletons per requests from YT and Discourse:
https://youtrack.jetbrains.com/issue/KT-32067
https://discuss.kotlinlang.org/t/lambdas-and-implicit-references-to-the-instance-of-the-enclosing-class/

### DIFF
--- a/pages/docs/reference/lambdas.md
+++ b/pages/docs/reference/lambdas.md
@@ -431,7 +431,7 @@ print(sum)
 </div>
 
 For lambdas that don't capture any variables (also know as _stateless_ lambdas), there is a built-in optimization:
-the compiler creates such lambdas as singleton classes. All references to them use the same instance of this class. 
+the compiler creates only one instance for each such lambda. So, all references to a stateless lambda use the same instance. 
 
 ### Function literals with receiver
 
--- a/pages/docs/reference/lambdas.md
+++ b/pages/docs/reference/lambdas.md
@@ -430,6 +430,9 @@ print(sum)
 
 </div>
 
+For lambdas that don't capture any variables (also know as _stateless_ lambdas), there is a built-in optimization:
+the compiler creates such lambdas as singleton classes. All references to them use the same instance of this class. 
+
 ### Function literals with receiver
 
 [Function types](#function-types) with receiver, such as `A.(B) -> C`, can be instantiated with a special form of function literals – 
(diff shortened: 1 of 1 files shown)


ilya-g:
I'd say that only one instance of that lambda is created without going into details that it's compiled to a singleton class.

pages/docs/reference/lambdas.md
```diff
@@ -430,6 +430,21 @@ print(sum)
...
 
+For lambdas that don't capture any variables (also know as _stateless_ lambdas), there is a built-in optimization:
+the compiler creates only one instance for each such lambda. So, all instances of a stateless lambda point to the same instance.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+ 
+ ```kotlin
+ val lambdaProvider = {
+    { Unit }  // returns a stateless lambda
+}
```
ilya-g: The example with a lambda stored in a property and returning another Unit-returning lambda is rather complicated to illustrate the concept of stateless lambda.

An alternative could be a function returning either a stateless or stateful lambda:
```kotlin
fun searchPredicate(expected: String? = null): (String?) -> Boolean {
    if (expected == null)
        return { value -> value == null } // stateless
    else
        return { value -> value == expected } // capturing
}
```

pages/docs/reference/lambdas.md
```diff
@@ -430,6 +430,21 @@ print(sum)
 
 </div>
 
+For lambdas that don't capture any variables (also know as _stateless_ lambdas), there is a built-in optimization:
+the compiler creates only one instance for each such lambda. So, all instances of a stateless lambda point to the same instance.
```
ilya-g: > all instances of a stateless lambda point to the same instance.

This confuses the matter: on one hand there is the single instance, on the other—there are many ('all') of them.

p7nov:
The described optimization appears to be a platform-specific optimization and is not a part of the design. So, we shouldn't mention it in the common public doc.

udalov:
Why not just add "On JVM, ..."?

p7nov:
As discussed with authors of the Kotlin language specification (https://github.com/Kotlin/kotlin-spec), this is an implementation detail that can be changed and is not guaranteed in the future. 
Currently, we've added the description of this optimization to the spec as a platform-specific note: https://github.com/Kotlin/kotlin-spec/blob/master/docs/kotlin.jvm/lambda-expressions.md (not sure where they will put it, now it's just a new file with this note).

PR #1419, opened 2019-07-01
KT-31748 Fixed example and updated coroutines version

vanyochek:
https://youtrack.jetbrains.com/issue/KT-31748

It would be nice if coroutines library has a variable for the last version like Kotlin version has (or maybe I didn't find it).
```{{ site.data.releases.latest.version }} ```


p7nov:
Thanks, a similar fix is already merged.

PR #1424, opened 2019-07-07
feat: add gradle multiplatform experimental syntax

matejdro:
At the moment, page only mentions the "normal" gradle project syntax, which is not compatible with multiplatform projects.

### DIFF
--- a/pages/docs/reference/experimental.md
+++ b/pages/docs/reference/experimental.md
@@ -143,6 +143,38 @@ tasks.withType<KotlinCompile>().all {
 </div>
 </div>
 
+If your Gradle module is a multiplatform module, use the `useExperimentalAnnotation` method:
+
+<div class="multi-language-sample" data-lang="groovy">
+<div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
+
+```groovy
+sourceSets {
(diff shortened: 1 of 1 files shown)


pages/docs/reference/experimental.md
```diff
@@ -140,6 +140,18 @@ tasks.withType<KotlinCompile>().all {
 }
 ```
 
+If your Gradle module is multiplatform module, you have to use `useExperimentalAnnotation` method:
```
p7nov:
```suggestion
If your Gradle module is a multiplatform module, use the `useExperimentalAnnotation` method:
```

pages/docs/reference/experimental.md
```diff
@@ -140,6 +140,18 @@ tasks.withType<KotlinCompile>().all {
...
+
+```groovy
+sourceSets {
```
p7nov: Please add a multilanguage sample (as in the example above) with all `<div>`s and language names. 
In Kotlin DSL, use the `"kotlin.Experimental"` instead of `'kotlin.Experimental'` (double quotation marks instead of single).

matejdro:
Sorry for the delay. I have made requested changes.

PR #1430, opened 2019-07-16
Delegated properties: Missing import; Lazy on the same Thread

peter-kehl:
Clarified requirements for non-synchronized `by lazy {...}`


p7nov:
Thanks!

PR #1431, opened 2019-07-17
Update building-mpp-with-gradle.md

msink:
Added 32-bit Windows target


p7nov:
Thanks for the addition!

PR #1434, opened 2019-07-23
chore: remove unneeded comparisons with java 

p7nov:
Removed references to Java from common doc pages. 
Added content where the descriptions were not enough
Replaced unreadable examples with foo(), bar(), etc.

This is the first part of the large task of removing Java mentions and comparisons from the common Kotlin doc.

### DIFF
--- a/pages/docs/reference/basic-syntax.md
+++ b/pages/docs/reference/basic-syntax.md
@@ -7,7 +7,8 @@ title: "Basic Syntax"
 
 # Basic Syntax
 
-## Package definition and imports <a name="defining-packages"></a>
+<a name="defining-packages"></a>
+## Package definition and imports
 
 Package specification should be at the top of the source file:
 
@@ -41,7 +42,8 @@ fun main() {
 
 </div>
 
-## Functions <a name="defining-functions"></a>
+<a name="defining-functions"></a>
+## Functions
 
 Function having two `Int` parameters with `Int` return type:
 
@@ -116,7 +118,8 @@ fun main() {
 
 See [Functions](functions.html).
 
-## Variables <a name="defining-variables"></a>
+<a name="defining-variables"></a>
+## Variables
 
 Read-only local variables are defined using the keyword `val`. They can be assigned a value only once.
 
@@ -208,7 +211,8 @@ and ends here. */
 
 See [Documenting Kotlin Code](kotlin-doc.html) for information on the documentation comment syntax.
 
-## String templates <a name="using-string-templates"></a>
+<a name="using-string-templates"></a>
+## String templates
 
 <div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
 
@@ -231,7 +235,8 @@ fun main() {
 
 See [String templates](basic-types.html#string-templates) for details.
 
-## Conditional expressions <a name="using-conditional-expressions"></a>
+<a name="using-conditional-expressions"></a>
+## Conditional expressions
 
 <div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
 
@@ -272,7 +277,8 @@ fun main() {
 
 See [*if*{: .keyword }-expressions](control-flow.html#if-expression).
 
-## Nullable values and *null*{: .keyword } checks <a name="using-nullable-values-and-checking-for-null"></a>
+<a name="using-nullable-values-and-checking-for-null"></a>
+## Nullable values and *null*{: .keyword } checks
 
 A reference must be explicitly marked as nullable when *null*{: .keyword } value is possible.
 
@@ -364,7 +370,8 @@ fun main() {
 
 See [Null-safety](null-safety.html).
 
-## Type checks and automatic casts <a name="using-type-checks-and-automatic-casts"></a>
+<a name="using-type-checks-and-automatic-casts"></a>
+## Type checks and automatic casts
 
 The *is*{: .keyword } operator checks if an expression is an instance of a type.
 If an immutable local variable or property is checked for a specific type, there's no need to cast it explicitly:
@@ -456,7 +463,8 @@ fun main() {
 
 See [Classes](classes.html) and [Type casts](typecasts.html).
 
-## `for` loop <a name="using-a-for-loop"></a>
+<a name="using-a-for-loop"></a>
+## `for` loop
 
 <div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
 
@@ -493,7 +501,8 @@ fun main() {
 
 See [for loop](control-flow.html#for-loops).
 
-## `while` loop <a name="using-a-while-loop"></a>
+<a name="using-a-while-loop"></a>
+## `while` loop
 
 <div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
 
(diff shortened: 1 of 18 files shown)


pages/docs/reference/basic-types.md
```diff
@@ -14,20 +14,74 @@ In this section we describe the basic types used in Kotlin: numbers, characters,
...
+
+```kotlin
+val pi = 3.14 //Double
+val e = 2.7182818284 //Double
+val eFloat = 2.7182818284f //Float, actual value is 2.7182817
+```
+
+</div>
+
+Note that unlike some other languages, there are no implicit widening conversions for numbers in Kotlin.
```
dsavvinov: Shouldn't we refer to `Explicit Conversions` section here, or explain in some other way how to workaround this issue?

pages/docs/reference/classes.md
```diff
@@ -295,72 +294,77 @@ class MyView : View {
...
 
 ```kotlin
-interface Foo {
-    val count: Int
+interface Department {
+    val employeesCount: Int
 }
 
-class Bar1(override val count: Int) : Foo
+class ITService(override val count: Int) : Department //constant number of employees
```
dsavvinov: Minor: i'm not sure if it is in the Kotlin code-style, but we usually insert space between `//` and comment (so, it would be `// constant number of employees`)

pages/docs/reference/functions.md
```diff
@@ -158,14 +171,15 @@ When a function is called with both positional and named arguments, all the posi
...
 ```kotlin
-fun foo(vararg strings: String) { ... }
+fun foo(vararg strings: String) { /*...*/ }
 
 ```
 </div>
 
-Note that the named argument syntax cannot be used when calling Java functions, because Java bytecode does not
+Note that the named argument syntax cannot be used on the JVM when calling Java functions because Java bytecode does not
```
abreslav: This is another platform-specific note, see my comment above

pages/docs/reference/nested-classes.md
```diff
@@ -62,7 +62,7 @@ window.addMouseListener(object : MouseAdapter() {
 
 </div>
 
-If the object is an instance of a functional Java interface (i.e. a Java interface with a single abstract method),
+_Note_: on the JVM, if the object is an instance of an functional Java interface (i.e. a Java interface with a single abstract method),
```
abreslav:
```suggestion
_Note_: on the JVM, if the object is an instance of a functional Java interface (i.e. a Java interface with a single abstract method),
```

abreslav: Another platform-specific note

pages/docs/reference/packages.md
```diff
@@ -10,20 +10,22 @@ title: "Packages and Imports"
...
 ```
 </div>
 
 All the contents (such as classes and functions) of the source file are contained by the package declared.
-So, in the example above, the full name of `baz()` is `foo.bar.baz`, and the full name of `Goo` is `foo.bar.Goo`. 
+So, in the example above, the full name of `printMessage()` is `org.example.printMessage`,
+and the full name of `Message` is `org.example.Message`. 
  
-If the package is not specified, the contents of such a file belong to "default" package that has no name.
+If the package is not specified, the contents of such a file belong to `default` package that has no name.
```
abreslav: This change is misleading because `default` looks like an identifier which it is not

pages/docs/reference/typecasts.md
```diff
@@ -93,16 +98,18 @@ Usually, the cast operator throws an exception if the cast is not possible. Thus
...
 ```kotlin
 val x: String = y as String
 ```
 </div>
 
 Note that *null*{: .keyword } cannot be cast to `String` as this type is not [nullable](null-safety.html),
-i.e. if `y` is null, the code above throws an exception.
-In order to match Java cast semantics we have to have nullable type at cast right hand side, like:
+i.e. if `y` is null, the code above throws an exception. 
+To make such code correct for null values, use the nullable type at the right hand side of the cast:
```
abreslav:
```suggestion
To make such code correct for null values, use the nullable type on the right hand side of the cast:
```

pages/docs/reference/basic-types.md
```diff
@@ -14,20 +14,74 @@ In this section we describe the basic types used in Kotlin: numbers, characters,
...
-| Byte	 | 8        |
+All variables initialized with integer values not exceeding the maximum value of `Int`
+have the inferred type `Int`. If the initial value exceeds this value, then the type is `Long`.
+To specify the `Long` value explicitly, append the suffix `l` or `L` to the value.
 
-Note that characters are not numbers in Kotlin.
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+val one = 1 //Int
```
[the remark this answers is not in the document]
p7nov: A space before the text looks better to me, so let's stick to this. I'll add spaces in other comments too.

pages/docs/reference/coding-conventions.md
```diff
@@ -20,25 +20,25 @@ This page contains the current coding style for the Kotlin language.
...
+In pure Kotlin projects, the recommended directory structure follows the package structure with
+the common root package omitted. For example,if all the code in the project is in the `org.example.kotlin` package and its
+subpackages, files with the `org.example.kotlin` package should be placed directly under the source root, and
+files in `org.example.kotlin.foo.bar` should be in the `foo/bar` subdirectory of the source root.
 
-In pure Kotlin projects, the recommended directory structure is to follow the package structure with
-the common root package omitted (e.g. if all the code in the project is in the "org.example.kotlin" package and its
-subpackages, files with the "org.example.kotlin" package should be placed directly under the source root, and
-files in "org.example.kotlin.foo.bar" should be in the "foo/bar" subdirectory of the source root).
+In JVM projects where Kotlin is used together with Java, Kotlin source files should reside in the same source root as the Java source files,
```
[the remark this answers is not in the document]
p7nov: For now we can maybe use the note formatting with platform name prefix like
>JVM: In projects where Kotlin is used together with Java...
The switch can be added later, I think, when we add a significant amount of such notes

ilya-g:
For the changed headings in long pages it may make sense to insert additional anchors with the names that were automatically generated for old headers to preserve anchor navigation from existing links.

p7nov:
> For the changed headings in long pages it may make sense to insert additional anchors with the names that were automatically generated for old headers to preserve anchor navigation from existing links.

Added anchors with old names.

pages/docs/reference/properties.md
```diff
@@ -170,7 +170,7 @@ public val table: Map<String, Int>
 ```
 </div>
 
-In all respects, this is just the same as in Java since access to private properties with default getters and setters is optimized so that no function call overhead is introduced.
+The access to private properties with default getters and setters is optimized so no function call overhead is introduced in this case.
```
ilya-g: Is it in accordance with kotlin spec and applied in all target platforms?

pages/docs/reference/basic-syntax.md
```diff
@@ -180,11 +194,21 @@ Just like Java and JavaScript, Kotlin supports end-of-line and block comments.
 
 </div>
 
-Unlike Java, block comments in Kotlin can be nested.
+Block comments in Kotlin can be nested.
```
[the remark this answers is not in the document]
p7nov: However, this does work in playground. I'll file an issue about incorrect highlighting.

PR #1439, opened 2019-07-29
Fix the Mapping Struct and Union Types from C tutorial

qiaoyuang:
Fix the Mapping Struct and Union Types from C tutorial. The navigation at the beginning of the tutorial should be Mapping Function Pointers from C instead of Mapping Struct and Union Types from C.


jonnyzzz:
Thanks :)

PR #1440, opened 2019-07-30
Typo

kotoko:
This example with 'val' does not make any sense for me.


p7nov:
We've rewritten the example, now it's correct, see https://kotlinlang.org/docs/reference/classes.html#overriding-properties

PR #1455, opened 2019-08-24
Moved the lambda function outside the parentheses

palkesh43:
Moved the lambda function outside the parentheses for IntArray for the best practice


p7nov:
Thanks!

PR #1458, opened 2019-08-26
Fix wrong type in sample, section "Local functions"

ark-1:
In Kotlin `Set` type denotes read-only set, while this function requires mutation.


p7nov:
Thanks!

PR #1460, opened 2019-08-27
Clarity to inheritance @param and @property tags

Astronaut4449:
Adds clarity to the relation of \@param and \@property as well as \@constructor and the class description.


p7nov:
Thanks for your suggestion.
However, we consider the current descriptions complete enough.

Astronaut4449:
> 
> 
> Thanks for your suggestion.
> However, we consider the current descriptions complete enough.

As far as I know it is not mentioned that using @param also documents @property and that @property serves as an override for @param. If I am mistaken please tell me where to find this. It took me a while to figure that out because I initially expected it to be the other way around (that you use @property by default and override it with @param).

PR #1461, opened 2019-08-28
Fix apparent error in 'Lambda Expression Syntax'

melston:
The original description of the 'full' syntax had less code than the 'simple' syntax shown later.


p7nov:
Thanks!

PR #1465, opened 2019-09-04
Update idioms.md

AndyMarchewka-rci:
Simple idiom to check a flag in a map


p7nov:
Hi! 
Thanks for your suggestion. It looks good for certain cases. 
However, such cases are special and rare (only for maps with boolean values), so hardly should we include it into the list of common idioms for everyday use.

PR #1467, opened 2019-09-06
remove AndroidRivers

falkorichter:
https://github.com/dodyg/AndroidRivers has not been touched in over four years and doesn´t even build with gradle


p7nov:
Thanks for noticing!

PR #1470, opened 2019-09-14
Small change to classes doc

BenChaimberg:
"many" is ambiguous and implies that there is a certain number of inherited classes required before the behaviour kicks in. On the other hand, "multiple" simply states that if a class inherits more than one class, then it must disambiguate the construction process.


p7nov:
Hi!
Thanks, _multiple_ definitely fits better here.

PR #1473, opened 2019-09-19
Fix kapt incremental annotation processing documentation.

Tapchicoma:
Mention that kapt incremental annotation processing is enabled by default starting version `1.3.50`.


p7nov:
Thanks!

PR #1475, opened 2019-09-28
Added Functional Programming in Kotlin

von-latinski:
Hi,

I thought this new early-access book on Kotlin might be a valuable resource for your list.

Thanks for considering it!

Branko


p7nov:
Hi!
Thanks for the suggestion. However, only a small part of the book is actually written.
We'll consider adding it once it's completed and published.

PR #1481, opened 2019-10-01
Added idiom TODO()

eekboom:
IMHO this is a very unambiguous idiom. (Can't think of any reason not do use it.)

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -381,3 +381,14 @@ a = b.also { b = a }
 ```
 </div>
 
+### TODO(): Marking code as incomplete
+ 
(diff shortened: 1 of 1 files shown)


pages/docs/reference/idioms.md
```diff
@@ -381,3 +381,14 @@ a = b.also { b = a }
 ```
 </div>
 
+### TODO(): Marking code as incomplete
+ 
+Kotlin's standard library has a `TODO()` function that will always throw a `NotImplementedError`.
+Its return type is `Nothing` so it can be used regardless of expected type.
+There's also an overload that accepts a reason parameter:
```
p7nov:
```suggestion
There's also an overload that accepts a String `reason` parameter:
```

p7nov:
Thanks for adding, this is really useful.

PR #1487, opened 2019-10-11
Adds demo for passing arguments to kts files

fowlerwill:
This update adds a sample for the reader to understand how to pass a command line argument containing a `-` to a kotlin script.


p7nov:
Thanks for the addition!

PR #1488, opened 2019-10-18
Remove broken link to a user group

frost13it:
Novosibirsk Kotlin User Group (https://vk.com/kotlin54) seems to be nonexistent.


p7nov:
Unfortunately, it is. Thanks for noticing.

PR #1489, opened 2019-10-21
feat: compiler reference documentation

p7nov:
Added a new page with available compiler options.

Staging build: https://branch-doc-compiler-reference.kotlin-web-site.labs.jb.gg/docs/reference/compiler-reference.html


p7nov:
Note: The Native part is not ready yet.
@ilmat192 please provide links to parts of the Native docs related to specific options like, for example, `-memory model` or `-endorsed-libs` etc

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+    ```
+    
+    </div>
+    
+    If the files reside in locations different from the current directory, use relative paths. 
+    
+    <div class="sample" markdown="1" mode="shell" theme="idea">
+    
+    ```bash
+    $ kotlinc @options\compiler.options hello.kt
```
udalov: `\` is an escape character in bash

```suggestion
    $ kotlinc @options/compiler.options hello.kt
```

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+## JVM compiler options
+
+Kotlin compiler for JVM compiles Kotlin source files into Java class files. 
+The command-line tools for Kotlin to JVM compilation are `kotlinc` and `kotlinc-jvm`.
+You can also use them for executing Kotlin script files.
+
+In addition to [common options](#common-options), Kotlin/JVM compiler has the options listed below.
+
+### `-classpath <path>` (`-cp <path>`)
+* Search for class files in the specified paths. Separate elements of the classpath with semicolons (**;**).
```
udalov:
```suggestion
* Search for class files in the specified paths. Separate elements of the classpath with system path separators (**;** on Windows, **:** on macOS/Linux).
```

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+### `-no-jdk`
+* Exclude the Java runtime from the classpath.
+
+### `-no-reflect`
+* Exclude `kotlin-reflect.jar` from the classpath.
+  
+### `-no-stdlib`
+* Exclude both `kotlin-stdlib.jar` and `kotlin-reflect.jar` from the classpath. 
+  
+### `-script <file>`
```
udalov: In fact, the `-script` argument is boolean, and it makes the compiler run the first script from the set of compiled files. (E.g. `kotlinc 1.kt 2.kts 3.kt -script` is valid, and will compile all files and run `2.kts`.)

Also note that since 1.3.70, this will be a common argument (https://github.com/JetBrains/kotlin/commit/f59e393e37d1d745261e6a2db9921b3fe5ccf7fc)

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+### `-module-kind {plain|amd|commonjs|umd}`
+
+- Kind of the JS module generated by the compiler:
+    - `plain` - a plain JS module;
+    - `commonjs` - a [CommonJS](http://www.commonjs.org/) module;
+    - `amd` - an [Asynchronous Module Definition](https://en.wikipedia.org/wiki/Asynchronous_module_definition) module;
+    - `umd` - a [Universal Module Definition](https://github.com/umdjs/umd) module.
+    
+    To learn more about the module kinds and dictinctions between them,
+    read [this](https://www.davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/) article.
```
udalov:
```suggestion
    see [this](https://www.davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/) article.
```

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+    
+### `-no-stdlib`
+
+- Don't use the bundled Kotlin standard library.
```
udalov: Sounds a bit ambiguous to me and too different from JVM's `-no-stdlib` option. Maybe something like this?
```
Exclude `kotlin-stdlib-js.jar` from the compilation dependencies.
```
(Maybe we should also change the description in `kotlinc-js -help`...)

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,354 @@
...
+
+### `-output <path>`
+
+- Place the output file into the specified location.
```
udalov: Please also explain that `<path>` is the path to the resulting .js file, and not a directory for example. (There's an error in the latter case.)

pages/docs/reference/compiler-reference.md
p7nov: Yes, I'll suggest some changes to option descriptions in the Kotlin repo as well.

pages/docs/reference/compiler-reference.md
```diff
@@ -0,0 +1,356 @@
...
+
+- Specify a name for the compilation module.
+
+### `-native-library <path>`(`-nl <path>`)
+
+- Include the native bitcode library.
+
+### `-no-default-libs`
+
+- Don't link the libraries from dist/klib automatically.
```
ilmat192: Description of this flag is obsolete and should be update at the compiler side. Currently it disables linking a user code with default platform libraries distributed with the compiler (More about platform libraries: https://kotlinlang.org/docs/reference/native/platform_libs.html).

udalov:
Ref: https://github.com/JetBrains/kotlin/pull/2717

pages/docs/reference/compiler-reference.md
p7nov: As discussed with Zalim, this option works quite different in JS. Since the JS compiler implicitly adds stdlib as a compilation dependency, the most accurate description for the option is "Don't add" or "Don't use". So I rephrase this closer to the original description.

PR #1490, opened 2019-10-21
Update server-overview.md with Micronaut

adamkobor:
Micronaut is a young prospect in the league of server-side frameworks, offering modern features and outstanding performance out of the box. I think the documentation should mention it, that's why I made this PR.


p7nov:
Thanks for your addition!

PR #1492, opened 2019-10-25
Added example of `apply` as an alternative to `with` in idioms.md

BikramjeetSingh:
The example given in the "Calling multiple methods on an object instance" section using the `with` method can also be achieved using the `apply` method.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -310,7 +310,7 @@ fun transform(color: String): Int = when (color) {
 ```
 </div>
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hi, thanks for the suggestion. 
However, according to Kotlin docs, this is not an idiomatic use of `apply`. See https://kotlinlang.org/docs/reference/scope-functions.html#apply.

You can still open a new PR with a new separate idiom for `apply`: configuring properties of an object.

Thanks again!
Regards, Pavel
Kotlin Technical Writer.

PR #1495, opened 2019-10-28
Update coroutines-basic-jvm.md with dependency version 1.3.2

edugof:
Update dependency version from https://github.com/Kotlin/kotlinx.coroutines


p7nov:
Thanks, we've already updated this.

PR #1498, opened 2019-10-29
Remove extra `,` on the function references

nitin-k:
Not a major change. Just removing an extra `,` that was mentioned in documents.
Not sure, if it was intentionally added.


p7nov:
Hi! Thanks for your feedback.
Anyway, the [serial comma](https://en.wikipedia.org/wiki/Serial_comma) (`,` placed before the conjunction) is a matter of style and we stick to American styleguides that recommend placing a comma after each word in series.

PR #1499, opened 2019-10-30
Added `apply` idiom

BikramjeetSingh:
Added an example for use of `apply` to configure properties of an object.


p7nov:
Thanks for the addition!

PR #1504, opened 2019-11-04
Update basic-types.md

sihyeon-kim:
Remove the following complaint in IntelliJ IDE.  
"Lambda argument should be moved out of parentheses"


p7nov:
Thanks!

PR #1507, opened 2019-11-05
Use Gradle Task Configuration API in Documentation

JLLeitschuh:
Updates documentation to use Gradle's lazy task configuration API to configure the `KotlinCompile` task.
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

CC: @big-guy @bamboo @eskatos


p7nov:
Hey, thanks for the update!

PR #1508, opened 2019-11-06
Remove ambiguity from infix function calls predence examples

kwahome:
The examples showing precedence of infix function calls have some ambiguity due to the use of the conjunction `and`, which is also a bitwise operation keyword in Kotlin, to join two equivalent expressions in a sentence. 

The example block is enclosed in indented quoting (>) which drowns the monospace font otherwise differentiating the conjunctive use of the `and` between the expressions. It all appears as though one long expressions and takes you a few minutes to get the gist of what's going on.

see:
image


h0tk3y:
Thanks!

PR #1512, opened 2019-11-13
feat: scratch tutorial update

p7nov:
Updated pictures in [scratch tutorial](https://kotlinlang.org/docs/tutorials/quick-run.html)
Added info about worksheets


pages/docs/tutorials/quick-run.md
```diff
@@ -48,25 +60,33 @@ IntelliJ IDEA can run your scratches automatically. To get the execution results
 
 ### Using modules
 
-To use classes or functions from a project in the scratch, import them into the scratch file as usual with the `import` statement. Then write your code and run it with the appropriate module selected in __Use classpath of module__ list. To rebuild the module automatically before running the scratch, select __Make before Run__.
+Worksheets automatically get access to classes and functions from the module where they reside.
+
+To use classes or functions from a project in a scratch, import them into the scratch file as usual with the 
+`import` statement. Then write your code and run it with the appropriate module selected in __Use classpath of module__ list.
+ To rebuild the module automatically before running the scratch, select __Make before Run__.
```
yalishevant: I would like to add a reason why/when module rebuild is needed. It's quite common wrong expectation, that a scratch/worksheet during its run relies on current version of source code. Actually it relies on compiled code and doesn't track changes in related sources. So when user modifies dependent sources and then runs the scratch/worksheet - one needs to compile them before. I it happens iteratively - "Make before Run" checkbox is helpful.

PR #1513, opened 2019-11-14
Deleted unused parameter p

MareikeLowinski:
The unused parameter p: Int is confusing. On the first sight the reader might get the impression that p - as the port - is needed, but is actually already set within the constructor.


p7nov:
Thanks for you fix!

PR #1516, opened 2019-11-19
The dependency plugin should be run before the compile phase

bdemers:
The dependency plugin should be run before the compile phase
This helps prevent potential plugin ordering issues, and is _more_ correct as the dependency plugin doesn't actually compile anything
it _could_ run in an earlier phase as well


bdemers:
Based on this example: https://github.com/JetBrains/kotlin/blob/master/libraries/examples/js-example/pom.xml

This doc is out of date, and the dep-plugin shouldn't be used

PR #1522, opened 2019-11-28
Use the `it` convention (from coding conventions)

elifarley:
Lambda parameters:
In lambdas which are short and not nested, it's recommended to use the `it` convention instead of declaring the parameter explicitly.

---
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/kotlin-web-site/1522)


p7nov:
Hi! Thanks for your addition.
Anyway, we better keep both options to show readers that they can use any name for the list element in the lambda.

elifarley:
Fair point. Thanks!

PR #1523, opened 2019-11-28
Method call formatting according to coding conventions

elifarley:
In long argument lists, put a line break after the opening parenthesis. Indent arguments by 4 spaces.

---
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/kotlin-web-site/1523)


p7nov:
Thanks again.
Actually, we don't think this line is long. It fits even in a small window. The default maximum line length in IDEA is 120 symbols - 2.5 times longer than this.

PR #1524, opened 2019-12-01
Edited wording for a sentence

ozejde:
Edited the grammar for a sentence:

changed  "... more that 6-7 decimal digits, it will be rounded" 
to "... more than 6-7 decimal digits, it will be rounded"


AlexanderPrendota:
Thanks!

PR #1539, opened 2019-12-12
Make servlet example runnable without IJ Ultimate

yole:


yole:
Found this old branch; need to check if the provided instructions work.

PR #1541, opened 2019-12-15

Ethan1983:
Add instructions for setting up gradlew in a new multiplatform project.

### DIFF
--- a/pages/docs/tutorials/mpp/multiplatform-library.md
+++ b/pages/docs/tutorials/mpp/multiplatform-library.md
@@ -267,6 +267,8 @@ class Base64Test {
 
 </div>
 
+Copy gradlew, gradlew.bat and gradle/wrapper/gradle-wrapper.jar from another project setup with gradle.
+
 Execute `./gradlew check` and you will see that the tests are run three times, on JVM, on JS, and on Native!
 
 If we want, we can add tests to a specific platform, then it will be executed only as part of these platform tests.
--- a/pages/docs/tutorials/mpp/multiplatform-library.md
+++ b/pages/docs/tutorials/mpp/multiplatform-library.md
@@ -267,7 +267,7 @@ class Base64Test {
 
 </div>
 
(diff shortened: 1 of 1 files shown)


pages/docs/tutorials/mpp/multiplatform-library.md
```diff
@@ -267,6 +267,8 @@ class Base64Test {
 
 </div>
 
+Copy gradlew, gradlew.bat and gradle/wrapper/gradle-wrapper.jar from another project setup with gradle.
+
 Execute `./gradlew check` and you will see that the tests are run three times, on JVM, on JS, and on Native!
```
Ethan1983: gradlew is missing in newly created kotlin multiplatform project using IntelliJ IDEA 2019.3 (Community Edition)

Build #IC-193.5233.102, built on November 27, 2019
Runtime version: 11.0.4+10-b520.11 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
macOS 10.15.1
Non-Bundled Plugins: jd-intellij, org.jetbrains.kotlin

pages/docs/tutorials/mpp/multiplatform-library.md
```diff
@@ -267,6 +267,8 @@ class Base64Test {
 
 </div>
 
+Copy gradlew, gradlew.bat and gradle/wrapper/gradle-wrapper.jar from another project setup with gradle.
```
qwwdfsad: Advising to copy something from something else as one of the first steps is not the best option for the tutorial.
It would be much better if we asked to [generate the wrapper instead](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper)

Ethan1983: Yup, just realized. Will update the PR. Any gradle above 4.7 did it for me.

qwwdfsad: Thanks!

PR #1546, opened 2019-12-27
Update Index

packtpartner:
Added new book titles


p7nov:
Sorry, we don't post third-party commercial offerings on the Kotlin documentation pages.
You can add your books to the *Books* page: https://kotlinlang.org/docs/books.html
Be sure to check if they are already listed here.

PR #1547, opened 2019-12-29
Fixed link to LINQ

Structed:
Skip redirection from MSDN to Microsoft Docs and link directly to the MS Docs article.


p7nov:
Thanks!

PR #1548, opened 2019-12-30
Update collection-aggregate.md

jlents:
making the word "return" plural, to be consistent with the other 3 bullet points


p7nov:
In this case, we write about two functions, so the plural is correct here.

PR #1552, opened 2020-01-07
new_value must be supertype not subtype.

shiehnpin:
Consider following example.

the type of new value in setValue() must be the same type as property or its supertype.
```
open class Parent
open class Child: Parent()

class DelegateA{

    //this function must return the same type as property (or its subtype).
    operator fun getValue(thisRef: Any?, property: KProperty<*>): Child {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }

    //new value — must be of the same type as the property or its supertype.
    operator fun setValue(thisRef: Any?, property: KProperty<*>, value: Parent) {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }
}

class Container{

    val valChild:Child by DelegateA()

    val valParent:Parent by DelegateA()

    var varChild:Child by DelegateA()

    var varParent:Parent by DelegateA()

}
```

### DIFF
--- a/pages/docs/reference/delegated-properties.md
+++ b/pages/docs/reference/delegated-properties.md
@@ -257,7 +257,7 @@ For a **mutable** property (a *var*{:.keyword}), a delegate has to _additionally
  
 * `thisRef` --- same as for `getValue()`;
 * `property` --- same as for `getValue()`;
-* `new value` --- must be of the same type as the property or its subtype.
+* `new value` --- must be of the same type as the property or its supertype.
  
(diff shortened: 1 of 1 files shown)


p7nov:
Hi! 
Thanks for your suggestion.
However, the actual value we pass can't be a supertype of a property's type. If you declare a `Child` property, the compiler won't let you assign a `Parent` object to it.
As I see, our docs are not clear enough about what type we can _declare_ (parameter type in the function declaration; can be a supertype) and what type we can _pass_ (argument type in the function call; must be the same or a subtype). I'll update this part of the page.

Regards, Pavel,
Kotlin Technical Writer.

PR #1558, opened 2020-01-14
Update basic-types.md to clarify integer division

georgeleege:
I have just started to learn Kotlin. While going through the basic-types section of the docs, it wasn't clear to me what the behavior of the division operator was. 

A quick search led me to https://discuss.kotlinlang.org/t/dividing-integers/1662

I understand any implied consistency with Java, but figured that it would be good to be more explicit about it in the docs since people might be coming to Kotlin from a variety of languages or be new to programming in general.

### DIFF
--- a/pages/docs/reference/basic-types.md
+++ b/pages/docs/reference/basic-types.md
@@ -235,9 +235,7 @@ val l = 1L + 3 // Long + Int => Long
 Kotlin supports the standard set of arithmetical operations over numbers (`+` `-` `*` `/` `%`), which are declared as members of appropriate classes (but the compiler optimizes the calls down to the corresponding instructions).
 See [Operator overloading](operator-overloading.html).
 
-Note that division between integers always returns an integer. Any fractional part is discarded.
-
-For example:
+Note that division between integers always returns an integer. Any fractional part is discarded. For example:
 
 ```kotlin
 val x = 5 / 2
--- a/pages/docs/reference/basic-types.md
+++ b/pages/docs/reference/basic-types.md
@@ -237,27 +237,38 @@ See [Operator overloading](operator-overloading.html).
 
 Note that division between integers always returns an integer. Any fractional part is discarded. For example:
 
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
 ```kotlin
 val x = 5 / 2
-println(x == 2.5) // ERROR: Operator '==' cannot be applied to 'Int' and 'Double'
-
-val y = 5 / 2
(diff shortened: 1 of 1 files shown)


pages/docs/reference/basic-types.md
```diff
@@ -232,7 +232,7 @@ val l = 1L + 3 // Long + Int => Long
 
 ### Operations
 
-Kotlin supports the standard set of arithmetical operations over numbers, which are declared as members of appropriate classes (but the compiler optimizes the calls down to the corresponding instructions).
+Kotlin supports the standard set of arithmetical operations over numbers (`+` `-` `*` `/` `%`), which are declared as members of appropriate classes (but the compiler optimizes the calls down to the corresponding instructions). Note that calling `/` between an `Int` and an `Int` results in integer division and returns an `Int`, e.g. `5 / 2` returns `2`. For floating-point division between two arguments of type `Int`, convert one of the arguments to a floating-point type, e.g. `5 / 2.toDouble()` returns `2.5` as a `Double`.
```
p7nov: Hey! Thanks for the addition.
I suggest some improvements:

1. Let's leave the first paragraph as is and add this thought in a separate one.
2. To be precise, this works not only on `Int` but on _integer types_ - `Int` and `Byte`; the division results in `Int` for both.
3. Please use _for example_ instead of _e.g._ Not a big issue, just our writing style.
4. Even better to move the example into a runnable code sample. You know, developers love code :) In this case, the text can contain a more general explanation, like "..._the fractional part is dicsarded_" or smth.

georgeleege: Thanks for the suggestions.
It's been a while since I have been on GitHub.
I have rebased my commit and made the changes you suggested.

p7nov: Thanks! 
One more thing: please make the code samples runnable by wrapping them in 
```
<div class="sample" markdown="1" theme="idea">
fun main() {
//sampleStart
    your code
//sampleEnd
}
</div>
```
Find examples in the same file, *Representation* section.
If you want to show errors (your first sample), comment out the lines that produce them and leave explanations in the comment. Example: https://kotlinlang.org/docs/reference/collections-overview.html#collection-types.

georgeleege: Done.
Didn't even notice that the other samples were runnable.
That's pretty cool.

p7nov: Seems OK, will merge this.
Thanks again for your contribution!

PR #1562, opened 2020-01-20
Use MD5 digest instead of timestamp as asset version

hltj:
Because that the timestamps of the assets always changed after `yarn run build`.
Use MD5 digest can prevent that all generated pages have changed in each build even no asset source file change.


zoobestik:
thx

PR #1569, opened 2020-01-26
Update idioms.md

post2shyam:
Range example with 'in' was repeated. Hence removed it.


post2shyam:
Range example with 'in' was repeated. Hence removed it.

p7nov:
Hi!
Thanks for the suggestion, we'll leave this as it shows that ranges can also be used in `if`.

PR #1571, opened 2020-01-27
feat: navigation update

p7nov:
Fixed the order in left navigation pane and index page.


p7nov:
Sorry, wrong commits are here.

PR #1573, opened 2020-01-27
Print the results in order to provide feedback

chapeupreto:
Print the results in order to get more feedback from the samples.

This also fixes "variable 'adam' is never used" noticed messages.


p7nov:
Thanks!

PR #1576, opened 2020-02-03
Add ID to "Why Kotlin?" section

chrisparton1991:
I was writing up an email suggesting the use of Kotlin over Java for a project.

The "Why Kotlin?" section of the site is an excellent showcase of the benefits of Kotlin, so it would be great to be able to link directly there.

Thanks for the amazing work you're all doing with Kotlin!

### DIFF
--- a/templates/pages/index.html
+++ b/templates/pages/index.html
@@ -64,7 +64,7 @@
     <a id="try-kotlin"></a>
     {% include 'inc/pages/index/code-examples.html' %}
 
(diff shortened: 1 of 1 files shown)


zoobestik:
Yep, great idea.

**P.S.** I think, we will add icon for people who don't know about this PR :) something like this:
Screenshot 2020-02-04 at 13 20 53

chrisparton1991:
That sounds great @zoobestik, thanks :)

PR #1579, opened 2020-02-07
fix a typo

hltj:
The `sets` should be `maps` according to the context.


p7nov:
You're right, thanks!

PR #1581, opened 2020-02-09
Update properties.md

somsubhojit1988:
Grammatical correction of the 1st sentence of "Compile-Time Constants". 
Proposition -
"Properties the value of which is known at compile time ... " -> "Property values which are known at compile time .... "


p7nov:
Thanks!

PR #1592, opened 2020-02-15
Update links for Sydney Kotlin UG

gouline:
Sydney Kotlin User Group was consolidated with other related local meetups into SYDspace last year. This updates all the sydkotlin.space links to sydspace.org/kotlin links.


zoobestik:
Thanks!

PR #1596, opened 2020-02-18
chore: use JB Mono for code examples

zoobestik:


Takhion:
:tada:

@zoobestik could you add the dev. versions of the fonts as local, with highest priority?
So we can test them also on the website :)

Ah, has this been tested on languages with unsupported characters?

zoobestik:
Thanks, @Takhion,

Excuse me, i'm not sure, what you mean in "devs fonts as local" ☹️
It presents as for ex.:
```
/* .... */
src: local("JetBrains Mono Regular"),
    local("JetBrainsMono-Regular"),
    /* .... */
```
And it should works with local font if it's exist.
Is it not good enough for your OS?

Takhion:
Hey @zoobestik! :D

Sorry, what I meant is that sometimes we get early access to _development_ versions of the font, which we install on our local machine, and it would be awesome if those could be prioritized over the regular version of the font.

Does it make sense?

JetBrains Mono dev Regular

PR #1597, opened 2020-02-22
Update Exceptions reference page with more information

Nezteb:
1. Fix minor grammar issue to use "results in" instead of "results into".
2. Update linked blog post that no longer exists (using Wayback Machine).
3. Refer to the `@Throws` annotation in case readers want to explicitly declare exceptions for Java interop.

### DIFF
--- a/pages/docs/reference/exceptions.md
+++ b/pages/docs/reference/exceptions.md
@@ -72,7 +72,7 @@ Appendable append(CharSequence csq) throws IOException;
 
 What does this signature say? It says that every time I append a string to something (a `StringBuilder`, some kind of a log, a console, etc.)
 I have to catch those `IOExceptions`. Why? Because it might be performing IO (`Writer` also implements `Appendable`)...
-So it results into this kind of code all over the place:
+So it results in this kind of code all over the place:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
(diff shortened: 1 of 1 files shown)


pages/docs/reference/exceptions.md
```diff
@@ -96,6 +96,8 @@ Other citations of this sort:
 * [Java's checked exceptions were a mistake](http://radio-weblogs.com/0122027/stories/2003/04/01/JavasCheckedExceptionsWereAMistake.html) (Rod Waldhoff)
 * [The Trouble with Checked Exceptions](http://www.artima.com/intv/handcuffs.html) (Anders Hejlsberg)
 
+If your Kotlin code is being called from Java, you can always use the [`@Throws`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-throws/) annotation in order to alert callers of possible exceptions.
```
katia-energizer: Thank you for your helpful suggestion. Could you add 2 links? The `@Throws` annotation can be used for Swift and Objective-C as well. The link you suggested is also available in the linked section. 
```suggestion
If you want to alert callers of possible exceptions when calling Kotlin code from Java, Swift, or Objective-C, you can use the `@Throws` annotation. Read more about using this annotation [for Java](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#checked-exceptions) as well as [for Swift and Objective-C](https://kotlinlang.org/docs/reference/native/objc_interop.html#errors-and-exceptions).
```

PR #1610, opened 2020-03-05
Make code blocks visible if JS is disabled

smichel17:
**Not intended to be merged as-is.**

Right now, if JS is disabled, code blocks are not visible at all, and leave blank space on the page:

---

image

---

Obviously just removing the style is not sufficient, so this is more like an issue than a PR (I would have submitted a proper PR, but it's not a quick job for someone unfamiliar with the code base, so I figured I'd try this first).

### DIFF
--- a/static/css/playgrounds.scss
+++ b/static/css/playgrounds.scss
@@ -1,10 +1,5 @@
 @import './static/css/_config';
 
-// Hide plain text before loading runnable examples
-.sample {
-  visibility: hidden;
-}
-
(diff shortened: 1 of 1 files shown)


zoobestik:
I agree, it's better.

**P.S.** Yep, we are working with documentation.

smichel17:
You may need to apply the hidden style with js once the interactive bits load, otherwise they may appear twice. Or at least, that's what happened when I manually removed the style in the browser.

PR #1616, opened 2020-03-13
Restore missed divs

bashor:


zoobestik:
Yep, thanks. It has already merged in #1617

PR #1617, opened 2020-03-14
Add missing div end tags

JakeWharton:
This was causing the rest of the page to render in the code block despite the closing backticks.


JakeWharton:
See https://kotlinlang.org/docs/reference/using-gradle.html#targeting-javascript

PR #1621, opened 2020-03-17
Clarification of `Id` when choosing good names.

Enrico2:
Nit pick: When choosing good names for things, `ID` is often a point of contention. 
I was happy to see that the coding conventions doc has a specific section for acronyms and follows what I'd expect. I think it bares clarification that `Id` is not an acronym, it's an abbreviation.


pcarrier:
> I think it bares clarification that Id is not an acronym, it's an abbreviation.

It'd help if we agreed.

image
-- https://en.wiktionary.org/wiki/ID
image
-- https://en.wiktionary.org/wiki/id.

p7nov:
Hey!
Thanks for your PR. After searching in several style guides and discussions, I see that there is no consensus about the proper spelling of `id`/`Id`/`ID`. People keep arguing even in this exact pull request :) So we'll leave the choice to the reader.

Regards, Pavel,
Kotlin Technical Writer.

PR #1623, opened 2020-03-18
Improve readability about object class init timing

OrigamiDream:
It was very hard to look for explanations about initialization timing of instances which is declared with `object` keyword usage instead of `class` keyword since the explanations about 'lazy initialization' is located at bottom paragraph when excerpt the part of lazy init stuff.

Most of people who are looking for about initialization timing of object instances will excerpt the part of object class declaration paragraph, not to the end of the document.


p7nov:
Thanks!

PR #1626, opened 2020-03-19
Update data-science-overview.md

Lundez:
Adding a library that makes Smile easier to use from Kotlin by adding the Scala implicits in form of extension functions. Also added a trait as a interface.

The homepage actually makes use of these implicits, with this library it's easier to follow the documentation in the natural language part.

The repo is not in any way "official" for Smile but rather my own work.


p7nov:
Thanks for your addition!

PR #1632, opened 2020-03-24
Add space between for and opening parenthesis

jeanbarrossilva:
[As suggested in the documentation](https://kotlinlang.org/docs/reference/coding-conventions.html#horizontal-whitespace), `if`s, `when`s, `for`s and `while`s should have a space before its opening parenthesis.


SebastianAigner:
You are absolutely right!

PR #1633, opened 2020-03-24
Removed link from Android Studio description

DfM7:
Removed link in "Studio 3.0" words, as it's essentially the same as the download link.


p7nov:
I would also remove the `3.0` version since it's outdated. @dmitry-konchalenkov what do you think?

dmitry-konchalenkov:
> I would also remove the `3.0` version since it's outdated. @dmitry-konchalenkov what do you think?

@p7nov Agreed, please do.

PR #1634, opened 2020-03-25
Adding hyphen to lambda arrow (was ">")

schw1804:


SebastianAigner:
Thanks for the contribution. However, the current snippet is correct. It describes a comparison using an implicitly named parameter - filter for all elements that are greater than 10.

schw1804:
Oops. Sorry. Was apparently not reading very carefully.

PR #1639, opened 2020-03-29
Grammar: "its" instead of "it's"

evjan:
"it's" means "it is"


SebastianAigner:
Well spotted, thanks! 🎉

PR #1643, opened 2020-03-30
Added Alpas to list of Kotlin frameworks

armgitaar:


antonarhipov:
Instead of adding new frameworks to the list will better rework the chapter and point at https://kotlin.link community maintained list of frameworks, tools, and libraries. Alpas framework is listed at kotlin.link web page already.

PR #1649, opened 2020-04-02
KT-37943 Use abstract return type as the name of factory function

stevecstian:
Factory functions can have same name as the abstract return type.


p7nov:
Thank you for fixing this!

PR #1653, opened 2020-04-02
Remove redundant .toLong() call

arnavb:
Running .await on a `Deferred` returns the type inside directly.


p7nov:
Hi!
Thanks for your suggestion, but in this case the returned value is incorrect. The expected value (which you get with `toLong()`) is `500000500000`.

PR #1654, opened 2020-04-03
Improve example for using when without an argument

AfzalivE:
In the current example:
Both branches are for x, it makes sense to use `when(x)` in it.

But if we switch the example to have two variables, then it makes sense to just use when without arguments.

### DIFF
--- a/pages/docs/reference/control-flow.md
+++ b/pages/docs/reference/control-flow.md
@@ -146,8 +146,8 @@ If no argument is supplied, the branch conditions are simply boolean expressions
 ```kotlin
 when {
     x.isOdd() -> print("x is odd")
-    x.isEven() -> print("x is even")
-    else -> print("x is funny")
+    y.isEven() -> print("y is even")
+    else -> print("x and y are both funny")
 }
 ```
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hi!
Thanks for your PR.
How would you use `when(x)` for this case?

AfzalivE:
You wouldn't. 

This is more to highlight a case where using when without an argument makes more sense over `when(x)`

pages/docs/reference/control-flow.md
```diff
@@ -146,8 +146,8 @@ If no argument is supplied, the branch conditions are simply boolean expressions
 ```kotlin
 when {
     x.isOdd() -> print("x is odd")
-    x.isEven() -> print("x is even")
-    else -> print("x is funny")
+    y.isEven() -> print("y is even")
+    else -> print("x and y are both funny")
```
p7nov, line 150:
```suggestion
    else -> print("x+y is even.")
```

p7nov:
Ok, I see. Just one more suggestion to make the last line meaningful.

p7nov:
Thanks!

PR #1662, opened 2020-04-09
Update GitHub contributors numbers

hamzaslama:


p7nov:
Thanks for your suggestion!
However, the Kotlin team (about 100 people) also commits to GitHub, so 250 is really close to the number of _external_ contributors. Let's leave it as is.

PR #1664, opened 2020-04-11
Kotlin Gradle DSL snippet in inline-classes.md

sebokopter:
This commit adds the Kotlin Gradle DSL code snippet for "Enabling inline classes in Gradle"


p7nov:
Thanks for your addition!

PR #1667, opened 2020-04-12
Rename kotlintest to kotest

ashishkujoy:
Since 4.0 release kotlintest has been renamed to [kotest](https://github.com/kotest/kotest) so changing that name here.


p7nov:
Thanks for the update!

PR #1669, opened 2020-04-13
chore: added binary configuration to MPP DSL reference

p7nov:


pages/docs/reference/mpp-dsl-reference.md
```diff
@@ -270,6 +268,99 @@ kotlin {
 
 </div>
 
+For binaries configuration, the following parameters are available:
+
+|**Name**|**Description**| 
+| --- | --- |
+|`compilation`|The compilation from which the binary is built. By default, `test` binaries are based on the `test` compilation while other binaries - on the `main` compilation.|
+|`linkerOpts`|Options passed to a system linker during binary building.|
```
katia-energizer, line 276: Is there any list of these options we could link to? Do developers know them?

pages/docs/reference/mpp-dsl-reference.md
```diff
@@ -270,6 +268,99 @@ kotlin {
...
+|`baseName`|Base name for the output file.|
```
katia-energizer, line 277: Don't you want to add information about the difference between the base and final output file names?

pages/docs/reference/mpp-dsl-reference.md
```diff
@@ -270,6 +268,99 @@ kotlin {
...
+|`entryPoint`|The entry point function for executable binaries. By default, it's `main()` in the root package.|
+|`outpuFile`|Access to the output file.|
```
katia-energizer, line 279:
```suggestion
|`outputFile`|Access to the output file.|
```
Typo

pages/docs/reference/mpp-dsl-reference.md
p7nov: Doesn't seem so, system linkers are out of our scope. We just know that they have configuration options. All info is in their help/manuals.

pages/docs/reference/mpp-dsl-reference.md
```diff
@@ -270,6 +268,99 @@ kotlin {
...
+
+|**Name**|**Description**| 
+| --- | --- |
+|`compilation`|The compilation from which the binary is built. By default, `test` binaries are based on the `test` compilation while other binaries - on the `main` compilation.|
+|`linkerOpts`|Options passed to a system linker during binary building.|
+|`baseName`|Custom base name for the output file. The final file name will be formed by adding system-dependent prefix and postfix to this base name.|
+|`entryPoint`|The entry point function for executable binaries. By default, it's `main()` in the root package.|
+|`outputFile`|Access to the output file.|
+|`linkTask`|Access to the link task.|
+|`runTask`|Access to the run task.|
```
ilmat192, line 281: It is available for executables only. Also this property returns `null` if the binary target is other than `linuxX64`, `macosX64` or `mingwX64`.

PR #1672, opened 2020-04-14
Comments added : sample code for reference types

amritlalsahu5:
Comments added in the sample code for both null references and non-null references


p7nov:
Thanks!

PR #1676, opened 2020-04-25
docs: remove controversial advice about 'also'

ilya-g:
The advice causes confusion, as can be seen in:
- https://www.reddit.com/r/Kotlin/comments/g6oq6c/correct_usage_of_also_scope_function/
- https://stackoverflow.com/questions/61367281/whats-the-real-intention-behind-kotlins-also-scope-function

### DIFF
--- a/pages/docs/reference/scope-functions.md
+++ b/pages/docs/reference/scope-functions.md
@@ -499,9 +499,9 @@ Having the receiver as the return value, you can easily include `apply` into cal
 
 **The context object** is available as an argument (`it`). **The return value** is the object itself.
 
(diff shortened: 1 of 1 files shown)


andbed:
Now this is confusing. From previous clear usage advice, now it is smth. like: "use `also` for a specific reason (Use `also` for actions that need a reference rather to the object than to its properties and functions) unless there is `this` shadowing, all the usage guidance do not count, use it as `apply`. 
While reading code I need to analyse if `also` is put there because it was shadowing `this` or if it followed intended usage.
Not to mention that all code bases that applied previous rule for years become messy after starting to apply new version of the rule...

PR #1690, opened 2020-05-08
added a course

geek-2002:
added a course by udacity in the course section of the website


p7nov:
Hi!
Thanks for your PR.
However, currently we list only courses from the Kotlin team on this page.
There are many other courses, but we must review them before we can recommend them on the Kotlin website.
We hope to create another section with approved third-party courses in the future. We'll definitely consider Google's courses for this section.

Regards, Pavel
Kotlin Tech Writer

PR #1692, opened 2020-05-11
Update sam-with-receiver section

nglorioso:
Matching the groovy configuration to the example and Maven configuration.


p7nov:
Hey!
Thanks. It seems that we've already merged the same change from you in https://github.com/JetBrains/kotlin-web-site/pull/1386
Here it is on the page: https://kotlinlang.org/docs/reference/compiler-plugins.html#using-in-gradle-2

nglorioso:
Yikes, apologies for the dupe! I saw the loose branch, and github said it would merge cleanly. Let me clean that up.

PR #1694, opened 2020-05-12
Fix a grammar mistake

MasoodFallahpoor:


pages/docs/reference/delegation.md
```diff
@@ -13,7 +13,7 @@ Delegated properties are described on a separate page: [Delegated Properties](de
 
 ## Implementation by Delegation
 
-The [Delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern) has proven to be a good alternative to implementation inheritance,
+The [Delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern) has proven to be a good alternative to implement inheritance,
```
p7nov, line 16:
```suggestion
The [Delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern) has proven to be a good way to implement inheritance,
```

MasoodFallahpoor: I think replacing 'alternative' with 'way' somehow changes the meaning of the sentence.

p7nov: Thanks, I get your point.
Anyway, the sentence is grammatically correct since the words "implementation inheritance" are a commonly known name for an OOP concept and we want to highlight that the delegation concept is similar to it (_A is a good alternative to B_)

p7nov: To make this clear, we can add a link to https://kotlinlang.org/docs/reference/classes.html#inheritance under the words implementation inheritance_

MasoodFallahpoor: If "implementation inheritance" is grammatically correct then everything's fine and this issue could be closed.

p7nov: Yes, we're sure it's correct. 
Anyway, thanks for your contribution!

PR #1697, opened 2020-05-15
deprecated ByteArray.stringFromUtf8() is replaced with ByteArray.toKS…

jolasjoe:
deprecated ByteArray.stringFromUtf8() is replaced with ByteArray.toKS…


p7nov:
Thanks for the update!

PR #1706, opened 2020-05-27
chore: clarify object identity for nullable references

p7nov:
https://youtrack.jetbrains.com/issue/KT-39176


pages/docs/reference/basic-types.md
```diff
@@ -124,7 +124,7 @@ val bytes = 0b11010010_01101001_10010100_10010010
 On the Java platform, numbers are physically stored as JVM primitive types, unless we need a nullable number reference (e.g. `Int?`) or generics are involved. 
 In the latter cases numbers are boxed.
 
-Note that boxing of numbers does not necessarily preserve identity:
+Note that nullable references to the same number can be different objects:
```
ilya-g, line 127: Boxing happens not only when nullable references are involved, as it is said in the previous paragraph.

ilya-g: IMO, the new wording doesn't clear the confusion from the original issue.

PR #1708, opened 2020-05-29
Update collection-parts.md

MasoodFallahpoor:
Mention that functions chunked(), and windowed() are available since Kotlin 1.2

### DIFF
--- a/pages/docs/reference/collection-parts.md
+++ b/pages/docs/reference/collection-parts.md
@@ -75,7 +75,7 @@ fun main() {
 ```
 </div>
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hi! 
Thanks for your suggestion.
As an emerging technology, Kotlin develops really fast. We encourage our users to always use its latest version and keep the docs up with it. Thus, we avoid specific notes for users of older versions. (Yeah, you might find them somewhere in docs, by we remove them where possible).

Regards, Pavel
Kotlin Technical Writer

PR #1720, opened 2020-06-09
Use appropriate keyword for definition

sbagdat:
Defining an integer constant with an `L` suffix doesn't make sense. Because we can't change its value in the future. It will never be inside long type bounds.


p7nov:
Hi! 
Thanks for your suggestion.
However, there are cases when this makes sense, for example, when passing `1` to functions as a `Long` argument:
```
fun printLong(l: Long) { print(l) }
val i = 1
printLong(I) // Type mismatch!
```
If I write `val i = 1L`, this works.

Regards, Pavel
Kotlin Technical Writer

PR #1733, opened 2020-06-23
Fix in thread-based calculation.

Shakenbeer:
In order to provide a proper result, we need to wait for every thread to finish. Otherwise, it shows an arbitrary number instead of "500000500000". This could be confusing after a reader fixes the coroutine approach and get the right result.

### DIFF
--- a/pages/docs/tutorials/coroutines/coroutines-basic-jvm.md
+++ b/pages/docs/tutorials/coroutines/coroutines-basic-jvm.md
@@ -194,7 +194,7 @@ val c = AtomicLong()
 for (i in 1..1_000_000L)
     thread(start = true) {
         c.addAndGet(i)
-    }
+    }.join()
 
 println(c.get())
 ```


p7nov:
Hi! 
Could you provide more details or examples of returned arbitrary values?
I've never seen results other than "500000500000" when running this without `join`.

Shakenbeer:
@p7nov 
Let's take 10 threads for simplicity. Code, that I run:

```
fun main() {
    val c = AtomicLong(0)

    for (i in 1..10L)
        thread(start = true) {
            c.addAndGet(i)
            println(Thread.currentThread().name)
        }

    println(c.get())
}
```

The result should be 55, but it varies. And this is how log looks like:

```
Thread-0
Thread-2
Thread-3
8
Thread-4
Thread-1
Thread-8
Thread-7
Thread-5
Thread-9
Thread-6
```

Result (8) printed before all threads are finished. With `join()` everything works, as expected:

```
Thread-0
Thread-1
Thread-2
Thread-3
Thread-4
Thread-5
Thread-6
Thread-7
Thread-8
Thread-9
55
```

Hm... now I have that feeling, it's not async...

koshachy:
Thanks for your PR!
We've reworked the article and examples, so I close this PR.

PR #1737, opened 2020-06-27
Add info about class.javaObjectType

ileasile:


stachenov:
I was thinking to submit a PR about the same thing, but add more details, for example, describe the difference between `Int::class.javaObjectType` and `Integer.TYPE`: they refer to the same object at runtime, but have different compile time types: `Class<Int>` vs `Class<Integer>`, which is important for generics (type inference). Should I submit a separate PR?

And I'd also like to add a warning about not using `MyClass.javaClass` instead of `MyClass::class.java`. IDEA warns about it, though, so it's not that important, but it's nice to have in the docs nevertheless.

p7nov:
Thanks!

PR #1744, opened 2020-07-06
Rename POCO into POKO (was probably a typo)

LifeIsStrange:
I believe that POCO does not mean anything and that plain old Kotlin objects should be spelled POKO.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -9,7 +9,7 @@ title: "Idioms"
 
 A collection of random and frequently used idioms in Kotlin. If you have a favorite idiom, contribute it by sending a pull request.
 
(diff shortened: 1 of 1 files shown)


h0tk3y:
Note that POCO also means a 'plain old CLR object' in the .NET world. I'm not entirely sure it should read as `plain old Kotlin object' here in the title.

LifeIsStrange:
@h0tk3y interesting! But I still think that POKO would be a more important acronym to learn for someone learning Kotlin.

p7nov:
H!
Yep, @h0tk3y is right, it's a reference for those coming from other languages.
For now, there is no POKO term around, as far as I know.

Thanks for the suggestion, I'm closing this.

Best regards,
Pavel,
Kotlin Technical Writer

PR #1750, opened 2020-07-10
Add "when (true)" idiom

danieljsummers:
This lets you represent a series of `if` / `else if` / etc. / `else` statements very clearly and succinctly


h0tk3y:
But why use `when (true) {...}` instead of just `when {...}` with the same set of conditions and branches?

danieljsummers:
I completely missed that in the docs; my apologies.

If I just close this without merging will that clear it out?

p7nov:
Thanks, I'll close this.

PR #1754, opened 2020-07-14
feat: delegated properties improvements in 1.4

p7nov:
Documentation for new features of delegated properties in 1.4:
- [KT-27729](https://youtrack.jetbrains.com/issue/KT-27729) Inherit `ReadWriteProperty` from `ReadOnlyProperty`
- [KT-26494](https://youtrack.jetbrains.com/issue/KT-26494) Create an interface with provideDelegate()
- [KT-8658](https://youtrack.jetbrains.com/issue/KT-8658) Add property delegates which call get/set on the given KProperty instance, e.g. a property reference

### DIFF
--- a/pages/docs/reference/delegated-properties.md
+++ b/pages/docs/reference/delegated-properties.md
@@ -87,7 +87,7 @@ NEW has been assigned to ‘p’ in Example@33a17727.
 The specification of the requirements to the delegated object can be found [below](delegated-properties.html#property-delegate-requirements).
 
 Note that since Kotlin 1.1 you can declare a delegated property inside a function or code block, it shouldn't necessarily be a member of a class.
-Below you can find [the example](delegated-properties.html#local-delegated-properties-since-11).
+Below you can find [an example](delegated-properties.html#local-delegated-properties-since-11).
 
 ## Standard delegates
 
@@ -409,7 +409,7 @@ The Kotlin compiler provides all the necessary information about `prop` in the a
 to an instance of the outer class `C` and `this::prop` is a reflection object of the `KProperty` type describing `prop` itself.
 
 Note that the syntax `this::prop` to refer a [bound callable reference](reflection.html#bound-function-and-property-references-since-11)
-in the code directly is available only since Kotlin 1.1.  
+in the code directly has only been available since Kotlin 1.1.
 
 ### Providing a delegate
 
--- a/pages/docs/reference/delegated-properties.md
+++ b/pages/docs/reference/delegated-properties.md
@@ -155,15 +155,29 @@ The handler passed to the `vetoable` is called _before_ the assignment of a new 
 
 ## Delegating to another property
 
-A property can delegate its getter and setter to another property of the same class. In this case, use the property name
-with the `this::` qualifier as the delegate name.
+Since Kotlin 1.4, a property can delegate its getter and setter to another property. Such delegation is available for
+both top-level and class properties (member and extension). The delegate property can be:
+- a top-level property
+- a member or an extension property of the same class
+- a member or an extension property of another class
+
+To delegate a property to another property, use the proper `::` qualifier in the delegate name, for example, `this::delegate` or
+`MyClass::delegate`.
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
 ```kotlin
-    var delegate: Int
-    // ...
-    var delegatedProperty: Int by this::delegate
+
+var topLevelDelegate: Int
+class MyClass(var memberDelegate: Int) {
+    var delegatedToMember: Int by this::memberDelegate
+    var delegatedToTopLevel: Int by ::topLevelDelegate
+    val delegegateToAnotherClass: Int by ClassWithDelegate::anotherClassDelegate
+}
+class ClassWithDelegate(val anotherClassDelegate: Int)
+
+var MyClass.extDelegated: Int by ::topLevelDelegate
+
 ```
 
 </div>
@@ -258,7 +272,7 @@ class MutableUser(val map: MutableMap<String, Any?>) {
 
 </div>
 
-## Local delegated properties (since 1.1)
+## Local delegated properties
 
 You can declare local variables as delegated properties.
 For instance, you can make a local variable lazy:
@@ -343,17 +357,22 @@ class ResourceDelegate(private var resource: Resource = Resource()) {
 The latter is handy when you need to delegate property to an object which doesn't originally provide these functions.
 Both of the functions need to be marked with the `operator` keyword.
 
-You can create delegate classes by implementing interfaces `ReadOnlyProperty` and `ReadWriteProperty` from the Kotlin standard
-library. They provide the required methods: `getValue()` is declared in `ReadOnlyProperty`; `ReadWriteProperty`
+You can create delegates as anonymous objects without creating new classes using the interfaces `ReadOnlyProperty`
+and `ReadWriteProperty` from the Kotlin standard library.
+They provide the required methods: `getValue()` is declared in `ReadOnlyProperty`; `ReadWriteProperty`
 extends it and adds `setValue()`. Thus, you can pass a `ReadWriteProperty` whenever a `ReadOnlyProperty` is expected.
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only auto-indent="false">
 
 ```kotlin
-fun resourceDelegate(): ReadWriteProperty<Nothing?, Int> =
-   object : ReadWriteProperty<Nothing?, Int> {
-      //...
-   }
+fun resourceDelegate(): ReadWriteProperty<Any?, Int> =
+    object : ReadWriteProperty<Any?, Int> {
+        var curValue = 0 
+        override fun getValue(thisRef: Any?, property: KProperty<*>): Int = curValue
+        override fun setValue(thisRef: Any?, property: KProperty<*>, value: Int) {
+            curValue = value
+        }
+    }
 
 val readOnly: Int by resourceDelegate()  // ReadWriteProperty as val
 var readWrite: Int by resourceDelegate()
@@ -391,14 +410,13 @@ to an instance of the outer class `C` and `this::prop` is a reflection object of
 Note that the syntax `this::prop` to refer a [bound callable reference](reflection.html#bound-function-and-property-references-since-11)
 in the code directly is available only since Kotlin 1.1.  
 
-### Providing a delegate (since 1.1)
+### Providing a delegate
 
 By defining the `provideDelegate` operator you can extend the logic of creating the object to which the property implementation
 is delegated. If the object used on the right-hand side of `by` defines `provideDelegate` as a member or extension function,
 that function will be called to create the property delegate instance.
 
-One of the possible use cases of `provideDelegate` is to check property consistency when the property is created,
-not only in its getter or setter.
+One of the possible use cases of `provideDelegate` is to check the consistency of the property upon its initialization.
 
 For example, if you want to check the property name before binding, you can write something like this:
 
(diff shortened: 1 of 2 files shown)


pages/docs/reference/delegated-properties.md
```diff
@@ -153,7 +153,45 @@ fun main() {
 If you want to intercept assignments and "veto" them, use [`vetoable()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.properties/-delegates/vetoable.html) instead of `observable()`.
 The handler passed to the `vetoable` is called _before_ the assignment of a new property value has been performed.
 
-## Storing Properties in a Map
+## Delegating to another property
+
+A property can delegate its getter and setter to another property of the same class. In this case, use the property name
+with the `this::` qualifier as the delegate name.
```
ilya-g, line 159: Not only `::this` and not only the same class. See all possible options in this test: https://github.com/JetBrains/kotlin/blob/ea1e16e309429ec63c8e619ec5db2c6fe1485e0b/libraries/stdlib/test/properties/delegation/PropertyReferenceTest.kt#L27

pages/docs/reference/delegated-properties.md
```diff
@@ -305,28 +343,29 @@ class ResourceDelegate(private var resource: Resource = Resource()) {
...
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only auto-indent="false">
 
 ```kotlin
-interface ReadOnlyProperty<in R, out T> {
-    operator fun getValue(thisRef: R, property: KProperty<*>): T
-}
+fun resourceDelegate(): ReadWriteProperty<Nothing?, Int> =
+   object : ReadWriteProperty<Nothing?, Int> {
+      //...
```
ilya-g, line 355: Please add the overrides.

pages/docs/reference/delegated-properties.md
```diff
@@ -346,17 +385,20 @@ class C {
...
-By defining the `provideDelegate` operator you can extend the logic of creating the object to which the property implementation is delegated.
-If the object used on the right hand side of `by` defines `provideDelegate` as a member or extension function, that function will be
-called to create the property delegate instance.
+By defining the `provideDelegate` operator you can extend the logic of creating the object to which the property implementation
+is delegated. If the object used on the right-hand side of `by` defines `provideDelegate` as a member or extension function,
+that function will be called to create the property delegate instance.
 
-One of the possible use cases of `provideDelegate` is to check property consistency when the property is created, not only in its getter or setter.
+One of the possible use cases of `provideDelegate` is to check property consistency when the property is created,
+not only in its getter or setter.
```
ilya-g, line 401: Minor: the first part is about 'when' and the opposite is about 'where'. Perhaps, makes sense to reword both parts to be about the same category.
Also "property is created" sounds vague, because properties are not objects themselves.

pages/docs/reference/delegated-properties.md
```diff
@@ -153,7 +153,45 @@ fun main() {
 If you want to intercept assignments and "veto" them, use [`vetoable()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.properties/-delegates/vetoable.html) instead of `observable()`.
 The handler passed to the `vetoable` is called _before_ the assignment of a new property value has been performed.
 
-## Storing Properties in a Map
+## Delegating to another property
```
ilya-g, line 156: We can specify "(since 1.4)" as we did in the other headings here.

ilya-g: Though I'd devise a label or note for that, rather than putting that in a heading.

pages/docs/reference/delegated-properties.md
```diff
@@ -444,4 +487,21 @@ class C {
 
 </div>
 
-Note that the `provideDelegate` method affects only the creation of the auxiliary property and doesn't affect the code generated for getter or setter.
+Note that the `provideDelegate` method affects only the creation of the auxiliary property and doesn't affect the code
+generated for getter or setter.
+
+With the `PropertyDelegateProvider` interface from the standard library, you can create delegate providers as anonymous
+objects without creating new classes.
```
ilya-g, line 494: > as anonymous objects without creating new classes.

This note would be also valuable for implementing property delegates with `ReadOnlyProperty` and `ReadWriteProperty` delegates.

pages/docs/reference/delegated-properties.md
```diff
@@ -444,4 +487,21 @@ class C {
...
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only auto-indent="false">
+
+```kotlin
+fun makeDelegate() = PropertyDelegateProvider<Nothing?, Int> =
+    object : PropertyDelegateProvider<Nothing?, Int> {
```
ilya-g, line 500: Makes sense to showcase SAM constructor syntax being available due to the fact that `PropertyDelegateProider ` is a fun interface

pages/docs/reference/delegated-properties.md
```diff
@@ -444,4 +487,21 @@ class C {
...
+Note that the `provideDelegate` method affects only the creation of the auxiliary property and doesn't affect the code
+generated for getter or setter.
+
+With the `PropertyDelegateProvider` interface from the standard library, you can create delegate providers as anonymous
+objects without creating new classes.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only auto-indent="false">
+
+```kotlin
+fun makeDelegate() = PropertyDelegateProvider<Nothing?, Int> =
```
ilya-g, line 499: `<Nothing?, Int>` — that seems to be a quite strange combination of type arguments.

- `Nothing?` for the property owner type limits delegation to top-level properties only. If the property owner type doesn't matter, better to specify it as `Any?`.
- `Int`: here should be the delegate object type, not the property type. While it's possible to define delegation operators for `Int` as extensions, it's usually not the case.

pages/docs/reference/delegated-properties.md
ilya-g: See this example in the test related to `PropertyDelegateProvider`: https://github.com/JetBrains/kotlin/blob/de6154980d3f16ccf2bf0fa194ae74802dd51dbc/libraries/stdlib/test/properties/delegation/DelegationTest.kt#L71

pages/docs/reference/delegated-properties.md
p7nov: Yes, we're already getting rid of such notes in headings while moving to WebHelp.
We'll think up some kind of label for versions where they are important, and completely remove the old ones such as 1.1

pages/docs/reference/delegated-properties.md
```diff
@@ -153,7 +153,59 @@ fun main() {
...
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+
+var topLevelDelegate: Int
+class MyClass(var memberDelegate: Int) {
+    var delegatedToMember: Int by this::memberDelegate
+    var delegatedToTopLevel: Int by ::topLevelDelegate
+    val delegegateToAnotherClass: Int by ClassWithDelegate::anotherClassDelegate
```
ilya-g, line 175: typo
```suggestion
    val delegateToAnotherClass: Int by ClassWithDelegate::anotherClassDelegate
```

pages/docs/reference/delegated-properties.md
```diff
@@ -153,7 +153,59 @@ fun main() {
...
+- a member or an extension property of another class
+
+To delegate a property to another property, use the proper `::` qualifier in the delegate name, for example, `this::delegate` or
+`MyClass::delegate`.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+
+var topLevelDelegate: Int
```
ilya-g, line 171: IMO it's kinda confusing to call such property a delegate. What serves as a delegate here is a property reference, not the property itself.

pages/docs/reference/delegated-properties.md
ilya-g: Also this likely won't work because the receiver of an unbound property reference must be able to take the receiver of the delegated property, and `ClassWithDelegate` is not a supertype of `MyClass`.

Possible options:
- make `ClassWithDelegate` a supertype of `MyClass`. But then its properties will be inherited, and delegation with an unbound reference won't be required.
- use a bound reference: pass an instance of ClassWithDelegate somehow and delegate to `thatInstance::property`

pages/docs/reference/delegated-properties.md
```diff
@@ -89,7 +89,7 @@ The specification of the requirements to the delegated object can be found [belo
 Note that since Kotlin 1.1 you can declare a delegated property inside a function or code block, it shouldn't necessarily be a member of a class.
 Below you can find [the example](delegated-properties.html#local-delegated-properties-since-11).
```
B1ggDave, line 90: I would say "an example" 

as in: below you can find an example [of this]

pages/docs/reference/delegated-properties.md
```diff
@@ -242,7 +295,7 @@ fun example(computeFoo: () -> Foo) {
 The `memoizedFoo` variable will be computed on the first access only.
 If `someCondition` fails, the variable won't be computed at all.
 
-## Property Delegate Requirements
+## Property delegate requirements
 
 Here we summarize requirements to delegate objects.
```
B1ggDave, line 300: Here we summarize the requirements to delegate objects.

pages/docs/reference/delegated-properties.md
```diff
@@ -346,17 +405,19 @@ class C {
 
 </div>
 
-The Kotlin compiler provides all the necessary information about `prop` in the arguments: the first argument `this` refers to an instance of the outer class `C` and `this::prop` is a reflection object of the `KProperty` type describing `prop` itself.
+The Kotlin compiler provides all the necessary information about `prop` in the arguments: the first argument `this` refers
+to an instance of the outer class `C` and `this::prop` is a reflection object of the `KProperty` type describing `prop` itself.
 
-Note that the syntax `this::prop` to refer a [bound callable reference](reflection.html#bound-function-and-property-references-since-11) in the code directly is available only since Kotlin 1.1.  
+Note that the syntax `this::prop` to refer a [bound callable reference](reflection.html#bound-function-and-property-references-since-11)
+in the code directly is available only since Kotlin 1.1.
```
B1ggDave, line 412: in the code directly has only been available since Kotlin 1.1.

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,19 @@ title: "What's New in Kotlin 1.4"
 
 # What's New in Kotlin 1.4
 
+## Delegated properties improvements
+
+In 1.4, we add new features to improve your experience with delegated properties in Kotlin:
```
B1ggDave, line 11: In 1.4, we have added new features to improve your experience with delegated properties in Kotlin:

is it "delegated properties" or "delegate properties"?

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,19 @@ title: "What's New in Kotlin 1.4"
...
+- now a property can be delegated to another property
```
B1ggDave, line 12: - Now a property can be delegated to another property.
- A new interface `PropertyDelegateProvider` helps create delegate providers in a single declaration.
- `ReadWriteProperty` now extends `ReadOnlyProperty` so you can use both of them for read-only properties.

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,19 @@ title: "What's New in Kotlin 1.4"
...
+- new interface `PropertyDelegateProvider` helps create delegate providers in a single declaration
+- `ReadWriteProperty` now extends `ReadOnlyProperty` so that you can use both of them for read-only properties
+
+Aside from new API, we've done some optimizations that reduce the resulting bytecode size. These optimizations are described
```
B1ggDave, line 16: Aside from the new API, we've made some optimizations that reduce the resulting bytecode size. These optimizations are described in this blog post.

pages/docs/reference/whatsnew14.md
p7nov: it's "delegated properties"

PR #1756, opened 2020-07-16
feat: sam conversions for Kotlin interfaces

katia-energizer:
Added a new topic about Functional (SAM) interfaces; updated Java interop.


pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+layout: reference
+title: "Functional interfaces (SAM interfaces)"
+---
+
+# Functional (SAM) interfaces
+
+An interface with only one abstract method is called a _functional interface_, or a _Single Abstract 
+Method (SAM) interface_.
+
+Functional interfaces in Kotlin are marked with the `fun` keyword.
```
zarechenskiy, line 12: Here it'd be more correct to use the word `modifier`:
```
... are marked with the `fun` modifier
```

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
+---
+type: doc
+layout: reference
+title: "Functional interfaces (SAM interfaces)"
+---
+
+# Functional (SAM) interfaces
```
zarechenskiy, line 7: I propose adding a note somewhere to state that this functionality is available only since Kotlin 1.4

zarechenskiy: Also, let's put (SAM) at the end: Functional Interfaces (SAM)

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+
+An interface with only one abstract method is called a _functional interface_, or a _Single Abstract
```
zarechenskiy, line 9: For other parts of doc (data classes, sealed classes...), the description starts from use case explanation.
I think we have a good example here https://kotlinlang.org/docs/reference/java-interop.html#sam-conversions

Something like:
```
Since 1.4, Kotlin supports functional interface conversions (SAM conversions). This means that Kotlin function 
literals can be automatically converted into implementations of _functional_ interfaces with a single abstract 
method, as long as the parameter types of the interface method match the parameter types of the Kotlin 
function.

To declare a functional interface, one puts the `fun` modifier before the interface:

<example>
```

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+    fun run()
+}
+```
+
+</div>
+
+## SAM conversions
+
+For functional interfaces, you can use SAM conversions that help you make your code more concise and readable thanks to 
+lambda expressions.
```
zarechenskiy, line 27: I believe this sentence can be omitted if we extend the explanation in `Functional (SAM) interfaces`. Plus, note that "thanks to" expression is also used in the next paragraph

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+
+Instead of creating a class that implements a functional interface manually, you can use a lambda expression. 
+Thanks to a SAM conversion, Kotlin can transparently convert any lambda expression whose signature matches
```
zarechenskiy, line 30: The word `transparently` doesn't clear here to me :) Maybe we can omit it?

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+## SAM conversions
+
+For functional interfaces, you can use SAM conversions that help you make your code more concise and readable thanks to 
+lambda expressions.
+
+Instead of creating a class that implements a functional interface manually, you can use a lambda expression. 
+Thanks to a SAM conversion, Kotlin can transparently convert any lambda expression whose signature matches 
+the signature of the interface's single method into an instance of a class that implements the interface.
+
+For example, consider the following Kotlin functional interface:
```
zarechenskiy, line 33: I think this sentence also can be omitted (I have a bad habit putting it into my docs :))

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+fun interface Sum {
+    fun add(a: Int, b: Int): Int
+}
```
zarechenskiy, line 40: Not a very good example: `Sum` means some specific operation while `fun` interface should state some meta-functionality over two ints here. Maybe it'd be better to use this example:
```
fun interface IntPredicate {
    fun accept(i: Int): Boolean
}

val isEven = IntPredicate { it % 2 == 0 }
```

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+</div>
+
+If you don't use lambdas, you will need to write code like this:
```
zarechenskiy, line 45: To me something like "Before Kotlin 1.4. you have to write code like this..." sounds better

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,101 @@
...
+val sum = object : Sum {
+    override fun add(a: Int, b: Int): Int {
+        return a + b
+    }
+}
+```
+
+</div>
+
+That's a lot of unnecessary code, which is also not very readable. By leveraging Kotlin's SAM conversion
```
zarechenskiy, line 60: "That's a lot of unnecessary code, which is also not very readable"

- it should be clear from the example, so let's omit this sentence

pages/docs/reference/java-interop.md
```diff
@@ -752,9 +752,6 @@ executor.execute(Runnable { println("This runs in a thread pool") })
 Note that SAM conversions only work for interfaces, not for abstract classes, even if those also have just a single
 abstract method.
 
-Also note that this feature works only for Java interop; since Kotlin has proper function types, automatic conversion
-of functions into implementations of Kotlin interfaces is unnecessary and therefore unsupported.
```
zarechenskiy, line 756: Let's provide a link to `fun` interface here and say something like "Starting from version 1.4, it's possible to use `fun` interace for SAM conversions..."

pages/docs/reference/fun-interfaces.md
katia-energizer: > I propose adding a note somewhere to state that this functionality is available only since Kotlin 1.4

These docs will be published after 1.4 release. Our docs cover the latest version of Kotlin so we don't need to add _since Kotlin 1.4_. 
Also I would avoid using versions in documentation. It's ok to use them in What's new (I haven't prepared it yet). Our documentation already includes _since 1.1, since 1.2, since 1.3_. It looks very strange and is a bad practice (what's the purpose?). We will remove these _since_ when migrating to WH.

Docs describe the latest Kotlin version so our goal is to motivate people use it earlier. What's new covers when a specific feature was added consistently for all features.

katia-energizer: > Also, let's put (SAM) at the end: Functional Interfaces (SAM)

I had this initially but it was not gramatically correct. SAM is not the same as Functional interfaces, right?. _Functional interfaces (SAM interfaces)_ looks a bit long. I will consult with our proofreaders when I send them this content to proofread.

pages/docs/reference/fun-interfaces.md
katia-energizer: The more good realistic examples, the better. 
It's an example of a conversion. I would start with a functional interface with a good example and then explain conversions. Mixing one with another seems overloading and makes it harder to understand.

BTW, in the Internet, there are many materials about Kotlin that explain everything in easier ways with simple words. The easier we write, the more people will adopt it. 
Some language articles on our web site are too overloaded with words and complex grammatical constructions. I'm for simple language and explanation. :) 

As for explanation of a functional interface, what can we add (without conversions)?

pages/docs/reference/fun-interfaces.md
katia-energizer: Why do you think it's a bad habit? It's good to explain what you do. If I just put a code snippet after generic explanation, it would look strange. Docs should be easy to read and understand. We are not limited in a number of words but all words should be helpful.

pages/docs/reference/fun-interfaces.md
katia-energizer: removed.

pages/docs/reference/fun-interfaces.md
katia-energizer: updated the example. please take a look. 
it's strange that it doesn't work on playground but works in IDE.

pages/docs/reference/fun-interfaces.md
katia-energizer: Agree. removed.

pages/docs/reference/java-interop.md
katia-energizer: Added cross-links between articles. Thx! I forgot 🙈
But as I described above, I don't use versions except for What's new. 🙂

katia-energizer: BTW, in docs we use "you" according to the JB style guide (Google and MS style guides as well). It's better to avoid "it's possible, one can". We refer to the reader. This makes docs more personal and friendly. 🤗

pages/docs/reference/fun-interfaces.md
katia-energizer: thanks. updated.

pages/docs/reference/fun-interfaces.md
p7nov: I'd add a code comment that this is what you write _without SAM conversion_.
Our readers often scan the page stopping only on code snippets, so if you show a "bad" example, you better mark that it's "bad" **in the snippet** itself.

pages/docs/reference/fun-interfaces.md
katia-energizer: I have updated examples and wording.

pages/docs/reference/fun-interfaces.md
katia-energizer: Kept. Removed "thanks to" in one sentence 😊

pages/docs/reference/fun-interfaces.md
```diff
@@ -0,0 +1,127 @@
...
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.4-M1">
+
+```kotlin
+fun interface IntPredicate {
+    fun accept(i: Int): Boolean
+}
+
+val isEven = IntPredicate { it % 2 == 0 }
+
+fun main(){
```
zarechenskiy, line 82: Formatting: no space between `()` and `{`

pages/docs/reference/fun-interfaces.md
katia-energizer: ✅

pages/docs/reference/fun-interfaces.md
katia-energizer, line 82:
```suggestion
fun main() {
```

katia-energizer: ✅

PR #1758, opened 2020-07-16
update: 1.4 new -Xjvm-default modes

p7nov:
Doc update for new -Xjvm-default modes added in 1.4

### DIFF
--- a/pages/docs/reference/java-to-kotlin-interop.md
+++ b/pages/docs/reference/java-to-kotlin-interop.md
@@ -399,7 +399,7 @@ public class BB8 implements Robot {
 
 >**Note**: Prior to Kotlin 1.4, to generate default methods, you could use the `@JvmDefault` annotation on these methods.
 > Compiling with `-Xjvm-default=all` in 1.4 generally works as if you annotated all non-abstract methods of interfaces
-> with `@JvmDefault`and compiled with `-Xjvm-default=enabled`. However, there are cases when their behavior differs.
+> with `@JvmDefault`and compiled with `-Xjvm-default=enable`. However, there are cases when their behavior differs.
 > The detailed information about changes in default methods generation in Kotlin 1.4 is provided in [this post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/)
 > on the Kotlin blog.
 
--- a/pages/docs/reference/java-to-kotlin-interop.md
+++ b/pages/docs/reference/java-to-kotlin-interop.md
@@ -400,7 +400,7 @@ public class BB8 implements Robot {
 >**Note**: Prior to Kotlin 1.4, to generate default methods, you could use the `@JvmDefault` annotation on these methods.
 > Compiling with `-Xjvm-default=all` in 1.4 generally works as if you annotated all non-abstract methods of interfaces
 > with `@JvmDefault`and compiled with `-Xjvm-default=enable`. However, there are cases when their behavior differs.
-> The detailed information about changes in default methods generation in Kotlin 1.4 is provided in [this post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/)
+> Detailed information about the changes in default methods generation in Kotlin 1.4 is provided in [this post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/)
 > on the Kotlin blog.
 
 ### Compatibility mode for default methods
@@ -412,12 +412,12 @@ To avoid breaking the compatibility with such clients, compile your Kotlin code 
 the `-Xjvm-default=all-compatibility` compiler option. In this case, all the code that uses the previous version will 
  work fine with the new one. However, the compatibility mode adds some overhead to the resulting bytecode size.
 
-There is no need in compatibility for the new interfaces you add because no client code uses their previous versions.
-Thus, you can minimize the compatibility overhead by excluding these interfaces from the compatibility mode.
-To do this, annotate them with the `@JvmDefaultWithoutCompatibility` annotation. Such interfaces compile the same way as
-with `-Xjvm-default=all`. 
+There is no need to consider compatibility for new interfaces, as no clients have used them before.
+You can minimize the compatibility overhead by excluding these interfaces from the compatibility mode.
+To do this, annotate them with the `@JvmDefaultWithoutCompatibility` annotation. Such interfaces compile the same way as 
+with `-Xjvm-default=all`.
 
-Additionally, in the `all-compatibility` mode you can use  `@JvmDefaultWithoutCompatibility` to annotate all interfaces 
+Additionally, in the `all-compatibility` mode you can use `@JvmDefaultWithoutCompatibility` to annotate all interfaces
 which are not exposed in the public API and therefore aren’t used by the existing clients.
 
 ## Visibility
--- a/pages/docs/reference/java-to-kotlin-interop.md
+++ b/pages/docs/reference/java-to-kotlin-interop.md
@@ -351,8 +351,9 @@ interface Robot {
 ```
 </div>
 
-Compiling with `-Xjvm-default=all` has the same effect as annotating all non-abstract methods of interfaces with
-`@JvmDefault` and compiling with `-Xjvm-default=enabled`.
+>**Note**: Compiling with `-Xjvm-default=all` generally works as if you annotated all non-abstract methods of interfaces
+>with `@JvmDefault`and compiled with `-Xjvm-default=enabled`. This was the way to compile methods of Kotlin interfaces
+>to `default` in Kotlin prior to 1.4.
 
 The default implementation is available for Java classes implementing the interface.
 
@@ -402,12 +403,12 @@ public class BB8 implements Robot {
 
 ### Compatibility mode for default methods
 
-If there is client Java code that uses your Kotlin interfaces compiled without `-Xjvm-default`, then recompiling it with
-`-Xjvm-default=all` may break that Java code. 
+If there are clients that use your Kotlin interfaces compiled without `-Xjvm-default`, then they can be incompatible with
+ the same code compiled with `-Xjvm-default=all`. 
 
-To avoid breaking the Java code that uses your Kotlin code compiled without `-Xjvm-default`, use the `-Xjvm-default=all-compatibility`
-compiler option. In this case, all the interface implementations written for the previous versions will be fully compatible
-with the new version. However, the compatibility mode adds some overhead to the resulting bytecode size.
+To avoid breaking the compatibility with such clients, compile your Kotlin code in the _compatibility mode_ by specifying
+the `-Xjvm-default=all-compatibility` compiler option. In this case, all the code that uses the previous version will 
+ work fine with the new one. However, the compatibility mode adds some overhead to the resulting bytecode size.
 
 There is no need in compatibility for the new interfaces you add because no client code uses their previous versions.
 Thus, you can minimize the compatibility overhead by excluding these interfaces from the compatibility mode.
--- a/pages/docs/reference/java-to-kotlin-interop.md
+++ b/pages/docs/reference/java-to-kotlin-interop.md
@@ -333,19 +333,19 @@ making its getter and setter methods static members in that object or the class 
 > Default methods are available only for targets JVM 1.8 and above.
 {:.note}
 
-> The `@JvmDefault` annotation is experimental in Kotlin 1.3. Its name and behavior may change, leading to future incompatibility.
-{:.note}
-
 Starting from JDK 1.8, interfaces in Java can contain [default methods](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html).
-You can declare a non-abstract member of a Kotlin interface as default for the Java classes implementing it.
-To make a member default, mark it with the [`@JvmDefault`](/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/index.html) annotation.
+To make all non-abstract members of Kotlin interfaces default for the Java classes implementing them, compile the Kotlin 
+code with the `-Xjvm-default=all` compiler option.
+
 Here is an example of a Kotlin interface with a default method:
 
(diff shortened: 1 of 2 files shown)


pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,26 @@ public class BB8 implements Robot {
...
-   This includes adding the entire interface for your API.
-* `-Xjvm-default=compatibility` should be used if you are adding a `@JvmDefault` to the methods that were available in the API before.
-   This mode helps avoid compatibility breaks: all the interface implementations written for the previous versions will be fully compatible with the new version.
-   However, the compatibility mode may add some overhead to the resulting bytecode size and affect the performance.
-
-For more details about compatibility issues, see the `@JvmDefault` [reference page](/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/index.html).
+### Compatibility mode for default methods
 
-### Using in delegates
+If there is client Java code that uses your Kotlin interfaces compiled with Kotlin 1.3 or older, then recompiling it with
```
udalov, line 402: The version 1.3 is not important here, what's important is that these interfaces were compiled in the old DefaultImpls-based compilation mode, which is for example still the default mode in 1.4. Maybe we could word this somehow to avoid mentioning 1.3?

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,26 @@ public class BB8 implements Robot {
...
-
-fun main() {
-    val prod = ProducerImpl()
-    DelegatedProducer(prod).produce() // prints "interface method"
-}
-```
-</div>
+To avoid breaking the Java code that uses your Kotlin code compiled with 1.3 or older versions, use the `-Xjvm-default=all-compatibility`
+compiler option. In this case, all the interface implementations written for the previous versions will be fully compatible
+with the new version. However, the compatibility mode adds some overhead to the resulting bytecode size and affects the performance.
```
udalov, line 407: Let's remove this part. As far as I know it doesn't affect performance :)

```suggestion
with the new version. However, the compatibility mode adds some overhead to the resulting bytecode size.
```

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,26 @@ public class BB8 implements Robot {
...
 
-For more details about interface delegation in Kotlin, see [Delegation](/docs/reference/delegation.html).
+There is no need in compatibility for the new interfaces you add because no client code uses their previous versions.
+Thus, you can minimize the compatiblity overhead by excluding these interfaces from the compatibility mode.
+To do this, annotate them with the `@JvmDefaultWithoutCompatibility` annotation. Such interfaces compile the same way as
+with `-Xjvm-default=all`.
 
+Additionally, in the `all-compatibility` mode you can use  `@JvmDefaultWithoutCompatibility` to annotate all interfaces 
+not exposed in the public API and therefore aren’t used by the existing clients.
```
udalov, line 415:
```suggestion
which are not exposed in the public API and therefore aren’t used by the existing clients.
```

pages/docs/reference/java-to-kotlin-interop.md
p7nov: Is it correct to say _ compiled without -Xjvm-default option_? Will this be equal to _DefaultImpls-based compilation mode_?

max-kammerer: Main point of compatibility modes is binary compatibility, not source one. (But it also could be used as source one, but it's minor compare to binary one)

max-kammerer: And it's applies for Kotlin and Java clients

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -333,24 +333,27 @@ making its getter and setter methods static members in that object or the class
...
+
 interface Robot {
-    @JvmDefault fun move() { println("~walking~") }
+    fun move() { println("~walking~") }  // will be default in the Java interface
     fun speak(): Unit
 }
 ```
 </div>
 
+Compiling with `-Xjvm-default=all` has the same effect as annotating all non-abstract methods of interfaces with
```
max-kammerer, line 354: It's not totatlly true because of delegation by difference

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +400,26 @@ public class BB8 implements Robot {
...
-    }
-}
+### Compatibility mode for default methods
 
-class ProducerImpl: Producer {
-    override fun produce() {
-        println("class method")
-    }
-}
+If there is client Java code that uses your Kotlin interfaces compiled without `-Xjvm-default`, then recompiling it with
```
max-kammerer, line 405: Actually I'm not sure that we should write anything about compatibility mode in Calling Kotlin from Java' section

p7nov: Can you share your thoughts about better places for this content? 
I mean, it's about compatibility for this exact feature, hardly readers will look for this compatibility mode aside from default interfaces in general...
@udalov what do you think?

pages/docs/reference/java-to-kotlin-interop.md
p7nov: Since we're gonna deprecate `@JvmDefault` in 1.4.20, maybe we can skip this detail? Just add `In general` or `In most cases`.
Otherwise, we'll have to keep the entire section about delegation with `@JvmDefault` which gets deprecated really soon.

pages/docs/reference/java-to-kotlin-interop.md
udalov: Technically not, because there are old modes `-Xjvm-default=enable`, `-Xjvm-default=compatibility` which are still "DefaultImpls-based" for all methods which are not annotated with the old annotation `@JvmDefault`.

But maybe something like "compiled without the _new_ -Xjvm-default options" would work? Or "compiled with the old mode (with method bodies in DefaultImpls)"

pages/docs/reference/java-to-kotlin-interop.md
udalov: +1 for "in general" / "in most cases". Also, maybe clarify in a footnote, if footnotes are supported? :)

pages/docs/reference/java-to-kotlin-interop.md
udalov: Why not though? I think it's fine to have it here. Some users might read the section before this and decide that `-Xjvm-default=all` is a perfect choice for their project, only to find out later that they actually also needed to keep binary compatibility for some reason.

Having this section of course wouldn't completely solve this problem, but it's at least an attempt to present the full picture.

Of course, we need not delve into details about the actual problems that the all-compatibility mode has, for that we can add a link to the [blog post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/).

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -333,24 +333,28 @@ making its getter and setter methods static members in that object or the class
...
+
 interface Robot {
-    @JvmDefault fun move() { println("~walking~") }
+    fun move() { println("~walking~") }  // will be default in the Java interface
     fun speak(): Unit
 }
 ```
 </div>
 
+>**Note**: Compiling with `-Xjvm-default=all` generally works as if you annotated all non-abstract methods of interfaces
```
udalov, line 354: I think we should rather extract this into a separate section named `@JvmDefault` in the end of this chapter, which would clearly say that "previously, to generate default methods you had to use this annotation, but it's not needed anymore, -Xjvm-default=all does it for you; JvmDefault is deprecated now", etc.

pages/docs/reference/java-to-kotlin-interop.md
max-kammerer: I thinks it's bad example to use `all-compatibility` mode for source compatibility cases. It's best to switch Java code to direct default methods usage - it would be more elegant and right from Java point of view. 
We also forgot that `DefaultImpls` members are deprecated in compatibility mode!
And we plan that with time `all-compatibility` mode would be subsituted with 'all' one.

So I suggest to mention `all-compatibility` mode only in case of binary compatibility issues.

udalov: Those are good points. Maybe we should be more explicit with the fact that it only helps with binary, not source, compatibility. So instead of saying "If there is client Java code", say "If there is compiled code", and so on. And then clarify that although source compatibility is _kind of_ provided, it's recommended to rewrite Java code to use default methods, since `DefaultImpls` is going to be removed in a future version.

pages/docs/reference/java-to-kotlin-interop.md
p7nov: I don't really like the idea of creating a dedicated section for `@JvmDefault`. People can think that it's an equal alternative, not an experimental mechanism from the past. 
Additionally, we'll have to support its anchor URL after removing it when we deprecate the annotation (in a couple of months, right?).
I'd put the text you suggest in a note (so that readers distinguish it from the common text flow) and provide a link to Sveta's blog post about how this all changes in 1.4.

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,32 @@ public class BB8 implements Robot {
...
-
-```kotlin
-interface Producer {
-    @JvmDefault fun produce() {
-        println("interface method")
-    }
-}
+>**Note**: Prior to Kotlin 1.4, to generate default methods, you could use the `@JvmDefault` annotation on these methods.
+> Compiling with `-Xjvm-default=all` in 1.4 generally works as if you annotated all non-abstract methods of interfaces
+> with `@JvmDefault`and compiled with `-Xjvm-default=enabled`. However, there are cases when their behavior differs.
```
udalov, line 402: **`...=enable`**

pages/docs/reference/java-to-kotlin-interop.md
udalov: Great, thanks :)

> (in a couple of months, right?)

Yes, https://youtrack.jetbrains.com/issue/KT-40392.

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,32 @@ public class BB8 implements Robot {
...
-```kotlin
-interface Producer {
-    @JvmDefault fun produce() {
-        println("interface method")
-    }
-}
+>**Note**: Prior to Kotlin 1.4, to generate default methods, you could use the `@JvmDefault` annotation on these methods.
+> Compiling with `-Xjvm-default=all` in 1.4 generally works as if you annotated all non-abstract methods of interfaces
+> with `@JvmDefault`and compiled with `-Xjvm-default=enable`. However, there are cases when their behavior differs.
+> The detailed information about changes in default methods generation in Kotlin 1.4 is provided in [this post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/)
```
B1ggDave, line 403: Detailed information about the changes in default methods generation in Kotlin 1.4 is provided in [this post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/)

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,32 @@ public class BB8 implements Robot {
...
-    DelegatedProducer(prod).produce() // prints "interface method"
-}
-```
-</div>
+To avoid breaking the compatibility with such clients, compile your Kotlin code in the _compatibility mode_ by specifying
+the `-Xjvm-default=all-compatibility` compiler option. In this case, all the code that uses the previous version will 
+ work fine with the new one. However, the compatibility mode adds some overhead to the resulting bytecode size.
 
-For more details about interface delegation in Kotlin, see [Delegation](/docs/reference/delegation.html).
+There is no need in compatibility for the new interfaces you add because no client code uses their previous versions.
```
B1ggDave, line 415: not sure what this means "There is no need in compatibility for the new interfaces you add because no client code uses their previous versions."

There is no need to consider compatibility for new interfaces, as no clients have used it before

pages/docs/reference/java-to-kotlin-interop.md
```diff
@@ -397,53 +397,32 @@ public class BB8 implements Robot {
...
+Thus, you can minimize the compatibility overhead by excluding these interfaces from the compatibility mode.
```
B1ggDave, line 416: You can minimize the compatibility overhead by excluding these interfaces from the compatibility mode.
To do this, annotate them with the `@JvmDefaultWithoutCompatibility` annotation. Such interfaces compile the same way as with `-Xjvm-default=all`. 

Additionally, in the `all-compatibility` mode you can use  `@JvmDefaultWithoutCompatibility` to annotate all interfaces 
which are not exposed in the public API and therefore aren’t used by the existing clients.

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,19 @@ title: "What's New in Kotlin 1.4"
 
 # What's New in Kotlin 1.4
 
+## New modes for generating default methods
+
+When compiling Kotlin code to targets JVM 1.8 and above, you could compile non-abstract methods of Kotlin interfaces into
+Java's `default` methods. For this purpose, there was a mechanism that includes the `@JvmDefault` annotation for marking
+such methods and the `-Xjmv-default` compiler option that enables processing of this annotation.
+
+In 1.4, we add a new mode for generating default methods: `-Xjvm-default=all` compiles *all* non-abstract methods of Kotlin
```
B1ggDave, line 15: In 1.4, we've added a new mode for generating default methods:

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,19 @@ title: "What's New in Kotlin 1.4"
...
+interfaces to `default` Java methods. For compatibility with the code that uses the interfaces compiled without `default`, 
+we also add `all-compatibility` mode.
```
B1ggDave, line 17: we also added `all-compatibility` mode.

PR #1762, opened 2020-07-17
chore: clarify kotlin-jpa plugin introduction

Crow-EH:
One might wrongly understand that *kotlin-spring* is also wrapped on top of *no-arg*, which is not the case since it's a subplugin of *allopen*.

I'm not a native english speaker, so rephrasing suggestions are welcome.


yanex (approved):

p7nov:
Thanks for clarifying this!

PR #1764, opened 2020-07-19
Update java-to-kotlin-interop.md NullPointerException -> IllegalArgumentException

tciesla:
Calling kotlin function from java with required parameter as null actually returns `IllegalArgumentException`. Example: `Exception in thread "main" java.lang.IllegalArgumentException: Parameter specified as non-null is null: method MyUtils__OldutilsKt.doSomething, parameter circle``

### DIFF
--- a/pages/docs/reference/java-to-kotlin-interop.md
+++ b/pages/docs/reference/java-to-kotlin-interop.md
@@ -612,7 +612,7 @@ fun writeToFile() {
 
 When calling Kotlin functions from Java, nobody prevents us from passing *null*{: .keyword } as a non-null parameter.
 That's why Kotlin generates runtime checks for all public functions that expect non-nulls.
-This way we get a `NullPointerException` in the Java code immediately.
+This way we get a `IllegalArgumentException` in the Java code immediately.
 
 ## Variant generics
 


Wasabi375:
To be honest I think this is something that should have been changed. Sadly I think it's far to late because there is probably a lot of java code out there relying on this fact. Still I agree that this should be properly reflected in the docs.

p7nov:
Hi!
Thanks for your PR. Actually, this is one of the [breaking changes in Kotlin 1.4](https://kotlinlang.org/docs/reference/compatibility-guide-14.html#unify-exceptions-from-null-checks): _Starting from Kotlin 1.4, all runtime null checks will throw a java.lang.NullPointerException_ 
So I suppose this PR is not relevant anymore.
See [KT-22275](https://youtrack.jetbrains.com/issue/KT-22275) for details.

Regards, Pavel
Kotlin Technical Writer

p7nov:
If you use Kotlin 1.4+ and experience a behavior different from what I wrote above, please let us know here on in the [Kotlin issues tracker](https://youtrack.jetbrains.com/newissue?project=kt)

PR #1766, opened 2020-07-20
feat: new type inference algorithm & callable reference improvements

katia-energizer:
Added What's new about a new type inference algorithm & callable reference improvements

### DIFF
--- a/pages/docs/reference/whatsnew14.md
+++ b/pages/docs/reference/whatsnew14.md
@@ -6,6 +6,158 @@ title: "What's New in Kotlin 1.4"
 
 # What's New in Kotlin 1.4
 
+## More powerful type inference algorithm
+
+Kotlin 1.4 uses a new, more powerful type inference algorithm. You were already able to try this new algorithm with 
+Kotlin 1.3 by specifying a compiler option, and now it’s used by default. You can find the full list of issues fixed in 
+the new algorithm in [YouTrack](https://youtrack.jetbrains.com/issues/KT?q=Tag:%20fixed-in-new-inference%20). Here
+you can find some of the most noticeable improvements.
+
+### Inferring type automatically in more use-cases
+
+The new inference algorithm infers types for many cases where the old inference required you to specify them explicitly. 
+For instance, in the following example, the type of the lambda parameter `it` is correctly inferred to `String?`:
+
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.4">
+
+```kotlin
+//sampleStart
+val rulesMap: Map<String, (String?) -> Boolean> = mapOf(
+    "weak" to { it != null },
+    "medium" to { !it.isNullOrBlank() },
+    "strong" to { it != null && "^[a-zA-Z0-9]+$".toRegex().matches(it) }
+)
+//sampleEnd
+
+fun main() {
+    println(rulesMap.getValue("weak")("abc!"))
+    println(rulesMap.getValue("strong")("abc"))
+    println(rulesMap.getValue("strong")("abc!"))
+}
+```
+
+</div>
+
+In Kotlin 1.3, you needed to introduce an explicit lambda parameter, or replace `to` with a `Pair` constructor with 
+explicit generic arguments to make it work.
+
+### Smart casts for lambda’s last expression
+
+In Kotlin 1.3, the last expression inside lambda isn’t smart cast unless you specify the expected type. Thus, in the 
+following example, Kotlin 1.3 infers `String?` as the type of the `result` variable:
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+val result = run {
+    var str = currentValue()
+    if (str == null) {
+        str = "test"
+    }
+    str // the Kotlin compiler knows that str is not null here
+}
+// The type of 'result' is String? in Kotlin 1.3 and String in Kotlin 1.4
+```
+
+</div>
+
+In Kotlin 1.4, thanks to the new inference algorithm, the last expression inside lambda gets smart cast, and this new, 
+more precise type is used to infer the resulting lambda type. Thus, the type of the `result` variable becomes `String`.
+
+In Kotlin 1.3, you often needed to add explicit casts (either `!!` or type casts like `as String`) to make such cases work, 
+and now these casts have become unnecessary.
+
+### Smart casts for callable references
+
+In Kotlin 1.3, you couldn’t access a member reference of a smart cast type. Now you can:
+
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.4">
+
+```kotlin
+sealed class Animal
+class Cat : Animal() {
+    fun meow() {
+        println("meow")
+    }
+}
+
+class Dog : Animal() {
+    fun woof() {
+        println("woof")
+    }
+}
+
+//sampleStart
+fun perform(animal: Animal) {
+    val kFunction: KFunction<*> = when (animal) {
+        is Cat -> animal::meow
+        is Dog -> animal::woof
+    }
+    kFunction.call()
+}
+//sampleEnd
+
+fun main() {
+    perform(Cat())
+}
+```
+
+</div>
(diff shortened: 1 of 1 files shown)


pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+    A.foo(r1) {}  // Works now
+}
+```
+
+</div>
+
+#### Java SAM interfaces in Kotlin
+
+Starting from Kotlin 1.4, you can use Java SAM interfaces in Kotlin. Previously, you could declare such a function `foo`
+only in Java code.
```
zarechenskiy, line 183: Not very correct. Of course, you could declare such a function before that. The main improvement is that now it's possible to perform SAM conversion if such function is in Kotlin.

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+fun foo(r1: java.lang.Runnable) {}
```
zarechenskiy, line 188: Let's avoid using fully qualified names:
```
import java.lang.Runnable

fun foo(r: Runnable) {} 
...
```

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+```kotlin
+fun foo(i: Int = 0): String = "$i!"
+
+fun apply1(func: () -> String): String = func()
+fun apply2(func: (Int) -> String): String = func(42)
+
+fun main() {
+    println(apply1(::foo))
+    println(apply2(::foo))
+}
```
zarechenskiy, line 222: Let's simplify this example:
```
fun foo(i: Int = 0): String = "$i!"

fun apply(func: () -> String): String = func()

fun main() {
    println(apply(::foo))
}
```

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+
+</div>
+
+Previously, you should write additional overloads for the function `apply` to use the default argument values.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+// some new overload
+fun applyInt(func: (Int) -> String): String = func()
```
zarechenskiy, line 233: `func()` should have an argument, so it should be something like `func(0)`

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+// some new overload
+fun applyInt(func: (Int) -> String): String = func() 
+```
+
+</div>
+
+#### Function references in `Unit`-returning functions 
+
+Now you can use callable references to functions returning any type in `Unit`-returning functions. The Unit-returning 
+function `foo()` uses a callable reference to `returnsInt()` returning `Int` as a result.
```
zarechenskiy, line 241: We can add that it worked for lambda arguments and now it works for callable references as well. For example:
```
fun foo(f: () -> Unit) { }

fun returnsInt(): Int = 42

fun main() {
    foo { returnsInt() } // it was OK before 1.4
    foo(::returnsInt) // since Kotlin 1.4, it works as well
}
```

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.4">
+
+```kotlin
+fun foo(f: () -> Unit) { }
+fun returnsInt(): Int = 42
+
+fun main() {
+    println(foo(::returnsInt))
```
zarechenskiy, line 250: Let's drop `println` since it doesn't print anything meaningful

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+}
+```
+
+</div>
+
+#### References to functions with a variable number of arguments
+
+Starting from Kotlin 1.4, you can use callable references to functions with a variable number of arguments (`vararg`)
```
zarechenskiy, line 258: You could use them before 1.4, now it's possible to _adapt_ such callable references when passing them into functions with an arbitrary count of functional parameters at the end

pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,293 @@ title: "What's New in Kotlin 1.4"
...
+    use1(::foo) 
+    use2(::foo) 
+}
+```
+
+</div>
+
+#### Suspend conversion on callable references
+
+Kotlin 1.4 supports suspend conversion on callable references in Kotlin/JVM.
```
zarechenskiy, line 281: Same as with callable references, I think we can add that it worked with lambdas and now with it works with callable references as well:
```
fun call() {}
fun takeSuspend(f: suspend () -> Unit) {}
fun test() {
    takeSuspend { call() } // OK before 1.4
    takeSuspend(::call) // Now this case also works
}
```

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
```diff
@@ -35,6 +41,288 @@ fun main(){
...
+    use1(::foo) 
+    use2(::foo) 
+}
+```
+
+</div>
+
+#### Suspend conversion on callable references
+
+In addition to suspend conversion on lambdas, Kotlin 1.4 supports suspend conversion on callable references in Kotlin/JVM.
```
zarechenskiy, line 310: > callable references in Kotlin/JVM

Why only `Kotlin/JVM`? As I know, this kind of conversion works everywhere

pages/docs/reference/whatsnew14.md
```diff
@@ -132,6 +420,7 @@ Now you can also work with Pod libraries stored locally.
 Depending on your needs, you can add dependencies between:
 * A Kotlin project and Pod libraries stored remotely in the CocoaPods repository or stored locally on your machine.
 * A Kotlin Pod (Kotlin project used as a CocoaPods dependency) and an Xcode project with one or more targets.
+* Several Kotlin Pods and an Xcode project.
```
zarechenskiy, line 423: It seems this line is from another commit :)

katia-energizer: I merged changes. It's from a different branch. :)

pages/docs/reference/whatsnew14.md
katia-energizer: Removed.

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

PR #1769, opened 2020-07-21
feat: 1.4 new stable API, deprecations of old API

p7nov:
Update docs according to this list of deprecations and stabilizations in 1.4: https://jetbrains.quip.com/wmawAbqodCFk/14-Experimental-and-deprecated-API-review

### DIFF
--- a/pages/docs/reference/list-operations.md
+++ b/pages/docs/reference/list-operations.md
@@ -278,7 +278,7 @@ fun main() {
 ```
 </div>
 
-For removing the first and the last element, use functions [`removeFirst()`](/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/remove-first.html)
+For removing the first and the last element, there are handy shortcuts [`removeFirst()`](/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/remove-first.html)
 and [`removeLast()`](/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/remove-last.html). Note that on empty lists,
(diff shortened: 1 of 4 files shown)


pages/docs/tutorials/native/mapping-strings-from-c.md
```diff
@@ -248,7 +248,7 @@ fun main() {
 
     val useMe2 = copy_string(pinned.addressOf(0), pinned.get().size - 1)
     if (useMe2 != 0) throw Error("Failed to read string from C")
-    pinned.get().stringFromUtf8()
+    pinned.get().decodeToString()
```
ilya-g, line 251: There was a subtle difference between `stringFromUtf8` and `decodeToString()` around null-terminated strings. The former stops decoding on zero byte, whereas the latter decodes it normally as `nul` char.
The equivalent replacement would be `toKString()` from `kotlinx.cinterop` package.

pages/docs/reference/list-operations.md
```diff
@@ -278,6 +278,31 @@ fun main() {
 ```
 </div>
 
+For removing the first and the last element, use functions [`removeFirst()`](/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/remove-first.html)
```
ilya-g, line 281: I'd say it less imperatively:
```suggestion
For removing the first and the last element, there are handy shortcuts [`removeFirst()`](/api/latest/jvm/stdlib/kotlin.collections/-mutable-list/remove-first.html)
```

PR #1771, opened 2020-07-23
feat: IR BE what's new

katia-energizer:
Added what's new about IR BE and how to enable JVM IR BE.


pages/docs/reference/whatsnew14.md
```diff
@@ -35,6 +35,50 @@ fun main(){
 
 Learn more about [Kotlin functional interfaces and SAM conversions](fun-interfaces.html).
 
+## New compiler
+
+The new Kotlin compiler is going to be really fast; it will unify all the supported platforms and provide 
+an API for compiler extensions. It's a long-term project, and we've already completed several steps in Kotlin 1.4:
+
+* New type inference algorithm is enabled by default. 
+* New JVM and JS IR BE are available in the experimental mode. They will become the default once we stabilize them.
```
udalov, line 44: Let's replace BE with "back-ends" here

pages/docs/reference/whatsnew14.md
```diff
@@ -35,6 +35,50 @@ fun main(){
...
+share much logic and have a unified pipeline. This allows implementing most features, optimizations, and bugfixes 
+only once for all platforms.
+
+A common back-end infrastructure also opens the door for multiplatform compiler extensions. You will be able to plug into the 
+pipeline and add custom processing and transformations that will automatically work for all platforms. 
+
+In Kotlin 1.4, we do not provide a public API for such extensions yet, but we are working closely with our partners, 
+including JetPack Compose, who are already building their compiler plugins using our new back-end.
+
+Now you can opt into using the new JVM IR back-end, where we've already fixed many bugs.
```
udalov, line 62: Maybe something like this? To stress that we're looking for feedback

> Now you can opt into using the new JVM IR back-end. We've fixed many long-standing issues of the old back-end there and would be grateful for feedback.

pages/docs/reference/whatsnew14.md
```diff
@@ -35,6 +35,50 @@ fun main(){
...
+
+Specify an additional compilation option in your Gradle build script:
```
udalov, line 64: "compilation option" -> "compiler option" here and below (it's the term used here: https://kotlinlang.org/docs/reference/compiler-reference.html)

pages/docs/reference/whatsnew14.md
katia-energizer: ✅

pages/docs/reference/whatsnew14.md
katia-energizer: Added about feedback and our issue tracker. @udalov let me know if we have any additional channel for feedback, such as a public Slack channel.

pages/docs/reference/whatsnew14.md
katia-energizer: It's my fault. Thanks! Don't know why I used _compilation_ 🛠

PR #1773, opened 2020-07-23
we can't declare method that has no open modifier in base class with …

thisisxitij:
This line is not true.. 
If there is no *open*{: .keyword } modifier on a function, like `Shape.fill()`, declaring a method with the same signature in a subclass is illegal,
either with *override*{: .keyword } or without it.

whenever i am  trying to declare a method or function with same signature and name in subclass which has no open modifier in base class  compiler is complaining  that.. 
"<methodName> hides member of supertype <BaseClass> and needs 'override' modifier" and thus i need to add override before 'fun' keyword and if i do soo 
then we have to declare that function or method in base class as open..
so this line below  is totally wrong...

" If there is no open modifier on a function, like Shape.fill(), declaring a method with the same signature in a subclass is illegal, either with override or without it.  "

### DIFF
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -326,9 +326,7 @@ class Circle() : Shape() {
 
 </div>
 
-The *override*{: .keyword } modifier is required for `Circle.draw()`. If it were missing, the compiler would complain.
-If there is no *open*{: .keyword } modifier on a function, like `Shape.fill()`, declaring a method with the same signature in a subclass is illegal,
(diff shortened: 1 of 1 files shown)


thisisxitij:
i think its a major bug in language.. please check it

p7nov:
Hi!
Thanks for your PR.
Please clarify, what exactly do you find wrong in this line?
"If there is **no `open` modifier** on a function, **declaring a method with the same signature in a subclass is illegal**, either with override or without it."
As I see, it describes exactly what happens in your case: you don't add `open` on a superclass method and therefore can't declare the same method in its subclass. In other words, you're doing what is stated **illegal** and get error messages about it.
And then you fix it: you add `open` and then you can override the method.

Regards, Pavel
Kotlin Technical Writer

PR #1775, opened 2020-07-24
feat: mixing positional and named arguments

katia-energizer:
Added what's new about mixing named and positional arguments; updated functions documentation on this topic

### DIFF
--- a/pages/docs/reference/functions.md
+++ b/pages/docs/reference/functions.md
@@ -166,7 +166,21 @@ reformat(str, wordSeparator = '_')
 ```
 </div>
 
-When a function is called with both positional and named arguments, all the positional arguments should be placed before the first named one. For example, the call `f(1, y = 2)` is allowed, but `f(x = 1, 2)` is not.
+When calling a function with both named and positional arguments, you can mix them in any way. 
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+reformat(str,
+    true,
+    upperCaseFirstLetter = true,
+    false,
+    wordSeparator = '_'
+)
+```
+
+</div>
+
 
 [Variable number of arguments (*vararg*{: .keyword })](#variable-number-of-arguments-varargs) can be passed in the named form by using the **spread** operator:
 
--- a/pages/docs/reference/functions.md
+++ b/pages/docs/reference/functions.md
@@ -9,7 +9,7 @@ title: "Functions: infix, vararg, tailrec"
 
 ## Function declarations
 
-Functions in Kotlin are declared using the *fun*{: .keyword } keyword:
+Functions in Kotlin are declared using the `fun` keyword:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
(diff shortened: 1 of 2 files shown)


pages/docs/reference/functions.md
```diff
@@ -123,52 +128,48 @@ fun reformat(str: String,
...
+reformat(str, upperCaseFirstLetter = false, wordSeparator = '_')
 ```
+
 </div>
 
-When a function is called with both positional and named arguments, all the positional arguments should be placed before the first named one. For example, the call `f(1, y = 2)` is allowed, but `f(x = 1, 2)` is not.
 
-[Variable number of arguments (*vararg*{: .keyword })](#variable-number-of-arguments-varargs) can be passed in the named form by using the **spread** operator:
+You can pass a [variable number of arguments (`vararg`)](#variable-number-of-arguments-varargs) with names using the 
+`spread` operator:
```
Wasabi375, line 172: Not sure using a code block for the term *spread* is the best choice here. Maybe change it to italic if we don't like the bold version
```suggestion
*spread* operator:
```

pages/docs/reference/functions.md
```diff
@@ -177,10 +178,11 @@ fun foo(vararg strings: String) { /*...*/ }
 
 foo(strings = *arrayOf("a", "b", "c"))
 ```
+
 </div>
 
-> **On the JVM**: the named argument syntax cannot be used when calling Java functions because Java bytecode does not
-always preserve names of function parameters.
+> **On the JVM**: You can't use the named argument syntax when calling Java functions because Java bytecode does not
+always preserve names of function arguments.
```
Wasabi375, line 185: Parameter is the correct term here. Parameter refers to the definition of the variable in the function header while an argument is the value passed to a a function. 
```suggestion
always preserve names of function parameters.
```

Wasabi375:
Not sure what just happened. My requested changes seem to be displayed twice, at least on my end here. Maybe I somehow commited them twice? I had the PR open in multiple tabs. If so I apologize for the inconvenience 😉

pages/docs/reference/whatsnew14.md
```diff
@@ -53,6 +54,27 @@ described in  [this blog post](https://blog.jetbrains.com/kotlin/2019/12/what-to
...
+
+In Kotlin 1.4, there is no such a limitation anymore - you can mix positional and named arguments as you like.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+fun f(a: Int, b: Int, c: Int) {}
+
+fun main() {
+f(1, b = 2, 3)
```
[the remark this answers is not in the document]
katia-energizer: Changed the example

pages/docs/reference/functions.md
katia-energizer: We use this formatting .

pages/docs/reference/functions.md
katia-energizer: Let's keep parameters. Thanks!

PR #1799, opened 2020-08-03
events july_3 2020

TinaNEs:
third PR in July


templates/pages/books.html
```diff
@@ -7,6 +7,13 @@
         <section class="page-books__section page-books__featured g-9">
             <h2>Featured</h2>
 
+            <h3>Kotlin ile Asenkron Programlama</h3>
```
meilalina, line 10: @TinaNEs please move the book to the end of the list into the "General" section

templates/pages/books.html
```diff
@@ -7,6 +7,13 @@
...
+            <div class="book-lang">en</div>
```
meilalina, line 11: @TinaNEs this book is in Turkish

PR #1816, opened 2020-08-09
Update idioms.md

emendes28:
Update map example / description


p7nov:
Hi!
Thanks for your suggestion. However, your version doesn't keep the meaning: defaultValue is returned in two cases:
- the initial `value` is `null`
- the `transform result(value)`  is `null`
Thus, we can't merge this PR.

Regards, Pavel
Kotlin Technical Writer

PR #1817, opened 2020-08-10
feat: 1.4 whatsnew navigation

p7nov:
Add a TOC and **Back to top** links in **What's New in 1.4**


pages/docs/reference/whatsnew14.md
```diff
@@ -6,6 +6,76 @@ title: "What's New in Kotlin 1.4"
 
 # What's New in Kotlin 1.4
```
katia-energizer, line 7: `1.4.0`

PR #1820, opened 2020-08-11
Update android-overview.md

ssharanyab:
Updated android-overview.md by correcting a grammatical typo error.


p7nov:
Hi! Thanks for your suggestion, but the current text is grammatically correct as well (see [here](https://www.learnersdictionary.com/qa/not-only-but-also), for example). We'll keep the current wording as is because it's what the author wanted to say.

Regards, Pavel
Kotlin Technical Writer

PR #1821, opened 2020-08-11
update: added Kotlin Releases page

katia-energizer:
Added a Kotlin Releases page with a general description of releases and details of latest releases. The page is included in the footer for now.

### DIFF
--- a/pages/releases.md
+++ b/pages/releases.md
@@ -35,6 +35,8 @@ If you are migrating to the new feature release, Kotlin plugin's migration tools
 
 The following table lists details of latest Kotlin releases.
 
+You can also use [preview versions of Kotlin](eap/index.html#build-details).
+
 <table>
     <tr>
         <th>Build info</th>
--- a/pages/releases.md
+++ b/pages/releases.md
@@ -39,7 +39,7 @@ The following table lists details of latest Kotlin releases.
     <tr>
         <th>Build info</th>
         <th>Build highlights</th>
-        <th>kotlinx library versions</th>
+        <th>Recommended kotlinx library versions</th>
     </tr>
     <tr>
         <td><strong>1.4.0</strong>
@@ -100,3 +100,7 @@ The following table lists details of latest Kotlin releases.
         </td>
     </tr>    
 </table>
+
+> On the JVM, you usually can use library versions other than the recommended ones. However, we don't guarantee the compatibility
+>in such cases. 
+{:.note}
--- a/pages/releases.md
+++ b/pages/releases.md
@@ -15,8 +15,8 @@ performance improvements, and bug fixes.
 For example, for the feature release 1.3 we had several incremental releases including 1.3.10, 1.3.20, and 1.3.70.
 For 1.3.70, we had 2 bug fix releases – 1.3.71 and 1.3.72.
 
-For each incremental and feature release, we also ship several [preview versions](eap/index.html) that allow you to try 
-new features before they are released (Early Access Preview).
+For each incremental and feature release, we also ship several preview (_EAP_) versions for you to try 
+new features before they are released. See [Early Access Preview](eap/index.html) for details.
 
 Learn more about [types of Kotlin releases and their compatiblity](docs/reference/evolution/kotlin-evolution.html#feature-releases-and-incremental-releases). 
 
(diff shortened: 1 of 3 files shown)


templates/inc/footer.html
```diff
@@ -23,6 +23,8 @@
                 <div class="global-footer-row">
                         <p class="terms-para"><a class="terms-kit"
                               href="/contribute.html">Contributing to Kotlin</a></p>
+                    <p class="terms-para"><a class="terms-kit"
```
zoobestik, line 26: Do we need both links “ Releases” and “EAP” in footer?
May be EAP can accessed from new release page?

p7nov: I'd leave them both at least for some time, to avoid confusing people who know about the EAP page.

pages/releases.md
```diff
@@ -0,0 +1,105 @@
...
+                <li><a href="docs/reference/compatibility-guide-14.html" target="_blank">Compatibility Guide</a></li>
+                <li><a href="docs/reference/whatsnew14.html#migrating-to-kotlin-14" target="_blank">Migrating to Kotlin 1.4.0</a></li>
+            </ul>
+         </td>
+        <td>
+            <ul>
+                <li><a href="https://github.com/Kotlin/kotlinx.serialization" target="_blank"><strong>kotlinx.serialization</strong>
+                </a> version: <a href="https://bintray.com/kotlin/kotlinx/kotlinx.serialization.runtime/1.0.0-rc" target="_blank">1.0.0-rc</a></li>
+                <li><a href="https://github.com/Kotlin/kotlinx.coroutines" target="_blank"><strong>kotlinx.coroutines</strong></a>
+                version: <a href="https://bintray.com/kotlin/kotlinx/kotlinx.coroutines/1.3.9" target="_blank">1.3.9
```
erokhins, line 64: It is better to give links to the https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.3.8 pagesbecause there is a changelog. Same for all other libraries where there is corresponding releases pages:
https://github.com/Kotlin/kotlinx.atomicfu/releases
https://github.com/Kotlin/kotlinx.serialization/blob/master/CHANGELOG.md#0200--2020-03-04 for serialization
https://github.com/ktorio/ktor/releases
https://github.com/Kotlin/kotlinx.html/releases
(for node-js there is no fancy page, so bintray it is)

p7nov: Maybe it's better to give links to exact versions instead of `/releases` because with `/releases` they get inconsistent once a newer version is out?

erokhins: Yes, I mean that

erokhins:
I'm waiting for staging page to review the changes once again. Also we are unable to forecast link for kotlinx.serialization 1.4-RC release

templates/inc/footer.html
zoobestik: I think it's not so confusing, if when eap replaced by releases in same place in footer.
Much confuse if link just will be dropped? no?
Actually We just doesn't have much space for link in new WH footer.

pages/releases.md
```diff
@@ -0,0 +1,105 @@
...
+                <li><a href="http://blog.jetbrains.com/kotlin/2020/08/kotlin-1-4-released-with-a-focus-on-quality-and-performance/" target="_blank">Release blog post</a></li>
+                <li><a href="docs/reference/whatsnew14.html" target="_blank">What's new in Kotlin 1.4.0</a></li>
+                <li><a href="docs/reference/compatibility-guide-14.html" target="_blank">Compatibility Guide</a></li>
+                <li><a href="docs/reference/whatsnew14.html#migrating-to-kotlin-14" target="_blank">Migrating to Kotlin 1.4.0</a></li>
+            </ul>
+         </td>
+        <td>
+            <ul>
+                <li><a href="https://github.com/Kotlin/kotlinx.serialization" target="_blank"><strong>kotlinx.serialization</strong>
+                </a> version: <a href="https://github.com/Kotlin/kotlinx.serialization/blob/master/CHANGELOG.md#100-rc--2020-08-17" target="_blank">1.0.0-rc</a></li>
```
erokhins, line 62: Version 1.0.0-RC + link could be different, but it is ok, we could fix that later

templates/inc/footer.html
katia-energizer: Actually, I planned to replace the _EAP_ link with _Releases_ in our new footer. We discussed this with Denis Voronov. I agree that there should be one link in the footer.
I have also added one more link to EAP under _Release details_.

zoobestik: 👍

PR #1827, opened 2020-08-14
Feature new footer Kotlinlang + WebHelp

JuliaVolkova:


static/css/com/layout/footer.scss
```diff
@@ -3,103 +3,231 @@
...
+  .g-footer-layout {
+    min-width: 960px;
+    max-width: 960px;
+    width: 100%;
+    margin: auto;
+    display: flex;
+    justify-content: space-between;
+
+    @media (max-width: 960px) {
+      width: 100%;
```
zoobestik, line 35: exists in base selector for this rule and somethin else, does it need?

static/css/com/layout/layout.scss
```diff
@@ -11,11 +11,13 @@
 
 .global-layout {
     min-height: 100%;
-    margin-bottom: -$footer-height;
```
zoobestik, line 14: Oh, i know thats ugly, try to open [security.html](https://kotlinlang.org/security.html) or other short page.
I think it's better to improve body size :(

JuliaVolkova: Fixed it. Flex box forever ❤️

static/css/com/layout/footer.scss
JuliaVolkova: It's redundant, I've deleted it. Thanks!

PR #1834, opened 2020-08-17
Use the correct version for serialization compiler plugin on the serialization feature page

TheMrMilchmann:
According to the documentation of https://github.com/Kotlin/kotlinx.serialization, the version of the compiler plugin is still aligned with the Kotlin version. Additionally, version "1.0.0-RC" does not exist for the plugin. Therefore I assume that this is a mistake on the website.

I couldn't find any contribution guidelines so please let me know if there is anything else I should do.


p7nov:
Hi!
You're absolutely right! Thank you for fixing this.

PR #1835, opened 2020-08-18
Fix broken link to Kotlin serialization guide

michou:
The link to the Kotlin serialization guide was broken as it didn't include the branch in the path.


p7nov:
Thank you for noticing this! This is already fixed.

PR #1845, opened 2020-08-22
Update control-flow.md

danmaze:
The when expression replaces or is similar to the switch statement in C-like languages. The problem was with the "the switch operator of C-like languages". *switch* is not an operator, it is a statement. (And it is an expression in Kotlin, Haskell, Elm and the like. Which is actually quite interesting.)


p7nov:
Hi!
You're right, thank you for fixing this!

PR #1851, opened 2020-08-31
Fix broken sample in mixing named - position args section

AkshayChordiya:
The code sample shown in [mixing named and positional arguments section](https://kotlinlang.org/docs/reference/whatsnew14.html#mixing-named-and-positional-arguments) is broken.

Updated the code with the following changes to fix it:
* Replace single quote with double quote for first parameter in `reformat` function
* Replace type of 3rd parameter `Character` to `kotlin.Char`


p7nov:
Thanks!

PR #1852, opened 2020-08-31
fix: updated a way to create a security problem

katia-energizer:
updated a link to create a security issue and removed irrelevan information in Kotlin Security as agreed with Sergei Pecherkin.


zoobestik (approved):

specherkin (approved):
LGTM

PR #1855, opened 2020-09-01
Java class and primitives

stachenov:
Three separate things in a single PR because they are small enough, and yet, I believe they are important.

First, the fact that `ClassName.javaClass` is *not* the same as `ClassName::class.java`.

Second, the difference between `Integer::class.java` and `Int::class.javaObjectType` (and the same similar primitive wrapper types).

Third, the difference between `Int::class.java` and `Int::class.javaObjectType` (the previous change to this section only explained the latter).

### DIFF
--- a/pages/docs/reference/java-interop.md
+++ b/pages/docs/reference/java-interop.md
@@ -706,6 +706,8 @@ To access static members of a Java type that is [mapped](#mapped-types) to a Kot
 
 Java reflection works on Kotlin classes and vice versa. As mentioned above, you can use `instance::class.java`,
 `ClassName::class.java` or `instance.javaClass` to enter Java reflection through `java.lang.Class`.
+Do not use `ClassName.javaClass` for this purpose because it refers to `ClassName`'s companion object class,
+which is the same as `ClassName.Companion::class.java` and not `ClassName::class.java`.
 You may also use `ClassName::class.javaObjectType` for getting primitive types wrappers.
 
 Other supported cases include acquiring a Java getter/setter method or a backing field for a Kotlin property, a `KProperty` for a Java field, a Java method or constructor for a `KFunction` and vice versa.
--- a/pages/docs/reference/java-interop.md
+++ b/pages/docs/reference/java-interop.md
@@ -708,7 +708,10 @@ Java reflection works on Kotlin classes and vice versa. As mentioned above, you 
 `ClassName::class.java` or `instance.javaClass` to enter Java reflection through `java.lang.Class`.
 Do not use `ClassName.javaClass` for this purpose because it refers to `ClassName`'s companion object class,
 which is the same as `ClassName.Companion::class.java` and not `ClassName::class.java`.
-You may also use `ClassName::class.javaObjectType` for getting primitive types wrappers.
+For each primitive type, there are two different Java classes, and Kotlin provides ways to get both. For
+example, `Int::class.java` will return the class instance representing the primitive type itself,
+corresponding to `Integer.TYPE` in Java. To get the class of the corresponding wrapper type, use
+`Int::class.javaObjectType`, which is equivalent of Java's `Integer.class`.
 
 Other supported cases include acquiring a Java getter/setter method or a backing field for a Kotlin property, a `KProperty` for a Java field, a Java method or constructor for a `KFunction` and vice versa.
 


pages/docs/reference/java-interop.md
```diff
@@ -706,7 +706,13 @@ To access static members of a Java type that is [mapped](#mapped-types) to a Kot
 
 Java reflection works on Kotlin classes and vice versa. As mentioned above, you can use `instance::class.java`,
 `ClassName::class.java` or `instance.javaClass` to enter Java reflection through `java.lang.Class`.
-You may also use `ClassName::class.javaObjectType` for getting primitive types wrappers.
+Do not use `ClassName.javaClass` for this purpose because it refers to ClassName's companion object class,
+which is the same as `ClassName.Companion::class.java` and not `ClassName::class.java`.
+For primitive types, say, `Int::class.java` will yield the same object as `Integer.TYPE` in Java.
+To get the class of the corresponding boxed type, you may use `ClassName::class.javaObjectType`. Note, however,
```
udalov, line 712:
```suggestion
To get the class of the corresponding wrapper type, you may use `ClassName::class.javaObjectType`. Note, however,
```

pages/docs/reference/java-interop.md
```diff
@@ -706,7 +706,13 @@ To access static members of a Java type that is [mapped](#mapped-types) to a Kot
...
+that even though, say, `java.lang.Integer::class.java` and `Int::class.javaObjectType` refer to the same object,
+they have different compile time types: the former has `Class<java.lang.Integer>`, while the latter
+has `Class<Int>`, which is better for seamless integration with Kotlin code.
```
udalov, line 715: I'd remove this last sentence altogether because we don't recommend to use `java.lang.Integer` anywhere in Kotlin, and in fact doing so leads to a compiler warning

stachenov: And that's exactly what I was trying to explain here by “better for seamless integration”. I created this PR after I went through a lot of confusion trying to get my code to work using `java.lang.Integer::class.java` because I didn't know about the right construct.

Maybe it should be reworded better. Something along the lines of
“While it is also technically possible to get the same type using, for example, `java.lang.Integer::class.java`, doing so is not recommended for two reasons. First, using `java.lang.Integer` and similar types will lead to a compiler warning. Second, even though `java.lang.Integer::class.java` and `Int::class.javaObjectType` refer to *the same instance,* they have different *compile time* types: the former has `Class<java.lang.Integer>`, which leads to various compilation problems, and the latter has `Class<Int>`, which works just fine”.

udalov: My main wish is to avoid overloading users, especially beginners, with information on such (already complicated) topic. It's a documentation, not the language spec, so IMHO we should try to keep it easy to understand and on point.

Not everything needs to be documented here. Some things are better left to be as warnings in the compiler. Unless the confusion is so widespread, or warnings are not readable, that it makes sense to clarify it in the docs. But I don't see the evidence for it yet on this particular issue.

Also, if we're comparing class literal access in Kotlin to Java, then we could for example provide a 2x2 table (rows = primitive/wrapper, columns = Java/Kotlin) to make the correspondence more memorable. Or something else that would list the correct code patterns in an easy-to-understand way. But IMHO we should avoid speaking about the _incorrect_ code patterns such as `java.lang.Integer::class.java` in such manner that it could somehow be imprinted in a beginner's head and then when they come to use it, they'll doubt which form (`Integer::class.java`, `Int::class.java`, `Int::class.javaObjectType`) was correct because they saw all of them in the text but couldn't remember exactly which one was recommended and which wasn't.

It's just my opinion though. I'm not an expert on documentation and I don't maintain the Kotlin website :).

With all that said, I just checked and found out that the compiler doesn't report warnings on Java platform classes used in class literal position: https://youtrack.jetbrains.com/issue/KT-43557. I hope that after fixing that issue, the need for clarification here will be greatly reduced.

stachenov: Makes sense. Although I'm not an expert documentation writer either. I'll remove the last sentence then. I still feel, however, that the difference between `Int::class.java` and `Int::class.javaObjectType` needs to be explained, because a lot of people, myself included, seldom use `Integer.TYPE` and confuse it with `Integer.class` when see it the first time. Maybe this:

> For each primitive type, there are two different Java classes, and Kotlin provides ways to get both. For example, `Int::class.java` will return the class instance representing the primitive type itself, corresponding to `Integer.TYPE` in Java. To get the class of the corresponding wrapper type, use `Int::class.javaObjectType`, which is equivalent of Java's `Integer.class`.

This way only the right constructs are mentioned, and a clear correspondence between Kotlin and Java is explained.

p7nov: So I'm the only documentation writer here, I suppose :)
@stachenov thanks for these clarifications. Anyway, I agree with @udalov that we're trying to keep Kotlin docs concise and avoid writing about "bad" patterns unless we see that people generally have problems with them. 
I like the wording you suggest in your latest comment. Please replace the current version with it and I'll merge the PR.

stachenov:
@p7nov Reworded, rebased and force-pushed. Re-pushed twice because the first time I got the lines too long and they looked ugly in the web interface.

p7nov:
Thanks!

PR #1858, opened 2020-09-02
feature: add trailing commas

koshachy:
add trailing commas to Coding Conventions (coding-conventions.md)

### DIFF
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -105,7 +105,7 @@ class Customer(name: String) {
 </div>
 
 In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax.
-Also, use a [trailing comma](coding-conventions.html#trailing-commas) when declaring class properties:
+Also, the Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -104,12 +104,17 @@ class Customer(name: String) {
 
 </div>
 
-In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax:
+In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax.
+Also, use a [trailing comma](coding-conventions.md#trailing-commas) when declaring class properties:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
 ```kotlin
-class Person(val firstName: String, val lastName: String, var age: Int) { /*...*/ }
+class Person(
+    val firstName: String,
+    val lastName: String,
+    var age: Int, // trailing comma
+) { /*...*/ }
 ```
 
 </div>
@@ -129,7 +134,6 @@ class Customer public @Inject constructor(name: String) { /*...*/ }
 </div>
 
 For more details, see [Visibility Modifiers](visibility-modifiers.html#constructors).
-
 
 #### Secondary constructors
 
@@ -402,7 +406,7 @@ open class Base(val name: String) {
 
 class Derived(
     name: String,
-    val lastName: String
+    val lastName: String,
 ) : Base(name.capitalize().also { println("Argument for Base: $it") }) {
 
     init { println("Initializing Derived") }
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -114,17 +114,15 @@ class Person(val firstName: String, val lastName: String, var age: Int) { /*...*
 
 </div>
 
-The Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties.
-It simplifies reordering, deleting, and adding new properties. Using trailing commas are optional: without them your code will work as well.
+You can use a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
 ```kotlin
 class Person(
-    val name: String,
+    val firstName: String,
     val lastName: String,
-    val age: Int,
-    val gender: String, // trailing comma
+    val age: Int, // trailing comma
 ) { /*...*/ }
 ```
 
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -105,7 +105,7 @@ class Customer(name: String) {
 </div>
 
 In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax.
-Also, use a [trailing comma](coding-conventions.md#trailing-commas) when declaring class properties:
+Also, use a [trailing comma](coding-conventions.html#trailing-commas) when declaring class properties:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
--- a/pages/docs/reference/classes.md
+++ b/pages/docs/reference/classes.md
@@ -104,16 +104,27 @@ class Customer(name: String) {
 
 </div>
 
-In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax.
-Also, the Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties:
+In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax:
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+class Person(val firstName: String, val lastName: String, var age: Int) { /*...*/ }
+```
+
+</div>
+
+The Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties.
+It simplifies reordering, deleting, and adding new properties. Using trailing commas are optional: without them your code will work as well.
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
 ```kotlin
 class Person(
-    val firstName: String,
+    val name: String,
     val lastName: String,
-    var age: Int, // trailing comma
+    val age: Int,
+    val gender: String, // trailing comma
 ) { /*...*/ }
 ```
 
--- a/pages/docs/reference/coding-conventions.md
+++ b/pages/docs/reference/coding-conventions.md
@@ -840,7 +840,7 @@ fun print(
 val sum: (Int, Int, Int) -> Int = fun(
     x,
     y,
-    z// trailing comma
+    z, // trailing comma
 ): Int {
     return x + y + x
 }
@@ -1263,7 +1263,7 @@ Although the semantics are similar, there are some stylistic conventions on when
 Prefer a property over a function when the underlying algorithm:
 
 * does not throw
-* is cheap to calculate (or caсhed on the first run)
+* is cheap to calculate (or cached on the first run)
 * returns the same result over invocations if the object state hasn't changed
 
 ### Using extension functions
--- a/pages/docs/reference/coding-conventions.md
+++ b/pages/docs/reference/coding-conventions.md
@@ -729,7 +729,7 @@ foo {
 
 ### Trailing commas
 
-A trailing comma is a comma sign after the last item of the element series:
+A trailing comma is a comma symbol after the last item of a series of elements:
 
 <div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
 
@@ -745,13 +745,13 @@ class Person(
 
 Using trailing commas has several benefits:
 
-* It makes version-control diffs cleaner: all focus is on changed value.
-* It's easy to add or reorder elements: no need to add or delete the comma sign if you manipulate elements.
-* It simplifies code generation (for example, for object initializers). The last element can also have the comma sign.
+* It makes version-control diffs cleaner – as all the focus is on the changed value.
+* It makes adding and reordering elements – there is no need to add or delete the comma if you manipulate elements.
+* It simplifies code generation, for example, for object initializers. The last element can also have a comma.
 
-Trailing commas are optional: your code will work without them. Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
+Trailing commas are entirely optional – your code will still work without them. The Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
 
-To enable trailing commas for the IntelliJ formatter, go to __Settings | Editor | Code Style | Kotlin__, 
+To enable trailing commas in the IntelliJ IDEA formatter, go to __Settings | Editor | Code Style | Kotlin__, 
 open the __Other__ tab and select the __Use trailing comma__ option.
 
 Kotlin supports trailing commas in the following cases:
--- a/pages/docs/reference/coding-conventions.md
+++ b/pages/docs/reference/coding-conventions.md
@@ -752,7 +752,7 @@ Using trailing commas has several benefits:
 Trailing commas are optional: your code will work without them. Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
 
 To enable trailing commas for the IntelliJ formatter, go to __Settings | Editor | Code Style | Kotlin__, 
-open the __Other__ tab and enable the __Use trailing comma__ option.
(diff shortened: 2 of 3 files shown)


pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,213 @@ foo {
...
+    Foo, 
+    Bar, // trailing comma
+}
+```
+
+</div>
+
+#### Class parameters
+
+Place the parameters on different lines:
```
koshachy, line 787: I'm not quite sure about these repeated single lines before each code block. 
@p7nov do you have any thoughts on it?

p7nov: I don't see much value in these lines.

koshachy: Ok, I've eliminated them.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,195 @@ foo {
 
 </div>
 
+### Trailing commas
+
+Use trailing commas at declaration site, including multi-line enumerations such as argument and parameter lists, `when` entries, and components of destructuring declarations.
```
p7nov, line 732: I'm not sure that readers are familiar with the term "trailing comma". Perhaps we should add a definition.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,195 @@ foo {
...
+For call site trailing commas are optional.
```
p7nov, line 733: Isn't it optional at the declaration site as well?

p7nov: Maybe we should explain separately that it's _technically_ optional at both sites, but _as a convention_ we recommend (what we recommend :)).

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,195 @@ foo {
...
+* [Class parameters](#class-parameters)
+* [Function value parameters](#function-value-parameters)
+* [Parameters with optional type (including setters)](#parameters-with-optional-type-including-setters)
+* [Value arguments](#value-arguments)
+* [Indexing suffix](#indexing-suffix)
+* [Lambda parameters](#lambda-parameters)
+* [`when` entry](#when-entry)
+* [Collection literals (in annotations)](#collection-literals-in-annotations)
+
+> Kotlin doesn't support:
```
p7nov, line 752: Seems out of scope for coding conventions.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,195 @@ foo {
...
+
+```kotlin
+val colors = listOf(
+    "red",
+    "green",
+    "blue", // trailing comma
+)
+
+enum class Classes { 
+    Foo,
```
p7nov, line 776: We want our examples to be more meaningful. At least by avoiding adding new `Foo`s and `Bar`s.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,195 @@ foo {
...
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+fun foo(
+    x: Int,
+    y: Number, // trailing comma
+) {}
+
+constructor(
+    x: Comparable<Comparable<Number>>,
```
p7nov, line 807: Are you sure we need all these generics in generics for an example of trailing comma?

pages/docs/reference/functions.md
```diff
@@ -43,13 +43,16 @@ Stream().read() // create instance of class Stream and call read()
 
 ### Parameters
 
-Function parameters are defined using Pascal notation, i.e. *name*: *type*. Parameters are separated using commas. 
+Function parameters are defined using Pascal notation, i.e. *name*: *type*. Parameters are separated using commas. Use a [trailing comma](coding-conventions.html#trailing-commas) when declaring function parameters.
```
p7nov, line 46: Not sure we can just recommend it in general for all functions. There are hundreds of examples without it in our own docs and Kotlin codebase :)

p7nov (changes requested):

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,201 @@ foo {
...
+* It's easy to add or reorder elements: no need to add or delete the comma sign if you manipulate elements.
+* It simplifies code generation (for example, for object initializers). The last element can also have the comma sign.
+
+Using trailing commas are optional: your code will work without them. But Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
+
+To enable trailing commas for the IntelliJ formatter, go to __Settings | Editor | Code Style | Kotlin__, 
+open the __Other__ tab and enable the __Use trailing comma__ option.
+
+Kotlin supports trailing commas in the following cases:
+* [Lists and enumerations](#lists-and-enumerations)
```
PetukhovVictor, line 758: The support in Lists is the same as in value arguments.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,201 @@ foo {
...
+* It makes version-control diffs cleaner: all focus is on changed value.
+* It's easy to add or reorder elements: no need to add or delete the comma sign if you manipulate elements.
+* It simplifies code generation (for example, for object initializers). The last element can also have the comma sign.
+
+Using trailing commas are optional: your code will work without them. But Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
+
+To enable trailing commas for the IntelliJ formatter, go to __Settings | Editor | Code Style | Kotlin__, 
+open the __Other__ tab and enable the __Use trailing comma__ option.
+
+Kotlin supports trailing commas in the following cases:
```
PetukhovVictor, line 757: Not all supported cases are listed here. There is also support trailing comma in type arguments, type parameters and destructuring declaration.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,201 @@ foo {
...
+
+
+#### Collection literals (in annotations)
+
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+annotation class Anno(val x: IntArray)
+
+@Anno([1, 2, 3, 4,])
```
PetukhovVictor, line 919: I believe the elements should be placed on separate lines otherwise the trailing comma loses all its advantages.

pages/docs/reference/coding-conventions.md
p7nov: Actually, this list of supported cases seems like reference-style documentation. I personally don't think that Coding Conventions page is a place for describing the language syntax in this way. 
I'd write in in more general with words with several (3-5) **typical** examples.

pages/docs/reference/functions.md
```diff
@@ -43,13 +43,17 @@ Stream().read() // create instance of class Stream and call read()
 
 ### Parameters
 
-Function parameters are defined using Pascal notation, i.e. *name*: *type*. Parameters are separated using commas. 
+Function parameters are defined using Pascal notation, i.e. *name*: *type*. Parameters are separated using commas.
+The Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare function parameters.
```
p7nov, line 47: This is the most questionable thing as for me, so I'll ask again: can we really recommend trailing comma in function (and class) declarations **unconditionally**?
I mean, everybody declares small functions in a way like `fun sum(a: Int, b:Int): Int` - this sentence states that it's wrong now. Why? How does trailing comma make it better?

pages/docs/reference/coding-conventions.md
koshachy: Is it ok if I merge them into one chapter?

I mean
`### Lists, enumerations, value arguments`

pages/docs/reference/functions.md
koshachy: I don't see a contradiction here: 
Kotlin style **recommends**: recommendation != obligation

And this code line `fun sum(a: Int, b:Int): Int` is not wrong.

p7nov: I sure agree that it's not an obligation. Under "wrong" I mean "not recommended"/"contradicts with recommendations". 
The sentence says "recommend trailing comma in (any!) function parameters". So we directly say to the reader that the recommended way is different from this (my example). But this declaration is OK, isn't it?
I would expect more tailored recommendations - when is this comma useful? I really don't see a point in trailing comma in many cases (one-line declarations, for example).

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,253 @@ foo {
...
+    NORTH,
+    SOUTH,
+    WEST,
+    EAST, // trailing comma
+}
+
+fun sum(x: Any, y: Any) {}
+
+fun main() {
+    sum(
```
p7nov, line 790: I think `listOf()` is a more natural example for value parameters (and it doesn't require declaration and `main` :)
```
val numbers = listOf(
    "one",
    "two",
    "three", // trailing comma
)
```

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,253 @@ foo {
...
+</div>
+
+#### Class parameters
+
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+class Customer(
+    val name: String,
+    val surname: String, // trailing comma
```
p7nov, line 806: Nitpicking: _last name_ is preferable over _surname_ in American English.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,253 @@ foo {
...
+
+</div>
+
+#### Indexing suffix
+
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+class Surface {
+    operator fun get(x: Int, y: Int) = 2*x + 4*y - 10
```
p7nov, line 858: Kotlin coding conventions (aka this page :) ) says [put spaces around binary operators (a + b).](https://kotlinlang.org/docs/reference/coding-conventions.html#horizontal-whitespace)

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,253 @@ foo {
...
+
+</div>
+
+#### Collection literals (in annotations)
+
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+annotation class ApplicableFor(val services: Array<String>)
+@ApplicableFor([
```
p7nov, line 910: Pls add a blank line between annotation and annotated function.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,253 @@ foo {
...
+```
+
+</div>
+
+#### Destructuring declarations
+
+<div class="sample" markdown="1" theme="idea" mode="kotlin" data-highlight-only>
+
+```kotlin
+fun main() {
```
p7nov, line 948: You don't necessarily need `main` for non-runnable code blocks.

pages/docs/reference/functions.md
koshachy: I understand the concern; thus I rewrote this part.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,242 @@ foo {
...
+Using trailing commas has several benefits:
+
+* It makes version-control diffs cleaner: all focus is on changed value.
+* It's easy to add or reorder elements: no need to add or delete the comma sign if you manipulate elements.
+* It simplifies code generation (for example, for object initializers). The last element can also have the comma sign.
+
+Using trailing commas are optional: your code will work without them. But Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
+
+To enable trailing commas for the IntelliJ formatter, go to __Settings | Editor | Code Style | Kotlin__, 
+open the __Other__ tab and enable the __Use trailing comma__ option.
```
p7nov, line 755:
```suggestion
open the __Other__ tab and select the __Use trailing comma__ option.
```
See [MS Guide on UI description](https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui)

pages/docs/reference/coding-conventions.md
PetukhovVictor: I suggest to have only enumeration chapter and `value arguments` as `Lists` is subset of `value arguments`. `listOf(a, b, c)` is just a function call.

pages/docs/reference/coding-conventions.md
PetukhovVictor: OK, I just thought that here describes all of the supported cases (due to the phrase `Kotlin supports trailing commas in the following cases`). BTW it can be merged in the following way: `value and type arguments`, `value and type parameters`.

pages/docs/reference/coding-conventions.md
```diff
@@ -726,6 +727,242 @@ foo {
...
+
+</div>
+
+Using trailing commas has several benefits:
+
+* It makes version-control diffs cleaner: all focus is on changed value.
+* It's easy to add or reorder elements: no need to add or delete the comma sign if you manipulate elements.
+* It simplifies code generation (for example, for object initializers). The last element can also have the comma sign.
+
+Using trailing commas are optional: your code will work without them. But Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
```
p7nov, line 752:
```suggestion
Trailing commas are optional: your code will work without them. Kotlin style guide encourages the use of trailing commas at the declaration site and leaves it optional for the call site.
```
- With `Using`, the 'are' verb is incorrect.
- `But` doesn't seem necessary and starting a sentence with it is an arguable topic, so I'd just remove it.
- `optional` is used in two different meanings in one line. We should replace it with smth that conducts `up to you` meaning better.

pages/docs/reference/functions.md
```diff
@@ -53,6 +53,20 @@ fun powerOf(number: Int, exponent: Int) { /*...*/ }
 ```
 </div>
 
+The Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare function parameters.
+It simplifies reordering, deleting, and adding new parameters. Using trailing commas are optional: without them your code will work as well.
```
p7nov, line 57: Do we really need to repeat this on three pages? "Optional" is obvious from the example above and the `recommends` word, benefits are described in coding conventions. The same goes for the Classes page.

pages/docs/reference/classes.md
```diff
@@ -114,6 +114,22 @@ class Person(val firstName: String, val lastName: String, var age: Int) { /*...*
 
 </div>
 
+The Kotlin code style recommends using a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties.
+It simplifies reordering, deleting, and adding new properties. Using trailing commas are optional: without them your code will work as well.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
```
p7nov, line 122: Maybe it's better to rewrite the class from the example above to show only the syntax difference?
Now there are structure differences, so it's harder to find the difference we want to show.

pages/docs/reference/functions.md
koshachy: IIRC, you've mentioned above that you want to see here the benefits of using trailing commas ("when is this comma useful?"). Did I miss something?

pages/docs/reference/coding-conventions.md
koshachy: To recap.
No more `Lists and enumerations`: there are `Enumerations` and `Value arguments`.

pages/docs/reference/coding-conventions.md
koshachy: I've added `Type arguments and type parameters` and `Destructuring declarations` instead of merging

pages/docs/reference/functions.md
```diff
@@ -53,6 +53,18 @@ fun powerOf(number: Int, exponent: Int) { /*...*/ }
 ```
 </div>
 
+You can use a [trailing comma](coding-conventions.html#trailing-commas) when you declare class properties:
```
p7nov, line 56:
```suggestion
You can use a [trailing comma](coding-conventions.html#trailing-commas) when you declare function parameters:
```

p7nov (approved):
LGTM now.

PetukhovVictor (approved):

PR #1861, opened 2020-09-03
Improve documentation for JS DCE

SebastianAigner:
Emphasize DCE application only for production assets
Emphasize that names used as arguments should be in terms of the generated JS code
Add note (+ warning) about how to keep whole packages / modules from elimination

Fixes https://youtrack.jetbrains.com/issue/KT-41188

### DIFF
--- a/pages/docs/reference/javascript-dce.md
+++ b/pages/docs/reference/javascript-dce.md
@@ -19,7 +19,8 @@ Unused declarations can appear in cases like:
   adapters for DOM, and so on. All of this functionality would require about 1.3 MB as a JavaScript file. A simple "Hello, world" application only requires
   console routines, which is only few kilobytes for the entire file.
 
-The Kotlin/JS Gradle plugin handles DCE automatically when you build a production bundle, for example by using the `browserProductionWebpack` task. Development bundling tasks don't include DCE.
+> The Kotlin/JS Gradle plugin handles DCE automatically when you build a production bundle, for example by using the `browserProductionWebpack` task. Development bundling tasks (like `browserDevelopmentWebpack`) don't include DCE.
+{:.note}
 
 ## Excluding declarations from DCE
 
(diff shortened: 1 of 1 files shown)


pages/docs/reference/javascript-dce.md
```diff
@@ -19,7 +19,8 @@ Unused declarations can appear in cases like:
   adapters for DOM, and so on. All of this functionality would require about 1.3 MB as a JavaScript file. A simple "Hello, world" application only requires
   console routines, which is only few kilobytes for the entire file.
 
-The Kotlin/JS Gradle plugin handles DCE automatically when you build a production bundle, for example by using the `browserProductionWebpack` task. Development bundling tasks don't include DCE.
+> The Kotlin/JS Gradle plugin handles DCE automatically when you build a production bundle, for example by using the `browserProductionWebpack` task. Development bundling tasks (like `browserDevelopmentWebpack`) don't include DCE.
```
p7nov, line 22: I don't think this should be a note. It's an important part of the general explanation. Notes are usually for some side additions that don't fit into the common flow.

pages/docs/reference/javascript-dce.md
```diff
@@ -30,33 +31,47 @@ To keep certain declarations from elimination, add the `dceTask` block to your G
...
 ```
-
-</div>
 </div>
 
-<div class="multi-language-sample" data-lang="kotlin">
-<div class="sample" markdown="1" mode="kotlin" theme="idea" data-lang="kotlin" data-highlight-only>
+If you want to keep a whole package or module from elimination, you can use its fully qualified name as it appears in the generated JavaScript code.
+
+> Keeping whole packages or modules from elimination can prevent DCE from removing many declarations that otherwise would have been removed. Because of this, it is preferable to select individual declarations which should be excluded from DCE one by one.
```
p7nov, line 56:
```suggestion
> Keeping whole packages or modules from elimination can prevent DCE from removing many unused declarations. Because of this, it is preferable to select individual declarations which should be excluded from DCE one by one.
```
If we shorten the sentence like this, will it keep the meaning?

SebastianAigner: Yup, that still sounds good to me!

pages/docs/reference/javascript-dce.md
SebastianAigner: Thanks, I have removed the note tag from this, and added some bold emphasis to show the two differing cases instead.

katia-energizer (approved):
LGTM

p7nov (approved):

bashor (approved):

PR #1864, opened 2020-09-06
Fix grammatical errors

androiddevnotes:


p7nov:
Hi!
Thanks for your PR.
I don't think it's an error because `are` refers to multiple (two) nouns:
`The most important (ones) are ... template ... and type`

PR #1868, opened 2020-09-08
import as instead of private typealias

florianhof:
because "import ... as ..." can remain unknown when using "typealias".


p7nov:
Thanks!

PR #1873, opened 2020-09-12
Update classes.md

dusanhornik:


p7nov:
Thanks for your PR.
`var` is not a typo here. It shows that `age` can change in the future.

dusanhornik:
In that case, you want to change the `val` into `var` in the next paragraph as they both suppose to represent the same class with only formatting differences.

p7nov:
Yes, you're right, thank you. I'll reopen the PR, please commit that change if you want.

dusanhornik:
done

PR #1875, opened 2020-09-13
Fixed typo

ARK17:


p7nov:
Hi. 
Thanks for your PR.
Anyway, both options are grammatically correct and the one we use (_approach to doing smth_) seems more widely used in the US English.

PR #1881, opened 2020-09-17
Update operator-overloading.md

punzki:
This matches what's actually generated by the compiler, at least for the JVM backend (didn't verify on other backends). Using `a.inc()` would be incorrect because `a` could be an expression with side-effects and hence would be evaluated twice: Once during initialization of `a0`, and then again in that step.


p7nov:
Thank you for this fix!

PR #1884, opened 2020-09-18
update: add info about reflection dependencies

p7nov:
fix for [KT-41280 Reflection documentation is missing the dependencies coordinates](https://youtrack.jetbrains.com/issue/KT-41280)

### DIFF
--- a/pages/docs/reference/reflection.md
+++ b/pages/docs/reference/reflection.md
@@ -11,13 +11,57 @@ Reflection is a set of language and library features that allows for introspecti
 Kotlin makes functions and properties first-class citizens in the language, and introspecting them (i.e. learning a name or 
 a type of a property or function at runtime) is closely intertwined with simply using a functional or reactive style.
 
-> On the Java platform, the runtime component required for using the reflection features is distributed as a separate
-JAR file (`kotlin-reflect.jar`). This is done to reduce the required size of the runtime library for applications
-that do not use reflection features. If you do use reflection, please make sure that the .jar file is added to the
-classpath of your project.
+> On the JavaScript platform, only class references are currently supported. [Learn more about reflection in Kotlin/JS](js-reflection.html).
 {:.note}
 
-## Class References
+## JVM dependency
+
+On the JVM platform, the runtime component required for using the reflection features is distributed as a separate
+artifact. This is done to reduce the required size of the runtime library for applications
+that do not use reflection features. If you do use reflection, add the dependency on `kotlin-reflect` to your project:
+* In Gradle:
+  <div class="multi-language-sample" data-lang="groovy">
+  <div class="sample" markdown="1" theme="idea" mode="groovy" data-highlight-only>
+  
+  ```groovy
+  dependencies {
+      compile "org.jetbrains.kotlin:kotlin-reflect:{{ site.data.releases.latest.version }}"
+  }
+  ```
+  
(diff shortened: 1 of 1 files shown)


pages/docs/reference/reflection.md
```diff
@@ -11,13 +11,57 @@ Reflection is a set of language and library features that allows for introspecti
...
+On the JVM platform, the runtime component required for using the reflection features is distributed as a separate
+artifact. This is done to reduce the required size of the runtime library for applications
+that do not use reflection features. If you do use reflection, add the dependency on `kotlin-reflect` to your project:
+* In Gradle:
+  <div class="multi-language-sample" data-lang="groovy">
+  <div class="sample" markdown="1" theme="idea" mode="groovy" data-highlight-only>
+  
+  ```groovy
+  dependencies {
+      compile "org.jetbrains.kotlin:kotlin-reflect:{{ site.data.releases.latest.version }}"
```
udalov, line 28: AFAIK `compile` is deprecated, do we want to promote it? Maybe `implementation`, as in the kts version?

pages/docs/reference/reflection.md
```diff
@@ -11,13 +11,57 @@ Reflection is a set of language and library features that allows for introspecti
...
+  <dependencies>
+      <dependency>
+          <groupId>org.jetbrains.kotlin</groupId>
+          <artifactId>kotlin-reflect</artifactId>
+      </dependency>
+  </dependencies>
+  ```
+  </div>
+
+If you don't use Gradle or Maven, you can add the `kotlin-reflect.jar` file from the Kotlin compiler distribution to
```
udalov, line 61: If one doesn't use Gradle or Maven, they're likely using either JPS or CLI (or Ant), and in all those cases `kotlin-reflect` is added by default (there's a flag `-no-reflect` to prevent this). Maybe rephrase this to something like "ensure you have kotlin-reflect.jar in the classpath"?

pages/docs/reference/reflection.md
```diff
@@ -11,13 +11,60 @@ Reflection is a set of language and library features that allows for introspecti
...
+          <groupId>org.jetbrains.kotlin</groupId>
+          <artifactId>kotlin-reflect</artifactId>
+      </dependency>
+  </dependencies>
+  ```
+  </div>
+
+If you don't use Gradle or Maven, make sure you have `kotlin-reflect.jar` in the classpath of your project.
+In other supported cases (IntelliJ IDEA projects, using command-line compiler or Ant),
+it is added by default unless `-no-reflect` compiler option is specified.
```
udalov, line 65: Sorry, I should've been more clear. `-no-reflect` indeed works for CLI and Ant, but it won't work for IDEA projects, because the project structure there determines what dependencies will be used. By default, Kotlin modules are configured *with* kotlin-reflect in dependencies. To avoid having kotlin-reflect in the classpath, I guess the only way is to manually remove kotlin-reflect.jar from the configured library in project settings. Not sure if it needs to be documented though :D

p7nov: This part about JPS seems too much about **not using** reflection for a page that explain how to **use** reflection :) Let just drop the mentioning of IDEA projects here.

udalov: Yes, I agree :+1:

udalov (approved):

PR #1893, opened 2020-09-26
fixed typo in line no. 418

keshav304:


p7nov:
Hi! 
Thanks for the fix. 
However, this is not a typo: this code uses the [trailing comma](https://kotlinlang.org/docs/reference/coding-conventions.html#trailing-commas) available in Kotlin since 1.4.

PR #1895, opened 2020-09-29
Fix broken newline

imknown:
The [official website](https://kotlinlang.org/docs/reference/java-interop.html#java-arrays) is correct, but the [markdown format](https://github.com/JetBrains/kotlin-web-site/blob/master/pages/docs/reference/java-interop.md#java-arrays) is broken.


p7nov:
Thanks!

PR #1898, opened 2020-09-30
Update README.md

Vaibhavkatre005:


p7nov:
Hi!

Thank you for this PR. Unfortunately, we don't see a practical value in these changes, so we won't merge this PR to keep the Git history clean and simple.

Anyway, we're always grateful fo our contributors, so don't hesitate to submit more PRs that fix issues or suggest more practical improvements.

PR #1900, opened 2020-10-01
Update docs

Dhruv122:


p7nov:
This doesn't seem correct.
See [this reference](https://docs.docker.com/compose/compose-file/).

PR #1907, opened 2020-10-03
Improve gradle scripts

naftalmm:
This PR mainly focused on complete migration to [Task Configuration Avoidance API](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) in gradle buildscripts, which is a recommended way to work with tasks in modern Gradle. It is available since Gradle 4.9, while Kotlin Gradle plugin 1.4.10 requires Gradle 5.4+, so it could be used in examples without any additional notes (actually it was already used somethere).

Also moved to usage of some overloaded methods variants in Gradle API, making kotlin code more idiomatic. 
And fix a couple of groovy scripts to make them fully equivalent with their kotlin counterparts.


pages/docs/reference/inline-classes.md
```diff
@@ -199,7 +199,7 @@ To remove the warning you have to opt in to the usage of this feature by passing
 <div class="sample" markdown="1" theme="idea" mode="groovy" data-lang="groovy">
 
 ```groovy
-compileKotlin {
+tasks.withType(KotlinCompile).configureEach {
```
h0tk3y, line 202: Maybe we should replace this and other occurrences where a language feature is enabled in a task's compiler args with `kotlin.sourceSets.all { languageSettings.enableLanguageFeature("InlineClasses") }`.
@p7nov WDYT?

pages/docs/reference/using-gradle.md
```diff
@@ -610,8 +610,10 @@ It is also possible to configure all Kotlin compilation tasks in the project:
 <div class="sample" markdown="1" mode="groovy" theme="idea" data-lang="groovy">
 
 ```groovy
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
-    kotlinOptions { //... }
```
h0tk3y, line 613: Note that the snippet that uses the FQN inline is a bit easier to copy-paste, so I'm not sure we should switch to using imports.

h0tk3y (approved):

pages/docs/reference/inline-classes.md
p7nov: It's a good idea, let's do.

p7nov:
Thank you for this update!

PR #1908, opened 2020-10-04
Update interfaces.md

imranpasha-tech:
A method is abstract by default if the body is not present.

### DIFF
--- a/pages/docs/reference/interfaces.md
+++ b/pages/docs/reference/interfaces.md
@@ -125,7 +125,7 @@ class D : A, B {
 ```
 </div>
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hi!

Thanks for the PR.
The text we currently have is correct, you can see it in the example above it: 
> bar() is *not marked* abstract in A

```
interface A {
    fun foo() { print("A") }
    fun bar() // this is abstract by default but *not marked* as such - no `abstract` modifier here
}
```

Your change does not reflect what we have in the example.

If you find this wording unclear, feel free to suggest improvements that preserve the meaning. Thank you!

Regards, Pavel
Kotlin Technical Writer.

p7nov:
Please feel free to reopen the PR is you need further discussions. Thank you!

PR #1921, opened 2020-10-12
Fix in create anonymous objects and the supertype

Husseinhj:
To describe `ERROR: Unresolved reference 'x'` error you should remove `object` from the `publicFoo()` method.

Playground test based on wrong documentation:
```
https://pl.kotl.in/83dNsbq5e
```

Playground test based on new documentation:

```
https://pl.kotl.in/OjLcqFynC
```

### DIFF
--- a/pages/docs/reference/object-declarations.md
+++ b/pages/docs/reference/object-declarations.md
@@ -73,7 +73,7 @@ class C {
     }
 
     // Public function, so the return type is Any
-    fun publicFoo() = object {
+    fun publicFoo() = {
         val x: String = "x"
     }
 


Husseinhj:
@p7nov

p7nov:
Hi! 
Thanks for your PR.
However, if you test the exact class shown in the example (without any additions), you get the error described in the text.
Please see here: https://pl.kotl.in/VvDcHABV5

Husseinhj:
@p7nov

Husseinhj:
@p7nov 
It shows totally different behavior when you use this code in the main function and in the root of the file.

You can see my example when I use in 
Playground test based on wrong documentation:
```
https://pl.kotl.in/83dNsbq5e
```

Playground test based on new documentation:

```
https://pl.kotl.in/OjLcqFynC
```

When you removed `object` type from the sample it's work exact behavior

p7nov:
@Husseinhj you're right in your examples: if C is declared within `main`, the code compiles.
The important thing is that the text above the example says `anonymous object as a return type of a public function`. But inside `main`, the function is not **public** - it's **local**. So the explanation and example in docs don't contradict. They are just unrelated to the case you show.

Husseinhj:
> So the explanation and example in docs don't contradict. They are just unrelated to the case you show.

Yes, you're right, that was clear. Thanks for explaining this issue.
@p7nov

p7nov:
OK then, I'll close this. Don't hesitate to create more PRs if you find issues in the docs.
Have a nice Kotlin!

PR #1925, opened 2020-10-15
Improve the convention on expresion bodies

LouisCAD:
Replace the previous snippet with a better example,
and specify that we're talking about the first line of the expression body.

See misinterpretation of the previous convention wording in this PR: https://github.com/LouisCAD/kotlin-libraries-playground/pull/68#discussion_r505742093


p7nov:
Thanks, Louis!

PR #1929, opened 2020-10-20
Update Incremental compilation section

StefMa:
I found the section about `Incremental compilation` in the docs.
It was only documented **how** to change the *default* setting. But it was not mention *what* the default setting is.
I found [this blog](https://blog.jetbrains.com/kotlin/2017/03/kotlin-1-1-1-is-out/) and thought it make sense to mention that it is enabled by default.

I also slightly updated the whole section to reflect more the *is is enabled by default, here is how to disable it* state 🙂 .


p7nov:
Thank you for this addition!

PR #1935, opened 2020-10-30
Typo fix

shibasis0801:
I was going through this to build a library to be called from TypeScript/React. 
I got these errors. 

Unresolved reference: binaries
Please initialize the Kotlin/JS target. Use:
kotlin {
    js {
        // To build distributions for and run tests on browser or Node.js use one or both of:
        browser()
        nodejs()
    }
}

Moving binaries.executable to the js block fixed it. 
I may be wrong, but I thought it is a typo, so raising a PR for the same.


SebastianAigner:
Well spotted! Thanks for your contribution! 🎉

PR #1938, opened 2020-11-03
update: MPP publication structure

katia-energizer:
updated mpp publication structure; updated the topic structure; added about gradle-bintray-plugin

### DIFF
--- a/pages/docs/reference/mpp-publish-lib.md
+++ b/pages/docs/reference/mpp-publish-lib.md
@@ -39,7 +39,7 @@ or migrate to the [`maven-publish` plugin](https://docs.gradle.org/current/userg
 
 ## Structure of publications
 
-The `maven-publish` plugin automatically creates publications for each target that can be built on the current host, except for the Android target, 
+The Kotlin plugin when used with `maven-publish` automatically creates publications for each target that can be built on the current host, except for the Android target, 
 which needs an [additional step to configure publishing](#publish-an-android-library).
 
 Publications of a multiplatform library include an additional _root_ publication `kotlinMultiplatform` that stands for the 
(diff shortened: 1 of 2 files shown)


pages/docs/reference/mpp-publish-lib.md
```diff
@@ -33,20 +32,33 @@ publishing {
...
 
+You can also use [`gradle-bintray-plugin`](https://github.com/bintray/gradle-bintray-plugin) for publishing multiplatform libraries. 
+However, this plugin does not support publishing Gradle module metadata required for [hierarchical structure support](mpp-share-on-platforms.html#share-code-on-similar-platforms).
+Use [this workaround](https://github.com/bintray/gradle-bintray-plugin/issues/229#issuecomment-473123891) to enable metadata publishing 
+or migrate to the [`maven-publish` plugin](https://docs.gradle.org/current/userguide/publishing_maven.html). 
+
+## Structure of publications
 
-Publications are automatically created for each target that can be built on the current host, except for the Android target, 
+The `maven-publish` plugin automatically creates publications for each target that can be built on the current host, except for the Android target,
```
h0tk3y, line 42: It's not the `maven-publish` plugin that creates the publications. The Kotlin plugin does that when used with `maven-publish`. And `maven-publish`, in turn, usually requires creating publications manually, so Kotlin is a bit special when used with `maven-publish` as the publications appear automatically.

h0tk3y (approved):

pages/docs/reference/mpp-publish-lib.md
katia-energizer: I didn't realize. I want to add about `maven-publish` because we have decided to mention `gradle-bintray-plugin`. I've rephrased:
> The Kotlin plugin when used with `maven-publish` automatically creates publications for each target...

Is it OK? Doesn't it relate to `gradle-bintray-plugin` BTW?

h0tk3y: Yep, this phrasing is OK. In fact, the `gradle-bintray-plugin` is also based on `maven-publish` and reuses the same publications.

h0tk3y (approved):

p7nov (approved):

PR #1954, opened 2020-11-18
Update KVision framework website address

rjaros:


p7nov:
Thanks for the update!

PR #1956, opened 2020-11-20
Fix args for main()

jottinger:
Removing unnecessary args from the sample "Hello, world" - generates a warning that may confuse newbies

### DIFF
--- a/pages/docs/tutorials/command-line.md
+++ b/pages/docs/tutorials/command-line.md
@@ -67,7 +67,7 @@ $ sudo snap install --classic kotlin
    <div class="sample" markdown="1" theme="idea">
 
    ```kotlin
-   fun main(args: Array<String>) {
+   fun main() {
        println("Hello, World!")
    }
    ```


LeamHall:
As a confused newbie, I'd also recommend putting some visual separation in the Changelog page (https://github.com/JetBrains/kotlin/releases/tag/v1.4.10). Linux people who are new to Kotlin, and used to downloading and compiling, may not see kotlin-compiler but are habituated to keying off "linux". So they, and I mean I, will download the native tar.gz file thinking I'm getting the Linux version of what the web-site tutorial's "Manual Install" section references.

LeamHall:
> @LeamHall I get your point, it's really valid. Could you please submit this as an issue to Kotlin [YouTrack](https://youtrack.jetbrains.com/issues/KT) so that people from the team can discuss it together?

@p7nov I looked at YouTrack, but I haven't been sold on Kotlin enough to create an account, or to care. I want to, really, but I spent most of my five hours in the Kotlin IRC chat being told that I was a <1% edge case and no one else on the planet would do what I was doing. Which was reading the instructions. I did make an error, and a couple of folks helped me see that. While telling me that I was an edge case.  :(

jottinger:
@LeamHall I think you're mischaracterizing the discussion, but you're not wrong even so: it IS an edge case, but maybe the fact that you ran into it is worth discussing. @p7nov, what can *I* do to help? He was running Kotlin native and interpreted the "unused argument" as an error, and then ran into a separate issue onboarding from there.

LeamHall:
> @LeamHall I think you're mischaracterizing the discussion, but you're not wrong even so: it IS an edge case, but maybe the fact that you ran into it is worth discussing. @p7nov, what can _I_ do to help? He was running Kotlin native and interpreted the "unused argument" as an error, and then ran into a separate issue onboarding from there.

Hey @jottinger, good to "see" you again!

@p7nov, there are two *different* issues here.  The technical issue you have already fixed; the arguments to "main". The second was the UI bit. The tutorial was *correct* in which version to get, the mistake that led me to Kotlin native was based on visual habit.  Once I realized that, and fixed it, the rest of the task worked. 

I accept that I'm an "edge case".  ( I have an ex-wife; I've been called worse. )   Maybe it's because I view "server-side" differently; I work in it all day long (file parsing, OS poking, regex munging) and don't use a Graphical IDE.  If I started doing Kotlin seriously then I would probably take @jottinger's advice and learn an GIDE to move faster.  Before that, though, I would want to edit it in vim, run it on the command line, and really understand what was going on.

p7nov:
I believe this topic about artifacts naming is out of this PR's scope. I totally agree that these are two separate issues. This is why I merged the _docs improvement_ suggested by @jottinger and recommended moving the _artifacts naming discussion_ to YouTrack. This is where you can reach folks who are in charge of publishing Kotlin distribs.
As for docs, we can extend the introduction of [this tutorial](https://kotlinlang.org/docs/tutorials/command-line.html) so that it explains what is in the distrib.

PR #1972, opened 2020-12-01
update: clarify propagation of `copy` function

tochev:
Clarify that the default `copy` implementation performs a shallow copy.

For people coming from some languages this might not be immediately clear (in C++ the copy constructor usually performs a deep copy, in python there are `copy` and `deepcopy`).

Perhaps the example code is not needed, but I decided to include it in the pull request.

### DIFF
--- a/pages/docs/reference/data-classes.md
+++ b/pages/docs/reference/data-classes.md
@@ -119,6 +119,28 @@ val olderJack = jack.copy(age = 2)
 
 </div>
 
+Note that the `copy()` call performs shallow copy, meaning that it doesn't propagate recursively through attributes:
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+data class User(var name : String)
+
+data class Address(var user: User, var city: String)
+
+var userJack = User(name="jack")
+var address = Address(user = userJack, city = "London")
+var addressCopy = address.copy()
+
+address.city = "New York"
+address.user.name = "john"
+
+println(addressCopy.city)  // prints "London"
+println(addressCopy.user.name)  // prints "john", because `address.user === addressCopy.user`
+```
+
+</div>
+
 ## Data Classes and Destructuring Declarations
 
 _Component functions_ generated for data classes enable their use in [destructuring declarations](multi-declarations.html):
--- a/pages/docs/reference/data-classes.md
+++ b/pages/docs/reference/data-classes.md
@@ -119,7 +119,7 @@ val olderJack = jack.copy(age = 2)
 
 </div>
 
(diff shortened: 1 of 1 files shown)


pages/docs/reference/data-classes.md
```diff
@@ -119,6 +119,28 @@ val olderJack = jack.copy(age = 2)
 
 </div>
 
+Note that the `copy()` call performs shallow copy, meaning that it doesn't propagate recursively through attributes:
```
p7nov, line 122: I suggest adding a clearer description of how `copy` treats reference and primitive properties.
```suggestion
Note that the `copy()` call performs a shallow copy, meaning that the copy's reference properties point to the _same objects_ as the original object's references. However, the copy has its own primitive properties not linked to the original.
```

pages/docs/reference/data-classes.md
```diff
@@ -119,6 +119,28 @@ val olderJack = jack.copy(age = 2)
...
+
+var userJack = User(name="jack")
+var address = Address(user = userJack, city = "London")
+var addressCopy = address.copy()
+
+address.city = "New York"
+address.user.name = "john"
+
+println(addressCopy.city)  // prints "London"
+println(addressCopy.user.name)  // prints "john", because `address.user === addressCopy.user`
```
p7nov, line 139: We can make this same runnable for a better demonstration. I also changed the example a bit, please check that it's OK with you.
```suggestion
<div class="sample" markdown="1" theme="idea">

```kotlin
fun main() {
//sampleStart
    var userJack = User(name = "Jack")
    var address = Address(user = userJack, city = "London")
    var addressCopy = address.copy()

    address.city = "New York" 
    address.user.name = "John" // Propagates to addressCopy because they both point to userJack.
//sampleEnd
    println("addressCopy.city is ${addressCopy.city}")  // Prints "London"
    println("addressCopy.user.name is ${addressCopy.user.name}")  // Prints "John"
}
```

p7nov (commented):
Hey!
Your addition is really valuable, thank you!
Please see my comments and suggestions on improving this.

pages/docs/reference/data-classes.md
tochev: Don't you need the class definitions?
```
data class User(var name : String)
data class Address(var user: User, var city: String)
```

Also I was thinking of another way to rewrite the example a bit.

tochev: I added your suggestions in a5753bd and I also changed the code so that the copy is modified, which seems to me to be the more common mistake make by developers. Please review it.

P.S. Sorry for the long absence, I was away during the holidays. Happy New Year :)

p7nov: Seems fine now, thank you! 
Happy New Year to you too =)

PR #1977, opened 2020-12-03
Serialization docs fixed

KrevedAccky:
My main concern was that when using a field with a default value without `@Require` annotation, this field will not be added to the resulting JSON, so I fixed the code, and now running it will lead to the expected result. Also, I updated naming because in one place the variable was called `b` whereas in another place it was `str`.


p7nov:
Thank you for this improvement!

PR #1979, opened 2020-12-05
fix: adhere to coding conventions in tailrec example

tochev:
Make tailrec example function definition adhere to the ["Expression Body Formatting" from the coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#expression-body-formatting)


p7nov:
Thank you!

PR #1993, opened 2020-12-20
Unify executable creation for the browser target

loxal:
The idiomatic way to create JS binaries seems to be after the browser block instead of inside it.


SebastianAigner:
Hey @loxal and happy 2021! Thank you for this PR. You're right – binaries is part of the `KotlinJsTargetDsl`, not the `KotlinJsBrowserDsl`, so it should probably be lifted out to avoid confusion. The code as seen in the docs is what the wizard currently generates. I've added a PR to have that part changed as well, see here: https://youtrack.jetbrains.com/issue/KT-44285

SebastianAigner (approved):

PR #1995, opened 2020-12-21
Fix formatting for dependency code block

ben-xD:
The code should be in the code block 😅


p7nov:
Hi! Thanks for pointing this out.
Actually, it's a typo in the dependency name - it must be `org.jetbrains.kotlinx:kotlinx-serialization-json` in the description - matching the example. See the original kotlinx.serialization docs [here](https://github.com/Kotlin/kotlinx.serialization#dependency-on-the-json-library).
We'll fix this. 
Thanks again!

PR #1996, opened 2020-12-21
Update to use the proper kotlinx-html jar reference

pambrose:
An app will not compile with the kotlinx-html-js reference. It fails in the 
compileKotlinJs step. Using kotlinx-html fixes the problem.

### DIFF
--- a/pages/docs/tutorials/javascript/typesafe-html-dsl.md
+++ b/pages/docs/tutorials/javascript/typesafe-html-dsl.md
@@ -20,7 +20,7 @@ repositories {
 
 dependencies {
     implementation(kotlin("stdlib-js"))
-    implementation("org.jetbrains.kotlinx:kotlinx-html-js:0.7.1")
+    implementation("org.jetbrains.kotlinx:kotlinx-html:0.7.1")
     // ...
 }
 ```


p7nov:
Hey!
Sorry for the late response. I've just reproduced the tutorial steps, and everything works fine with Kotlin 1.4.20 and `org.jetbrains.kotlinx-html-js:0.7.1`. 
If I change the dependency to just `org.jetbrains.kotlinx:kotlinx-html:0.7.1` as you suggest, the project doesn't see anything from `kotlin.html.*` and the code fails to compile.
You probably have some other issues. Please feel free to ask for help in [YouTrack](https://youtrack.jetbrains.com/issues/KT) or [Kotlin Slack](https://kotlinlang.slack.com/) (get invite [here](https://surveys.jetbrains.com/s3/kotlin-slack-sign-up)).

pambrose:
Agreed, I no longer see the problem on my end, so I am not sure what the problem was. However, it is worth mentioning that the IJ "New Project" wizard produces code using:
```
dependencies {
    testImplementation(kotlin("test-js"))
    implementation("org.jetbrains.kotlinx:kotlinx-html:0.7.2")
}
``` 

Thanks for looking into it.

Paul

PR #1998, opened 2020-12-27
Update inline-classes.md

amatanat:


pages/docs/reference/inline-classes.md
```diff
@@ -38,12 +38,17 @@ This is the main feature of inline classes, which inspired the name "inline": da
 
 ## Members
 
-Inline classes support some functionality of regular classes. In particular, they are allowed to declare properties and functions:
+Inline classes support some functionality of regular classes. In particular, they are allowed to declare properties, functions and have *init*{: .keyword } blocks:
```
ilya-g, line 41:
```suggestion
Inline classes support some functionality of regular classes. In particular, they are allowed to declare properties, functions and, starting with Kotlin 1.4.30, to have *init*{: .keyword } blocks:
```

ilya-g (commented):
I propose to clarify since when this feature is supported. Otherwise, ok to be released with 1.4.30 site update.

p7nov:
Thanks for this update. We've published the same changes together with all documentation updates for Kotlin 1.4.30. See [here](https://kotlinlang.org/docs/reference/inline-classes.html).

PR #2003, opened 2021-01-10
easier to understand what is key and what is value

SoltauFintel:
I suggest to change https://kotlinlang.org/docs/reference/idioms.html

I'm a Java pro and a Kotlin newbie. It was hard for me to understand what is the key and what is the value.

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -118,7 +118,7 @@ val list = listOf("a", "b", "c")
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
-val map = mapOf("a" to 1, "b" to 2, "c" to 3)
+val map = mapOf("key-a" to "value-a", "key-b" to "value-b")
 ```
 </div>
 
@@ -126,8 +126,9 @@ val map = mapOf("a" to 1, "b" to 2, "c" to 3)
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
-println(map["key"])
-map["key"] = value
+val map = mutableMapOf("key-a" to "value-a", "key-b" to "value-b")
+map["key-c"] = "value-c"
+println(map["key-c"]) // prints "value-c"
 ```
 </div>
 


pages/docs/reference/idioms.md
```diff
@@ -118,16 +118,17 @@ val list = listOf("a", "b", "c")
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
-val map = mapOf("a" to 1, "b" to 2, "c" to 3)
+val map = mapOf("key-a" to "value-a", "key-b" to "value-b")
```
p7nov, line 121: Perhaps we can add an explanation in a comment. This will keep the code concise. What do you think?
```suggestion
val map = mapOf("a" to 1, "b" to 2) // "key to value" pairs
```

pages/docs/reference/idioms.md
```diff
@@ -118,16 +118,17 @@ val list = listOf("a", "b", "c")
...
 
 ### Accessing a map
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 ```kotlin
-println(map["key"])
-map["key"] = value
+val map = mutableMapOf("key-a" to "value-a", "key-b" to "value-b")
+map["key-c"] = "value-c"
+println(map["key-c"]) // prints "value-c"
```
p7nov, line 131: Not sure we should extend the idiom examples this much. This is just a short reminder, and the current example is pretty clear about what is key and what is value.
The complete runnable examples are provided in the [maps documentation](https://kotlinlang.org/docs/reference/map-operations.html).
I wouldn't change this part.

p7nov (commented):
Hey! 
Good point, thank you.
However, we keep the idiom examples as short as possible, so I suggest a shorter version of what you add. Please check it out.

Regards, 
Pavel,
Kotlin Technical Writer.

pages/docs/reference/idioms.md
SoltauFintel: I think my solution is more understandable for beginners.

pages/docs/reference/idioms.md
SoltauFintel: I wanted to show here that you can not only change an entry, but also add one. Unfortunately, the mutableMapOf line is necessary for this, since mapOf from the previous section would not work.

I don't think Kotlin pros really need this page. Isn't it especially for newbies? And the more understandable the site, the fewer newbies will be put off by Kotlin.

p7nov: Well, in some sense this page _is helpful_ for newbies but it's _not intended for teaching_ Kotlin concepts. It just shows how common things look in Kotlin without explaining how they work and how/when to use them.
This is why we're keeping code samples here **as short as possible**, as opposed to samples [here](https://kotlinlang.org/docs/reference/collections-overview.html#map) for instance.
In your suggestion, the `map` declaration is valid but it's not reflecting what's written in the heading (**Accessing a map**). The other two lines are the same as the current ones.

SoltauFintel: I wrote this in what I think is best. Ultimately, it is your decision to take it over, change it, or discard it. :-)

p7nov:
Hey! 
Sorry for not responding, we've been a bit busy with the Kotlin 1.4.30 release lately.
Anyway, thank you for the PR. For not we don't see that people generally have issues with these examples, so let's keep them as is. We'll surely return to them if we have more complaints.

Thank you and have a nice Kotlin!

PR #2004, opened 2021-01-10
instantiate abstract class (new idiom)

SoltauFintel:
I had an abstract class with 2 abstract methods. Now I wanted to instantiate this anonymously and implement the 2 methods. Android Studio did not provide any help on this. Frustrating. It is important to specify "object :".

The documentation (especially for Java pros) could be improved here. My contribution is a start. I would like to see a Classes overview page (above https://kotlinlang.org/docs/reference/classes.html ) where all the keywords around classes are mentioned shortly! (class, data class, open class and especially object!)

### DIFF
--- a/pages/docs/reference/idioms.md
+++ b/pages/docs/reference/idioms.md
@@ -161,6 +161,29 @@ object Resource {
 ```
 </div>
 
+### Instantiate abstract class
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+```kotlin
+abstract class AnAbstractClass {
+    abstract fun doSomething()
+    abstract fun sleep()
+}
+
+fun main() {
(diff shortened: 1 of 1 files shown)


SoltauFintel:
Works also for interfaces.

p7nov:
Thanks for this addition! 
Regarding the new overview pages: we're soon moving Kotlin docs to a new engine (see [this blog post](https://blog.jetbrains.com/kotlin/2020/12/revamped-kotlin-documentation/) for the details) where this is technically possible - there is a hierarchical page structure.
We will consider creating overview pages for all big sections. However, I can't promise that this will happen soon - there is still much work to do before.

Regards, Pavel
Kotlin Technical Writer.

SoltauFintel:
The page menu on the right side is good and really needed.

PR #2011, opened 2021-01-15
update: sealed interface doc

p7nov:
Doc update for:
- [KT-20423 Support sealed interfaces](https://youtrack.jetbrains.com/issue/KT-20423)
- [KT-13495 Sealed class hierarchy and nesting should allow more freedom](https://youtrack.jetbrains.com/issue/KT-13495)

[KEEP](https://github.com/Kotlin/KEEP/blob/sealed-interface-freedom/proposals/sealed-interface-freedom.md)

### DIFF
--- a/pages/docs/reference/sealed-classes.md
+++ b/pages/docs/reference/sealed-classes.md
@@ -7,24 +7,41 @@ title: "Sealed Classes"
 
 # Sealed Classes
 
-> **Sealed interfaces** are [Experimental](evolution/components-stability.html). They may be dropped or changed at any time.
-> Opt-in is required (see details [below](#experimental-status-of-sealed-interfaces-and-extended-inheritance-rules)). Use them only for evaluation purposes. We appreciate your feedback on them in [YouTrack](https://youtrack.jetbrains.com/issues/KT-42433).
-{:.note}
-
-_Sealed_ classes and interfaces represent restricted class hierarchies that provide more control over inheritance. 
+_Sealed_ classes represent restricted class hierarchies that provide more control over inheritance. 
 All subclasses of a sealed class are known at compile time. No other subclasses may appear after
 a module with the sealed class is compiled. For example, third-party clients can't extend your sealed class in their code.
-
-The same works for sealed interfaces and their implementations: once a module with a sealed interface is compiled, 
-no new implementations can appear.
-
 Thus, instances of sealed classes can have one of the types from a limited set closed for further extension.
 
 In some sense, sealed classes are similar to enum classes: the set of values
 for an enum type is also restricted, but each enum constant exists only as a _single instance_, whereas a subclass
 of a sealed class can have _multiple_ instances, each with its own state.
 
-To declare a sealed class or interface, put the `sealed` modifier before its name.
+To declare a sealed class, put the `sealed` modifier before its name.
+
+<div class="sample" markdown="1" theme="idea" data-highlight-only>
+
+```kotlin
+sealed class Expr
+
+data class Const(val number: Double) : Expr()
+data class Sum(val e1: Expr, val e2: Expr) : Expr()
+object NotANumber : Expr()
+```
+</div>
+
+A sealed class is [abstract](classes.html#abstract-classes) by itself, it cannot be instantiated directly and can have *abstract*{: .keyword } members.
+
+Sealed classes are not allowed to have non-*private*{: .keyword } constructors (their constructors are *private*{: .keyword } by default).
+
+## Sealed interfaces
+
+> Sealed interfaces are [Experimental](evolution/components-stability.html). They may be dropped or changed at any time.
+> Opt-in is required (see details [below](#experimental-status-of-sealed-interfaces-and-extended-inheritance-rules)). Use them only for evaluation purposes. We appreciate your feedback on them in [YouTrack](https://youtrack.jetbrains.com/issues/KT-42433).
+{:.note}
+
+Interfaces can be declared `sealed` as well as classes. The `sealed` modifier works on interfaces the same way:
+all implementations of a sealed interface are known at compile time. Once a module with a sealed interface is compiled,
+no new implementations can appear.
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
@@ -38,10 +55,6 @@ data class Sum(val e1: Expr, val e2: Expr) : MathExpr()
 object NotANumber : Expr
 ```
 </div>
-
-A sealed class is [abstract](classes.html#abstract-classes) by itself, it cannot be instantiated directly and can have *abstract*{: .keyword } members.
-
-Sealed classes are not allowed to have non-*private*{: .keyword } constructors (their constructors are *private*{: .keyword } by default).
 
 ## Inheritance rules
 
@@ -86,7 +99,7 @@ fun eval(expr: Expr): Double = when(expr) {
 
 ## Experimental status of sealed interfaces and extended inheritance rules
 
-Sealed interfaces and [extended inheritance rules](#extended-inheritance-rules) are [Experimental](evolution/components-stability.html).
+[Sealed interfaces](#sealed-interfaces) and [extended inheritance rules](#extended-inheritance-rules) are [Experimental](evolution/components-stability.html).
 To be able to use them in your code, switch to the language version `1.5`:
 * In Gradle, add the [compiler option](using-gradle.html#attributes-common-for-jvm-and-js) `languageVersion` with the value `1.5`.
 
--- a/pages/docs/reference/sealed-classes.md
+++ b/pages/docs/reference/sealed-classes.md
@@ -7,38 +7,60 @@ title: "Sealed Classes"
 
 # Sealed Classes
 
-Sealed classes are used for representing restricted class hierarchies, when a value can have one of the types from a
-limited set, but cannot have any other type. They are, in a sense, an extension of enum classes: the set of values
-for an enum type is also restricted, but each enum constant exists only as a single instance, whereas a subclass
-of a sealed class can have multiple instances which can contain state.
+_Sealed_ classes and interfaces represent restricted class hierarchies that provide more control over inheritance. 
+All subclasses of a sealed class are known at compile time. No other subclasses may appear after
+a module with the sealed class is compiled. For example, third-party clients can't extend your sealed class in their code.
 
-To declare a sealed class, you put the `sealed` modifier before the name of the class. A sealed class can have
-subclasses, but all of them must be declared in the same file as the sealed class itself. (Before Kotlin 1.1,
-the rules were even more strict: classes had to be nested inside the declaration of the sealed class).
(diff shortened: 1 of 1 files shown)


demiurg906 (approved):

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,64 @@ title: "Sealed Classes"
...
-
 A sealed class is [abstract](classes.html#abstract-classes) by itself, it cannot be instantiated directly and can have *abstract*{: .keyword } members.
 
 Sealed classes are not allowed to have non-*private*{: .keyword } constructors (their constructors are *private*{: .keyword } by default).
 
-Note that classes which extend subclasses of a sealed class (indirect inheritors) can be placed anywhere, not necessarily in
-the same file.
+## Inheritance rules
+
+Direct subclasses of sealed classes and interfaces must be declared in the same package. They may be top-level or nested
```
elizarov, line 48: This freedom is also experimental and opt-in for now. https://youtrack.jetbrains.com/issue/KT-13495

elizarov (changes requested):

pages/docs/reference/sealed-classes.md
p7nov: Please check the new version.

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,86 @@ title: "Sealed Classes"
...
-(The example above uses one additional new feature of Kotlin 1.1: the possibility for data classes to extend other
-classes, including sealed classes.)
-
 A sealed class is [abstract](classes.html#abstract-classes) by itself, it cannot be instantiated directly and can have *abstract*{: .keyword } members.
 
 Sealed classes are not allowed to have non-*private*{: .keyword } constructors (their constructors are *private*{: .keyword } by default).
 
-Note that classes which extend subclasses of a sealed class (indirect inheritors) can be placed anywhere, not necessarily in
-the same file.
+## Sealed interfaces
```
p7nov, line 36: Added a separate section.
@elizarov please check.

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,86 @@ title: "Sealed Classes"
...
 # Sealed Classes
 
-Sealed classes are used for representing restricted class hierarchies, when a value can have one of the types from a
-limited set, but cannot have any other type. They are, in a sense, an extension of enum classes: the set of values
-for an enum type is also restricted, but each enum constant exists only as a single instance, whereas a subclass
-of a sealed class can have multiple instances which can contain state.
+_Sealed_ classes represent restricted class hierarchies that provide more control over inheritance. 
+All subclasses of a sealed class are known at compile time. No other subclasses may appear after
+a module with the sealed class is compiled. For example, third-party clients can't extend your sealed class in their code.
+Thus, instances of sealed classes can have one of the types from a limited set closed for further extension.
```
katia-energizer, line 13: This phrase is difficult to read and follow. I read it three time and I'm still not sure that I understand it correctly:
> one of the types from a limited set closed for further extension

Do you mean `only one type`? `limited set closed` is difficult to read as well.

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,86 @@ title: "Sealed Classes"
...
 data class Const(val number: Double) : Expr()
 data class Sum(val e1: Expr, val e2: Expr) : Expr()
 object NotANumber : Expr()
 ```
 </div>
 
-(The example above uses one additional new feature of Kotlin 1.1: the possibility for data classes to extend other
-classes, including sealed classes.)
-
 A sealed class is [abstract](classes.html#abstract-classes) by itself, it cannot be instantiated directly and can have *abstract*{: .keyword } members.
```
katia-energizer, line 32: Do we keep `{: .keyword}` in old docs?

pages/docs/reference/sealed-classes.md
```diff
@@ -48,3 +96,19 @@ fun eval(expr: Expr): Double = when(expr) {
...
+
+## Experimental status of sealed interfaces and extended inheritance rules
+
+[Sealed interfaces](#sealed-interfaces) and [extended inheritance rules](#extended-inheritance-rules) are [Experimental](evolution/components-stability.html).
+To be able to use them in your code, switch to the language version `1.5`:
+* In Gradle, add the [compiler option](using-gradle.html#attributes-common-for-jvm-and-js) `languageVersion` with the value `1.5`.
+
+<div class="sample" markdown="1" mode="groovy" theme="idea">
+
+```groovy
```
katia-energizer, line 108: Does it work for Kotlin in the same way as in Groovy?

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,86 @@ title: "Sealed Classes"
...
+
+sealed class MathExpr(): Expr
+
+data class Const(val number: Double) : MathExpr()
+data class Sum(val e1: Expr, val e2: Expr) : MathExpr()
+object NotANumber : Expr
+```
+</div>
+
+## Inheritance rules
```
katia-energizer, line 59: This structure seems strange:
* Inheritance rules with information only about location. Is it the only rule? Do we need to call it `Inheritance rules`?
* Only one child heading.

`Extended` – is it our official naming? If we extend only one rule about location, it seems strange. Maybe use another word, such as additional or experimental.

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,86 @@ title: "Sealed Classes"
...
 
-Note that classes which extend subclasses of a sealed class (indirect inheritors) can be placed anywhere, not necessarily in
-the same file.
+## Sealed interfaces
+
+> Sealed interfaces are [Experimental](evolution/components-stability.html). They may be dropped or changed at any time.
+> Opt-in is required (see details [below](#experimental-status-of-sealed-interfaces-and-extended-inheritance-rules)). Use them only for evaluation purposes. We appreciate your feedback on them in [YouTrack](https://youtrack.jetbrains.com/issues/KT-42433).
+{:.note}
+
+Interfaces can be declared `sealed` as well as classes. The `sealed` modifier works on interfaces the same way:
```
katia-energizer, line 42: Don't you want to add info when sealed interfaces could be helpful compared to sealed classes? Not just specify how to use them, but also provide a guideline in what cases to use them.

katia-energizer (changes requested):
I've left a few comments and questions. Please take a look.

pages/docs/reference/sealed-classes.md
p7nov: Not sure we should care about it right before completely migrating to WH.

pages/docs/reference/sealed-classes.md
p7nov: It's basically the same as how 'usual' interfaces complement 'usual' classes: contract definition, multiple inheritance, etc. I don't think that this is the place to describe how interfaces extend capabilities for creating type hierarchies.
So the thought is exactly "Interfaces can be sealed as well". How to use interfaces seems out of scope here.

pages/docs/reference/sealed-classes.md
p7nov: Good catch, thanks. I'll rename the sections.
Regarding one child heading: I think it's fine here, because:
- the subsection extends the location topic, so it's strange to put it outside this section.
- it must be separated because it describes an "advanced/future" aspect of this topic.

pages/docs/reference/sealed-classes.md
p7nov: Yes, this line is the same in both languages.

pages/docs/reference/sealed-classes.md
katia-energizer: OK if there is nothing helpful to add.

pages/docs/reference/sealed-classes.md
katia-energizer: OK if it works

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,83 @@ title: "Sealed Classes"
...
 # Sealed Classes
 
-Sealed classes are used for representing restricted class hierarchies, when a value can have one of the types from a
-limited set, but cannot have any other type. They are, in a sense, an extension of enum classes: the set of values
-for an enum type is also restricted, but each enum constant exists only as a single instance, whereas a subclass
-of a sealed class can have multiple instances which can contain state.
+_Sealed_ classes represent restricted class hierarchies that provide more control over inheritance.
+All subclasses of a sealed class are known at compile time. No other subclasses may appear after
+a module with the sealed class is compiled. For example, third-party clients can't extend your sealed class in their code.
+Thus, each instance of a sealed class belongs a type from a limited set that is known when this class is compiled.
```
katia-energizer, line 13: Seems like a preposition is missing in `belongs to`. However, `has a type` seems more natural to me.

pages/docs/reference/sealed-classes.md
```diff
@@ -48,3 +93,19 @@ fun eval(expr: Expr): Double = when(expr) {
 }
 ```
 </div>
+
+## Try sealed interfaces and additional location for direct subclasses
```
katia-energizer, line 97: `Try additional location` seems unnatural. Maybe something like `Try placing direct subclasses in the same package` or `using direct subclasses from the same package`.

pages/docs/reference/sealed-classes.md
```diff
@@ -48,3 +93,19 @@ fun eval(expr: Expr): Double = when(expr) {
...
+
+[Sealed interfaces](#sealed-interfaces) and [extended inheritance rules](#additional-location-the-same-package) are [Experimental](evolution/components-stability.html).
```
katia-energizer, line 99: This sentence still includes `extended inheritance rules.`

katia-energizer (changes requested):
A few comments on wording and naming. Overall, looks fine 👍

pages/docs/reference/sealed-classes.md
```diff
@@ -7,38 +7,83 @@ title: "Sealed Classes"
...
 # Sealed Classes
 
-Sealed classes are used for representing restricted class hierarchies, when a value can have one of the types from a
-limited set, but cannot have any other type. They are, in a sense, an extension of enum classes: the set of values
-for an enum type is also restricted, but each enum constant exists only as a single instance, whereas a subclass
-of a sealed class can have multiple instances which can contain state.
+_Sealed_ classes represent restricted class hierarchies that provide more control over inheritance.
+All subclasses of a sealed class are known at compile time. No other subclasses may appear after
+a module with the sealed class is compiled. For example, third-party clients can't extend your sealed class in their code.
+ iThus, each instance of a sealed class has a type from a limited set that is known when this class is compiled.
```
katia-energizer, line 13:
```suggestion
 Thus, each instance of a sealed class has a type from a limited set that is known when this class is compiled. 
```
Typo

katia-energizer (approved):
LGTM. Just minor comments.

PR #1931, opened 2020-10-22
Clarified the usage of fields in kotlin

Wasabi375:
This is based on the forum discussion here: https://discuss.kotlinlang.org/t/property-vs-field/8179/10?u=wasabi375

PR #2012, opened 2021-01-17
Fix link titles referring to the "Properties" page

yvolk:
After we changed the target page title in the merge request #1931 (based on the discussion here: https://discuss.kotlinlang.org/t/property-vs-field/8179/10?u=yvolk ), I noticed that we forgot to change titles of the links to that page.
The target page is: https://kotlinlang.org/docs/reference/properties.html


p7nov:
Thank you!

PR #2013, opened 2021-01-17
Update js-interop.md -- Dukat link

reeve567:
The previous link for dukat gave a 404 error, just added the right(?) link to fix it


SebastianAigner:
Yup, seems like this one slipped through the cracks – thanks a lot for your contribution, merged! 🎉

PR #2018, opened 2021-01-26
update: update inline classes to beta

koshachy:


docs/topics/inline-classes.md
```diff
@@ -1,19 +1,29 @@
...
 overhead due to additional heap allocations. Moreover, if the wrapped type is primitive, the performance hit is terrible, 
 because primitive types are usually heavily optimized by the runtime, while their wrappers don't get any special treatment. 
 
-To solve such issues, Kotlin introduces a special kind of class called an *inline class*, which is declared by placing an 
-`inline` modifier before the name of the class:
+To solve such issues, Kotlin introduces a special kind of class called an *inline class*. 
+Inline classes are a subset of value-based classes. They don't have identity and hold values only.
+
+To declare an inline class, use `inline` or `value` modifier before the name of the class.
+Kotlin also requires the `@JvmInline` annotation before class declaration for the JVM backend:
```
ilmirus, line 17: The annotation is needed only before `value` classes, but the wording suggests, that one have to annotate `inline` classes as well.

koshachy: I've updated the description — wrote about `value` explicitly.

docs/topics/inline-classes.md
```diff
@@ -1,19 +1,29 @@
...
-`inline` modifier before the name of the class:
+To solve such issues, Kotlin introduces a special kind of class called an *inline class*. 
+Inline classes are a subset of value-based classes. They don't have identity and hold values only.
+
+To declare an inline class, use `inline` or `value` modifier before the name of the class.
+Kotlin also requires the `@JvmInline` annotation before class declaration along with the `value` modifier for the JVM backend:
 
 ```kotlin
 inline class Password(val value: String)
+
```
p7nov, line 21: we should separate these classes into different code blocks. This code block is incorrect because it can't be used as a whole.
I'd also suggest putting first and second class after the sentence about modifiers, then the sentence about @JvmInline, then the class with annotation.

docs/topics/inline-classes.md
```diff
@@ -135,10 +153,26 @@ fun compute(x: UInt) { }
 To mitigate such issues, functions using inline classes are *mangled* by adding some stable hashcode to the function name. 
 Therefore, `fun compute(x: UInt)` will be represented as `public final void compute-<hashcode>(int x)`, which solves the clash problem.
 
-> `-` is an *invalid* symbol in Java, meaning that it's not possible to call functions which accept inline classes from Java.
+> Mangling schemes in Kotlin 1.4 and 1.4.30 are different.
```
p7nov, line 156: What do you mean by 1.4? Can we specify one exact version when the scheme changed?

docs/topics/inline-classes.md
```diff
@@ -135,10 +153,26 @@ fun compute(x: UInt) { }
...
 {type="note"}
 
+### Calling from Java code
+
+You can call functions which accept inline classes from Java code. To do that, you should disable mangling manually:
+add the `@JvmName` annotation before the function declaration:
+
+```kotlin
+@JvmInline
+value class UInt(val x: Int)
```
p7nov, line 168: Not sure it's a good example since there is a built-in class with the same name: https://kotlinlang.org/docs/basic-types.html#unsigned-integers
Please check.

p7nov (changes requested):

docs/topics/inline-classes.md
koshachy: It also was in the previous doc version.

@ilmirus should we change the example?

docs/topics/inline-classes.md
koshachy: Fixed.
I also thought about another sentence for the annotation :)

docs/topics/inline-classes.md
p7nov: I suspect it was written before the unsigned types appeared :) 🦣

koshachy: So, we should also renew the example in KEEP https://github.com/Kotlin/KEEP/blob/master/proposals/inline-classes.md

ilmirus: The class is not built-in, but rather standard. Given that it is implemented as an inline class, it is a perfect example of inline class, with JetBrains seal of approval.

So, I see no reason to change the example.

p7nov (approved):

ilmirus (approved):

PR #2028, opened 2021-01-28
A small addition to the docs for ZSH shell installs

AthulMuralidhar:
The SDKMan install step does not mention about anything about the ZSH shell installs, while this is documented in the main site. I just thought it would be helpful for other ZSH peeps coming this way to install Kotlin

### DIFF
--- a/pages/docs/tutorials/command-line.md
+++ b/pages/docs/tutorials/command-line.md
@@ -24,6 +24,11 @@ Simply run the following in a terminal and follow any instructions:
 ```bash
 $ curl -s https://get.sdkman.io | bash
 ```
+The above command works for ZSH shells as well. Following it up with the `source` linnking should get `sdkman` working in your current terminal shell
+
+```bash
+source "$HOME/.sdkman/bin/sdkman-init.sh"
+```
 
 </div>
 


pages/docs/tutorials/command-line.md
```diff
@@ -24,6 +24,11 @@ Simply run the following in a terminal and follow any instructions:
 ```bash
 $ curl -s https://get.sdkman.io | bash
 ```
+The above command works for ZSH shells as well. Following it up with the `source` linnking should get `sdkman` working in your current terminal shell
```
p7nov, line 27: We want to keep this instruction focused on Kotlin, so let's not overload the instruction with details about SDKMAN.
Instead of the lines you add, I suggest mentioning ZSH right after the list of supported OS. Something like this:
"An easier way to install Kotlin on UNIX based systems such as OS X, Linux, Cygwin, FreeBSD and Solaris is by using SDKMAN!. It also works in Bash and ZSH shells."

p7nov (changes requested):
Hey!
This is a good addition, thank you.
I just want to keep the instruction concise and don't see much value in duplicating SDKMAN's instructions. Please see my suggestion in the comment below.

Regards, Pavel
Kotlin Technical Writer

p7nov:
We've just moved our docs to the new engine. Now we're closing all open PRs because they don't match the new structure and can't be merged. We appreciate your contribution; we've transferred these changes to the new docs manually to ensure that nothing is lost in transition.
Thanks again and have a nice Kotlin!

AthulMuralidhar:
> We've just moved our docs to the new engine. Now we're closing all open PRs because they don't match the new structure and can't be merged. We appreciate your contribution; we've transferred these changes to the new docs manually to ensure that nothing is lost in transition.

in light of your moving the docs repo, could you also please add your suggestions to the commits as well? as i do not see how i can do that due to the closure of this issue

p7nov:
Last week, we've been migrating the docs, so this PR would become invalid. To save it from being lost, I've added the [thing I suggested here](https://github.com/JetBrains/kotlin-web-site/pull/2028#discussion_r572606226) in the new doc file by myself.
I've seen your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-45024). Please don't consider the closing of this PR as "being shot down". The thing you add is really useful; it was just the specific moment when we had to migrate all changes from open PRs by ourselves.
I've checked that [the same instruction](https://kotlinlang.org/docs/command-line.html#sdkman) works with ZSH on macOS; the sdkman docs say [the same](https://sdkman.io/install). I also took the liberty to adjust the text to our documentation guidelines as I suggested above.
We always welcome contributions from the community and try to polish every PR so that it follows the Kotlin docs style. Sorry if this situation turned you down; it's really unusual – it was the only doc migration in the entire Kotlin's history.

PR #2031, opened 2021-01-28
Adding kscript as a way to execute Kotlin files

AthulMuralidhar:
This is another small suggestion to get the Kotlin files running as soon as possible

### DIFF
--- a/pages/docs/tutorials/command-line.md
+++ b/pages/docs/tutorials/command-line.md
@@ -105,6 +105,19 @@ $ sudo snap install --classic kotlin
 
     </div>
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hey!
Thank you for this PR.
However, like many other Kotlin-related software, `kscript` is a product of its authors who aren't affiliated with the Kotlin team at JetBrains. It's being developed independently, we don't test its or track its updates.
Thus, we're not ready to recommend `kscript` as a part of the Kotlin ecosystem in its official documentation.

Regards, Pavel
Kotlin Technical Writer

PR #2038, opened 2021-01-30
Reference to enum classes in sealed-classes.md

utgarda:
Enum classes are mentioned at the beginning of the article. Made it a reference.


p7nov:
Hi! Good point, thanks.

However, the link you add isn't entirely correct; it should point to https://kotlinlang.org/docs/reference/enum-classes.html instead. Could you please update your branch and insert the correct link?

Regards, Pavel
Kotlin Technical Writer

p7nov:
We've just moved our docs to the new engine. Now we're closing all open PRs because they don't match the new structure and can't be merged. We appreciate your contribution; we've transferred these changes to the new docs manually to ensure that nothing is lost in transition.
Thanks again and have a nice Kotlin!

PR #2051, opened 2021-02-03
Fix inverted directory path in documentation

sleclercq:
Result of `gradle nativeBinaries` is actually `build/bin/native` and not `build/native/bin`. This pull request changes that.


p7nov:
Thank you for this fix!

PR #2052, opened 2021-02-04
update: update spring boot tutorial

koshachy:


docs/topics/jvm/jvm-spring-boot-restful.md
```diff
@@ -0,0 +1,177 @@
...
+   The project has the following structure: 
+   ![The Spring Boot project structure](spring-boot-project-structure.png)
+   
+    There are packages and classes under the `main/kotlin` folder that belong to the application. The entry point to the application is the `main()` method of the `DemoApplication.kt` file.
+
+## Explore the project build file
+
+Open the `build.gradle.kts` file:
+
+```kotlin
```
p7nov, line 35: Not sure about the value of the whole build script here. What do you want to say with it? Can we avoid listing third-party code? It will be hard to support it: there are a lot of versions (including Kotlin 1.4.21 which is outdated in scope of our docs), and any other changes can happen too.

docs/topics/jvm/jvm-spring-boot-restful.md
```diff
@@ -0,0 +1,177 @@
...
+}
+```
+
+Comparing to Java, the application file has the following differences:
+* As Spring Boot looks for a public static `main()` method, the Kotlin uses a [top-level function](functions.md#function-scope) defined outside `DemoApplication` class.
+* The `DemoApplication` class is not declared as `open`, since the [kotlin-spring](all-open-plugin.md#spring-support) Gradle plugin does that automatically.
+
+## Create a data class and a controller
+
+To create an endpoint, you need to create a [data class](data-classes.md) and a controller:
```
p7nov, line 103: to create you need to create :) Maybe `add` or something similar?

p7nov: 'you need' is also redundant

p7nov: Or even rephrase to: "To create an endpoint, you need a data class..."

docs/topics/jvm/jvm-spring-boot-restful.md
```diff
@@ -0,0 +1,177 @@
...
+1. In the `DemoApplication.kt` file, create a `Message` data class with two properties: `id` and `text`:
+
+  ```kotlin
+   data class Message(val id: String?, val text: String)
+   ```
+
+2. In the same file, create a `MessageResource` class which will serve the requests and return a JSON document representing a collection of `Message` objects:
+
+  ```kotlin
+   @RestController
```
p7nov, line 114: Don't we need short explanations or doc links for Spring annotations? This seems a pretty basic tutorial. I doubt that all readers know these.

docs/topics/jvm/jvm-spring-boot-restful-db.md
```diff
@@ -0,0 +1,186 @@
...
+2. Execute all POST requests. Use the green **Run** icon in the gutter to the request declaration.
+   These requests write the text messages to the database.
+    
+    ![Run HTTP POST requests](spring-boot-run-http-request.png)
+
+3. Execute the GET request and see the result in the **Run** tool window:
+
+    ![Run HTTP GET request](spring-boot-output-2.png)
+
+> You can also use any other HTTP client or cURL command-line tool. For example, you can run the following commands in the terminal to get the same result:
```
p7nov, line 170: Maybe make this a separate section? Seems like a complete topic - an alternative way to execute requests. Too broad for a note.

p7nov (changes requested):
Please consider the comment about the build script. The other ones are minor.

PR #2062, opened 2021-02-06
Update functions.md

elec60:
string literals in single quotation issue fixes


p7nov:
Thank you for spotting this!

PR #2063, opened 2021-02-07
update: Collection operations (sumOf, maxOf, minOf)

shadrina:
Not sure if we need to keep `sumBy` and `sumByDouble` when `sumOf` handles different types. 🤔


shadrina:
TBD: Remove `sumBy` and `sumByDouble` descriptions.

p7nov (approved):
LGTM

PR #2075, opened 2021-02-10
chore: split "Classes and inheritance" onto two separate pages

p7nov:


docs/topics/inheritance.md
```diff
@@ -0,0 +1,218 @@
...
+
+```kotlin
+class MyView : View {
+    constructor(ctx: Context) : super(ctx)
+
+    constructor(ctx: Context, attrs: AttributeSet) : super(ctx, attrs)
+}
+```
+
+## Override methods
```
koshachy, line 42: Actually, I see nothing wrong when using gerunds as a part of the concept title (when we are not waiting for an actual action from the user) — overriding methods, implementing interfaces, delegating properties, calling Java from Kotlin, etc. 
So maybe we should leave "overriding"?

koshachy (commented):
LGFM.
However, there is a point requiring discussion.

docs/topics/inheritance.md
p7nov: Yep, I was in doubt when changes these titles: does it actually make them better? Let's leave the gerunds if there is no obvious value in changing them to verbs.

koshachy (approved):

PR #2076, opened 2021-02-10
update: remove bintray from EAP config instruction

p7nov:
Update instructions on configuring build for using EAP Kotlin versions: remove Bintray and replace with mavenCentral


p7nov:
Closing due to complete doc migration to WebHelp.
The same changes for WebHelp are in https://github.com/JetBrains/kotlin-web-site/pull/2079

PR #2079, opened 2021-02-12
update: remove Bintray/JCenter from EAP instruction

p7nov:


katia-energizer (approved):
I've pushed fixes and corrections right to the branch.

PR #2083, opened 2021-02-14
fix typo

manuelcarriernunes:
The context suggests the type should have been omitted.

### DIFF
--- a/docs/topics/properties.md
+++ b/docs/topics/properties.md
@@ -41,7 +41,7 @@ The initializer, getter and setter are optional. Property type is optional if it
 (or from the getter return type, as shown below).
 
 ```kotlin
-var allByDefault: Int? // error: explicit initializer required, default getter and setter implied
+var allByDefault // error: explicit initializer required, default getter and setter implied
 var initialized = 1 // has type Int, default getter and setter
 ```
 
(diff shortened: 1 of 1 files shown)


docs/topics/properties.md
```diff
@@ -41,7 +41,7 @@ The initializer, getter and setter are optional. Property type is optional if it
 (or from the getter return type, as shown below).
 
 ```kotlin
-var allByDefault: Int? // error: explicit initializer required, default getter and setter implied
+var allByDefault // error: explicit initializer required, default getter and setter implied
 var initialized = 1 // has type Int, default getter and setter
```
p7nov, line 45: The current example is a bit hard to understand because it starts from broken code and continues with a good one. Let's swap the lines and comment out the incorrect one to separate it from the good code.
```suggestion
var initialized = 1 // has type Int, default getter and setter
// var allByDefault // ERROR: explicit initializer required, default getter and setter implied
```

p7nov (changes requested):
Hi!
Thanks for this PR.
I suggest improving this example even more, please see my comment.

docs/topics/properties.md
manuelcarriernunes: I think your suggestion makes the example clearer.

When I read the example the first time, I thought it was crafted so that I could paste it into my IDE to see an error (see attached screenshot).
image

p7nov (approved):
All good now!

PR #2084, opened 2021-02-14
update: Specify when the `..` range is empty

shadrina:


p7nov:
Generally, we don't describe "incorrect" usage of the language in website docs unless we find a common error. 
The [issue](https://youtrack.jetbrains.com/issue/KT-33783) about this case is open for more than a year and has no upvotes; the [stackoverflow question](https://stackoverflow.com/questions/57856102/can-i-rely-on-a-backwards-range-in-kotlin-not-executing-for-loop) doesn't have much activity or many views too.
I suggest leaving this page as is. Adding details in the function's KDoc should be enough.

shadrina:
Agree. I'll add the explanation to the ticket.

PR #2092, opened 2021-02-16
update: removed unused variables

katia-energizer:


p7nov (approved):
Ok, seems like other variables aren't used anywhere.

PR #2096, opened 2021-02-17
Updated: Compile Kotlin and Java sources

GlenKPeterson:
I think this is the modern way to run Kotlin and Java together in the same project.  IntelliJ IDEA warns if you try to do it the old way.  The new is also simpler; you don't even need a maven-compiler-plugin section.

I also added a note to use one of: jar-file-plugin or assembly-plugin or maven-shade-plugin

### DIFF
--- a/docs/topics/maven.md
+++ b/docs/topics/maven.md
@@ -83,9 +83,8 @@ The Kotlin Maven Plugin needs to be referenced to compile the sources:
 
 ## Compile Kotlin and Java sources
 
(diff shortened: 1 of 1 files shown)


GlenKPeterson:
See also related JetBrains YouTrack issue: https://youtrack.jetbrains.com/issue/IDEA-262874

GlenKPeterson:
I was wrong.  My patch was the wrong way to do it, so I'm closing this pull request.

PR #2110, opened 2021-02-21
Set code snippet language to Kotlin

alialbaali:


p7nov:
Thank you for spotting this. We've already merged the same fix in https://github.com/JetBrains/kotlin-web-site/pull/2105.

PR #2111, opened 2021-02-21
Improve documentation ToC organization

ianbrandt:
I had trouble intuitively locating the Operator Overloading page under the new ToC organization. While operator overloading is useful for DSLs, it seems more of a core language feature with general applicability. I propose categorizing it under Functions, as overloads are defined via `operator fun`.

With Type-safe Builders being the only page left under the DSL category, it could be flattened.

The Annotations page is kind of off on its own, whereas it might make sense under the Classes category since they're defined via `annotation class` (similar to `enum class`).

### DIFF
--- a/docs/kr.tree
+++ b/docs/kr.tree
@@ -69,6 +69,7 @@
             <toc-element id="generics.md"/>
             <toc-element id="nested-classes.md"/>
             <toc-element id="enum-classes.md"/>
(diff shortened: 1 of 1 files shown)


koshachy:
Hey Ian,
We've rearranged the structure in this PR – https://github.com/JetBrains/kotlin-web-site/pull/2427
Thanks for your suggestion!

PR #2116, opened 2021-02-24
update: create and publish lib tutorial

katia-energizer:
* Actualized the Create and publish a multiplatform library tutorial
* Added links to Create a mulitplatform library, Publish a multiplatform library


docs/topics/mpp/multiplatform-library.md
```diff
@@ -1,152 +1,161 @@
...
-but it's pretty straightforward and follows Base64 format description without any optimizations:
-
-```kotlin
-private val BASE64_ALPHABET: String = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
-private val BASE64_MASK: Byte = 0x3f
-private val BASE64_PAD: Char = '='
-private val BASE64_INVERSE_ALPHABET = IntArray(256) {
-    BASE64_ALPHABET.indexOf(it.toChar())
-}
+Unfortunately, there is no third-party implementation available for Kotlin/Native, so you need to write it yourself.
```
p7nov, line 123: I suppose there _are_ implementations for Kotlin/Native targets, but no _generic_ one for all of them. Perhaps we can formulate this more precisely.

docs/topics/mpp/multiplatform-library.md
```diff
@@ -240,33 +273,103 @@ class Base64JvmTest {
...
 
 ```groovy
 apply plugin: 'maven-publish'
 group 'org.jetbrains.base64'
 version '1.0.0'
 ```
 
-Now check it with the command `./gradlew publishToMavenLocal` and you should see a successful build. 
-That's it, our library is now successfully published and any Kotlin project can depend on it, whether it is another common library, JVM, JS, or Native application.
+```kotlin
```
p7nov, line 294: Why does the Kotlin sample have `multiplatform` plugin and the Groovy one doesn't?

KaterinaPetrova: I suggest adding multiplatform plugin to the Groovy example as well (though it's a section about publishing, you need both of the plugins to make it work) and update the plugins block [to the new style](https://stackoverflow.com/questions/32352816/what-the-difference-in-applying-gradle-plugin#:~:text=The%20plugins%20block%20is%20the,a%20plugin%20to%20your%20build.) so it would be consistent with Kotlin sample:

So the Groovy sample should look like this:
```

plugins {
    id 'org.jetbrains.kotlin.multiplatform' version '1.4.30'
    id 'maven-publish'
}

group = 'org.jetbrains.base64'
version = '1.0.0'

```

KaterinaPetrova (approved):

docs/topics/mpp/multiplatform-library.md
```diff
@@ -1,152 +1,161 @@
...
 
-```kotlin
-expect object Base64Factory {
-    fun createEncoder(): Base64Encoder
-}
-```
+You can complete this tutorial on any operating system.
+Download and install the [latest version of IntelliJ IDEA](http://www.jetbrains.com/idea/download/index.html) with the [latest Kotlin plugin](releases.md).
+
+## Create a project
```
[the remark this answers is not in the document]
katia-energizer: I don't want to send a user back and forth. It's not good experience, especially for tutorials. A tutorial should describe a process e2e. And only 3 steps are the same.
If we need to reuse content in more places, we can leverage reuse capabilities of WH.

docs/topics/mpp/multiplatform-library.md
```diff
@@ -1,152 +1,161 @@
...
+
+The wizard will create a sample multiplatform library with the following structure:
 
-Our factory is marked with the `expect` keyword. `expect` is a mechanism to define a requirement, which every platform should provide in order for the common part to work properly.
-So on each platform we should provide the `actual` `Base64Factory` which knows how to create the platform-specific encoder.
-You can read more about platform specific declarations [here](mpp-connect-to-apis.md).
+![Multiplatform library structure](mpp-lib-structure.png){width=250}
 
-## Platform-specific implementations
+## Make code cross-platform
```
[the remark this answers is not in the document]
katia-energizer: Agree. You're right. I've renamed titles for consistency, but confused two tutorials in my head.

docs/topics/mpp/multiplatform-library.md
```diff
@@ -1,152 +1,161 @@
...
+* [Native](#native)
 
 ### JVM
-We are starting with an implementation for the JVM. Let's create a file `Base64.kt` in `jvmMain/kotlin/jetbrains/base64` folder and provide a simple implementation, which delegates to `java.util.Base64`:
 
-```kotlin
-actual object Base64Factory {
-    actual fun createEncoder(): Base64Encoder = JvmBase64Encoder
-}
+1. In the `jvmMain/kotlin` directory, create the `org/jetbrains/base64` package.
```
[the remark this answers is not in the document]
katia-energizer: Fixed.

docs/topics/mpp/multiplatform-library.md
katia-energizer: Thanks. Rephrased.

docs/topics/mpp/multiplatform-library.md
katia-energizer: Thanks @KaterinaPetrova and @p7nov ! updated.

p7nov (approved):
Look good now 👍

docs/topics/mpp/multiplatform-library.md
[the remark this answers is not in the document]
mestir: @katia-energizer

PR #2119, opened 2021-02-25
fix: fix links with backticks

koshachy:


koshachy:
As far as I can see, the WH team fixed the issue. 
I close the PR.

PR #2129, opened 2021-03-05
Fix linkText(visibility) 404 not found

LeoAndo:
# Problem
The following link text is invalid.

https://kotlinlang.org/docs/sealed-classes.html#additional-location-the-same-package
スクリーンショット 2021-03-05 16 29 35

# Before
https://github.com/Kotlin/kotlinx.coroutines/blob/master/visibility-modifiers.html

# After
https://kotlinlang.org/docs/visibility-modifiers.html


p7nov:
Thank you for this fix!

PR #2133, opened 2021-03-08
Update classes.md

emmagency:


docs/topics/classes.md
```diff
@@ -71,7 +71,7 @@ class Customer(name: String) {
 In fact, for declaring properties and initializing them from the primary constructor, Kotlin has a concise syntax:
 
 ```kotlin
-class Person(val firstName: String, val lastName: String, var age: Int) { /*...*/ }
+class Person(val firstName: String = "Foo", val lastName: String = "Bar", var age: Int = 27) { /*...*/ }
```
p7nov, line 74: This is a good addition. Thanks.
However, I'd suggest adding another sentence and this example after the current one, not instead of it. 
This will separate two steps: _declaring properties as constructor parameters_ and _setting their default values_.
What do you think?

One more thing: please use a real name, not `Foo/Bar`.

docs/topics/classes.md
```diff
@@ -101,7 +101,7 @@ Learn more about [visibility modifiers](visibility-modifiers.md#constructors).
 The class can also declare _secondary constructors_, which are prefixed with `constructor`:
 
 ```kotlin
-class Person {
+class Person() {
```
p7nov, line 104: Thanks, this example was a bit incorrect and we've already reworked it. Check the new one [here](https://kotlinlang.org/docs/classes.html#secondary-constructors).

Please note that the changes you suggest also have an issue: 
`constructor(parent: Person) {` doesn't work without a primary constructor call (see the paragraph after this example).

p7nov (changes requested):
Hey! 
Thanks for this PR.
Please see my suggestions.

Regards, Pavel
Kotlin Technical Writer

PR #2137, opened 2021-03-10
update: kotlin-test as a single dependency

PetrakovichVictoria:


docs/topics/gradle.md
```diff
@@ -258,33 +258,27 @@ kotlin.stdlib.default.dependency=false
...
-Add the corresponding dependencies on test libraries:
-
-* For `commonTest`, add the `kotlin-test-common` and `kotlin-test-annotations-common` dependencies.
-* For JVM targets, use `kotlin-test-junit` or `kotlin-test-testng` for the corresponding asserter implementation and annotations mapping.
-* For Kotlin/JS targets, add `kotlin-test-js` as a test dependency. 
+Add the dependency `kotlin-test` to the `commonTest` source set, and the Gradle plugin will infer the corresponding 
+test dependencies for each source set:
+* `kotlin-test-common` and `kotlin-test-annotations-common` for common source sets
+* `kotlin-test-junit`/`kotlin-test-junit-5`/`kotlin-test-testng` for JVM source sets
+* `kotlin-test-js` for Kotlin/JS source sets
```
ilya-g, line 265: I don't think that details of dependency names for each platform are useful to readers given that now all should work with the single dependency.

docs/topics/gradle.md
```diff
@@ -316,10 +299,43 @@ kotlin{
...
+    useTestNG()
+    // or
+    // enable JUnit Platform (a.k.a. JUnit 5) support
+    useJUnitPlatform()
+}
+```
+
+ </tabs>
+
+You can disable the automatic selection of a testing framework by adding the line `kotlin.test.infer.jvm.variant=false`
```
ilya-g, line 336: It's not a testing framework that is selected automatically, but a variant of `kotlin-test` library.

docs/topics/gradle.md
PetrakovichVictoria: But may it worth to be mentioned for the common understanding? Let's ask @p7nov 's opinion

docs/topics/gradle.md
```diff
@@ -316,10 +299,43 @@ kotlin{
...
+        // enable TestNG support
+        useTestNG()
+        // or
+        // enable JUnit Platform (a.k.a. JUnit 5) support
+        useJUnitPlatform()
+    }
+}
+```
+
+```kotlin
```
p7nov, line 324: If Groovy and Kotlin snippets are the same, you can leave only one. We usually leave Groovy/

docs/topics/gradle.md
```diff
@@ -258,33 +258,27 @@ kotlin.stdlib.default.dependency=false
...
 
-Add the corresponding dependencies on test libraries:
-
-* For `commonTest`, add the `kotlin-test-common` and `kotlin-test-annotations-common` dependencies.
-* For JVM targets, use `kotlin-test-junit` or `kotlin-test-testng` for the corresponding asserter implementation and annotations mapping.
-* For Kotlin/JS targets, add `kotlin-test-js` as a test dependency. 
+Add the dependency `kotlin-test` to the `commonTest` source set, and the Gradle plugin will infer the corresponding 
+test dependencies for each source set:
+* `kotlin-test-common` and `kotlin-test-annotations-common` for common source sets
+* `kotlin-test-junit`/`kotlin-test-junit-5`/`kotlin-test-testng` for JVM source sets
```
p7nov, line 264: Please avoid using `/` as a conjunction. It's often unclear what you're replacing with it: and, or, something else?
I'd mention only `kotlin-test-junit` in the first sentence because it is actually inferred with `kotlin-test`. Perhaps we can add a second sentence about JUnit 5  and TestNG.

docs/topics/gradle.md
p7nov: This may be useful for readers who add these dependencies manually now. They'll learn what exactly is replaced with `kotlin("test")`. But I agree that it's not necessary.

docs/topics/gradle.md
```diff
@@ -258,33 +258,27 @@ kotlin.stdlib.default.dependency=false
...
+* `kotlin-test-js` for Kotlin/JS source sets
 
 Kotlin/Native targets do not require additional test dependencies, and the `kotlin.test` API implementations are built-in.
 
+Additionaly, you can use the `kotlin-test` dependency in any shared or platform-specific source set.
+
+For Kotlin/JVM Gradle uses JUnit 4 by default; therefore, the `kotlin('test')` dependency resolves to
+the variant for JUnit 4, namely `kotlin-test-junit`:
```
p7nov, line 272: `the variant for` – is this part necessary? the sentence seems OK if we just drop these words.
Additionally, `variant` is a term that was never introduced in our docs.

docs/topics/gradle.md
```diff
@@ -258,33 +258,27 @@ kotlin.stdlib.default.dependency=false
...
+test dependencies for each source set:
+* `kotlin-test-common` and `kotlin-test-annotations-common` for common source sets
+* `kotlin-test-junit`/`kotlin-test-junit-5`/`kotlin-test-testng` for JVM source sets
+* `kotlin-test-js` for Kotlin/JS source sets
 
 Kotlin/Native targets do not require additional test dependencies, and the `kotlin.test` API implementations are built-in.
 
+Additionaly, you can use the `kotlin-test` dependency in any shared or platform-specific source set.
+
+For Kotlin/JVM Gradle uses JUnit 4 by default; therefore, the `kotlin('test')` dependency resolves to
```
p7nov, line 271: I'd move this sentence right before the one about JUnit5 and TestNG. Otherwise, the info about JVM frameworks is broken into two parts

docs/topics/gradle.md
```diff
@@ -316,10 +299,43 @@ kotlin{
...
+    useTestNG()
+    // or
+    // enable JUnit Platform (a.k.a. JUnit 5) support
+    useJUnitPlatform()
+}
+```
+
+ </tabs>
+
+You can disable the automatic selection of a variant of `kotlin-test` library by adding the line `kotlin.test.infer.jvm.variant=false`
```
p7nov, line 336: And what to do is you disable it?

docs/topics/gradle.md
PetrakovichVictoria: Should I also do sth with ```/``` in [what's new](https://github.com/JetBrains/kotlin-web-site/pull/2192/files#r607146909)?

docs/topics/gradle.md
PetrakovichVictoria: Not necessary but may be useful – then I'll leave it here if nobody is against.

docs/topics/gradle.md
PetrakovichVictoria: Do you mean it is split by the note about the shortcut?
>You can use shorthand for a dependency on a Kotlin module, for example, kotlin('test') 

Or do you also mean the code below which demonstrates this sentence?

I'll deploy to a staging and put a link to it here; let's look at this page on the staging and decide whether these sentences are not alongside.

docs/topics/gradle.md
PetrakovichVictoria: Should I fix it in [what's new](https://github.com/JetBrains/kotlin-web-site/pull/2192/files#r607194270) too?:)

docs/topics/gradle.md
PetrakovichVictoria: Supplemented the sentence, but maybe it became too long now.

docs/topics/gradle.md
```diff
@@ -256,13 +256,12 @@ kotlin.stdlib.default.dependency=false
...
 ### Set dependencies on test libraries
 
-The [`kotlin.test` API](https://kotlinlang.org/api/latest/kotlin.test/) is available for testing different Kotlin projects. 
-
-Add the corresponding dependencies on test libraries:
-
-* For `commonTest`, add the `kotlin-test-common` and `kotlin-test-annotations-common` dependencies.
-* For JVM targets, use `kotlin-test-junit` or `kotlin-test-testng` for the corresponding asserter implementation and annotations mapping.
-* For Kotlin/JS targets, add `kotlin-test-js` as a test dependency. 
+The [`kotlin.test` API](https://kotlinlang.org/api/latest/kotlin.test/) is available for testing Kotlin projects on different Kotlin platforms.
```
p7nov, line 259: "different Kotlin platforms" doesn't tell exactly what platforms you mean. Please write precisely. "All supported platforms" I suppose.

docs/topics/gradle.md
```diff
@@ -256,13 +256,12 @@ kotlin.stdlib.default.dependency=false
...
+Add the dependency `kotlin-test` to the `commonTest` source set, and the Gradle plugin will infer the corresponding 
+test dependencies for each source set:
```
p7nov, line 261: Do we mean test source sets here? If we do, then the current wording is not completely true.
If we really mean _each_ source set in the project, then the behavior is strange =)

docs/topics/gradle.md
```diff
@@ -273,18 +272,7 @@ kotlin{
...
-        }
-        jvmTest {
-            dependencies {
-                implementation kotlin('test-junit')
-            }
-        }
-        jsTest {
-            dependencies {
-                implementation kotlin('test-js')
+                implementation kotlin('test') // This brings all the platforms' dependencies automatically
```
p7nov, line 275:
```suggestion
                implementation kotlin('test') // This brings all the platform dependencies automatically
```

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
 </tabs>
 
-> You can use shorthand for a dependency on a Kotlin module, for example, kotlin("test") for "org.jetbrains.kotlin:kotlin-test".
+> You can use shorthand for a dependency on a Kotlin module, for example, kotlin('test') for "org.jetbrains.kotlin:kotlin-test".
 >
 {type="note"}
 
+Additionally, you can use the `kotlin-test` dependency in any shared or platform-specific source set.
+
+For Kotlin/JVM Gradle uses JUnit 4 by default; therefore, the `kotlin('test')` dependency resolves to JUnit 4.
```
p7nov, line 302: _kotlin("test")_ (double quotes) is better because it works in both Groovy and Kotlin scripts.

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
+
+You can choose JUnit 5 or TestNG by calling 
+[`useJUnitPlatform()`]( https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/Test.html#useJUnitPlatform) 
+or [`useTestNG()`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/Test.html#useTestNG).
```
p7nov, line 306: by calling them where?

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
 
 </tabs>
 
-> You can use shorthand for a dependency on a Kotlin module, for example, kotlin("test") for "org.jetbrains.kotlin:kotlin-test".
+> You can use shorthand for a dependency on a Kotlin module, for example, kotlin('test') for "org.jetbrains.kotlin:kotlin-test".
```
p7nov, line 296: This shorthand has limitations if I remember correctly. If it does, let's write exactly when it's available. Please check

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
 {type="note"}
 
+Additionally, you can use the `kotlin-test` dependency in any shared or platform-specific source set.
+
+For Kotlin/JVM Gradle uses JUnit 4 by default; therefore, the `kotlin('test')` dependency resolves to JUnit 4.
+
+You can choose JUnit 5 or TestNG by calling 
+[`useJUnitPlatform()`]( https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/Test.html#useJUnitPlatform) 
+or [`useTestNG()`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/testing/Test.html#useTestNG).
+The following example is for a JVM project:
```
p7nov, line 307: Perhaps we should add an MPP example since we started this section from MPP.
```
kotlin {
    jvm {
        testRuns["test"].executionTask.configure {
            useJUnitPlatform()
        }
    }
}
```

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
+```
+
+```kotlin
+dependencies {
+    testImplementation(kotlin("test"))
+}
+
+tasks {
+    test {
+        useJUnitPlatform()
```
p7nov, line 328: Different frameworks in Groovy and Kotlin tabs. Please choose one.

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
+    }
+}
+```
+
+</tabs>
+
+See [the tutorial on how to test code using JUnit in JVM](#jvm-test-using-junit.md).
```
p7nov, line 335:
```suggestion
[Learn how to test code using JUnit on the JVM](jvm-test-using-junit.md).
```
"Learn how to ..." is our common wording for doc page links.

docs/topics/gradle.md
```diff
@@ -316,10 +293,51 @@ kotlin{
...
+
+If the standard variants described above don't cover you needs, you can disable the automatic selection
```
p7nov, line 337: In general, I'd start this sentence from a more exact case description such as: "If you need to use a different JVM test framework, you can ... and add this framework as a dependency ..."

docs/topics/gradle.md
p7nov: This seems to require an effort and out of this feature's scope. Let's do this separately later: https://youtrack.jetbrains.com/issue/KT-45974

docs/topics/gradle.md
ilya-g: ok

docs/topics/gradle.md
ilya-g: > the `kotlin('test')` dependency resolves to JUnit 4.

This new wording looks confusing to me. `kotlin("test")` doesn't resolve to JUnit 4, it resolves to the artifact of `kotlin-test` for JUnit 4. We called it "variant" before as it is called in the standard Gradle terminology of variant-aware dependency resolution.

docs/topics/gradle.md
PetrakovichVictoria: Supported by whom? 
"all platforms supported by Kotlin" I would say.

docs/topics/gradle.md
PetrakovichVictoria: Thank you! I'm going to return back "the `kotlin("test")` dependency resolves to the variant for 
JUnit 4, namely `kotlin-test-junit`"

docs/topics/gradle.md
PetrakovichVictoria: Just to recheck:
@ilya-g, does Gradle infer the corresponding test dependencies for each **test** source set?

ilya-g: It depends on where `kotlin("test")` dependency is added. If in `commonTest`, then all test source sets actualizing `commonTest` also get the respective variant of `kotlin-test` dependency.

docs/topics/gradle.md
PetrakovichVictoria: Thank you. Just for the history: as we discussed in DM, I would have noticed this mistake by my own, but the review has been made on the draft PR without a call to do it:)

docs/topics/gradle.md
PetrakovichVictoria: Fixed.

docs/topics/gradle.md
PetrakovichVictoria: As I remember, on the call we've discussed that there are many server-side users to whom we target the code too. You told it will be cool if there is a ready piece of code that includes as adding the dependency, as calling usage of some platform. Did I understand you incorrectly or did you change your mind? Or maybe you offer to add the additional example next to the existing one?

docs/topics/gradle.md
PetrakovichVictoria: I fixed but I'm afraid that the new sentence is too long and complex.

docs/topics/gradle.md
```diff
@@ -256,13 +256,13 @@ kotlin.stdlib.default.dependency=false
...
 ### Set dependencies on test libraries
 
-The [`kotlin.test` API](https://kotlinlang.org/api/latest/kotlin.test/) is available for testing different Kotlin projects. 
-
-Add the corresponding dependencies on test libraries:
-
-* For `commonTest`, add the `kotlin-test-common` and `kotlin-test-annotations-common` dependencies.
-* For JVM targets, use `kotlin-test-junit` or `kotlin-test-testng` for the corresponding asserter implementation and annotations mapping.
-* For Kotlin/JS targets, add `kotlin-test-js` as a test dependency. 
+The [`kotlin.test` API](https://kotlinlang.org/api/latest/kotlin.test/) is available for testing Kotlin projects on
```
PetrakovichVictoria, line 259: See how this link looks: https://branch-kotlin-test-as-the-single-dependency.kotlin-web-site.labs.jb.gg/docs/gradle.html#set-dependencies-on-test-libraries

Should I edit the formatting somehow?

docs/topics/gradle.md
p7nov: "**Kotlin** projects on all supported platforms" implies that platforms are supported by Kotlin. No need for more words here.

docs/topics/gradle.md
p7nov: Yes, let's move `API` out of the link.

docs/topics/gradle.md
p7nov: Yes, you're right.
I suggest adding **one more example** because calling these tasks in MPP is not so trivial as in JVM. The copy-pastable piece of config can save a lot of time for our readers :)

docs/topics/gradle.md
```diff
@@ -320,6 +298,49 @@ kotlin{
...
+}
+```
+
+</tabs>
+
+[Learn how to test code using JUnit on the JVM](jvm-test-using-junit.md).
+
+If you need to use a different JVM test framework, you can disable the automatic selection of a testing framework by 
+adding the line `kotlin.test.infer.jvm.variant=false` to the project’s `gradle.properties` and add this framework as a 
+dependency to your build script.
```
p7nov, line 342: This sentence is indeed too long. Let's think about splitting it.
Just an idea:
```suggestion
If you need to use a different JVM test framework, disable the automatic selection of a testing framework by 
adding the line `kotlin.test.infer.jvm.variant=false` to the project’s `gradle.properties`. Once you do this, add this this framework as a Gradle dependency.
```

p7nov (commented):
A couple of minor comments.

docs/topics/gradle.md
PetrakovichVictoria: Indeed) Thank you!

docs/topics/gradle.md
PetrakovichVictoria: Added the example.

p7nov (approved):
LGTM now 👍

ilya-g (approved):

PR #2139, opened 2021-03-11
update: add Compose as desktop development option in FAQ

dayanruben:


p7nov:
Hey! 
Thank you for this suggestion. Our colleagues keep working on Compose for Desktop, but it's still in an early stage of development, so we don't think it's ready to be recommended as a ready solution.
We'll sure list Compose for Desktop on kotlinlang.org once it reaches Beta at least.

PR #2142, opened 2021-03-11
I added your new Desktop Framework - Compose

Yourun-proger:
I think it will be useful for other community members to know about your new framework)) For example, it inspired me to learn Kotlin and learn how to work with this framework!


p7nov:
Hey! 
Thank you for this suggestion. Our colleagues keep working on Compose for Desktop, but it's still in an early stage of development, so we don't think it's ready to be recommended as a ready solution.
We'll sure list Compose for Desktop on kotlinlang.org once it reaches Beta at least.

PR #2144, opened 2021-03-14
Add missing double quote

happy-nut:
I added missing double quote in docs/topics/jvm/jvm-test-using-junit.md


koshachy:
Thanks for your PR!

PR #2145, opened 2021-03-15
update: update web-root and og-image properties in buildprofiles.xml

koshachy:


koshachy:
All necessary changes in https://github.com/JetBrains/kotlin-web-site/pull/2448/

PR #2152, opened 2021-03-17
Add Extension properties ordering

zayo:
I've noticed that detekt is treating extension properties as regular properties.
Opened issue on detekt: https://github.com/detekt/detekt/issues/3532
Lead to proposal: https://github.com/detekt/detekt/pull/3559

Yet that change would be unclear as this is not part of the spec at all. Thus proposing the following change.

### DIFF
--- a/docs/topics/coding-conventions.md
+++ b/docs/topics/coding-conventions.md
@@ -72,6 +72,8 @@ Do not sort the method declarations alphabetically or by visibility, and do not 
 from extension methods. Instead, put related stuff together, so that someone reading the class from top to bottom can 
 follow the logic of what's happening. Choose an order (either higher-level stuff first, or vice versa) and stick to it.
 
+Extension properties are effectively methods, thus in terms of order they are treated as methods.
+
 Put nested classes next to the code that uses those classes. If the classes are intended to be used externally and aren't
 referenced inside the class, put them in the end, after the companion object.
 


BraisGabin (approved):

BraisGabin:
This is not stale. I think that it should be merged or at least denied by a human telling why it should not be merged.

p7nov:
Hey! We're currently testing some automation to help us handle PR from contributors in time. Sorry for not responding for so long. We'll review your suggestion in a couple of days.

elizarov:
This addition contradicts the spirit of the Kotlin style guide which explicitly does not tries to mandate any order and says: " "Choose an order (either higher-level stuff first, or vice versa) and stick to it." It is up to specific projects and tools to decide upon any specific ordering requirements if they wish to enforce them.

zayo:
Thats true, but guys from detekt told me, that this has unless it's in doc, they won't approve it.
The problem is, that extension properties are treated as regular properties, which I believe is wrong. Yet the doc does not states what the ext property is.
Any suggestion how to rephrase it to fit both needs?

elizarov:
Kotlin style guide does not enforce the ordering for a reason. We have found that it is hard to come up with rock-solid ordering rules even for our own code, so we've bailed out of trying to solve it. 

It is up to detekt team to decide upon such rules if they wish to tackle this issue to have ordering enforced automatically.

zayo:
I got your point. I don't require the doc to contain info about order.
I would like the doc to state what the extension property is. Is it property? Is it function? Is it neither one of them?

elizarov:
I don't see how that kind of information is related to Kotlin's style guide.

zayo:
You cannot say to not force order on properties unless you start calling them "properties". The doc says nothing about extensions. It doesn't even give a hint how they should be treated. I don't require to tell where they should be, but current state gives everyone chnace to mix them with methods(which they effectively are), or treating them like properties(which they look like). Which gives too much power and can be even harmful and error prone from my POV.
If that ends up with mixing the extension properties into regular properties, it's super confusing for the reader.

So I don't want to state how to order them, but how not to order them in a first place. To avoid code being error prone and confusing for the reader. Hope this is clear explanation why.

elizarov:
This kind of decision is simply out of the scope of this particular document. It does not belong here.

chao2zhang:
I believe this is within the scope of this document, particularly about clarifications of the existing doc `1. Property declarations and initializer blocks`. The proposal here is to clarify the `property` as non-extension properties because extension properties do get parsed as `KtProperty`.

On the other hand, I agree with you that it makes sense to not put any restrictions on extension properties in the official documentation.

PR #2156, opened 2021-03-19
fix: KT-33201 rewrite unusable sample of secondary constructor

p7nov:
Fix for https://youtrack.jetbrains.com/issue/KT-33201

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -101,10 +101,11 @@ Learn more about [visibility modifiers](visibility-modifiers.md#constructors).
 The class can also declare _secondary constructors_, which are prefixed with `constructor`:
 
 ```kotlin
-class Person {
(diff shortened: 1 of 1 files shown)


docs/topics/classes.md
```diff
@@ -101,10 +101,9 @@ Learn more about [visibility modifiers](visibility-modifiers.md#constructors).
 The class can also declare _secondary constructors_, which are prefixed with `constructor`:
 
 ```kotlin
-class Person {
-    var children: MutableList<Person> = mutableListOf()
-    constructor(parent: Person) {
-        parent.children.add(this)
+class Pet {
```
koshachy, line 104: Do we suppose that we have got already declared `Person` class with the collection of Pet objects?

If we don't need any explanation of this (or it was somewhere above) it looks good.

koshachy (approved):
LGTM

PR #2159, opened 2021-03-19
update: clarified the place where to put a dependency

PetrakovichVictoria:
https://youtrack.jetbrains.com/issue/KT-44911

### DIFF
--- a/docs/topics/mpp/mpp-add-dependencies.md
+++ b/docs/topics/mpp/mpp-add-dependencies.md
@@ -1,7 +1,7 @@
 [//]: # (title: Add dependencies)
 
 To add a dependency on a library, set the dependency of the required [type](gradle.md#dependency-types) (for example, `implementation`) in the 
-`dependencies` block of the source sets DSL.
+`dependencies` block in your [Gradle](gradle.md) build script.
 
 <tabs>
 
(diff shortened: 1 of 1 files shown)


docs/topics/mpp/mpp-add-dependencies.md
```diff
@@ -1,7 +1,7 @@
 [//]: # (title: Add dependencies)
 
 To add a dependency on a library, set the dependency of the required [type](gradle.md#dependency-types) (for example, `implementation`) in the 
-`dependencies` block of the source sets DSL.
+`dependencies` block in your [Gradle](gradle.md) build script.
```
koshachy, line 4: TBH, I have a concern about `in` doubling in this sentence 🤔

PetrakovichVictoria: Thank you, I will ask for a proofread.

p7nov: Wouldn't this link be better here? https://kotlinlang.org/docs/mpp-dsl-reference.html#dependencies
It's MPP- and dependencies-specific, as opposed to just gradle.md.

PetrakovichVictoria: Thank you, Pasha! I've got an idea – what if to leave the link to Gradle, but also add the link you proposed? See the result in the diff.

Regarding the proofread: there is a couple of minor changes, I put them in the code too, but proofreaders didn't have any concerns regarding doubling `in`.

koshachy (approved):
👍

p7nov (approved):
All good, thanks.

PR #2165, opened 2021-03-22
Fix sample in when 'else if' documentation

caalador:
The sample uses `x` for both odd and even checks
where as the even check should be for `y`


p7nov:
Thanks for spotting!

PR #2166, opened 2021-03-22
update: update inline classes to Kotlin 1.5.0

koshachy:
Updated the Inline classes description to Stable.


ilmirus (approved):

docs/topics/inline-classes.md
```diff
@@ -1,23 +1,13 @@
...
 To solve such issues, Kotlin introduces a special kind of class called an _inline class_. 
-Inline classes are a subset of value-based classes. They don't have an identity and can only hold values.
-
-To declare an inline class, use an `inline` or `value` modifier before the name of the class:
+Inline classes are a subset of [value-based classes](https://github.com/Kotlin/KEEP/blob/master/notes/value-classes.md). They don't have an identity and can only hold values.
 
-```kotlin
-inline class Password(val value: String)
-```
+To declare an inline class, use a `value` modifier before the name of the class:
```
p7nov, line 10:
```suggestion
To declare an inline class, use the `value` modifier before the name of the class:
```

docs/topics/inline-classes.md
```diff
@@ -31,6 +21,10 @@ To declare an inline class for the JVM backend, use the `value` modifier along w
 value class Password(private val s: String)
 ```
 
+> The `inline` modifier is deprecated with the `WARNING` level.
```
p7nov, line 24: This can sound scary for those who write inline functions :) Let's be more precise and add "on classes".

I'm also not sure we need the detail about the level. Users will see the warning anyway, nothing will break, but we will need to update this sentence for the next steps of the deprecation cycle.

p7nov (approved):
A couple of minor comments.

docs/topics/inline-classes.md
koshachy: Yes, that may sound intimidating :)

Maybe we should remove information about this altogether? I have checked other features, and I haven't found such a description of the levels.

koshachy: I've updated the description :)

PR #2168, opened 2021-03-23
Update typecasts.md

lukandor:
Add more verbosity to type safety example

### DIFF
--- a/docs/topics/typecasts.md
+++ b/docs/topics/typecasts.md
@@ -131,10 +131,6 @@ generic type itself, *its* type arguments are still erased.
 
 ```kotlin
 //sampleStart
-class Greeter{
-    fun sayHello() = "Hello!"
-}
-
 inline fun <reified A, reified B> Pair<*, *>.asPairOf(): Pair<A, B>? {
     if (first !is A || second !is B) return null
     return first as A to second as B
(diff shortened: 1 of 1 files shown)


docs/topics/typecasts.md
```diff
@@ -141,14 +145,18 @@ val somePair: Pair<Any?, Any?> = "items" to listOf(1, 2, 3)
 val stringToSomething = somePair.asPairOf<String, Any>()
 val stringToInt = somePair.asPairOf<String, Int>()
 val stringToList = somePair.asPairOf<String, List<*>>()
-val stringToStringList = somePair.asPairOf<String, List<String>>() // Breaks type safety!
+val stringToStringList = somePair.asPairOf<String, List<Greeter>>() // Breaks type safety!
+
+stringToStringList?.second?.forEach(Greeter::sayHello) // This will throw ClassCastException as Int is not Greeter
```
p7nov, line 150: This doesn't work because executed outside main. It also breaks the sample.

p7nov (commented):
Hey!
Thanks, this is a valid point. However, it seems there are easier ways that don't add so much complexity to the example. 
The custom `Greeter` class is basically the same as String in this example.

Thus, you can add just a single line with a similar call of a String function/property. 
See my suggestions.

Regards, Pavel
Kotlin technical writer

PR #2175, opened 2021-03-28
Update object-declarations.md

j2esu:
Add visibility note on companion object's private fields


p7nov:
Thanks for this addition!

PR #2183, opened 2021-03-31
Improve nullability annotation handling section

shadrina:


docs/topics/jvm/java-interop.md
```diff
@@ -124,33 +124,39 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!> { ... }
+fun <T> collectNames(elements: (Mutable)Collection<T!>) : (Mutable)Collection<String!> { ... }
 ```
 
-Annotating type arguments works with Java 8 target or higher and requires the nullability annotations to support the
+Kotlin also supports: 
+- Nullability annotations on bounds of type parameters: `Collection<T extends @Nullable Number>` becomes 
+`(Mutable)Collection<T : Number?>!`
+- Nullability annotations on type arguments of base classes and interfaces
```
shadrina, line 152: TODO: Add example

docs/topics/jvm/java-interop.md
```diff
@@ -124,33 +124,39 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
 @NotNull
-Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ... }
+<T> Collection<@NotNull String> collectNames(@NotNull Collection<@NotNull T> elements) { ... }
 ```
 
 It leads to the following signature seen in Kotlin:
 
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String>) : (Mutable)Set<String> { ... }
+fun <T> collectNames(elements: (Mutable)Collection<T>) : (Mutable)Collection<String> { ... }
```
PetukhovVictor, line 140: Return type isn't exact `(Mutable)Collection<String>`. We don't have not null annotation on Java method's return type itself so in Kotlin it should be `(Mutable)Collection<String>!` (flexible).

PetukhovVictor: Moreover, ideally it should be:
```
fun <T> collectNames(elements: (Mutable)Collection<T!!>) : (Mutable)Collection<String>! { ... }
```
`T!!` means defenetely not null T (because we have the corresponding annotation on the Java side), but for now it isn't so. But with the special compiler flag `-XXLanguge:+ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated` the behaviour turns into it.

docs/topics/jvm/java-interop.md
```diff
@@ -124,33 +124,39 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
 ```
 
-Note the `@NotNull` annotations on `String` type arguments. Without them, you get platform types in the type arguments:
+Note the `@NotNull` annotations on `T` type parameter and `String` type argument. Without them, you get platform types:
 
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!> { ... }
+fun <T> collectNames(elements: (Mutable)Collection<T!>) : (Mutable)Collection<String!> { ... }
```
PetukhovVictor, line 146: Also missed `!` at the end on types: `(Mutable)Collection<T!>` and `(Mutable)Collection<String!>`.

docs/topics/jvm/java-interop.md
```diff
@@ -124,33 +124,39 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
 ```
 
-Annotating type arguments works with Java 8 target or higher and requires the nullability annotations to support the
+Kotlin also supports: 
+- Nullability annotations on bounds of type parameters: `Collection<T extends @Nullable Number>` becomes 
+`(Mutable)Collection<T : Number?>!`
+- Nullability annotations on type arguments of base classes and interfaces
+
+Annotating type arguments and parameters works with Java 8 target or higher and requires the nullability annotations to support the
```
PetukhovVictor, line 154: Suggest to replace "parameters' to "type parameters" not to clash with "value parameters" (usually we by default perceive "parameters" as "value parameters").

docs/topics/jvm/java-interop.md
shadrina: > Note the `@NotNull` annotations on `T` type parameter and `String` type argument. Without **them**, you get platform types:

We do not mention outer annotations here. To be honest, I also misunderstood this part at first. I'll rephrase it.

docs/topics/jvm/java-interop.md
shadrina: I wrote the example and checked the return value type in IDEA with and without specifying `@NotNull`:
- with `@NotNull` annotation on return type, I get `(Mutable)Collection<String>`
- with `@NotNull` annotation on return type, I get platform type `(Mutable)Collection<String>!`

🤔

docs/topics/jvm/java-interop.md
shadrina: Here we specify the attention on type arguments (in the new version).

docs/topics/jvm/java-interop.md
shadrina: Thank you, fixed.

docs/topics/jvm/java-interop.md
```diff
@@ -124,33 +124,39 @@ Kotlin types. The compiler supports several flavors of nullability annotations,
...
+Note the `@NotNull` annotations on `T` type parameter and `String` type argument. Without them, you get platform types:
 
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!> { ... }
+fun <T> collectNames(elements: (Mutable)Collection<T!>) : (Mutable)Collection<String!> { ... }
 ```
 
-Annotating type arguments works with Java 8 target or higher and requires the nullability annotations to support the
+Kotlin also supports: 
+- Nullability annotations on bounds of type parameters: `Collection<T extends @Nullable Number>` becomes
```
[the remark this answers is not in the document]
shadrina: Mentioned in the new version.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+
+By default, plain type parameters both in Kotlin and Java have undefined nullability. From Java, you can specify 
+it with nullability annotations. Let's annotate the type parameter of the `Base` class:
+
+```java
+public class Base<@NotNull T>
+```
+
+In Kotlin, inheriting from `Base` expects a non-nullable type argument or type parameter. Thus, the following Kotlin code
+produces warning:
```
PetukhovVictor, line 187: I'd say "warning by default for now" and that it's going to become an error in the future (in 1.6.0, but I think it's not worth specifying a specific version), because it may seem to a user that the compiler will report warnings permanently.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!> { ... }
+fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!>
+```
+
+Kotlin also takes into account nullability annotations on type arguments of base classes and interfaces. For example,
+there are 2 Java classes with the signatures provided below.
+
+```java
+public class Base<T>
```
PetukhovVictor, line 160: Minor: I don't know if there is a requirement for inserting Java code pieces to be valid, but if so, then here and below it is missing `{}`.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+In Kotlin, inheriting from `Base` expects a non-nullable type argument or type parameter. Thus, the following Kotlin code
+produces warning:
+
+```kotlin
+class Derived<K> : Base<K> // warning: K has undefined nullability
 ```
 
-Annotating type arguments works with Java 8 target or higher and requires the nullability annotations to support the
-`TYPE_USE` target (`org.jetbrains.annotations` supports this in version 15 and above).
+You can fix it by specifying the lower bound `K : Any`.
```
PetukhovVictor, line 193: > lower bound
It's an upper bound actually

docs/topics/jvm/java-interop.md
shadrina: @p7nov WDYT? Do we have such a requirement? Can I add a snippet with the signatures only?

p7nov: It's always better to have valid copy-pastable snippets. Exception: if this requires a lot of irrelevant code :)
In this case it's not a problem at all, let's add the braces.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
 It leads to the following signature seen in Kotlin:
 
 ```kotlin
-fun toSet(elements: (Mutable)Collection<String>) : (Mutable)Set<String> { ... }
+fun toSet(elements: (Mutable)Collection<String>) : (Mutable)Set<String>
 ```
 
 Note the `@NotNull` annotations on `String` type arguments. Without them, you get platform types in the type arguments:
```
p7nov, line 150: No need in such an intro, we've already introduced the annotations just above.

```suggestion
Without  the `@NotNull` annotations on type arguments, you get platform types in the type arguments:
```

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+public class Derived extends Base<@Nullable String>
+```
+
+In the Kotlin code, passing the instance of `Derived` where the `Base<String>` assumed produces the warning.
+
+```kotlin
+fun takeBaseOfNotNullStrings(x: Base<String>) {}
+
+fun main() {
+    takeBaseOfNotNullStrings(Derived()) // warning
```
p7nov, line 173: Please add details about the warning. If its message is short, you can just copy it. Otherwise, conduct its meaning in 3-5 words.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+
+```java
+public class Base<T>
+```
+
+```java
+public class Derived extends Base<@Nullable String>
+```
+
+In the Kotlin code, passing the instance of `Derived` where the `Base<String>` assumed produces the warning.
```
p7nov, line 167: Why the warning is produced? What to do if I see it?

p7nov: It would be great to have an explanation similar to the one in the next section. I like that one :)

docs/topics/jvm/java-interop.md
p7nov: Our docs reflect an actual state for the latest stable version, and we avoid announcing our plans for the future.
There are general reasons:
- This info complicates the text while being irrelevant for some readers.
- Plans can change.
- "for now" raises more questions than it gives answers =)

Aside from that, the planned change (warning > error) will require a doc update anyway, so no doubt that we'll return to this content when it's time.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+
+Kotlin translates it just as follows:
+
+```kotlin
+class BaseWithBound<T : Number>
+```
+
+So passing nullable type as a type argument or type parameter produces a warning.
+
+Annotating type arguments and type parameters works with Java 8 target or higher and requires the nullability annotations
```
p7nov, line 209: There are two conditions, let's split into two sentences.

docs/topics/jvm/java-interop.md
```diff
@@ -137,20 +144,76 @@ Set<@NotNull String> toSet(@NotNull Collection<@NotNull String> elements) { ...
...
+
+```kotlin
+class BaseWithBound<T : Number>
+```
+
+So passing nullable type as a type argument or type parameter produces a warning.
+
+Annotating type arguments and type parameters works with Java 8 target or higher and requires the nullability annotations 
+to support the `TYPE_USE` target (`org.jetbrains.annotations` supports this in version 15 and above). 
+For the cases where Kotlin code uses wrong nullability according to the nullability annotations from Java, you can turn on
```
p7nov, line 211: Too complicated, can we rephrase?

docs/topics/jvm/java-interop.md
shadrina: Postponed to 1.5.20

docs/topics/jvm/java-interop.md
shadrina: Postponed to 1.5.20

docs/topics/jvm/java-interop.md
shadrina: Split, thanks!

docs/topics/jvm/java-interop.md
shadrina: Rephrased. Please, take a look.

p7nov (approved):
All good to me now 👍

p7nov:
Please don't forget to proofread the new part.

PR #2185, opened 2021-03-31
update: stable sealed interfaces and inheritance

p7nov:
Doc update reflecting the sealed class changes in 1.5.0:
- Sealed interfaces are stable
- package-wide sealed hierarchies by default
- `when` on `expect sealed` classes is not exhaustive

### DIFF
--- a/docs/topics/sealed-classes.md
+++ b/docs/topics/sealed-classes.md
@@ -8,7 +8,7 @@ Thus, each instance of a sealed class has a type from a limited set that is know
 The same works for sealed interfaces and their implementations: once a module with a sealed interface is compiled,
 no new implementations can appear.
 
-In some sense, sealed classes are similar to [`enum` classes](enum-classes.md): the set of values
+In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes: the set of values
 for an enum type is also restricted, but each enum constant exists only as a _single instance_, whereas a subclass
 of a sealed class can have _multiple_ instances, each with its own state.
 
@@ -31,7 +31,7 @@ Sealed classes are not allowed to have non-`private` constructors (their constru
 ## Location of direct subclasses
 
 Direct subclasses of sealed classes and interfaces must be declared in the same package. They may be top-level or nested
-inside any number of other named classes, named interfaces, or named objects. Subclasses can have any [visibility](visibility-modifiers.html)
+inside any number of other named classes, named interfaces, or named objects. Subclasses can have any [visibility](visibility-modifiers.md)
 as long as they are compatible with normal inheritance rules in Kotlin.
 
 Subclasses of sealed classes must have a proper qualified name. They can't be local nor anonymous objects.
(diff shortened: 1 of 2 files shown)


docs/topics/sealed-classes.md
```diff
@@ -43,20 +24,14 @@ data class Sum(val e1: Expr, val e2: Expr) : MathExpr()
...
 ```
 
-## Location of direct subclasses
-
-All direct subclasses of a sealed class must be declared in the same file as this class itself. Classes that extend
-direct subclasses of a sealed class (indirect inheritors) can be placed anywhere, not necessarily in the same file.
+A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
 
-### Additional location: the same package
+Sealed classes are not allowed to have non-`private` constructors (their constructors are `private` by default).
```
demiurg906, line 29: Actually, constructors of `sealed` classes are `protected` by default,  and it's allowed to use `private` or `protected` visibility on them. `public` and `internal` are prohibited

demiurg906:
```kotlin
sealed class Base() {
    private constructor(x: Int) : this()
    protected constructor(x: Double) : this()
    <!NON_PRIVATE_OR_PROTECTED_CONSTRUCTOR_IN_SEALED!>internal<!> constructor(x: String) : this()
    <!NON_PRIVATE_OR_PROTECTED_CONSTRUCTOR_IN_SEALED!>public<!> constructor(x: Char) : this()

    class A0 : Base()
    class A1 : Base(1)
    class A2 : Base(1.0)
    class A3 : Base("")
    class A4 : Base('a')
}
```

demiurg906 (changes requested):

docs/topics/sealed-classes.md
p7nov: Thanks. Corrected the text and added an example.

demiurg906 (approved):

docs/topics/sealed-classes.md
```diff
@@ -1,37 +1,18 @@
...
-
-> Sealed interfaces are [Experimental](components-stability.md). They may be dropped or changed at any time.
-> Opt-in is required (see the details [below](#try-sealed-interfaces-and-package-wide-hierarchies-of-sealed-classes)), and you should use them only for evaluation purposes.  We would appreciate your feedback on them in [YouTrack](https://youtrack.jetbrains.com/issue/KT-42433).
->
-{type="warning"}
-
-Interfaces can be declared `sealed` as well as classes. The `sealed` modifier works on interfaces the same way:
-all implementations of a sealed interface are known at compile time. Once a module with a sealed interface is compiled,
-no new implementations can appear.
+To declare a sealed class or interface, put the `sealed` modifier before its name.
```
koshachy, line 15: It seems that we need to agree on whether we should put a colon in such cases or not :)

koshachy (approved):

docs/topics/sealed-classes.md
p7nov: Good idea, let's state this in our guidelines. 
I've checked docs of some popular languages (Java, Python, C#). They all use a colon between such explanations and examples.

PR #2194, opened 2021-04-05
update: KT-44361 deprecate Gradle options noStdlib, noReflect

p7nov:
[KT-44361 Gradle: deprecate options includeRuntime, noStdlib, noReflect](https://youtrack.jetbrains.com/issue/KT-44361)


udalov (changes requested):
They're only deprecated in Gradle, as per the issue you've linked. In CLI & Ant, they're still very useful and we have no plans to deprecate them there, so the corresponding docs shouldn't be touched.

udalov (approved):

PR #2200, opened 2021-04-07
Update basic-types.md

j2esu:
It seems like current example of a raw string with '$' symbol works nice even without ${...}, cause compiler can detect that there is no identifier after it. To make it fail during compilation we need to add, for example '_'


p7nov:
Thanks for this addition!

PR #2205, opened 2021-04-08
update: Add firstNotNullOf() description

shadrina:


docs/topics/collection-elements.md
```diff
@@ -123,6 +123,25 @@ fun main() {
...
+
+`firstNotNullOf()` throws the `NoSuchElementException` if the resulting collection doesn't have a non-null element. 
+Use the counterpart [`firstNotNullOfOrNull()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first-not-null-of-or-null.html) 
+to return null in this case.
+
+```kotlin
+fun main() {
+//sampleStart
+    val data = listOf("0", "true", "false")
+    println(data.firstNotNullOf(String::toBooleanStrictOrNull))
```
p7nov, line 139: let's make a simple example without callable references and unfamiliar functions. A well-known string operation or smth similar.

docs/topics/collection-elements.md
```diff
@@ -71,7 +71,7 @@ fun main() {
 ```
 
-## Retrieving by condition
+## Retrieving with selector
```
p7nov, line 74: Why do you think it's better?
"by condition" seems more familiar for the general audience.

shadrina: Because I put in this section `firstNotNullOf()`, which takes the mapping selector, not a condition.

We use the term "selector" quite often in the documentation, although I agree that "condition" is more familiar. I can make the separate section, but then `firstNotNullOf` would be the only item there. WDYT?

p7nov: I'd create a separate new section because most of this section remains about "by condition".

ilya-g: +1 for a separate section.

Also note the section titles, the first one is "Retrieve ...", the second is "Retrieving ...".

docs/topics/collection-elements.md
ilya-g: I propose to show how it could play nicely with `takeIf`

```
val stuff = listOf<Any>(0, true, false)
// converts each element to string and returns the first one that has required length
val longEnough = stuff.firstNotNullOf { it.toString().takeIf { it.length >= 4 } } 
```

shadrina: Thanks! Added this sample.

docs/topics/collection-elements.md
shadrina: Made the separate section "Retrieve with selector" and unified the form of titles.

p7nov (approved):
LGTM

docs/topics/collection-elements.md
```diff
@@ -123,6 +123,29 @@ fun main() {
...
+`firstNotNullOf()` throws the `NoSuchElementException` if the resulting collection doesn't have a non-null element. 
+Use the counterpart [`firstNotNullOfOrNull()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first-not-null-of-or-null.html) 
+to return null in this case.
+
+```kotlin
+fun main() {
+//sampleStart
+   val list = listOf("one", "two", "three")
+   // Returns the first string that has required length
+   val longEnough = list.firstNotNullOf { it.takeIf { it.length > 4 } }
```
ilya-g, line 142: This way it can be replaced with just a plain 'first` with predicate because no mapping is done here. Let's preserve the mapping in the example.

shadrina: Oh, I see. Thank you, will fix.

shadrina: I preserved the initial example but gave a name to an outer implicit `it` not to produce shadowing

ilya-g (approved):

PR #2210, opened 2021-04-11
Update getting-started.md

harsh1000-hub:


koshachy:
Thanks for your PR!

There is no need to capitalize the "applications" word.
Also, if you look through our documentation you'll find out that we use **sentence case** for titles.

PR #2227, opened 2021-04-19
update: update all code blocks with deprecated APIs

koshachy:


docs/topics/basic-types.md
```diff
@@ -433,7 +433,7 @@ If a value of character variable is a digit, you can explicitly convert it to an
 fun decimalDigitValue(c: Char): Int {
     if (c !in '0'..'9')
         throw IllegalArgumentException("Out of range")
-    return c.toInt() - '0'.toInt() // Explicit conversions to numbers
+    return c.code - '0'.code // Explicit conversions to numbers
```
ilya-g, line 436: This is not a conversion anymore, it's rather getting `code` property. 
To convert a digit char to a number it's better to use the dedicated function now: `Char.digitToInt`.

I'd say this example is overall very misleading. It's fine for illustrating `digitToInt`, but not for the `code` property.

docs/topics/whatsnew13.md
```diff
@@ -403,7 +403,7 @@ efficient and easy to explore alternative: `keys.associateWith { getValue(it) }`
 fun main() {
 //sampleStart
     val keys = 'a'..'f'
-    val map = keys.associateWith { it.toString().repeat(5).capitalize() }
+    val map = keys.associateWith { it.toString().repeat(5).replaceFirstChar { it.uppercase() } }
```
ilya-g, line 406: This may not work with `kotlin-min-compiler-version="1.3"` and also not appropriate as an example of what's new in 1.3, because there were no such functions `uppercase` and `replaceFirstChar` in 1.3. Better to find another transform.

koshachy: I've decided just to make this block not runnable — this doc is about 1.3, so let leave it with the code that works for that version.

docs/topics/basic-types.md
koshachy: I've change `code` properties to `digitToInt()`, now it should work fine.

ilya-g: No, I mean the entire function `decimalDigitValue` better to be replaced by `Char.digitToInt()`

docs/topics/whatsnew13.md
ilya-g: ok, but `capitalize` is not essential for this example, can be replaced with something else.

docs/topics/basic-types.md
koshachy: Oh, I see 🤦 
FIxed it.

ilya-g (approved):

PR #2231, opened 2021-04-21
update: add issue about JVM test frameworks and Gradle to compatibility guide 1.5

PetrakovichVictoria:


docs/topics/compatibility-guide-15.md
```diff
@@ -308,4 +308,28 @@ perspective
 > **Deprecation cycle**:
 >
 > - 1.4: deprecate `Double.toShort()/toByte()` and `Float.toShort()/toByte()` and propose replacement
-> - 1.5.0: raise the deprecation level to error
\ No newline at end of file
+> - 1.5.0: raise the deprecation level to error
+
```
PetrakovichVictoria, line 312: Is an empty line obligatory?

docs/topics/compatibility-guide-15.md
```diff
@@ -308,4 +308,35 @@ perspective
...
+> variants for different testing frameworks.
+>
+> **Deprecation cycle**:
+>
+> - < 1.5: having several mutually exclusive `kotlin-test` variants for different testing frameworks is allowed
+> - \>= 1.5: behavior changed,  
+> Gradle throws an exception like "Cannot select module with conflict on capability...". Possible solutions:
+>    * consider using the same `kotlin-test` variant and the corresponding testing framework as the transitive dependency brings.
+>    * consider finding another variant of the dependency that doesn't bring the `kotlin-test` variant transitively, so you can use the testing framework you would like to use.
+>    * consider finding another variant of the dependency that brings another `kotlin-test` variant transitively, which uses the same testing framework you would like to use.
```
p7nov, line 334: If you list a number of possible solutions, then the reader already considers them all :) 
```suggestion
>    * use the same `kotlin-test` variant and the corresponding testing framework as the transitive dependency brings.
>    * find another variant of the dependency that doesn't bring the `kotlin-test` variant transitively, so you can use the testing framework you would like to use.
>    * find another variant of the dependency that brings another `kotlin-test` variant transitively, which uses the same testing framework you would like to use.
```

docs/topics/compatibility-guide-15.md
p7nov: Why do you ask? I don't know if it's technically required but it's always good to have one before each heading.

p7nov (approved):
Just one minor comment

PR #2234, opened 2021-04-22
feat: KT-45842 sealed class inheritance in MPP

p7nov:
Update doc as per https://youtrack.jetbrains.com/issue/KT-45842

### DIFF
--- a/docs/topics/sealed-classes.md
+++ b/docs/topics/sealed-classes.md
@@ -56,7 +56,7 @@ In [multiplatform projects](mpp-intro.md), there comes one more restriction: dir
 reside in the same source set. It applies to sealed classes without the [`expect`/`actual`](mpp-connect-to-apis.md) modifiers.
 
 A sealed class can be declared as `expect` in a common source set and have `actual` implementations in platform source sets.
-In this case, both `expect` and `actual` classes can have subclasses in their source sets. Moreover, if you use [hierarchical structure](mpp-share-on-platforms.md#share-code-on-similar-platforms),
+In this case, both `expect` and `actual` classes can have subclasses in their source sets. Moreover, if you use a [hierarchical structure](mpp-share-on-platforms.md#share-code-on-similar-platforms),
 you can create subclasses in any source set between the `expect` and `actual` declarations. 
 
 [Learn more about the hierarchical structure of multiplatform projects](mpp-share-on-platforms.md#share-code-on-similar-platforms). 
--- a/docs/topics/sealed-classes.md
+++ b/docs/topics/sealed-classes.md
@@ -50,6 +50,18 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
 >
 {type="note"}
 
+### Inheritance in multiplatform projects
+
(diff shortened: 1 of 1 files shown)


wild-lynx (approved):

docs/topics/sealed-classes.md
```diff
@@ -50,6 +50,18 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
 >
 {type="note"}
 
+### Inheritance in multiplatform projects
+
+In [multiplatform projects](mpp-intro.md), there comes one more restriction: direct subclasses of sealed classes must
```
koshachy, line 55: I prefer more simplified wording in such cases, like "There is an inheritance restriction in multiplatform projects:", but yours is also good!

docs/topics/sealed-classes.md
```diff
@@ -50,6 +50,18 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
...
+reside in the same source set. It applies to sealed classes without the [`expect`/`actual`](mpp-connect-to-apis.md) modifiers.
+
+A sealed class can be declared as `expect` in a common source set and have `actual` implementations in platform source sets.
+In this case, both `expect` and `actual` classes can have subclasses in their source sets. Moreover, if you use [hierarchical structure](mpp-share-on-platforms.md#share-code-on-similar-platforms),
```
koshachy, line 59:
```suggestion
In this case, both `expect` and `actual` classes can have subclasses in their source sets. Moreover, if you use a [hierarchical structure](mpp-share-on-platforms.md#share-code-on-similar-platforms),
```

docs/topics/sealed-classes.md
```diff
@@ -50,6 +50,18 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
 >
 {type="note"}
 
+### Inheritance in multiplatform projects
+
+In [multiplatform projects](mpp-intro.md), there comes one more restriction: direct subclasses of sealed classes must
+reside in the same source set. It applies to sealed classes without the [`expect`/`actual`](mpp-connect-to-apis.md) modifiers.
```
koshachy, line 56: Maybe a conjunction suits here better than the backslash sign.

docs/topics/sealed-classes.md
```diff
@@ -50,6 +50,18 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
...
+
+A sealed class can be declared as `expect` in a common source set and have `actual` implementations in platform source sets.
```
koshachy, line 58: The second paragraph is a bit unclear for me — it seems to be a little out of touch with the beginning, where the restriction is explained.
I cannot offer a better structure here yet, but I think we can use this one (I mean yours) and return to it in the future. WDYT?

docs/topics/sealed-classes.md
koshachy: You can omit this if it was suggested by the copy editors team.

docs/topics/sealed-classes.md
p7nov: I've changed the intro a bit. Not sure this is worth returning to, let's see if we have complaints.

PR #2235, opened 2021-04-22
Fix typo in control-flow.md

yunxing:
iterator function should return `Iterator<>` instead.


yunxing:
Ping on this -- it's a simple change!

PR #2237, opened 2021-04-22
Add operator keyword to Array<T> class for generics example

plastic-karma:
If I copy and paste the example, I get the exception

```
'operator' modifier is required on 'set' in 'Array'
```

To have a running example, the operator keyword needs to be added.


djj1109 (approved):

p7nov:
Thanks!

PR #2241, opened 2021-04-26
update: add links to videos to whatsnew15.md

p7nov:


koshachy (approved):
👍 

Next time we should try another place for videos and check the users' behavior :)

For example, maybe "Learn more" is better to be placed after the video 🤔

p7nov:
> For example, maybe "Learn more" is better to be placed after the video 🤔

I thought about this option but decided that "Learn more" will be less visible then. 
Let's check next time anyway.

PR #2248, opened 2021-04-28
chore: move the scope function selection table up

p7nov:


koshachy (approved):

koshachy:
There is room for improvement, but I think that it is out of scope for this PR.

PR #2251, opened 2021-04-29
minor change in introduction to variance

venkatasaikarthikeya:
List<? super String> accepts objects of type List of either String type, or supertype of String type, but not a supertype of the object type. But it is written in such a way that it accepts supertype of Object which is fundamentally wrong. So, it should be "List< String >" instead of "List< Object >". That is "List<? super String>" is a supertype of "List< String >".

### DIFF
--- a/docs/topics/generics.md
+++ b/docs/topics/generics.md
@@ -82,7 +82,7 @@ In other words, the wildcard with an _extends_\-bound (_upper_ bound) makes the 
 The key to understanding why this trick works is rather simple: if you can only _take_ items from a collection, 
 then using a collection of `String`s and reading `Object`s from it is fine. Conversely, if you can only _put_ items 
 into the collection, it's OK to take a collection of `Object`s and put `String`s into it: there is 
-`List<? super String>` in Java, a _supertype_ of `List<Object>`.
+`List<? super String>` in Java, a _supertype_ of `List<String>`.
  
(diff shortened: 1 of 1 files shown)


p7nov:
Hi! 
>List<? super String> accepts objects of type List of either String type, or supertype of String type, but not a supertype of the object type. But it is written in such a way that it accepts supertype of Object

If there were supertypes of `Object`, then you would be able to assign a list of those to `List<? super String>` as well.

```
        // Child extends Parent extends GrandParent
        // read id as:
        // String extends Object extends (whatever supertype if it existed)
        List<? super Child> l1 = new ArrayList<Child>(); //List<String>
        List<? super Child> l2 = new ArrayList<Parent>(); // List<Object>
        List<? super Child> l3 = new ArrayList<GrandParent>(); //List<?supertype of object?>
```

>So, it should be "List< String >" instead of "List< Object >". That is "List<? super String>" is a supertype of "List< String >".

Whenever you refer to `List<? super String>` (I mean writing to it, of course), you can use `List<Object>` instead of it:`String` is an `Object`, so `List<Object>` accepts it.
But not vice versa: if you write anything to `List<Object>` then you can't pass `List<? super String>` here because only `String` fit in that list, not any `Object`.
So, in the writing context, any `List<Object>` is a `List<? super String>` but the reverse is wrong. This means that the current doc wording is correct: `List<? super String>` _is a supertype of_ `List<Object>`.

PR #2265, opened 2021-05-07
Inline classes are stable now

vkotovv:
See:
- https://blog.jetbrains.com/kotlin/2021/02/new-language-features-preview-in-kotlin-1-4-30/#inline-value-classes-stabilization
- https://blog.jetbrains.com/kotlin/2021/05/kotlin-1-5-0-released/#stable-language-features


koshachy:
Thanks!

PR #2275, opened 2021-05-13
Clarify fun interfaces vs. type aliases section on which one to use

elizarov:
The current text creates an impression that functional interfaces should be preferred to typealises, since they are more flexible, which is not really the case. Clarification is added to explain in which circumstances the use of functional interfaces is warranted, and in which a plain functional type shall be used.

### DIFF
--- a/docs/topics/fun-interfaces.md
+++ b/docs/topics/fun-interfaces.md
@@ -91,8 +91,13 @@ You can also use [SAM conversions for Java interfaces](java-interop.md#sam-conve
 
 Functional interfaces and [type aliases](type-aliases.md) serve different purposes. Type aliases are just names for
 existing types – they don't create a new type, while functional interfaces do.
+You can provide extensions that are specific to a particular functional interace, so that they would be inapplicable for
+plain functions or for typealiases to plain functions. 
 
 Type aliases can have only one member, while functional interfaces can have multiple non-abstract members and one abstract member.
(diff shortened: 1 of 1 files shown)


AlexanderPrendota:
Bad samples
Snippets:
docs/topics/fun-interfaces.md

```kotlin
fun interface SuspendRunnable {
   suspend fun invoke()
}
class Listener {
   fun setOnClickListener(r: SuspendRunnable) {
       GlobalScope.launch { r.invoke() }
   }
}
class Button(private val listener: Listener) {
   fun process() {
       listener.setOnClickListener {
           addText()
       }
   }
   suspend fun addText() {
       // ...
   }
}
```

(5:7, 5:18) **ERROR** Unresolved reference: GlobalScope
(5:30, 5:36) **ERROR** Suspend function 'invoke' should be called only from a coroutine or another suspend function
    
------

elizarov:
I also proposed to drop an example with `SuspendRunnable` for many reasons:

* It is not well formatted (no separating empty lines)
* It shows a bad example of unstructured concurrency using `GlobalScope`
* It is is generally useless (no benefit in declaring fun interface here)
* It is quite long, compared to other, short, to-the-point, and easy-to-read examples.

I cannot quicky come up with an good example as a replacement and I don't see much reason to try. The section has enough content already.

koshachy (approved):

koshachy (approved):

PR #2283, opened 2021-05-14
Add emphasis for heading requirement

LouisCAD:
It makes it harder to miss the most important part of the section.


p7nov:
Thanks! I've adjusted the highlighting a bit to follow our style guide.

PR #2288, opened 2021-05-19
chore: Change anchor text

shadrina:


AlexanderPrendota:
Bad samples
Snippets:
docs/topics/whatsnew11.md

```kotlin
import kotlin.coroutines.experimental.*

//sampleStart
    val seq = buildSequence {
      for (i in 1..5) {
          // yield a square of i
          yield(i * i)
      }
      // yield a range
      yieldAll(26..28)
    }

    // print the sequence
  println(seq.toList())
//sampleEnd
}
```

(0:25, 0:37) **ERROR** Unresolved reference: experimental
(2:9, 2:13) **WARNING** Parameter 'args' is never used
(4:14, 4:27) **ERROR** Using 'buildSequence(noinline suspend SequenceScope<T>.() -> Unit): Sequence<T>' is an error. Use 'sequence { }' function instead.
    
------

koshachy (approved):
I've fixed the runnable snippet.

Overall, it is pretty good!

I think that we can avoid the `here` word in the future: it can be omitted (# See how that works) or rephrased (# Available version listed here -> See the list of available versions).

shadrina:
Thank you!

PR #2294, opened 2021-05-21
Update object-declarations.md

pichuzhkin:
I think "Multiple" should be used here instead of "Many"


p7nov:
Yes, you're right. Thank you!

PR #2297, opened 2021-05-21
KT-45870

kalibrov520:
`binaries.executable()` can't be called for `BOTH`


p7nov:
Thanks, Ilya. 
Since this example is about showing available modes, not about creating some binaries, we better remove the conflicting `binaries.executable()` and leave `BOTH` as an option.

PR #2298, opened 2021-05-21
feat: KT-46614 add K/JS frameworks overview

p7nov:


p7nov:
@dromanov-jb please check. I can't add you as a reviewer for some reason

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
+
+Modern web development benefits significantly from frameworks that simplify building web applications.
+Here are examples of popular web frameworks for Kotlin/JS written by different authors:
+
+### KVision
+
+_KVision_ is an object-oriented web framework that makes it possible to write applications in Kotlin/JS without having to
+write HTML or CSS. Instead, it provides ready-to-use components that can be used as building blocks for your application’s
+user interface. You can use both reactive and imperative programming models to build your frontend, use connectors for Ktor,
+Spring Boot, and other frameworks to integrate it with your server-side applications, and share code using Kotlin Multiplatform.
```
koshachy, line 72: I think that we can provide a link to Multiplatform.

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
+
+Visit [https://www.fritz2.dev](https://www.fritz2.dev) for documentation, tutorials, and examples.
+
+Join the [#fritz2](https://kotlinlang.slack.com/messages/fritz2) and [#javascript](https://kotlinlang.slack.com/archives/C0B8L3U69)
+channels in the [Kotlin Slack](https://surveys.jetbrains.com/s3/kotlin-slack-sign-up)to stay up to date with the latest
+news about this framework.
+
+### Doodle
+
+_Doodle_ is a vector-based UI framework for Kotlin/JS. Apps built using it do not rely on DOM, CSS, or Javascript. Instead,
```
koshachy, line 93: > Apps built using it do not rely on DOM, CSS, or Javascript.

It's really hard to read and understand this. I'd recommend rephrasing.

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
 means that you can interact with the JavaScript ecosystem in a structured manner.
 
+## Kotlin/JS frameworks
+
+Modern web development benefits significantly from frameworks that simplify building web applications.
+Here are examples of popular web frameworks for Kotlin/JS written by different authors:
+
+### KVision
+
+_KVision_ is an object-oriented web framework that makes it possible to write applications in Kotlin/JS without having to
```
koshachy, line 69: Seems like a bizarre order of sentences...
Is it necessary to say what this framework doesn't provide/have first?

I see this approach in other framework descriptions (like Doodle), and I am not sure that this is a good way.

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
+Visit [https://www.fritz2.dev](https://www.fritz2.dev) for documentation, tutorials, and examples.
+
+Join the [#fritz2](https://kotlinlang.slack.com/messages/fritz2) and [#javascript](https://kotlinlang.slack.com/archives/C0B8L3U69)
+channels in the [Kotlin Slack](https://surveys.jetbrains.com/s3/kotlin-slack-sign-up)to stay up to date with the latest
+news about this framework.
+
+### Doodle
+
+_Doodle_ is a vector-based UI framework for Kotlin/JS. Apps built using it do not rely on DOM, CSS, or Javascript. Instead,
+they use the browser’s graphics capabilities to draw user interfaces, giving you precise control over the rendering of
```
koshachy, line 94: Again, why don't we use the following order: Application use this, and don't rely on that?

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
+### KVision
+
+_KVision_ is an object-oriented web framework that makes it possible to write applications in Kotlin/JS without having to
+write HTML or CSS. Instead, it provides ready-to-use components that can be used as building blocks for your application’s
+user interface. You can use both reactive and imperative programming models to build your frontend, use connectors for Ktor,
+Spring Boot, and other frameworks to integrate it with your server-side applications, and share code using Kotlin Multiplatform.
+
+Visit [https://kvision.io](https://kvision.io) for documentation, tutorials, and examples.
+
+For updates and discussions about the framework, join [#kvision](https://kotlinlang.slack.com/messages/kvision) and
```
koshachy, line 76: We can make these descriptions about channels more consistent.

Also, don't you think that it's better to use the "from goal to action" way?
`To get the result, do this`. I mean, before we ask users to join the channels, we should explain the benefits (why they should do that). This sentence is OK, but others use different word order.

docs/topics/js/js-overview.md
```diff
@@ -59,6 +59,61 @@ at the expense of type safety.
...
+this framework.
+
+### Compose for Web
+
+_Compose for Web_ brings [Google’s Jetpack Compose UI toolkit](https://developer.android.com/jetpack/compose) to your browser.
+It allows you to build reactive web user interfaces using the concepts introduced by Jetpack Compose. It provides a DOM API
+to describe your website, as well as an experimental set of multiplatform layout primitives. Compose for Web also gives
+you the option to share parts of your UI code and logic across Android, desktop, and web.
+
+Compose for Web is in _Technology Preview_, which means it is ready for experiments but not yet ready for production use.
```
koshachy, line 110: This information is extremely difficult to keep up to date.

Maybe give a link with the text "pay attention to the status of the feature" or smth like that?

https://developer.android.com/jetpack/androidx/versions/all-channel

koshachy (changes requested):
Nice job!
See the comments below. 

Feel free to start a discussion if you disagree with them.

docs/topics/js/js-overview.md
p7nov: Indeed it is. But I think it's crucial to have a note about unstable/non-production stage right here because we list Compose among other frameworks that don't have such concerns.

When Compose for Web graduates, we'll have to adjust our docs anyway (add examples, provide link to complete docs, whatever), so I don't think this point is worth removing important info about the current state.

p7nov:
Thanks @koshachy. I've fixed all things you point to except the note about Compose for Web stability. Please see my answer in the discussion

docs/topics/js/js-overview.md
koshachy: OK, sounds reasonable. So let's don't miss the stable version of Compose :)

koshachy (approved):
👍

PR #2300, opened 2021-05-22
Add `value` in Modifiers order

BraisGabin:
`value` was introduced in `1.5` and because is very related with `inline` it seems reasonable to have the same priority.


cortinico (approved):

koshachy (approved):

koshachy:
Hey,
Thanks for your PR!

PR #2313, opened 2021-05-31
Object Declarations: Change fields to members

YektaDev:
Class members can access any private member of the corresponding companion object, not necessarily fields (which still would be replaced with the word "properties").

An example would be:

```kotlin
class A {
  companion object B {
    private val c = "World!"
    private fun foo() = "Hello"
  }

  init {
    println("${A.foo()} ${A.c}") // Hello World!
  }
}
```


shadrina (approved):
Thanks!

PR #2321, opened 2021-06-03
Java2kotlin idioms guide

PetrakovichVictoria:


docs/topics/idioms.md
```diff
@@ -49,6 +51,8 @@ if ("jane@example.com" !in emailsList) { ... }
 println("Name $name")
 ```
 
+Learn the difference between [Java and Kotlin string concatenation](java-to-kotlin-idioms-strings.md#concatenate-strings).
```
PetrakovichVictoria, line 54: I wonder if I can call concatenation so here because here the section is called "String interpolation".

docs/topics/jvm/java-to-kotlin-idioms-strings.md
```diff
@@ -0,0 +1,390 @@
...
+```
+{id="join-strings-11-java"}
+
+In Java 15, [text blocks](https://docs.oracle.com/en/java/javase/15/text-blocks/index.html) appeared. 
+There is one thing to keep in mind: if you print a multiline string and the triple-quote is on the next line, 
+there will be an extra empty line:
+
+```java
+// Java
+   String result = """
```
koshachy, line 335: Do we really need this indent?

koshachy (commented):
Just a couple of comments.

docs/topics/idioms.md
PetrakovichVictoria: @koshachy what do you think?

koshachy: LGTM

docs/topics/jvm/java-to-kotlin-idioms-strings.md
PetrakovichVictoria: Oh, I missed it. Thank you, fixed.

PR #2325, opened 2021-06-04
Update examples for list and set

adamu:
This PR updates the examples given for a list and a set in this article: https://kotlinlang.org/docs/collections-overview.html

I think that an alphabet is a bad example for a set, because an alphabet is ordered. We even have the term _alphabetical order_. So I changed it to a lottery ticket.

I also changed the example for list, because while a sentence is a collection of words, it also has spaces and punctuation, so I think that example was unclear. I used the example of a telephone number instead, which is just a collection of ordered numbers.

I'm happy to change the examples here,  these were just the better examples I thought of. The main thing that I think should change is the usage of an alphabet as an example of a set.

### DIFF
--- a/docs/topics/collections-overview.md
+++ b/docs/topics/collections-overview.md
@@ -13,10 +13,11 @@ calculate their average age.
 The  following collection types are relevant for Kotlin:
 
 * _List_ is an ordered collection with access to elements by indices – integer numbers that reflect their position. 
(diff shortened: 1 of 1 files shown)


adamu:
bump

p7nov:
Thanks for this improvement! Sorry it took us so long to respond.

PR #2335, opened 2021-06-10
feat: JS IR migration guide

p7nov:
@kalibrov520 @dromanov-jb please check the draft too.

### DIFF
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -16,7 +16,7 @@ may require adjusting your code. On this page, we've compiled a list of known mi
 
 Note that this guide may change over time as we fix issues and find new ones. Please help us keep it complete –
 report any issues you encounter when switching to the IR compiler by submitting them to our issue tracker [YouTrack](https://kotl.in/issue)
-.<!-- or filling out [this Google form](TODO: link). -->
+or filling out [this form](https://surveys.jetbrains.com/s3/ir-be-migration-issue).
 
 ## Convert JS- and React-related classes and interfaces to external interfaces
 
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -5,26 +5,26 @@
 >
 {type="warning"}
 
-[The IR-based Kotlin/JS compiler](js-ir-compiler.md) comes to replace the current compiler for various reasons from unifying
-the Kotlin behavior on all platforms to enabling new JS-specific optimizations.
-You can learn more about the internal differences of two compilers in the blog post 
+We are replacing the current Kotlin/JS compiler with [the IR-based compiler](js-ir-compiler.md) in order to unify
+Kotlin’s behavior on all platforms and to make it possible to implement new JS-specific optimizations, among other reasons.
+You can learn more about the internal differences between the two compilers in the blog post
 [Migrating our Kotlin/JS app to the new IR compiler](https://dev.to/kotlin/migrating-our-kotlin-js-app-to-the-new-ir-compiler-3o6i)
 by Sebastian Aigner.
 
-Due to the significant internal differences, switching your Kotlin/JS project from the old backend to the new one
-may require adjusting your code. On this page, we've collected known migration issues along with suggested solutions.
+Due to the significant differences between the compilers, switching your Kotlin/JS project from the old backend to the new one
+may require adjusting your code. On this page, we've compiled a list of known migration issues along with suggested solutions.
 
-Note that this guide can change with time: we're fixing issues and finding new ones. Please help us keep it complete –
-report any issues you face when switching to the IR compiler to our issue tracker [YouTrack](https://kotl.in/issue)
-<!-- or [this Google form](TODO: link) -->.
+Note that this guide may change over time as we fix issues and find new ones. Please help us keep it complete –
+report any issues you encounter when switching to the IR compiler by submitting them to our issue tracker [YouTrack](https://kotl.in/issue)
+.<!-- or filling out [this Google form](TODO: link). -->
 
-## Convert JS- and React-related classes and interfaces to external interfaces 
+## Convert JS- and React-related classes and interfaces to external interfaces
 
 **Issue**: Using Kotlin interfaces and classes (including data classes) that derive from pure JS classes, such as React's `RState` and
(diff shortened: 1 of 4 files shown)


docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+```kotlin
+// Replace this
+val ktApp = AppPropsImpl("App2") // Kotlin object
+```
+
+```kotlin
+// With this
+val jsApp = js("{name: 'App1'}") as AppProps // or jsObject {}
```
SebastianAigner, line 173: `jsObject` is a helper function that comes from the `kotlin-wrappers` – https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-extensions/src/main/kotlin/kotlinext/js/Helpers.kt

It might be useful to link this / mention that this code can be used (it's stand-alone, and not really react-specific)

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+**Solution 2**: create objects with `kotlin.js.json()`:
+
+```kotlin
+val jsonApp = kotlin.js.json(Pair("name", "App3")) as AppProps
```
SebastianAigner, line 179: Is this a good idea? I'm not judging, I just haven't ever seen this before

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+## Create plain JS objects for interoperability
+
+**Issue**: Properties of a Kotlin object that implements an external interface are not _enumerable_. This means that they are not 
+visible for operations that iterate over the object's properties, such as:
+* `for (var name in obj)`
+* `console.log(obj)`
+* `JSON.stringify(obj)`
+* And other
```
SebastianAigner, line 140: I'd get rid of the "and others", and replace "such as" with "for example" before the enumeration

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+```kotlin
+// With this
+external interface ComponentProps: RProps {
+    var visible: Boolean?
+}
+```
+
+## Convert functions with receiver in external interfaces to regular functions 
+
+**Issue**: external declarations can't contain functions with receiver, such as extension functions or properties with corresponding
```
SebastianAigner, line 115: I think it should be "functions with receivers"

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+```kotlin
+val props = js("{}") as ComponentProps
+props.isInitialized = true
+// visible is not initialized - OK in JS – means it's false
+if (props.isInitialized && props.visible) {} // not OK in Kotlin
+```
+
+However, this code with throw a `ClassCastException` because `visible` is `null` while it's type doesn't allow this.
```
SebastianAigner, line 95: "will throw", and "its type" :)

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+    var isInitialized: Boolean
+    var visible: Boolean
+}
+```
+
+```kotlin
+val props = js("{}") as ComponentProps
+props.isInitialized = true
+// visible is not initialized - OK in JS – means it's false
+if (props.isInitialized && props.visible) {} // not OK in Kotlin
```
SebastianAigner, line 92: Is there another workaround for this that's worth mentioning? Like checking against `undefined`? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/undefined.html

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+// With this
+external interface CustomComponentState : RState {
+    var name: String
+}
+```
+
+## Make boolean properties nullable in external interface
+
+**Issue**: in JavaScript, the `null` or undefined value of a boolean variable equals to `false`. So, boolean properties can be used
+in expressions without being defined.
```
SebastianAigner, line 79: I think this sentence needs to rephrased. It reads like "boolean properties can be used without being defined (that's ok)". Maybe bring some of the content from L95 up here?

docs/topics/js/js-ir-migration.md
```diff
@@ -0,0 +1,186 @@
...
+may require adjusting your code. On this page, we've collected known migration issues along with suggested solutions.
+
+Note that this guide can change with time: we're fixing issues and finding new ones. Please help us keep it complete –
+report any issues you face when switching to the IR compiler to our issue tracker [YouTrack](https://kotl.in/issue)
+<!-- or [this Google form](TODO: link) -->.
+
+## Convert JS- and React-related classes and interfaces to external interfaces 
+
+**Issue**: Using Kotlin interfaces and classes (including data classes) that derive from pure JS classes, such as React's `RState` and
+`RProps`, can cause `ClassCastException`. Such exceptions appear because the compiler attempts to work with instances of
```
SebastianAigner, line 24: "can cause a ClassCast..."

SebastianAigner (changes requested):
Added some points and thoughts to this – thanks for writing it up in this form :)

docs/topics/js/js-ir-migration.md
p7nov: @ilgonmic what do you think?

docs/topics/js/js-ir-migration.md
p7nov: @ilgonmic any ideas you could recommend?

docs/topics/js/js-ir-migration.md
ilgonmic: I saw this earlier in user projects, and it will work
But don’t know which way will be preferable for users to create plain JS objects
This one is good, because not require additional dependencies

docs/topics/js/js-ir-migration.md
ilgonmic: So, seems that this problem not in checking in if, but when this value is provided to Kotlin overridden method.
So in fact this problem can be reproduced if this value will be provided to button from react-dom
Something like
```
button {
   attrs {
       autoFocus = props.visible // will be class cast exception in depth of this action
   }
}
```

PR #2338, opened 2021-06-13
Update server-overview.md

sanity:


koshachy:
Thanks for your PR! 

From now on we add new server-side framework links only to https://kotlin.link/ 
There is a link to kweb.

Sorry for inconvenience

sanity:
Thank you, I understand.

One thing to note however is that the "Kotlin for server side" page is the top search result if you Google [kotlin web framework](https://www.google.com/search?q=kotlin+web+framework), so I suspect a lot of people are landing on this page rather than https://kotlin.link/ when looking for Kotlin web development options.

koshachy:
@sanity Oh, I see.

We should discuss how to organize such sections with frameworks/libraries — it's hard to keep a list of "not abandoned projects". TBH, I think that there is no silver bullet here, but we'll look for ways to organize the process for community libs.

PR #2340, opened 2021-06-14
1 5 20 lombok and caching annotation processors classloaders

PetrakovichVictoria:


docs/topics/kapt.md
```diff
@@ -266,3 +291,20 @@ fun encodeList(options: Map<String, String>): String {
 }
 ```
 
+## Making Java's annotation processors working
```
zveznicht, line 294: It is better to rephrase this part.
AP always works. Kapt simply allows to run them on all sources java+kotlin.

So I would go for something like this:

## Keep javac annotation processors

By default, kapt assumes that all annotation processors work through it and disables Java's compiler's annotation processors.  Some AP should be run by javac (for example lombok). To enable javac annotation processing use flag keepJavacAnnotationProcessors, so kapt won't disable javac annotation processing.

docs/topics/lombok.md
```diff
@@ -0,0 +1,142 @@
...
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-7112).
+>
+{type="warning"}
+
+Since 1.5.20, Kotlin supports the experimental Lombok Kotlin compiler plugin. This plugin allows the use of
+Java's Lombok-generated declarations in mixed Kotlin/Java sources.
+If Java's Lombok-generated declarations are in another module than the Kotlin code that uses these declarations, 
+then you don't need to use this plugin.
+
+[Lombok](https://projectlombok.org/) does Java code generation itself, not the Lombok Kotlin compiler plugin.
```
zveznicht, line 14: I would have added 'You still need to configure lombok as usual, when using this plugin'

zveznicht (approved):

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+> Generation of `@Builder` annotation won't be supported in the current prototype. However, you can give your feedback
+> about the plugin and vote for [`@Builder` in YouTrack](https://youtrack.jetbrains.com/issue/KT-46959).
+>
+{type="warning"}
+
+> Kotlin compiler ignores Lombok annotations if you use them in Kotlin code.
+>
+{type="note"}
+
+Please read below about how to apply the Lombok Kotlin compiler plugin to your Gradle/Maven build.
```
koshachy, line 32: This line seems redundant because it doesn't affect anything.

What if we just rename Gradle and Maven headings to "Using in/with Gradle", "Using in/with Maven"? Does it make sense?

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+kotlinLombok {
+    lombokConfigurationFile(file("lombok.config"))
+}
+```
+
+Learn the [test project with examples of Lombok compiler plugin and `lombok.config` usages](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_gradle/withconfig).
+
+## Maven
+
+Here's how to use Lombok Kotlin compiler plugin with Maven:
```
koshachy, line 76: To apply/use Lombok ... plugin, add to the pom file/ `pom.xml`:

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+    </configuration>
+</plugin>    
+```
+
+> The Lombok compiler plugin works successfully along with [kapt](kapt.md) if annotations processors don't depend
+> on the code generated by Lombok.
+>
+{type="note"}
+
+Learn the test project examples of kapt and Lombok compiler plugins usages:
```
koshachy, line 146: You can use "look through", "see", etc.

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+</plugin>
+```
+
+Please refer to the [Gradle](#gradle) section for detailed information about the Lombok Kotlin compiler plugin configuration.
+
+Learn the [test project example of Lombok compiler plugin and `lombok.config` usages](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_maven/nokapt).
+
+## Using the plugin along with kapt
+
+If you use [kapt](kapt.md) compiler plugin, note that Java's Lombok should be launched via annotation processor of Java's compiler.
```
koshachy, line 107: Couldn't we just say "Java Lombok"?

koshachy: Is it ok to say "If you use the kapt compiler plugin, you should launch Java Lombok via Java compiler annotation processor"?

koshachy (changes requested):

docs/topics/kapt.md
```diff
@@ -98,6 +100,29 @@ To use the Gradle worker API for parallel execution of kapt tasks, add this line
...
+>
+{type="warning"}
+
+This experimental feature makes it possible to cache the classloaders of annotation processors in kapt and so 
+increases the speed of kapt for many consecutive Gradle runs.
+
+To enable this feature, use the following properties in your `gradle.properties` file:
+
+```groovy
+#positive value will enable caching
```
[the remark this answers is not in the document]
PetrakovichVictoria: I absolutely agree. It seemed to me that I saw somewhere the usage without spaces. But maybe I mixed up sth because I didn't find any examples without whitespaces. Will fix.

docs/topics/kapt.md
```diff
@@ -266,3 +291,17 @@ fun encodeList(options: Map<String, String>): String {
 }
 ```
 
+## Keeping Java compiler's annotation processors
+
+By default, kapt assumes that all annotation processors work through it and disables Java's compiler's annotation processors.
```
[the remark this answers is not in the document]
PetrakovichVictoria: >Work through it

Is "work via it" ok or do you mean it is better to rephrase this sentence at all like:
"By default, all annotation processors work through kapt and kapt disables javac's annotation processors"

docs/topics/kapt.md
```diff
@@ -266,3 +291,17 @@ fun encodeList(options: Map<String, String>): String {
 }
 ```
 
+## Keeping Java compiler's annotation processors
+
+By default, kapt assumes that all annotation processors work through it and disables Java's compiler's annotation processors. 
+Some annotation processors should be run by javac, for example, [Lombok](https://projectlombok.org/).
```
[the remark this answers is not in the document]
PetrakovichVictoria: I like it, thanks.

docs/topics/kapt.md
```diff
@@ -266,3 +291,17 @@ fun encodeList(options: Map<String, String>): String {
...
+Some annotation processors should be run by javac, for example, [Lombok](https://projectlombok.org/). 
+To enable javac's annotation processing in the Gradle, use the flag `keepJavacAnnotationProcessors` in your `build.gradle` file:
+
+```groovy
+kapt {
+    keepJavacAnnotationProcessors = true
+}
+```
+
+If you use the Maven build system, you need to specify a concrete plugin's settings.
```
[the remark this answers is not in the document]
PetrakovichVictoria: No, it isn't. 
>**a** concrete plugin's settings

docs/topics/kapt.md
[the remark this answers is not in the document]
koshachy: "via" is much better, but I'll recommend finding another verb.

docs/topics/kapt.md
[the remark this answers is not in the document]
koshachy: I mean "plugin settings" looks good for me. Let's see what copyeditors say.

Anyway, do we need an article here?

docs/topics/jvm/java-interop.md
```diff
@@ -702,3 +702,9 @@ var myProperty: String
 ```
 
 Behind the scenes, this will create two functions `getMyProperty` and `setMyProperty`, both marked as `external`.
+
+## Using Lombok-generated declarations in Kotlin
+
+If Java's Lombok-generated declarations are in another module than the Kotlin code that uses these declarations,
```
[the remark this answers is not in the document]
p7nov: IMO this lacks some intro, Lombok and its declarations just come out of the blue.

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
+[//]: # (title: Lombok compiler plugin)
+
+> The Lombok compiler plugin is [Experimental](components-stability.md). 
+> It may be dropped or changed at any time. Use it only for evaluation purposes. 
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-7112).
+>
+{type="warning"}
+
+Since 1.5.20, Kotlin supports the experimental Lombok Kotlin compiler plugin. This plugin allows the use of
+Java's Lombok-generated declarations in mixed Kotlin/Java sources.
```
p7nov, line 10: I'll repeat my concern from above: let's be as specific as possible. If it's about using Lombok-generated declarations from **Kotlin sources within the same module**, we should say exactly this.

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+If Java's Lombok-generated declarations are in another module than the Kotlin code that uses these declarations, 
+then you don't need to use this plugin.
+
+[Lombok](https://projectlombok.org/) does Java code generation itself, not the Lombok Kotlin compiler plugin. 
+So, you still need to configure Lombok as usual, when using this plugin. 
+Learn more about [how to make the plugin seeing Lombok's config](#defining-the-place-of-lombok-config).
+
+## Support of annotations
```
p7nov, line 18: I'd call this "Supported Lombok annotations"

docs/topics/lombok.md
p7nov: Please don't use slash as "or"

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+
+Learn which annotations the plugin supports and the
+[current state of Lombok Kotlin compiler plugin](https://github.com/JetBrains/kotlin/blob/master/plugins/lombok/lombok-compiler-plugin/README.md).
+
+> Generation of `@Builder` annotation won't be supported in the current prototype. However, you can give your feedback
```
p7nov, line 23: Not sure this is a good place for this note. We don't say a word about what is supported but explicitly say "We don't support Builder!" - this can cause negative perception.

docs/topics/lombok.md
p7nov: > What if we just rename Gradle and Maven headings to "Using in/with Gradle", "Using in/with Maven"? Does it make sense?

Actually we have such headings ("Gradle" and "Maven")for all other compiler plugins. Not sure that it's better than your idea, but it's consistent now.

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+The Lombok itself should be set up as the Java annotation processor (not via kapt) or as the plugin as below:
+
+```groovy
+plugins {
+    id "io.freefair.lombok" version "5.3.0"
+}
+```
+Learn the [test project with examples of Lombok compiler plugin usage](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_gradle/nokapt).
```
p7nov, line 59: "Learn the project" sounds awkward to me

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+Alternatively, you can use the `apply plugin` syntax:
+
+```groovy
+apply plugin: 'kotlin-plugin-lombok'
+```
+
+The Lombok itself should be set up as the Java annotation processor (not via kapt) or as the plugin as below:
```
p7nov, line 52: This line is hard to follow: you say what to do, but it's an advanced operation without any links or instructions.
Also, "should be set up" is unnecessary passive and also violates the recommendation to avoid "should".

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+plugins {
+    id "io.freefair.lombok" version "5.3.0"
+}
+```
+Learn the [test project with examples of Lombok compiler plugin usage](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_gradle/nokapt).
+
+### Defining the place of lombok.config
+
+If you use the file `lombok.config` for [configuring Lombok](https://projectlombok.org/features/configuration),
+Kotlin Lombok compiler plugin should know where this file situates. Use the following syntax in your `build.gradle` file:
```
p7nov, line 64: "Kotlin Lombok compiler plugin should know" lacks actor. Please give a direct instruction: what the user should do.
smth like "provide a path to it to the plugin". 
Don't forget to add `(.kts)` after `build.gradle`

docs/topics/lombok.md
p7nov: "note that Java's Lombok should be launched..." - just "launch Java Lombok..."
"should" is also a no-no: https://helpserver.labs.jb.gg/help/stardust/1.0/style-guide.html#modal-verbs

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+
+Learn the [test project example of Lombok compiler plugin and `lombok.config` usages](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_maven/nokapt).
+
+## Using the plugin along with kapt
+
+If you use [kapt](kapt.md) compiler plugin, note that Java's Lombok should be launched via annotation processor of Java's compiler.
+By default, kapt assumes that all annotation processors work through it and disables Java's compiler's annotation processors.
+To make Lombok working, you need to change this behavior.
+
+If you use the Gradle build system, add the additional flag to the `build.gradle` file to make Java's compiler's annotations
```
p7nov, line 111: "the Gradle build system" why not "Gradle"?
"add the additional" - repetition
"flag" - config property or option

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+processors working:
+
+```groovy
+kapt {
+    keepJavacAnnotationProcessors = true
+}
+```
+
+If you use Maven, use the following settings to make Lombok launched by Java's compiler:
```
p7nov, line 120: "use Maven, use the following settings" - "use" repetition
"to make Lombok launched" - unnecessary passive

docs/topics/lombok.md
```diff
@@ -0,0 +1,149 @@
...
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+            </annotationProcessorPath>
+        </annotationProcessorPaths>
+    </configuration>
+</plugin>    
+```
+
+> The Lombok compiler plugin works successfully along with [kapt](kapt.md) if annotations processors don't depend
```
p7nov, line 141: not sure  "successfully" is a good word here, let's check with copyeditors. In my head it usually stands for some event that happened and everything is OK, not for describing compatibility.

p7nov (changes requested):

docs/topics/kapt.md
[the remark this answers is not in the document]
PetrakovichVictoria: Seems you're right about both the apostrophe and the article.

docs/topics/lombok.md
PetrakovichVictoria: About this phrase – I stole it from another plugin's page)
Ok, will delete it.

>Please don't use slash as "or"
Ok.

>Actually we have such headings ("Gradle" and "Maven")for all other compiler plugins. 
This is why I used "Gradle" instead of "Using in Gradle", though had the same thought. So, I'll leave these headings as they are now.

docs/topics/lombok.md
PetrakovichVictoria: I thought we try to use only "learn", probably I was mistaken.

docs/topics/lombok.md
PetrakovichVictoria: >Couldn't we just say "Java Lombok"?

I'm not sure, let's wait for a proofread.

docs/topics/jvm/java-interop.md
```diff
@@ -702,3 +702,9 @@ var myProperty: String
...
+everything works out of the box. If you need to use Java's Lombok-generated declarations in mixed Kotlin/Java
```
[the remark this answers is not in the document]
PetrakovichVictoria: You understood correctly. I've rephrased this place.

docs/topics/kapt.md
```diff
@@ -85,7 +85,9 @@ kapt {
 }
 ```
 
-## Running kapt tasks in parallel (since 1.2.60)
+## Improving the speed of builds that use kapt
+
+### Running kapt tasks in parallel (since 1.2.60)
```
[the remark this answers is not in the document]
PetrakovichVictoria: >## Compile avoidance for kapt (since 1.3.20)

Should I do so here too?

docs/topics/kapt.md
```diff
@@ -98,6 +100,29 @@ To use the Gradle worker API for parallel execution of kapt tasks, add this line
...
+This experimental feature makes it possible to cache the classloaders of annotation processors in kapt and so 
+increases the speed of kapt for many consecutive Gradle runs.
+
+To enable this feature, use the following properties in your `gradle.properties` file:
+
+```groovy
+#positive value will enable caching
+#use the same value as the number of modules that use kapt
+kapt.classloaders.cache.size=5
+kapt.classloaders.cache.disableForProcessors=[specify annotation processors full names to disable cache for them]
```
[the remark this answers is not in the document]
PetrakovichVictoria: Replied in DM.

docs/topics/kapt.md
```diff
@@ -266,3 +291,17 @@ fun encodeList(options: Map<String, String>): String {
 }
 ```
 
+## Keeping Java compiler's annotation processors
+
+By default, kapt assumes that all annotation processors work through it and disables Java's compiler's annotation processors. 
+Some annotation processors should be run by javac, for example, [Lombok](https://projectlombok.org/). 
+To enable javac's annotation processing in the Gradle, use the flag `keepJavacAnnotationProcessors` in your `build.gradle` file:
```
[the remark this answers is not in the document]
PetrakovichVictoria: Actaully, we do:

gradle.md:
>If you do not need a standard library at all, you can add the opt-out flag to the `gradle.properties`:

js-external-declarations-with-dukat.md:
>You can use the flag `kotlin.js.generate.externals` in your `gradle.properties`

js-project-setup.md:
>set the `cssSupport.enabled` flag in the Gradle build file in the `commonWbpackConfig`

the kapt.md itself:
>Add the additional flag to the `build.gradle` file to enable error type inferring in stubs:

and so on.

>And "the" with Gradle.

Didn't catch what do you mean.

docs/topics/jvm/java-interop.md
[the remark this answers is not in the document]
PetrakovichVictoria: Does it work?

>You can use Java's Lombok-generated declarations in Kotlin code. If these declarations are in another module 
than the Kotlin code that uses these declarations, everything works out of the box. 
If you need to generate and use such declarations in the same multilingual module, learn how to do this on the [Lombok compiler plugin's page](lombok.md).

docs/topics/lombok.md
PetrakovichVictoria: Does it work?

>The Lombok compiler plugin allows the generation and use of Java's Lombok declarations within the same multilingual module. 
If such declarations are generated in another module than the Kotlin code that uses these declarations, 
then you don't need to use this plugin.

PetrakovichVictoria: Also, I've put a suggestion to fix the same in What's new Google doc.

docs/topics/lombok.md
PetrakovichVictoria: But will it then be valid to leave about Builder here?

docs/topics/lombok.md
PetrakovichVictoria: >"Kotlin Lombok compiler plugin should know" lacks actor. Please give a direct instruction: what the user should do.

I'm trying to explain the underlying motivation for actions. Does it work?

>If you use the file `lombok.config` for [configuring Lombok](https://projectlombok.org/features/configuration),
Kotlin Lombok compiler plugin should know where this file situates. Provide a `lombok.config`'s path to the plugin in your `build.gradle` file:


>Don't forget to add (.kts) after build.gradle

But this is the Groovy file. I researched our compiler plugin's docs, in many (or even most of them) instructions in Groovy are used.

docs/topics/lombok.md
PetrakovichVictoria: >"add the additional" - repetition

Ha-ha, even didn't catch it from the first time – was thinking where is the repetition and what's wrong)) Personally, I don't see anything bad in _add the additional_. They are just words with similar letters, no?

Actually, we have such wording in our kapt.md:
>Add the additional flag to the `build.gradle` file to enable error type inferring in stubs:

Will fix it too.

I need to say that my experience tells me that developers usually call things with true/false values flags.

docs/topics/lombok.md
PetrakovichVictoria: Can replace with "properly"...

docs/topics/lombok.md
p7nov: > Java's Lombok should be launched by
This can be read in multiple meanings, this is why we're avoiding should and passive.
The exact instruction for a reader can be something like "Use a Java AP to launch Lombok".

docs/topics/jvm/java-interop.md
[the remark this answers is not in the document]
p7nov: This looks good, but "multilingual modules" arent' defined anywhere in our docs. Let's use a simpler wording.

docs/topics/kapt.md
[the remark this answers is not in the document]
p7nov: Yes, please

docs/topics/kapt.md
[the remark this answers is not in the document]
p7nov: Well, we shouldn't :)
This is what Google style guide says: https://developers.google.com/style/word-list#letter-f
To your another response about this: of course developers do use 'flag', but they also have a lot of other professional jargon. Let's keep the set of terms we use to the minimum.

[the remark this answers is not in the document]
p7nov: > Didn't catch what do you mean.

I mean don't use the article "the" with "Gradle".

docs/topics/lombok.md
p7nov: See my answer above about "multilingual module". Otherwise good.

docs/topics/lombok.md
p7nov: IIRC I've left a comment that I don't see it valid in this section in general.

p7nov: This one: https://github.com/JetBrains/kotlin-web-site/pull/2340#discussion_r653505151

docs/topics/lombok.md
p7nov: > I'm trying to explain the underlying motivation for actions. Does it work?

I think it's unnecessary here. "If you use a config file, provide a path to it" is enough.

> But this is the Groovy file. 

It can be Kotlin, can't it? I can't remember cases where we restrict the usage of Kotlin for Gradle build scripts.
Compiler plugin docs were written some years ago when Kotlin support in build.gradle was poor and maybe even unofficial. We should update these docs as well.

docs/topics/lombok.md
p7nov: > They are just words with similar letters, no?

No, these are cognate words - they share the same root. **add**itional is what was **add**ed.

docs/topics/jvm/java-interop.md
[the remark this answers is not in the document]
PetrakovichVictoria: Does it work?

>in the same mixed Java/Kotlin module

docs/topics/lombok.md
PetrakovichVictoria: Ah, I've skipped this comment by a mistake, sorry. In the DM we decided to turn this note just to the text: "Currently, we don't have plans on supporting the @Builder annotation. However, we can consider this if you vote
for [`@Builder` in YouTrack](https://youtrack.jetbrains.com/issue/KT-46959)."

> We don't say a word about what is supported
We do this because the source of truth is the README under the link, and there would be redundant work to keep our doc synced with this README.

If to leave this heading as it's now: "Support of annotations", it seems to me it's ok to say here about @Builder. Otherwise, I don't know where to say about this but not in annotations, it can be negatively perceived everywhere.

Do you have any ideas where to put it?

docs/topics/lombok.md
PetrakovichVictoria: Ok, I'll fix to .kts in lombok.md.

>We should update these docs as well.

Ok, but until that time I'll use the example in Groovy in kapt.md for consistency.

docs/topics/kapt.md
```diff
@@ -98,6 +100,29 @@ To use the Gradle worker API for parallel execution of kapt tasks, add this line
...
 
+### Caching for annotation processors' classloaders (since 1.5.20)
+
+> Caching for annotation processors' classloaders in kapt is [Experimental](components-stability.md).
+> It may be dropped or changed at any time. Use it only for evaluation purposes.
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-28901).
+>
+{type="warning"}
+
+This experimental feature makes it possible to cache the classloaders of annotation processors in kapt and so
```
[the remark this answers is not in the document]
PetrakovichVictoria: Used
>Caching for annotation processors' classloaders increases the speed of kapt if you run many Gradle tasks consecutively.

docs/topics/kapt.md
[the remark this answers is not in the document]
PetrakovichVictoria: Used
>By default, kapt launches all annotation processors and disables javac's annotation processors.

docs/topics/jvm/java-interop.md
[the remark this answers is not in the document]
PetrakovichVictoria: Has it become better or both problems still persist?

>You can use Java's Lombok-generated declarations in Kotlin code. If you generate these declarations in a Java module and 
use them in another Kotlin module, everything works out of the box. 
If you need to generate and use such declarations in the same mixed Java/Kotlin module, 
learn how to do this on the [Lombok compiler plugin's page](lombok.md).

docs/topics/lombok.md
PetrakovichVictoria: Does it work?

>By default, [kapt](kapt.md) compiler plugin launches all annotation processors and disables javac's annotation processors.
You need to change this behavior: enable Java's Lombok.

>If you use Gradle, add the option to the `build.gradle.kts` file to make javac's annotations
processors working:

PR #2350, opened 2021-06-18
fix: KT-47324 fix attribute for enabling old backend in Gradle

koshachy:


docs/topics/gradle.md
```diff
@@ -607,7 +607,7 @@ The complete list of options for the Gradle tasks is the following:
 | `jdkHome` | Include a custom JDK from the specified location into the classpath instead of the default JAVA_HOME |  |  |
 | `jvmTarget` | Target version of the generated JVM bytecode | "1.6" (DEPRECATED), "1.8", "9", "10", "11", "12", "13", "14", "15", "16" | "%defaultJvmTargetVersion%" |
 | `noJdk` | Don't automatically include the Java runtime into the classpath |  | false |
-| `useIR` | Use the IR backend |  | false |
+| `useOldBackend` | Use the old JVM backend |  | false |
```
p7nov, line 610: Maybe some clarification that it was default before 1.5? Otherwise it's unclear what is "old"

p7nov (approved):

docs/topics/gradle.md
koshachy: I guess we could provide a link here to our 1.5.0 what's new, WDYT?

p7nov: Yes, this will work, I think.

PR #2351, opened 2021-06-18
Add a simple example of abstract classes

markallanson:
The current documentation has no simple example for abstract base classes and instead jumps directly into the syntax for overriding an open method in an open class to convert it to abstract in an inheriting classes.

This can confuse users experience for users new to the language especially since the example showed a rectangle class overriding and abstracting a method in the Polygon class.

This change introduces

1. A simple example for an abstract base class
2. A modified version of overriding and converting to abstract an open method in an open class.

This provides a more simple example to abstract classes for newer users.

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -217,7 +217,7 @@ abstract class Polygon {
     abstract fun draw()
 }
 
-class Rectangle : Polygon {
+class Rectangle : Polygon() {
     override fun draw() {
         // draw the rectangle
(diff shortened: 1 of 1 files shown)


nano-labs:
I'm learning Kotlin and that example confused me a lot. My first thought was "Oh, so kotlin's concept of abstract class works the other way around as I would expect from other languages?"

p7nov:
Thank you for this addition!

PR #2352, opened 2021-06-18
KTL-392 Kotlinlang - Consistent UI - update footer styles

nikpachoo:
- [YT](https://youtrack.jetbrains.com/issue/KTL-392)
- [UI](https://branch-ktl-392-update-footer.kotlin-web-site.labs.jb.gg/)


scripts/react-renderer/compile.mjs
```diff
@@ -3,8 +3,27 @@ import { renderToString } from 'react-dom/server';
...
-    console.log(renderToString(createElement(Component, props)));
+    let res = '';
+
+    try {
+        res = renderToString(createElement(Component, props))
+    } catch (e) {
+        res = e;
+    }
+
+    console.log(res);
```
krutilov, line 28: Is it needed?

krutilov (approved):

scripts/react-renderer/compile.mjs
zoobestik: Actually, yes. [Python takes](https://github.com/JetBrains/kotlin-web-site/blob/master/src/ktl_components.py#L60) node SSR from stdout.

static/js/ktl-component/footer/index.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import GlobalFooter from '@jetbrains/kotlin-web-site-ui/dist/footer.js';
+import '@jetbrains/kotlin-web-site-ui/dist/footer.css';
```
zoobestik, line 3: Is not possible case when footer/header not will have styles when js is disabled?

zoobestik (changes requested):

static/js/ktl-component/footer/index.jsx
nikpachoo: There is nothing to worry about. We use the "ExtractCssPlugin" that extracts all styles into a separate css file ("global.css").

zoobestik (approved):

PR #2353, opened 2021-06-18
Lombok and caching ap 2

PetrakovichVictoria:


docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
+[//]: # (title: Lombok compiler plugin)
+
+> The Lombok compiler plugin is [Experimental](components-stability.md). 
+> It may be dropped or changed at any time. Use it only for evaluation purposes. 
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-7112).
+>
+{type="warning"}
+
+The Lombok compiler plugin allows the generation and use of Java's Lombok declarations within the same mixed Java/Kotlin module.
+If you generate these declarations in a Java module and use them in another Kotlin module,
```
p7nov, line 10: The sentence needs a couple of improvements to be strictly true:
- "If you generate" - not necessarily "you", it can be anyone. Author of a library you use, for example.
- "in a Java module" - what if this module also contains code in Kotlin or other JVM languages?

p7nov: To avoid passive, you can start from "If ... declarations come from .."

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
+[//]: # (title: Lombok compiler plugin)
+
+> The Lombok compiler plugin is [Experimental](components-stability.md). 
+> It may be dropped or changed at any time. Use it only for evaluation purposes. 
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-7112).
+>
+{type="warning"}
+
+The Lombok compiler plugin allows the generation and use of Java's Lombok declarations within the same mixed Java/Kotlin module.
```
p7nov, line 9: When you introduce the plugin, please define it as precisely as possible. In this case "The Lombok compiler plugin" has nothing that connects it with Kotlin. It may be obvious (like - we're in Kotlin docs, right?), but it's bad for SEO and may be unclear to those who come from Google.
You can say, for example, "Kotlin Lombok compiler plugin" or "Lombok plugin for Kotlin compiler", or ..?

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+If you generate these declarations in a Java module and use them in another Kotlin module, 
+then you don't need to use this plugin.
+
+[Lombok](https://projectlombok.org/) generates Java code itself, not the Kotlin Lombok compiler plugin. 
+Thus, you still need to configure Lombok as usual, when using this plugin. 
+Learn more about [how to make the plugin seeing Lombok's config](#defining-the-place-of-lombok-config).
+
+## Support of annotations
```
p7nov, line 17: My concern about this whole section in the previous PR was that it seems unusable:
-  It's "Support of annotations" (should be "for", by the way). OK, let's see which annotations do we support.
- No supported annotations here! WTF? OK, there is a link. Let's go check here. Fine, I've found them after leaving the Kotlin website. Will I return to continue reading?
- If I return (hardly) I learn about what is not supported. It's a thought that continues what I've read on GitHub. Why is it here?

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+Currently, we don't have plans on supporting the @Builder annotation. However, we can consider this if you vote
+for [`@Builder` in YouTrack](https://youtrack.jetbrains.com/issue/KT-46959).
+
+> Kotlin compiler ignores Lombok annotations if you use them in Kotlin code.
+>
+{type="note"}
+
+## Gradle
+
+### Main settings
```
p7nov, line 31: Why do you need this heading?

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+plugins {
+    id ("io.freefair.lombok") version "5.3.0"
+}
+```
+
+See the [test project with examples of Lombok compiler plugin usage](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_gradle/nokapt).
+
+### Defining the place of lombok.config
+
+If you use the file `lombok.config` for [configuring Lombok](https://projectlombok.org/features/configuration),
```
p7nov, line 53:
```suggestion
If you use [Lombok configuration file](https://projectlombok.org/features/configuration)`lombok.config`,
```

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+provide a path to it to the plugin. Add the following code to your `build.gradle.kts` file:
+
+```kotlin
+kotlinLombok {
+    lombokConfigurationFile(file("lombok.config"))
```
p7nov, line 58: Can we make a more complete example with some directories, like "config/lombok.config"

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-maven-lombok</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+    </dependencies>
+</plugin>
+```
+
+Please refer to the [Gradle](#gradle) section for detailed information about the Lombok compiler plugin's configuration.
```
p7nov, line 91: I don't think that the Gradle instruction will actually help... It doesn't show what to write in maven.

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+
+See the [test project example of Lombok compiler plugin and `lombok.config` usages](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_maven/nokapt).
+
+## Using the plugin along with kapt
+
+By default, [kapt](kapt.md) compiler plugin launches all annotation processors and disables javac's annotation processors.
+You need to change this behavior: enable Java's Lombok.
```
p7nov, line 98: This reads a bit awkward. Can we explain that Lombok works only from javac and say how to use javac's APs?
"enable Java's Lombok" doesn't strictly correspond to what you write below. You just let javac APs works as they are supposed.

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+
+If you use Gradle, add the option to the `build.gradle.kts` file to make javac's annotations processors working:
+
+```kotlin
+kapt {
+    keepJavacAnnotationProcessors = true
+}
+```
+
+If Maven, use the following settings to launch Lombok with Java's compiler:
```
p7nov, line 108: If Maven?

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+            </annotationProcessorPath>
+        </annotationProcessorPaths>
+    </configuration>
+</plugin>    
+```
+
+> The Lombok compiler plugin works successfully along with [kapt](kapt.md) if annotations processors don't depend
```
p7nov, line 129:
```suggestion
> The Lombok compiler plugin works correctly along with [kapt](kapt.md) if annotations processors don't depend
```

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+> It may be dropped or changed at any time. Use it only for evaluation purposes. 
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-7112).
+>
+{type="warning"}
+
+The Lombok compiler plugin allows the generation and use of Java's Lombok declarations within the same mixed Java/Kotlin module.
+If you generate these declarations in a Java module and use them in another Kotlin module, 
+then you don't need to use this plugin.
+
+[Lombok](https://projectlombok.org/) generates Java code itself, not the Kotlin Lombok compiler plugin.
```
p7nov, line 13: The beginning is a bit strange: "Lombok generates Java code itself". It's better to be plugin-centric since this is the plugin doc.
Smth like: "The compiler plugin (does what?) while Lombok generates the declarations as it usually does in Java projects."

docs/topics/lombok.md
```diff
@@ -0,0 +1,137 @@
...
+
+```kotlin
+plugins {
+    id ("io.freefair.lombok") version "5.3.0"
+}
+```
+
+See the [test project with examples of Lombok compiler plugin usage](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_gradle/nokapt).
+
+### Defining the place of lombok.config
```
p7nov, line 51: Actually it's "specifying". But maybe a simpler version will work. Smth like "Using the Lombok configuration file".

p7nov (changes requested):

docs/topics/lombok.md
PetrakovichVictoria: >"in a Java module" - what if this module also contains code in Kotlin or other JVM languages?

Don't know anything about other JVM languages, do we consider any mixes in our docs?

>what if this module also contains code in Kotlin

Isn't "a Java module" enough to define that this module contains only code in Java? I'll add "a pure Java module".

>Author of a library you use, for example.

Here we're talking about code generation from sources and some peculiarities of processing such code with annotation processors. When using libraries, we use already generated code (re-asked this from the developer and he's agreed).

docs/topics/lombok.md
PetrakovichVictoria: Ok, I'd erased it earlier but will return back😂

docs/topics/lombok.md
PetrakovichVictoria: Because I wanted to divide into Main and "Defining the place of lombok.config". lombok.config is optional. Should I erase this heading?

docs/topics/lombok.md
PetrakovichVictoria: Thanks! I'll apply it manually. BTW, you forgot the space before lombok.config))

docs/topics/lombok.md
PetrakovichVictoria: True. I've added more details, you'll see them in the PR. But maybe we should delete this description at all and leave pure XML code as it is on other plugins pages.

docs/topics/lombok.md
PetrakovichVictoria: We can erase this title and about @Builder, so this stuff will be a part of the previous common section. But the external link will anyway stay even in this case.

docs/topics/lombok.md
PetrakovichVictoria: Replied in DM.

docs/topics/lombok.md
PetrakovichVictoria: >Can we explain that Lombok works only from javac

It was in the initial version just in other wording: "Java's Lombok should be launched via annotation processor of Java's compiler." For me "works only from javac" sounds strange too.

But there were problems with should and passive usage. But it's the exact sense...

>say how to use javac's APs?

Ok.

docs/topics/lombok.md
PetrakovichVictoria: >The compiler plugin (does what?)

He does only what is described above:
>The Kotlin Lombok compiler plugin allows the generation and use of Java's Lombok declarations within the same mixed Java/Kotlin module.

docs/topics/lombok.md
```diff
@@ -0,0 +1,164 @@
...
+        </dependency>
+    </dependencies>
+</plugin>
+```
+
+See the [test project example of Lombok compiler plugin and `lombok.config` usages](https://github.com/kotlin-hands-on/kotlin-lombok-examples/tree/master/kotlin_lombok_maven/nokapt).
+
+## Using the plugin along with kapt
+
+By default, [kapt](kapt.md) compiler plugin launches all annotation processors and disables javac's annotation processors.
```
zveznicht, line 124: I would changes 'launches' to 'runs'.

docs/topics/lombok.md
```diff
@@ -0,0 +1,164 @@
...
+You need javac's annotation processors working to launch Lombok.
```
zveznicht, line 125: For me it sounds weird. Maybe something like 'To run lombok along with kapt you need to tell it to keep javac's annotation processors'

docs/topics/lombok.md
zveznicht: 'and disables annotations processing by javac'

docs/topics/lombok.md
PetrakovichVictoria: Does it work?
>To run [Lombok](https://projectlombok.org/) along with kapt, you need to set up kapt to keep javac's annotation processors working.

docs/topics/lombok.md
```diff
@@ -0,0 +1,133 @@
...
+* `@Value`
+
+Currently, we don't have plans on supporting the `@Builder` annotation. However, we can consider this if you vote
+for [`@Builder` in YouTrack](https://youtrack.jetbrains.com/issue/KT-46959).
+
+> Kotlin compiler ignores Lombok annotations if you use them in Kotlin code.
+>
+{type="note"}
+
+We continue to work on this plugin. To find out the detailed current state, visit the [Lombok compiler plugin's README](https://github.com/JetBrains/kotlin/blob/master/plugins/lombok/lombok-compiler-plugin/README.md).
```
p7nov, line 34: Let's place this right below the list.

PR #2354, opened 2021-06-18
Unsigned types: removed Inline Class warning

A-Roso:
In the section unsigned types of the basic types chapter there is a line that "Unsigned types are implemented using feature that's not yet stable, namely inline classes.".
I removed it because inline classes are stable as of Kotlin 1.5.10


shadrina:
Thank you!

PR #2355, opened 2021-06-19
Correct function parameter naming notation

jlafshari:
Function parameters start with a lowercase letter, but the documentation says they use Pascal notation, which is contradictory. Pascal notation starts with an uppercase letter. So my edit is to change the doc to say "lower camel case".

### DIFF
--- a/docs/topics/functions.md
+++ b/docs/topics/functions.md
@@ -24,7 +24,7 @@ Stream().read() // create instance of class Stream and call read()
 
 ### Parameters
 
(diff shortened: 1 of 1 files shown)


shadrina:
Thank you for the report. The word "notation" is indeed misleading here. Actually, we do not mean the way we write identifiers, it is about the syntactic structure of parameter declaration, which has the form of `name: type` just as in Pascal language.

I've created a [YouTrack ticket](https://youtrack.jetbrains.com/issue/KT-47352).

PR #2357, opened 2021-06-21
update: Add information about JSpecify and type arguments bounds support

shadrina:


docs/topics/jvm/java-interop.md
```diff
@@ -153,6 +154,29 @@ When the `@NotNull` annotation is missing from a type argument, you get a platfo
...
+
+```kotlin
+fun takeBaseOfNotNullStrings(x: Base<String>) {}
+
+fun main() {
+    takeBaseOfNotNullStrings(Derived()) // warning: nullability mismatch
+}
+```
+
+The upper bound of `Derived` is set to `Base<String?>`, which is not a subtype of `Base<String>`.
```
shadrina, line 178: @p7nov In the previous release, you had a concern about the warning description. Is it now clear for a reader what is the warning and why does it appear?

docs/topics/jvm/java-interop.md
```diff
@@ -153,6 +154,29 @@ When the `@NotNull` annotation is missing from a type argument, you get a platfo
 fun toSet(elements: (Mutable)Collection<String!>) : (Mutable)Set<String!> { ... }
 ```
 
+Kotlin also takes into account nullability annotations on type arguments of base classes and interfaces. For example, 
+there are 2 Java classes with the signatures provided below.
```
p7nov, line 158:
```suggestion
there are two Java classes with the signatures provided below.
```

[MS style guide](https://docs.microsoft.com/en-us/style-guide/numbers): spell out whole numbers from zero through nine

docs/topics/jvm/java-interop.md
p7nov: Yes, this is good. However, variance is an advanced subject, so a link would be helpful. I'd add:
> Learn more about the [variance of generic classes](https://kotlinlang.org/docs/generics.html#variance).

p7nov (approved):
Looks good. Just a couple of small improvements.

docs/topics/jvm/java-interop.md
shadrina: Thank you!

shadrina:
@p7nov Thank you for pointing out the variance theme. I've made a mistake writing "is not a subtype of", because the `Base` class is invariant, so it doesn't matter if the argument type is a subtype or not. Fixed the wording, added the link to [Java generics In Kotlin](https://kotlinlang.org/docs/java-interop.html#java-generics-in-kotlin) section.

I'll send it for proofreading and merge.

PR #2380, opened 2021-07-02
Add emphasis in code of conduct

LouisCAD:


koshachy:
Hey Louis,
I don't think that we should emphasize only this part of the whole list... TBH, I don't think that we should emphasize anything in this list of rules...

Could you explain your motivation, please?

PR #2382, opened 2021-07-05
Make clear copy executes init block

hfhbd:
Make clear, copy is no "stupid" memory coping, but instead a shortcut for creating a new instance, calling the `init` block.

This would fail:
```kotlin
data class TestingCopy(val s: Int) {
    init {
        require(s != 42)
    }
}

val one = TestingCopy(s = 1)
val second = one.copy(s = 42)
```

### DIFF
--- a/docs/topics/data-classes.md
+++ b/docs/topics/data-classes.md
@@ -77,8 +77,9 @@ fun main() {
 
 ## Copying
   
-To copy an object for changing _some_ of its properties, but keeping the rest unchanged, use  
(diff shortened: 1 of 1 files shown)


shadrina:
Hi, Philip.

The example that goes right after this paragraph shows that generated `copy` implementation creates a new class instance:
```kotlin
fun copy(name: String = this.name, age: Int = this.age) = User(name, age)
                                                          ^^^^^^^^^^^^^^
```

Initializer blocks are executed during the initialization of an instance, so I don't think we need to state it explicitly.

PR #2387, opened 2021-07-07
Improve Gradle Kotlin example for lombok plugin

DRSchlaubi:
The rest of the Kotlin documentation uses the built-in `PluginDependencySpec.kotlin()` method to add Kotlin plugins the lombok plugin should do this too.
Also it's more common to not place a space between the `id` and the arguments as seen in the official Gradle plugin portal

### DIFF
--- a/docs/topics/lombok.md
+++ b/docs/topics/lombok.md
@@ -48,8 +48,8 @@ plugins {
 
 ```kotlin
 plugins {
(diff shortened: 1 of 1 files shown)


PetrakovichVictoria:
Hi, @DRSchlaubi! Thanks for your PR.

PetrakovichVictoria (approved):

PR #2391, opened 2021-07-08
chore: MKT-16782 Proofread Kotlin docs 3

p7nov:
Except generics.md.
@shadrina please add this file when you're done with it.

### DIFF
--- a/docs/topics/functions.md
+++ b/docs/topics/functions.md
@@ -4,7 +4,7 @@ Kotlin functions are declared using the `fun` keyword:
 
 ```kotlin
 fun double(x: Int): Int {
-   return 2 * x
+    return 2 * x
 }
 ```
 
@@ -35,8 +35,8 @@ You can use a [trailing comma](coding-conventions.md#trailing-commas) when you d
 
 ```kotlin
 fun powerOf(
-   number: Int,
-   exponent: Int, // trailing comma
+    number: Int,
+    exponent: Int, // trailing comma
 ) { /*...*/ }
 ```
 
@@ -47,9 +47,9 @@ of overloads:
 
 ```kotlin
 fun read(
-   b: ByteArray,
-   off: Int = 0,
-   len: Int = b.size,
+    b: ByteArray,
+    off: Int = 0,
+    len: Int = b.size,
 ) { /*...*/ }
 ```
 
@@ -60,11 +60,11 @@ When overriding a method that has default parameter values, the default paramete
 
 ```kotlin
 open class A {
-   open fun foo(i: Int = 10) { /*...*/ }
+    open fun foo(i: Int = 10) { /*...*/ }
 }
 
 class B : A() {
-   override fun foo(i: Int) { /*...*/ }  // No default value is allowed.
+    override fun foo(i: Int) { /*...*/ }  // No default value is allowed.
 }
 ```
 
@@ -73,8 +73,8 @@ the function with [named arguments](#named-arguments):
 
 ```kotlin
 fun foo(
-   bar: Int = 0,
-   baz: Int,
+    bar: Int = 0,
+    baz: Int,
 ) { /*...*/ }
 
 foo(baz = 1) // The default value bar = 0 is used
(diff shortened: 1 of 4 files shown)


docs/topics/functions.md
```diff
@@ -68,20 +68,20 @@ class B : A() {
...
+    bar: Int = 0,
     baz: Int,
 ) { /*...*/ }
 
 foo(baz = 1) // The default value bar = 0 is used
 ```
 
-If the last argument after default parameters is a [lambda](lambdas.md#lambda-expression-syntax), you can pass it 
-either as a named argument or [outside the parentheses](lambdas.md#passing-trailing-lambdas):
+If the last argument after default parameter, or  collection of default parameters,s is a [lambda](lambdas.md#lambda-expression-syntax),
```
shadrina, line 83:
```suggestion
If the last argument after default parameters is a [lambda](lambdas.md#lambda-expression-syntax),
```

docs/topics/functions.md
```diff
@@ -232,13 +233,13 @@ val list = asList(-1, 0, *a, 4)
...
-Functions marked with the `infix` keyword can also be called using the infix notation (omitting the dot and the parentheses 
+Functions marked with the `infix` keyword can also be called using the infix notation (omitting the dot and the parentheses
 for the call). Infix functions must meet the following requirements:
 
 * They must be member functions or [extension functions](extensions.md).
 * They must have a single parameter.
-* The parameter must not [accept variable number of arguments](#variable-number-of-arguments-varargs) and must have 
-no [default value](#default-arguments).
+* The parameter must not [accept variable number of arguments](#variable-number-of-arguments-varargs) and must have
+  no [default value](#default-arguments).
```
shadrina, line 242:
```suggestion
no [default value](#default-arguments).
```

docs/topics/generics.md
```diff
@@ -171,17 +174,17 @@ fun demo(x: Comparable<Number>) {
 }
 ```
 
-The words _in_ and _out_ seem to be self-explaining (as they were successfully used in C# for quite some time already), 
-thus the mnemonic mentioned above is not really needed, and one can rephrase it for a higher purpose:
+The words _in_ and _out_ seem to be self-explanatory (as they’ve already been used successfully in C# for quite some time),
+and so the mnemonic mentioned above is not really needed.  It can in fact be rephrased at a higher level of abstraction:
 
-**[The Existential](https://en.wikipedia.org/wiki/Existentialism) Transformation: Consumer in, Producer out\!** :-)
+**[The Existential](http://en.wikipedia.org/wiki/Existentialism) Transformation: Consumer in, Producer out\!** :-)
```
shadrina, line 180:
```suggestion
**[The Existential](https://en.wikipedia.org/wiki/Existentialism) Transformation: Consumer in, Producer out\!** :-)
```

docs/topics/object-declarations.md
```diff
@@ -78,11 +78,11 @@ class C {
 ```
 
 If this function or property is public or private inline, its actual type is:
-* `Any` if the anonymous object doesn't have a declared supertype
-* the declared supertype of the anonymous object if there is exactly one such type 
-* the explicitly declared type if there is more than one declared supertype
+* `Any` if the anonymous object doesn't have a declared supertype.
```
shadrina, line 81:
```suggestion
* `Any` if the anonymous object doesn't have a declared supertype
```

docs/topics/object-declarations.md
```diff
@@ -78,11 +78,11 @@ class C {
...
+* The declared supertype of the anonymous object, if there is exactly one such type.
```
shadrina, line 82:
```suggestion
* The declared supertype of the anonymous object, if there is exactly one such type
```

docs/topics/object-declarations.md
```diff
@@ -78,11 +78,11 @@ class C {
...
+* The explicitly declared type if there is more than one declared supertype.
```
shadrina, line 83:
```suggestion
* The explicitly declared type if there is more than one declared supertype
```

shadrina (approved):

PR #2396, opened 2021-07-09
Update data-science-overview.md

garethkoh:
NM Dev is a data science company that has libraries for classical mathematics, that run orders of magnitude faster than existing mathematical libraries. We hope to share our algorithms with the data science and mathematics communities.


koshachy (approved):

koshachy:
Thanks for your PR!

PR #2409, opened 2021-07-16
Update Unsigned types section and restructure the parent one

shadrina:


docs/topics/unsigned-types.md
```diff
@@ -0,0 +1,98 @@
...
+val a2 = 0xFFFF_FFFF_FFFFu // ULong: no expected type provided, constant doesn't fit in UInt
+```
+
+* `uL` and `UL` explicitly tag literal as unsigned long.
+
+```kotlin
+val a = 1UL // ULong, even though no expected type provided and constant fits into UInt
+```
+
+## Use cases
```
shadrina, line 64: @elizarov Please review only this "Use cases" section content

docs/topics/numbers.md
```diff
@@ -0,0 +1,289 @@
...
+
+All variables initialized with integer values not exceeding the maximum value of `Int`
+have the inferred type `Int`. If the initial value exceeds this value, then the type is `Long`.
+To specify the `Long` value explicitly, append the suffix `L` to the value.
+
+```kotlin
+val one = 1 // Int
+val threeBillion = 3000000000 // Long
+val oneLong = 1L // Long
+val oneByte: Byte = 1
```
elizarov, line 23: This example shows literal type inferred from the context, but the explanatory text does not mention this Kotlin's feature.

docs/topics/numbers.md
```diff
@@ -0,0 +1,289 @@
...
+val bytes = 0b11010010_01101001_10010100_10010010
+```
+
+## Numbers representation on the JVM
+
+On the JVM platform, numbers are stored as primitive types: `int`, `double`, and so on.
+Exceptions are cases when you create a nullable number reference such as `Int?` or use generics.
+In these cases numbers are boxed in Java classes `Integer`, `Double`, and so on.
+
+Note that nullable references to the same number can be different objects:
```
elizarov, line 110: Consider replacing `can be different objects` -> `can refer to different objects`

docs/topics/unsigned-types.md
```diff
@@ -0,0 +1,93 @@
...
+val yellow = Color(0xFFCC00CCu)
+```
+
+Or to initialize byte arrays without explicit `toByte()` literal casts:
+
+```kotlin
+val byteOrderMarkUtf8 = ubyteArrayOf(0xEFu, 0xBBu, 0xBFu)
+```
+
+Another yet use case is interoperability with native APIs. Kotlin allows representing native declarations that contain
```
elizarov, line 81: `another yet` -> `yet`

docs/topics/numbers.md
shadrina: Agree. Added a separate page for type inference and linked it in the paragraph above this snippet.

docs/topics/unsigned-types.md
```diff
@@ -0,0 +1,93 @@
+[//]: # (title: Unsigned types)
```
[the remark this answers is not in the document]
shadrina: Thanks, fixed.

docs/kr.tree
```diff
@@ -55,7 +55,14 @@
...
         <toc-element toc-title="Types">
-            <toc-element id="basic-types.md"/>
+            <toc-element toc-title="Basic types" id="basic-types.md">
+                <toc-element id="numbers.md"/>
+                <toc-element id="booleans.md"/>
+                <toc-element id="characters.md"/>
+                <toc-element id="strings.md"/>
+                <toc-element id="arrays.md"/>
+            </toc-element>
+            <toc-element id="unsigned-types.md"/>
```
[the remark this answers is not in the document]
shadrina: Added to the "Basic types" group

docs/topics/numbers.md
shadrina: It turned out the separate page for type inference doesn't fit in the current ToC. 

If we don't dive deep into the details about constraint systems, not so much left to say, so it needs to be grouped with smth. I tried to group it with the _Smart casts_ section (as in the specification), but smart casts are already described in the section about type casts. I tried to group it with our new article about builder inference but builder inference itself cannot be described before describing at least functions.

So I just changed the wording in the paragraph above to emphasize that the type is inferred. 
BTW, the feature itself is shortly mentioned in the _Basics_ section.

docs/topics/unsigned-types.md
```diff
@@ -0,0 +1,93 @@
...
+In addition to [integer types](numbers.md#integer-types), Kotlin provides the following types for unsigned integer numbers:
+
+* `UByte`: an unsigned 8-bit integer, ranges from 0 to 255
+* `UShort`: an unsigned 16-bit integer, ranges from 0 to 65535
+* `UInt`: an unsigned 32-bit integer, ranges from 0 to 2^32 - 1
+* `ULong`: an unsigned 64-bit integer, ranges from 0 to 2^64 - 1
+
+Unsigned types support most of the operations of their signed counterparts.
+
+>Changing type from unsigned type to signed counterpart (and vice versa) is a *binary incompatible* change.
```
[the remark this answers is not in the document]
shadrina: It is a note from the original version of this page, which was written by another technical writer. Do you want to remove it or clarify the connection between signed and unsigned integers?

docs/topics/generics.md
```diff
@@ -328,16 +328,109 @@ At runtime, the instances of generic types do not hold any information about the
...
+}
+```
+
+Similarly, when you already have the type arguments of an instance checked statically (at compile time),
+you can make an `is`-check or a cast that involves the non-generic part of the type. Note that
+angle brackets are omitted in this case:
+
+```kotlin
+fun handleStrings(list: List<String>) {
+    if (list is ArrayList) {
```
ilya-g, line 349: Funny that this is an example of a [type hole with variance](https://youtrack.jetbrains.com/issue/KT-7972) in Kotlin. It doesn't manifest itself with `String` type argument because it is final, but imagine if `list` was of `List<CharSequence>`: the check for `list is ArrayList` doesn't imply that that array list was created with `CharSequence` argument, therefore casting it to `ArrayList<CharSequence>` would be unsafe.

So I propose at least to declare  list as `MutableList<String>`, so that the example remains valid even when that bug is eventually fixed.

docs/topics/numbers.md
ilya-g: This is a specific type of inference that is inherent only to integer literals, so I think it's worth to be explained here.

docs/topics/typecasts.md
```diff
@@ -94,114 +94,3 @@ val x: String? = y as? String
 ```
 
 Note that despite the fact that the right-hand side of `as?` is a non-null type `String`, the result of the cast is nullable.
-
-## Type erasure and generic type checks
```
ilya-g, line 98: Makes sense to give a link to the new location of the moved text, so that the users who come here by an anchored link don't feel themselves lost.

docs/topics/unsigned-types.md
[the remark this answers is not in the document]
ilya-g: I propose to clarify, e.g. "While unsinged numbers are implemented as inline value classes with the single storage property of the corresponding singed counterpart type of the same width, changing type from unsigned type to signed counterpart ..."

docs/topics/basic-types.md
```diff
@@ -3,613 +3,5 @@
...
-// e.g. initialise the values in the array with a constant
-// Array of int of size 5 with values [42, 42, 42, 42, 42]
-val arr = IntArray(5) { 42 }
-
-// e.g. initialise the values in the array using a lambda
-// Array of int of size 5 with values [0, 1, 2, 3, 4] (values initialised to their index value)
-var arr = IntArray(5) { it * 1 } 
-```
+In this section we describe the basic types used in Kotlin: [numbers](numbers.md), [booleans](booleans.md), 
+[characters](characters.md), [strings](strings.md), and [arrays](arrays.md).
```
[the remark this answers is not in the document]
ilya-g: Although, they include advanced concepts such as arrays and ranges, so maybe it's ok to omit them from basic types list here. Those who are curious, could get there from ToC.

docs/topics/unsigned-types.md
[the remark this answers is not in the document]
shadrina: Thanks for the suggestion! [See the commit](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/739f6e14239d19a3ebf2e8e50bc3a8082752b425).

docs/topics/numbers.md
```diff
@@ -0,0 +1,289 @@
...
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+```
+
+## Floating-point types
+
+For real numbers, Kotlin provides floating-point types `Float` and `Double`.
+According to the [IEEE 754 standard](https://en.wikipedia.org/wiki/IEEE_754),
+floating point types differ by their _decimal place_, that is, how many decimal digits they can store.
+`Float` reflects the IEEE 754 _single precision_, while `Double` provides _double precision_.
```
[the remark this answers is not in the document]
shadrina: Thanks for the suggestion! [Fixed.](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/a3eb5242ca7a988b553bec7567bea85d2c8f309d)

docs/topics/basic-types.md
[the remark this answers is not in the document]
shadrina: I think it's better to add rather than to omit. We decided to place it inside the "Basic types" section, so why not include it here too. Thanks, [added](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/992bb5faa756409f778f1369c2fe0bd9eed4054b).

docs/topics/generics.md
shadrina: Thanks! [Fixed.](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/ba1930d6ca5035e9d44c667804d9ae473d1553ba)

docs/topics/typecasts.md
shadrina: Added a section with a link as we do in [some other pages](https://kotlinlang.org/docs/exceptions.html#java-interoperability). [Commit.](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/81bda1cf69b851bb954fa3ba6d3ac8fad552a06a)

docs/topics/numbers.md
shadrina: [Added a note about widening.](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/105a4721e029537fd031751c1c89f3bb917b9a9b) Does it work? I can also add a link to the corresponding section in the specification. WDYT?

docs/topics/unsigned-integer-types.md
```diff
@@ -9,7 +9,9 @@ In addition to [integer types](numbers.md#integer-types), Kotlin provides the fo
 
 Unsigned types support most of the operations of their signed counterparts.
 
->Changing type from unsigned type to signed counterpart (and vice versa) is a *binary incompatible* change.
+> Unsigned numbers are implemented as [inline classes](inline-classes.md) with the single storage property of the corresponding 
+> signed counterpart type of the same width. Thus, changing type from unsigned type to signed counterpart (and vice versa)
```
ilya-g, line 13: "Thus" doesn't seem logical here. I was proposing to connect these facts as a contrast.

shadrina: Now got it that this is the reason why someone could consider them compatible. Sorry, misunderstood. 
Does "Nevertheless" work here? [Commit.](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/297c3ff81fa91e1c1958044583a6516e0e7b13ba)

docs/topics/numbers.md
```diff
@@ -0,0 +1,295 @@
...
+For integer numbers, there are four types with different sizes and, hence, value ranges.
+
+| Type	 |Size (bits)| Min value| Max value|
+|--------|-----------|----------|--------- |
+| `Byte`	 | 8         |-128      |127       |
+| `Short`	 | 16        |-32768    |32767     |
+| `Int`	 | 32        |-2,147,483,648 (-2<sup>31</sup>)| 2,147,483,647 (2<sup>31</sup> - 1)|
+| `Long`	 | 64        |-9,223,372,036,854,775,808 (-2<sup>63</sup>)|9,223,372,036,854,775,807 (2<sup>63</sup> - 1)|
+
+If you initialize a variable with an integer value not exceeding the maximum value of `Int`, the compiler will
```
ilya-g, line 15: "not exceeding the range of" or something like that, because it could exceed (be greater than) max value or (be less than) min value.

docs/topics/numbers.md
```diff
@@ -0,0 +1,295 @@
...
+val one = 1 // Int
+val threeBillion = 3000000000 // Long
+val oneLong = 1L // Long
+val oneByte: Byte = 1
+```
+
+> Inside the compiler, there is a priority between built-in integer types. If there is no explicit type specification
+> (such as `val oneByte: Byte = 1`), an integer literal that is in the value range of multiple built-in integer types 
+> receives the type of the highest priority. Thus, the integer literal `1` in the first line of the example receives 
+> the `Int` type, although it can be either `Byte` or `Short`.
```
ilya-g, line 29: I would say it more shortly without mentioning internal priority in the compiler:
> If there is no explicit type specification and no explicit `L` suffix, the type is inferred automatically as `Int` or `Long` based on the value of an integer literal: the type with the smallest range enough to represent the value is chosen.

ilya-g: At this point it merely restates the text before the example in other words. Probably Roman proposed to say something about the situation when the expected type is specified explicitly? I.e. when the expected type is Byte, the literal becomes the literal of type Byte and is checked not to exceed the range of Byte type.

docs/topics/numbers.md
shadrina: Fixed

docs/topics/numbers.md
shadrina: Thanks, I removed the note and tried to compose all this information in the paragraph above.
Please see [the commit](https://github.com/JetBrains/kotlin-web-site/pull/2409/commits/8b4528422f9620a7fcc76c179dc780166ed29bbd).

shadrina:
@koshachy Andrey, I suppose the PR is ready for your review. Perhaps we'll change the details about type inference (see the unresolved thread above) but the rest is approved.

ilya-g (approved):

docs/topics/numbers.md
```diff
@@ -0,0 +1,289 @@
...
+
+All number types support conversions to other types:
+
+* `toByte(): Byte`
+* `toShort(): Short`
+* `toInt(): Int`
+* `toLong(): Long`
+* `toFloat(): Float`
+* `toDouble(): Double`
+* `toChar(): Char`
```
ilya-g, line 184: `toChar` is deprecated, so it makes sense to omit this one.

See https://youtrack.jetbrains.com/issue/KT-50859

shadrina: Thank you, I removed the item. The remaining update requested in the ticket will be in a separate PR.

koshachy:
I will proceed working with the PR in a week.

elizarov (approved):
@koshachy Ping

koshachy (approved):
Such an amazing rework.
Huge thanks to @shadrina for this!

PR #2410, opened 2021-07-16
fix: high order function definition

jlguenego:


jlguenego:
I proposed to align the high order function definition with the one of https://en.wikipedia.org/wiki/Higher-order_function

shadrina:
Hi! Thanks for the suggestion.

However, I suppose version "takes functions as parameters" is pretty close to "takes at least one function as parameter". At the same time, it is simpler and informative enough to understand the topic.

So let's leave it as is.

PR #2412, opened 2021-07-17
Fix extension method call in example

sandyiscool:
In the example code in section **Declaring extensions as members**, `printConnectionString()` does not accept any arguments.


p7nov:
Thank you for spotting this!

sandyiscool:
> Thank you for spotting this!

Sure, no problem!

PR #2413, opened 2021-07-17
Update Note on Visibilty section of extensions

sandyiscool:
`protected` members of the receiver type are not accessible either, to an extension method declared outside its receiver type


p7nov:
You're absolutely right, thanks for the addition!

PR #2419, opened 2021-07-23
fix: fix broken rendering of the list in gradle.md

PetrakovichVictoria:
See how it was: https://kotlinlang.org/docs/gradle.html#incremental-compilation

Before:
Screenshot 2021-07-23 at 17 18 22


p7nov (approved):
👍

PR #2454, opened 2021-08-12
KTL-32 Add "online" mark to Kotlin talks description on kotlinlang.org

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-32

adding `<online>true</online>` prop for an event in the events.xml add green badge to the list:

Screenshot 2021-08-12 at 17 36 56

### DIFF
--- a/data/cities.yml
+++ b/data/cities.yml
@@ -1489,11 +1489,11 @@
 - geo:
     lat: 35.7795897
     lng: -78.6381787
-  name: Raleigh, NC, US
+  name: Raleigh, NC, USA
 - geo:
     lat: 42.0333607
     lng: -88.0834059
-  name: Schaumburg, IL, US
+  name: Schaumburg, IL, USA
 - geo:
     lat: -26.2041028
     lng: 28.0473051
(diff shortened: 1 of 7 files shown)


data/cities.yml
```diff
@@ -1489,11 +1489,11 @@
 - geo:
     lat: 35.7795897
     lng: -78.6381787
-  name: Raleigh, NC, US
+  name: Raleigh, NC, USA
```
nikpachoo, line 1492: Most events in the states contain "USA" in their names, so I renamed the rest accordingly.

krutilov (approved):

static/js/page/events/map/Map.js
```diff
@@ -67,19 +67,24 @@ export default class Map {
     });
 
     emitter.on(EVENTS.EVENT_HIGHLIGHTED, (event) => {
+      if (!event.hasCity()) return;
```
zoobestik, line 70: may be we just don't be [emmit event](https://github.com/JetBrains/kotlin-web-site/blob/4604c9bb6623685b20121bbddae74a0d06252b2b/static/js/page/events/events-list/EventsList.js#L73) for this online events?

nikpachoo: There are two listeners subscribed to the "EVENT_HIGHLIGHTED" event: a list and a map. In the list, the event is used for the related item to highlight it. But not on the map. So I had to make the condition in the map listener.

PR #2457, opened 2021-08-13
feat: New default publishing setup for Android artifacts

shadrina:


docs/topics/mpp/mpp-publish-lib.md
```diff
@@ -122,40 +122,14 @@ kotlin {
...
-> ```
->
-></tabs>
->
-{type="note"}
-
-Similarly, a library consumer needs to provide matching fallbacks for custom product flavors if some are missing in the 
-library publications.
+The default publishing setup is the following:
+* If all Android variants that the project publishes have the same build type attribute, then the published variants
```
h0tk3y, line 126: I'm not sure that mentioning the attribute values here makes things clearer, as attributes are a lower-level detail of dependency resolution than an ordinary build author usually cares about.

Maybe we can somehow reformulate this. Instead of describing the attribute values we can tell about compatibility rules, like:

> If all published Android variants have the same build type (`debug`, `release` etc.), they will be compatible with any consumer's build type.
> If the published variants have different build types, then the `release` variants will also be compatible with any consumer's build type that is absent among the published variants, while all other variants (like `debug`) will only match the same build type on the consumer side, unless the consumer project specifies the matching fallbacks. 

So the part about matching fallbacks might still be relevant for the second case. However, matching fallbacks are no longer necessary specifically for MPP, so maybe a link to the AGP docs would be enough.

shadrina: Thank you! Just a little bit modified your option and changed the wording in the last sentence on how to opt out. Please take a look.

h0tk3y: Looks good to me. 👍

h0tk3y (approved):

PR #2460, opened 2021-08-17
Replace dependency on Node Buffer API

erikhuizinga:
This fixes an issue: https://kotlinlang.slack.com/archives/C0B8L3U69/p1628876549226400

The issue is likely cause by the project being created by IDEA, which produces

```
js(LEGACY) {
    browser {
        commonWebpackConfig {
            cssSupport.enabled = true
        }
    }
}
```

in build.gradle.kts. Note that this uses a `browser` instead of `nodejs` target. Because of this, running the tests as described in the document (before this commit) results in failure:

```
ReferenceError: Buffer is not defined
```

This commit fixes that issue by replacing the used Node API for a browser API: `btoa`.

### DIFF
--- a/docs/topics/mpp/multiplatform-library.md
+++ b/docs/topics/mpp/multiplatform-library.md
@@ -7,7 +7,7 @@ This library converts raw data – strings and byte arrays – to the [Base64](h
 You will use different ways to implement the conversion to the Base64 format on different platforms:
 
 * For JVM – the [`java.util.Base64` class](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html).
-* For JS – the [`WindowOrWorkerGlobalScope.btoa()` API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/btoa).
+* For JS – the [`btoa()` function](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/btoa).
 * For Kotlin/Native – your own implementation.
 
 You will also test your code using common tests, and then publish the library to your local Maven repository.
@@ -100,7 +100,7 @@ The JS implementation will be very similar to the JVM one.
 
 1. In the `jsMain/kotlin` directory, create the `org.jetbrains.base64` package.
 2. Create the `Base64.kt` file in the new package.
-3. Provide a simple implementation of the `Base64Factory` object that delegates to the `WindowOrWorkerGlobalScope.btoa()` API:
+3. Provide a simple implementation of the `Base64Factory` object that delegates to the `btoa()` function.
 
     ```kotlin
     package org.jetbrains.base64
--- a/docs/topics/mpp/multiplatform-library.md
+++ b/docs/topics/mpp/multiplatform-library.md
@@ -7,7 +7,7 @@ This library converts raw data – strings and byte arrays – to the [Base64](h
 You will use different ways to implement the conversion to the Base64 format on different platforms:
 
 * For JVM – the [`java.util.Base64` class](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html).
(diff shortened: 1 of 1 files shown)


docs/topics/mpp/multiplatform-library.md
```diff
@@ -7,7 +7,7 @@ This library converts raw data – strings and byte arrays – to the [Base64](h
 You will use different ways to implement the conversion to the Base64 format on different platforms:
 
 * For JVM – the [`java.util.Base64` class](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html).
-* For JS – the [Buffer API](https://nodejs.org/docs/latest/api/buffer.html).
+* For JS – the [`WindowOrWorkerGlobalScope.btoa()` API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/btoa).
```
SebastianAigner, line 10: Since the function is available on the global object (that's essentially what `WindowOrWorkerGlobalScope` expresses), I'd prefer we just call it the "btoa() function"

docs/topics/mpp/multiplatform-library.md
```diff
@@ -100,7 +100,7 @@ The JS implementation will be very similar to the JVM one.
 
 1. In the `jsMain/kotlin` directory, create the `org.jetbrains.base64` package.
 2. Create the `Base64.kt` file in the new package.
-3. Provide a simple implementation of the `Base64Factory` object that delegates to the NodeJS `Buffer` API:
+3. Provide a simple implementation of the `Base64Factory` object that delegates to the `WindowOrWorkerGlobalScope.btoa()` API:
```
SebastianAigner, line 103: Again, I suggest simplifying to "the `btoa()` function"

docs/topics/mpp/multiplatform-library.md
```diff
@@ -111,9 +111,9 @@ The JS implementation will be very similar to the JVM one.
     
     object JsBase64Encoder : Base64Encoder {
         override fun encode(src: ByteArray): ByteArray {
-            val buffer = js("Buffer").from(src)
-            val string = buffer.toString("base64") as String
-            return ByteArray(string.length) { string[it].toByte() }
+            val string = src.decodeToString()
+            val encodedString = js("btoa(string)") as String
```
SebastianAigner, line 115: This can be rewritten without the need for a `js()` invocation:

```kotlin
import kotlinx.browser.window

override fun encode(src: ByteArray): ByteArray {
    val string = window.btoa(src.decodeToString())
    return string.encodeToByteArray()
}
```

SebastianAigner (changes requested):
Thanks for making this suggestion! Indeed, this is a good catch. The refactored solution and explanation can still be cleaned up a bit, please check my comments.

docs/topics/mpp/multiplatform-library.md
erikhuizinga: I've accepted your suggestion, with a minor difference: I've kept the three lines from my original logic. I believe that it would help the beginner Kotlin/JS user, because it clearly shows the three steps the code takes:

1. Convert to the original string.
2. Encode the string using `btoa`.
3. Encode the encoded string as a byte array.

To me it seems that this clearly separates what parts are from the Kotlin stdlib and what from the Kotlin/JS. Let me know what you think! I'd be happy to accept your code as suggested, or just commit it yourself to my PR.

SebastianAigner: Thanks, I like your approach, and your reasoning seems sound. This looks good to me.

SebastianAigner (approved):
Thanks for incorporating these changes. This is clearer and correcter than before! We appreciate the contributions! Ready to merge from my end.

p7nov:
Thanks for the update, @erikhuizinga!

PR #2462, opened 2021-08-18
feature: more iOS-authentic way to get objects and companion objects

koshachy:


docs/topics/native/native-objc-interop.md
```diff
@@ -173,8 +173,31 @@ with an additional receiver parameter. These types include:
 
 Kotlin singleton (made with an `object` declaration, including `companion object`)
 is imported to Swift/Objective-C as a class with a single instance.
-The instance is available through the factory method, i.e. as
-`[MySingleton mySingleton]` in Objective-C and `MySingleton()` in Swift.
+
+The instance is available through the `shared` and `companion` properties.
```
SvyatoslavScherbina, line 177: The same should be reflected in the table at the beginning of the document.

docs/topics/native/native-objc-interop.md
```diff
@@ -173,8 +173,31 @@ with an additional receiver parameter. These types include:
...
+}
+
+class MyClass {
+    companion object {
+        val x = "Some value"
+    }
+}
+```
+
+Get the objects as follows:
```
SvyatoslavScherbina, line 193: Replacing `get` by `access` might make this more understandable.

docs/topics/native/native-objc-interop.md
```diff
@@ -173,8 +173,31 @@ with an additional receiver parameter. These types include:
...
+```
+
+Get the objects as follows: 
+
+```swift
+MyObject.shared
+MyObject.shared.x
+MyClass.companion
+MyClass.Companion.shared
+```
```
SvyatoslavScherbina, line 200: Btw, the old variant (`[MyObject myObject]` in Objective-C and `MyObject()` in Swift) still works.
What do you think, does it make sense to keep it in the documentation (as an alternative or "deprecated" way)?

docs/topics/native/native-objc-interop.md
koshachy: I've updated the table,
please, take a look

docs/topics/native/native-objc-interop.md
koshachy: I've returned it back

docs/topics/native/native-objc-interop.md
```diff
@@ -32,7 +32,7 @@ The table below shows how Kotlin concepts are mapped to Swift/Objective-C and vi
 | `suspend` -> | `completionHandler:`/ `async` | `completionHandler:` | [note](#errors-and-exceptions) [note](#suspending-functions) |
 | `@Throws fun` | `throws` | `error:(NSError**)error` | [note](#errors-and-exceptions) |
 | Extension | Extension | Category member | [note](#extensions-and-category-members) |
-| `companion` member <- | Class method or property | Class method or property |  |
+| `companion` member <- | `shared` or `companion` property | `shared` or `companion` property | [note](#kotlin-singletons)  |
```
SvyatoslavScherbina, line 35: This row of the table described how Objective-C class methods and properties are imported to Kotlin -- as companion members. It didn't change, and the new version is incorrect. Additionally, the link to `#kotlin-singletons` is unrelated.

So please update the row below instead:
```
| `Singleton` | `Singleton()` | `[Singleton singleton]` | [note](#kotlin-singletons) |
```

SvyatoslavScherbina (approved):

PR #2475, opened 2021-08-23
KTL-504 Kotlinlang - update PostCSS version

nikpachoo:
Issue: [KTL-504](https://youtrack.jetbrains.com/issue/KTL-504)


postcss.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   plugins: [
-    require('autoprefixer')({ browsers: ['last 2 versions'] }),
+    require('autoprefixer'),
```
nikpachoo, line 3: We have a browser list in the package.json instead.

webpack.config.js
```diff
@@ -71,10 +71,7 @@ module.exports = (params = {}) => {
               }
             },
             {
-              loader: 'resolve-url-loader',
-              options: {
-                keepQuery: true
-              }
+              loader: 'resolve-url-loader'
```
nikpachoo, line 74: The keepQuery behaviour is now the default, the keepQuery option has been removed.

zoobestik (approved):

PR #2485, opened 2021-08-25
Fixes and clarifications regarding the use of `vararg` usage with primitive type arrays

AzimMuradov:
Add additional usage information for `vararg` with primitive type arrays
Fix possible documentation bug related to `vararg` and primitive type arrays


p7nov:
Thanks for this update!
I've adjusted the text to our docs style a bit.

PR #2496, opened 2021-08-30
Update idioms.md

ShoaibKakal:
Added joinToString method.


PetrakovichVictoria:
Hi @ShoaibKakal! Thanks for your suggestion. However, we have to decline it because we show the joinToString method usage in [Migrating from Java to Kotlin: Strings](https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#create-a-string-from-collection-items).

PR #2497, opened 2021-08-31
fix: kt-48004 sealed hierarchy example

p7nov:
Provide real-life use case for sealed hierarchies as per request in:
https://youtrack.jetbrains.com/issue/KT-48004
https://youtrack.jetbrains.com/issue/KT-48242

UPD: clarify rules of indirect inheritance: https://youtrack.jetbrains.com/issue/KT-48542

### DIFF
--- a/docs/topics/sealed-classes.md
+++ b/docs/topics/sealed-classes.md
@@ -12,16 +12,23 @@ In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes: 
 for an enum type is also restricted, but each enum constant exists only as a _single instance_, whereas a subclass
 of a sealed class can have _multiple_ instances, each with its own state.
 
+As an example, consider a hierarchy of possible errors in a library. There can be classes for networking, database, and
+other errors, as well as code that uses them, such as handling and logging.
+If a root interface of the error hierarchy is exposed in the public API, client classes can inherit it, creating
+new error types. They are declared outside the library, so it cannot handle them in specific ways. With a sealed hierarchy,
+library authors can be sure that they know all possible error types and no other ones can appear.
+
 To declare a sealed class or interface, put the `sealed` modifier before its name:
 
 ```kotlin
-sealed interface Expr
+sealed interface Error
 
-sealed class MathExpr(): Expr
+sealed class IOError(): Error
 
-data class Const(val number: Double) : MathExpr()
-data class Sum(val e1: Expr, val e2: Expr) : MathExpr()
-object NotANumber : Expr
+class FileReadError(val f: File): IOError()
+class DatabaseError(val source: DataSource): IOError()
+
+object RuntimeError : Error
 ```
 
 A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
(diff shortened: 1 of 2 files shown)


demiurg906 (approved):

docs/topics/sealed-classes.md
```diff
@@ -50,6 +56,16 @@ Subclasses of sealed classes must have a proper qualified name. They can't be lo
 >
 {type="note"}
 
+These restrictions don't apply to indirect subclasses. If a direct subclass of a sealed class is not marked as sealed,
+it can be extended in any ways that its modifiers allow.
```
koshachy, line 60:
```suggestion
it can be extended in any ways that its modifiers allow:
```

docs/topics/sealed-classes.md
```diff
@@ -69,10 +85,10 @@ If it's possible to verify that the statement covers all cases, you don't need t
...
 ```kotlin
-fun eval(expr: Expr): Double = when(expr) {
-    is Const -> expr.number
-    is Sum -> eval(expr.e1) + eval(expr.e2)
-    NotANumber -> Double.NaN
+fun log(e: Error) = when(e) {
+    is FileReadError -> { println("Error while reading file ${e.file}") }
+    is DatabaseError -> { println("Error while reading from database ${e.source}") }
+    RuntimeError ->  { println("Runtime error") }
     // the `else` clause is not required because we've covered all the cases
```
koshachy, line 92:
```suggestion
    // the `else` clause is not required because all the cases are covered
```

docs/topics/sealed-classes.md
```diff
@@ -12,16 +12,23 @@ In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes:
 for an enum type is also restricted, but each enum constant exists only as a _single instance_, whereas a subclass
 of a sealed class can have _multiple_ instances, each with its own state.
 
+As an example, consider a hierarchy of possible errors in a library. There can be classes for networking, database, and
```
koshachy, line 15: It's a bit confusing: classes for ..., ..., and other errors — these entities don't have the same meaning, IDK can we enumerate them in that way?

docs/topics/sealed-classes.md
```diff
@@ -12,16 +12,23 @@ In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes:
...
+other errors, as well as code that uses them, such as handling and logging.
+If a root interface of the error hierarchy is exposed in the public API, client classes can inherit it, creating
+new error types. They are declared outside the library, so it cannot handle them in specific ways. With a sealed hierarchy,
```
koshachy, line 18: `it` is too ambiguous here

docs/topics/sealed-classes.md
```diff
@@ -12,16 +12,23 @@ In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes:
...
+library authors can be sure that they know all possible error types and no other ones can appear.
```
koshachy, line 19: To me, this whole text paragraph is too complicated and inconsequent. Let's try to rewrite it?

koshachy (changes requested):
As I also wrote — it's a bit hard to understand the new para.
I don't have a solution for it now, but we could discuss it later.

docs/topics/sealed-classes.md
p7nov: Tried =) Please check.

koshachy (approved):

PR #2507, opened 2021-09-05
Remove dataclass comment about generating setters

mhansen:
The code below doesn't actually generate setters, because it uses `val`:
https://pl.kotl.in/puHlJh3AI

If I try to use setters: 

```
    val c = Customer("Alice", "alice@bob.com", "Jetbrains")
    c.setName("Bob")
    c.name = "Bob"
```

I get errors:

```
Unresolved reference: setName
Val cannot be reassigned
```

### DIFF
--- a/kotlin-features/concise.md
+++ b/kotlin-features/concise.md
@@ -1,7 +1,7 @@
 <div class="sample" markdown="1" mode="kotlin" theme="idea" data-highlight-only="1" auto-indent="false">
 ```kotlin
 /*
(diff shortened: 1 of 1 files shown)


koshachy:
@nikpachoo I guess the new examples that are on the way will fix this problem.
Take a look

nikpachoo (approved):

nikpachoo:
@koshachy you are right, the new examples will fix this problem, but I'm not sure that we can update examples on this week.
I would like to accept @mhansen changes.

PR #2508, opened 2021-09-06
feat: add synchronization for Kotlin/Groovy tabs

p7nov:


docs/topics/configure-build-for-eap.md
```diff
@@ -74,20 +79,25 @@ Here is an example.
...
 ```
 
-```kotlin
+</tab>
+
+```groovy
 dependencies {
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:%coroutinesEapVersion%")
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:%coroutinesEapVersion%"
```
shadrina, line 96: Minor. Seems like we usually use single-quoted strings in Groovy samples. Here are no templates inside, so maybe it is worth changing.

docs/topics/gradle.md
```diff
@@ -36,40 +41,50 @@ require the `kotlin-multiplatform` plugin. [Learn more about the plugin](mpp-dis
...
 ```
 
-```kotlin
+</tab>
+
+```groovy
 plugins {
-    kotlin("jvm") version "%kotlinVersion%"
+    id "org.jetbrains.kotlin.jvm" version "%kotlinVersion%"
```
shadrina, line 83: The same here

shadrina (approved):

docs/topics/configure-build-for-eap.md
shadrina: I see double quotes were already there, so it is kind of out of this PR context :)

p7nov: Yes, diffs look weird here because I've swapped Groovy and Kotlin code.

PR #2514, opened 2021-09-09
Change parseInt (deprecated) to .toInt()

kevinpita:
Following [api](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/parse-int.html) parseInt is deprecated and .toInt() should be used in place


p7nov:
Thanks for the update!

PR #2517, opened 2021-09-09
feature: add redirects for KMM portal

koshachy:


redirects/kmm.yml
```diff
@@ -0,0 +1,66 @@
+- from:
+   - /docs/mobile/home.html
```
zoobestik, line 2: index pages like ‘/docs/mobile’ required roo.

koshachy: I've added `/docs/mobile`
should I add `/docs/mobile/` too (with a slash sign)?

PR #2520, opened 2021-09-10
fix: internal modifiers retain original visibility when overriden

danil-pavlov:


docs/topics/visibility-modifiers.md
```diff
@@ -57,16 +57,16 @@ For members declared inside a class:
 >
 {type="note"}
 
-If you override a `protected` member and do not specify the visibility explicitly, the overriding member will also have
-`protected` visibility.
+If you override a `protected` or a `private` member and do not specify the visibility explicitly, the overriding member
```
elizarov, line 60: You cannot override a `private` method. You can override `public`, `protected`, and `internal`.

docs/topics/visibility-modifiers.md
```diff
@@ -80,6 +80,7 @@ class Subclass : Outer() {
     // Nested and e are visible
 
     override val b = 5   // 'b' is protected
+    override val c = 7   // 'c' is private
```
elizarov, line 83: internal

elizarov (changes requested):

docs/topics/visibility-modifiers.md
danil-pavlov: Sorry, mixed up the names. Fixed

elizarov (approved):

PR #2526, opened 2021-09-15
update: different minor changes

PetrakovichVictoria:
Some edits regarding collections.

### DIFF
--- a/docs/topics/collection-aggregate.md
+++ b/docs/topics/collection-aggregate.md
@@ -77,8 +77,8 @@ fun main() {
 //sampleStart
     val numbers = listOf(5, 2, 10, 4)
 
-    val sum = numbers.reduce { sum, element -> sum + element }
-    println(sum)
+    val simpleSum = numbers.reduce { sum, element -> sum + element }
+    println(simpleSum)
     val sumDoubled = numbers.fold(0) { sum, element -> sum + element * 2 }
     println(sumDoubled)
 
--- a/docs/topics/collection-transformations.md
+++ b/docs/topics/collection-transformations.md
@@ -144,7 +144,7 @@ fun main() {
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 For building maps with collection elements as values, there is the function [`associateBy()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate-by.html).
-It takes a function that returns a key based on an element's value. If two elements are equal, only the last one remains
+It takes a function that returns a key based on an element's value. If two elements' keys are equal, only the last one remains
 in the map.
 
 `associateBy()` can also be called with a value transformation function.
--- a/docs/topics/collections-overview.md
+++ b/docs/topics/collections-overview.md
@@ -46,7 +46,7 @@ Although, if you try to reassign a `val` collection, you'll get a compilation er
 fun main() {
 //sampleStart
     val numbers = mutableListOf("one", "two", "three", "four")
-    numbers.add("five")   // this is OK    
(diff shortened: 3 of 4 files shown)


docs/topics/collection-aggregate.md
```diff
@@ -77,8 +77,8 @@ fun main() {
 //sampleStart
     val numbers = listOf(5, 2, 10, 4)
 
-    val sum = numbers.reduce { sum, element -> sum + element }
-    println(sum)
+    val simpleSum = numbers.reduce { sum, element -> sum + element }
```
p7nov, line 80: Sorry, I don't see a point in this change. It's just a usual sum, no need to complicate.

docs/topics/collection-transformations.md
```diff
@@ -144,7 +144,7 @@ fun main() {
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 For building maps with collection elements as values, there is the function [`associateBy()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate-by.html).
-It takes a function that returns a key based on an element's value. If two elements are equal, only the last one remains
+It takes a function that returns a key based on an element's value. If two elements' keys are equal, only the last one remains
```
p7nov, line 147: Good point, thanks!

p7nov: Perhaps we can show this case in the example as well. Just an idea, I don't insist.

docs/topics/collections-overview.md
```diff
@@ -46,7 +46,7 @@ Although, if you try to reassign a `val` collection, you'll get a compilation er
 fun main() {
 //sampleStart
     val numbers = mutableListOf("one", "two", "three", "four")
-    numbers.add("five")   // this is OK    
+    println(numbers.add("five"))   // this is OK
```
p7nov, line 49: This prints `true`, not a really demonstrative output.
If you want to show the change, print the whole collection in a separate call of println

p7nov (commented):
Thanks for the update, although some changes don't seem obvious; see my comments.

docs/topics/collection-aggregate.md
PetrakovichVictoria: This is to avoid the warning when launching it:

Screenshot 2021-09-16 at 14 42 51

docs/topics/collections-overview.md
PetrakovichVictoria: I added this printing for the purpose that if launching this example, then there's some result, but not only the progress and then nothing. I suppose we should either make such examples with no output not launchable or, if to leave launchable, to print something.
We can print the whole collection, I'll ad this.

docs/topics/collection-aggregate.md
p7nov: Good catch, I haven't checked this.

docs/topics/collection-transformations.md
PetrakovichVictoria: Good idea, but maybe later) Could you approve this PR, please?

p7nov (approved):
👍

docs/topics/collections-overview.md
```diff
@@ -46,10 +46,11 @@ Although, if you try to reassign a `val` collection, you'll get a compilation er
 fun main() {
 //sampleStart
     val numbers = mutableListOf("one", "two", "three", "four")
-    numbers.add("five")   // this is OK    
-    println(numbers)   // this is OK
```
PetrakovichVictoria: Oops, thank you, Pasha!😂

PR #2527, opened 2021-09-15
update: verbose logging works only when Gradle debug log enabled

PetrakovichVictoria:


docs/topics/gradle.md
```diff
@@ -662,7 +662,7 @@ Here is a complete list of options for Gradle tasks:
 |------|-------------|-----------------|--------------|
 | `allWarningsAsErrors` | Report an error if there are any warnings |  | false |
 | `suppressWarnings` | Don't generate warnings |  | false |
-| `verbose` | Enable verbose logging output |  | false |
+| `verbose` | Enable verbose logging output; works only when the Gradle debug log level enabled: `./gradlew <YOUR_TASKS> --debug` |  | false |
```
PetrakovichVictoria, line 665: For tech writers: I'm not sure whether we should write the command `./gradlew ...` this way because we never mention such launches on this page. However, I feel we should express clear where to use `--debug`. Maybe we should add some commands like `gradle build` or so on. I'm surprised that there are no such commands on this page, because it's some kind of basis. It can be found it Gradle docs, however, maybe we should add some examples too.

Tapchicoma (approved):

Tapchicoma (approved):

docs/topics/gradle.md
```diff
@@ -662,7 +662,7 @@ Here is a complete list of options for Gradle tasks:
 |------|-------------|-----------------|--------------|
 | `allWarningsAsErrors` | Report an error if there are any warnings |  | false |
 | `suppressWarnings` | Don't generate warnings |  | false |
-| `verbose` | Enable verbose logging output |  | false |
+| `verbose` | Enable verbose logging output; works only when the [Gradle debug log level enabled](https://docs.gradle.org/current/userguide/logging.html) |  | false |
```
koshachy, line 665: Why did you decide not to write a separate sentence?

PetrakovichVictoria, line 665:
```suggestion
| `verbose` | Enable verbose logging output. Works only when the [Gradle debug log level enabled](https://docs.gradle.org/current/userguide/logging.html) |  | false |
```

koshachy (approved):

docs/topics/gradle.md
PetrakovichVictoria: Discussed in DM.

PR #2545, opened 2021-10-01
Clarify the navigation to code conventions

mira01:
As there is no "Settings" section on MacOS, mention that on this operating system people should go to the Preferences instead


p7nov:
Hi! 
This is a good addition, thank you.
I've adjusted the exact look to be consistent with the IntelliJ IDEA docs – they write `Settings/Preferences`([example](https://www.jetbrains.com/help/idea/code-style-vue.html)).

PR #2546, opened 2021-10-01
MutableLists should be stored in vals, not vars

lowasser:
Mutation to MutableLists should be done not by reassigning the property, but by modifying the list.  Otherwise, you can get nasty ambiguity errors with e.g. +=.


p7nov:
Good catch, thank you for spotting!

PR #2547, opened 2021-10-03
Update collection-aggregate.md

homchom:
Update collection-aggregate.md to fix the incorrectly documented relationship between "by" max/min functions (`maxByOrNull`) and "of" max/min functions (`maxOfOrNull`). The current documentation says that `maxByOrNull` and `maxOf` are the same except for how they handle empty collections. I believe this is incorrect because `maxOf` also returns a selector return value, not a collection element. In addition, the current documentation does not mention some functions, such as `maxOfOrNull` and `maxOfWithOrNull`.

This is my first time contributing to the docs; please let me know if I made a mistake or am misunderstanding something.


homchom:
Update: Apparently, maxBy and minBy are deprecated, with a message saying to use maxByOrNull and minByOrNull instead. My commit references them, which is a mistake. However, now I am confused; maxOf and minOf are not deprecated, but they also have maxOfOrNull and minOfOrNull counterparts. Are there plans to deprecate maxOf and minOf as well? And why exactly were maxBy and minBy deprecated? I can make another commit if needed to remove mentions of maxBy and minBy, but I want to make sure I understand the situation fully before doing so.

p7nov:
Hi @homchom!

Thank you for spotting this.
Here is the situation: `max`/`min` and `maxBy`/`minBy` were introduced in early versions of stdlib. Now their behavior on empty collections doesn't fit into the modern naming scheme of stdlib: they return `null` on empty collections but don't have the `orNull` suffix. This is why we deprecated them in favor of their new `orNull` counterparts. Later we'll remove these functions and reintroduce them with the correct behavior after the deprecation cycle. You can read about this in [this blog post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-is-out-standard-library-changes/).

Bottom line: 
- you're completely right that `maxByOrNull` and `maxOf` aren't alternatives 👍 
- adding `maxOfOrNull()` and `maxOfWithOrNull()` is correct 👍 
- in the paragraph about alternatives we need to list only relevant (non-deprecated) functions. 

You can update the PR yourself or I can help with it. 
Thanks again!

p7nov (changes requested):
Please see my comment in the conversation.

homchom:
Thanks for the response! That makes a lot of sense; I see now.

As for your requested change, I am not sure I understand. In my later commits (31 Oct), I believe I removed references to the deprecated functions in the alternatives paragraph. Did I miss some, or am I misunderstanding? If you want, you can go ahead and just make any changes you want.

p7nov:
Ah, sorry, I also got confused with all these functions. `maxOf/minOf` are fine; I thought they were deprecated too.
Now everything is fine, I've just updated the formatting a bit.

p7nov (approved):

PR #2549, opened 2021-10-04
Update contribute.md

Rocksnake:
I found a link error in 'contribute.html' about documentation source code, and I think it should be changed.


p7nov:
Yes, you're completely right, this URL is outdated. Thank you for fixing it!
I've added `/topics` at the end of the new URL to ease the navigation to actual MD files.

PR #2560, opened 2021-10-09
docs: Add an observation about the copy() function

lrmiguel:
Since in line 56 is declared that properties outside the primary constructor are not being used in the `copy()` implementation, it is clearer if we add a disclaimer in the docs.

### DIFF
--- a/docs/topics/data-classes.md
+++ b/docs/topics/data-classes.md
@@ -77,7 +77,7 @@ fun main() {
 
 ## Copying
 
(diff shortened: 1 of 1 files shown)


p7nov:
Hi!
Thanks for the PR.
In the paragraph you want to modify, we refer to the `User` class with `age` already in the primary constructor, not the `Person` class with `age` declared in body. So the current wording is correct.

PR #2564, opened 2021-10-11
feat: Constructing collections with builder functions

danil-pavlov:


ilya-g:
There are still many unrelated commits in this branch. Please try rebasing it on top of the target branch: `1-6-doc-update`.
Currently it seems to be merged with that branch, and not rebased onto it.

docs/topics/constructing-collections.md
```diff
@@ -31,6 +31,24 @@ initialization fluent here.
 val numbersMap = mutableMapOf<String, String>().apply { this["one"] = "1"; this["two"] = "2" }
 ```
 
+## Create with collection builder functions
+
+Another way of creating a collection is to call a builder function –
+[`buildList()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-list.html), [`buildSet()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-set.html),
+or [`buildMap()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-map.html). They create a new
```
koshachy, line 38: I feel awkward whenever I use pronouns for such objects — functions, classes, etc.
I prefer to use "These functions create". But maybe it looks redundant. 
WDYT?

docs/topics/constructing-collections.md
```diff
@@ -31,6 +31,24 @@ initialization fluent here.
...
+mutable collection of the corresponding type, populate it using [write operations](collection-write.md),
+and return a read-only collection with the same elements.
```
koshachy, line 40: It's better to use a colon if the following code block shows an example.

koshachy (approved):
Looks good.
Start a discussion if you have any questions or comments.

docs/topics/constructing-collections.md
danil-pavlov: I think it's OK to use such pronouns even with inanimate objects. It looks simpler and more colloquial
I can double-check with copyeditors

docs/topics/constructing-collections.md
danil-pavlov: Thanks, fixed

PR #2568, opened 2021-10-12
Compatibility guide for Kotlin 1.6

ilya-g:


ilya-g:
@qurbonzoda 
> CharSequence.isBlank(), String.compareTo(), String?.equals(ignoreCase = true) -- behavioral

IMO, too minor to be mentioned and the current stability mode doesn't require to do it

> AbstractMutableCollection.toJSON(), open fun HashMap.createEntrySet() -- source

Makes sense to add, is there an issue?

qurbonzoda:
@ilya-g 
> Makes sense to add, is there an issue?

No dedicated issue for them alone, but there is an issue where the changes are mentioned: https://youtrack.jetbrains.com/issue/KT-48587

PR #2597, opened 2021-10-20
Fix "if expression" idiom example returning Unit

stewSquared:
`foo` is composed of a single assignment statement, and the `result` value is never used.

I think the intent was to demonstrate with a single-expression function.

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -225,14 +225,13 @@ fun test() {
 ## if expression
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


p7nov:
Hey!
Thank you for this PR.
I think we can actually remove the function declaration to keep the focus on `if`. A variable assignment is shorter and demonstrative enough (see my commit)

Thanks again!

stewSquared:
Makes sense! Thanks 😄

PR #2618, opened 2021-10-29
chore(deps): bump path-parse from 1.0.6 to 1.0.7 in /scripts/react-renderer

Chinay-Domitrix:


nikpachoo:
Thank you for your efforts! We have removed package.json from this directory, so there is no need to fix this.

PR #2621, opened 2021-10-31
feat: Add documentation on builder inference

shadrina:


docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,253 @@
+[//]: # (title: Using builders with builder type inference)
+
+Kotlin supports _builder type inference_ (or builder inference), which is a type inference flavour that comes in handy 
+when writing generic builders. It helps the compiler infer the type arguments of a builder call based on the type information
```
petukhovv, line 4: > when writing generic builders

Actually, it comes in handy when _using_ generic builders. BI works at call site.

petukhovv: Maybe it's possible to use something common like "when working with generic builders"

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,253 @@
...
+1. It should use the type arguments that builder inference is supposed to infer. For example:
+   ```kotlin
+   fun <V> buildList(builder: MutableList<V>.() -> Unit) { ... }
+   ```
+   
+   > Note that passing type parameter's type directly like `fun <T> myBuilder(builder: T.() -> Unit)` is not yet supported.
+   > 
+   {type="note"}
+
+2. It should provide public members that contain the corresponding type parameters in their signature. For example:
```
petukhovv, line 49: Sorry, it seems I haven't mentioned it in the internal doc, but extensions are also possible to use:
```
class ItemHolder<T> {}

fun <T> ItemHolder<T>.foo(x: T) {}

fun <T> itemHolderBuilder(builder: ItemHolder<T>.() -> Unit): ItemHolder<T> =
    ItemHolder<T>().apply(builder)

fun test(s: String) {
    val itemHolder1 = itemHolderBuilder { // Type of itemHolder1 is ItemHolder<String>
        foo(s)
    }
}
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,253 @@
...
+fun someMap() = mutableMapOf<CharSequence, String>()
+
+fun <E> MutableMap<E, String>.f(x: MutableMap<E, String>) { ... }
+
+fun main() {
+    val x: Map<in String, String> = buildMap {
+        put("", "")
+        f(someMap()) // Type mismatch (required String, found CharSequence)
+        // Type mismatch appears because builder inference didn't work here,
+        // the expected Map<in String, String> type is specified outside a builder lambda
```
petukhovv, line 249: For clarity, we can add "...and builder lambda is being analysed with already fixed receiver's type"

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,253 @@
...
+    val y: String = x
+} // result has the List<String> type inferred
+```
+
+After the postponed type variable gets assigned to a variable of the `String` type, builder inference can infer `E` into
+`String`.
+
+Note that you can always call `equals()`, `hashCode()`, and `toString()` functions with a postponed type variable as a 
+receiver. Once the postponed type variable gets associated with some concrete type, you can use it as a call argument, 
+call available extensions, and access available properties.
```
petukhovv, line 158: > Once the postponed type variable gets associated with some concrete type, you can use it as a call argument, 
call available extensions, and access available properties.

Haven't got it. Inside a builder lambda, we have _always_ not fixed yet type variable (postponed type variable). What is the difference between the state "associated with some concrete type" and "not associated yet"? Associating is an adding constraints technically, but not variable fixation. We fix postponed variables only after _full_ analysis of a builder lambda. States "added some constraints" and "there aren't any added constraints yet" actually the same.

shadrina: Thank you, I indeed got it wrong. Changed the wording in multiple places of this section. [Please take a look at the last commit with review fixes.](https://github.com/JetBrains/kotlin-web-site/pull/2621/commits/f4e87dcde0888c7a1cfff7f69a22dc4cb13ed26e) I removed the sentence you cited here because now I get that it's about contributing type information, which we describe in the next section.

docs/topics/using-builders-with-builder-inference.md
shadrina: Yeap, changed to "when working with generic builders"

docs/topics/using-builders-with-builder-inference.md
shadrina: [See the commit.](https://github.com/JetBrains/kotlin-web-site/pull/2621/commits/63cc850b733dd2fccf8b250c9064ce0bf3628b49)

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,277 @@
...
+    // Considering E is Number or a subtype of Number
+    val n: Number? = getOrNull(0)
+    // Considering E is Int or a supertype of Int
+    add(1)
+    // E gets inferred into Int
+} // result has the List<Int> type
+```
+
+So the resulting type is the most specific type that corresponds to the type information collected during the analysis.
+If given type information is contradictory and cannot be merged, the compiler reports an error.
```
shadrina, line 257: @petukhovv This is the additional info you requested in DM. I tried to translate terms like upper or lower bounds into sub- and supertypes.

petukhovv: Looks good to me

petukhovv (approved):

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+   val myMap = buildMap {
+       putAll(baseMap)
+       if (additionalEntry != null) {
+           put(additionalEntry.first, additionalEntry.second)
+       }
+   }
+}
+```
+
+We don't have enough type information here to infer type arguments in a regular way, but builder inference can
```
shadrina, line 21:
```suggestion
There is not enough type information here to infer type arguments in a regular way, but builder inference can
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+analyze the calls inside the lambda argument. Based on the type information about `putAll()` and `put()` calls,
+the compiler can automatically infer type arguments of the `buildMap()` call into `String` and `Number`. 
+Builder inference allows us to omit type arguments while using generic builders.
```
shadrina, line 24:
```suggestion
Builder inference allows to omit type arguments while using generic builders.
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+analysis ends with the result of inferring the type argument `E` into `String`.
+
+Note that you can always call `equals()`, `hashCode()`, and `toString()` functions with a postponed type variable as a
+receiver.
+
+### Contributing to builder inference results
+
+Builder inference can collect different varieties of type information that contribute to the analysis result.
+It considers:
+* Calling methods on a lambda's receiver that use the type parameter's type:
```
shadrina, line 170:
```suggestion
* Calling methods on a lambda's receiver that use the type parameter's type
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+  ```kotlin
+  val result = buildList {
+      // Type argument is inferred into String based on the passed "value" argument
+      add("value")
+  } // result has the List<String> type inferred
+  ```
+* Specifying the expected type for calls which return type parameter's type:
```
shadrina, line 177:
```suggestion
* Specifying the expected type for calls that return the type parameter's type
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+  fun <K> myBuilder(builder: Foo<K>.() -> Unit): Foo<K> = Foo<K>().apply(builder)
+
+  fun main() {
+      val result = myBuilder {
+          val x: List<CharSequence> = items
+          // ...
+      } // result has the Foo<CharSequence> type
+  }
+  ```
+* Passing postponed type variables' types into methods which expect concrete types:
```
shadrina, line 198:
```suggestion
* Passing postponed type variables' types into methods that expect concrete types
```

docs/topics/using-builders-with-builder-inference.md
```diff
@@ -0,0 +1,276 @@
...
+          val isLong = x.isMoreThat3()
+      // ...
+      } // result2 has the List<String> type
+  
+      val result3 = buildList {
+          takeListOfStrings(this)
+      } // result3 has the List<String> type
+  }
+  ```
+* Taking a callable reference to the lambda receiver's member:
```
shadrina, line 223:
```suggestion
* Taking a callable reference to the lambda receiver's member
```

PR #2629, opened 2021-11-02
update: rules for expect/actual declarations

danil-pavlov:


docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,23 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
 The main rules regarding expected and actual declarations are:
 * An expected declaration is marked with the `expect` keyword; the actual declaration is marked with the `actual` keyword.
 * `expect` and `actual` declarations have the same name and are located in the same package (have the same fully qualified name).
-* `expect` declarations never contain any implementation code.
+* `expect` declarations never contain any implementation code and are abstract by default.
+
+> In interfaces, functions in actual declarations can be non-abstract and have a
+> body. To indicate that in `expect` declarations, explicitly mark functions as `open`:
```
dsavvinov, line 78: "that" is a little bit unclear for me here: for example, I might assume that this is a way to give expect-functions bodies, while it is in fact only about changing their modality. I'd rephrase it in a more specific way, just as an example:

"In interfaces, functions in actual declarations can be non-abstract and have a body, which allows inheritors to not implement this function. `expect`-functions can not have bodies, so to let common-inheritors know that it's not necessary to implement particular function, one can mark it as `open`. Naturally, all `actual`-counterparts will be required to have bodies as well"

(still far from ideal IMO, I'm not that good at this stuff :) )

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,23 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
...
+>
+> ```kotlin
+> expect interface Mascot {
+>     open fun name()
+> }
+>
+> actual interface Mascot {
```
dsavvinov, line 85: It would be useful to indicate somehow where the common code ends and where the platform code begins. For example, [this](https://kotlinlang.org/docs/kmm-connect-to-platform-specific-apis.html#example-send-and-receive-messages-from-a-websocket) article just uses separate code blocks with code-comments, like `// Common code` and `// Platform code`

I don't know whether we have a better way of representing this

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,23 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
...
 * An expected declaration is marked with the `expect` keyword; the actual declaration is marked with the `actual` keyword.
 * `expect` and `actual` declarations have the same name and are located in the same package (have the same fully qualified name).
-* `expect` declarations never contain any implementation code.
+* `expect` declarations never contain any implementation code and are abstract by default.
+
+> In interfaces, functions in actual declarations can be non-abstract and have a
+> body. To indicate that in `expect` declarations, explicitly mark functions as `open`:
+>
+> ```kotlin
+> expect interface Mascot {
```
dsavvinov, line 81: Please, add the usage of this interface in common code, otherwise it's completely not clear why this feature is even needed. Example:

```kotlin
expect interface Mascot {
    open fun name() 
}

class MascotImpl : Mascot {
    // fine not to implement 'name': all 'actual's are guaranteed to have a default implementation for it
}
```

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,23 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
...
+>     open fun name()
```
dsavvinov, line 82: This is certainly minor, but I guess it a little bit contradicts the general sense of Kotlin language: `name` sounds as a property for me. One could rewrite it as `getName` to make it sound more like a function, but then we explicitly recommend to not do that and use properties instead.

The reason I'm bringing this up is that sometimes some users are very nitpicky about examples we use in our public content, so it's not pointless to be extra-careful here

dsavvinov (changes requested):

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -73,22 +73,30 @@ The main rules regarding expected and actual declarations are:
...
+}
+
+class MascotImpl : Mascot {
+    val name: String = "TBD"
+    // it's ok not to implement `display()`: all `actual`s are guaranteed to have a default implementation
+}
+
+// Platform-specific
+actual interface Mascot {
+    fun display(output: MascotImpl): MascotImpl {
```
dsavvinov, line 95: This code doesn't compiles for me, because `display`-declaration in `actual` is quite different from the one in `expect`. If it works for you, I'd be curious to see a sample project (might be a bug)

dsavvinov (changes requested):

docs/topics/mpp/mpp-connect-to-apis.md
danil-pavlov: No, I was just trying something out. It wasn't ready for review. Now I simplified the text a bit more, as well as the example

dsavvinov: Ah, ok, sorry then!

dsavvinov (approved):

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,30 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
 The main rules regarding expected and actual declarations are:
 * An expected declaration is marked with the `expect` keyword; the actual declaration is marked with the `actual` keyword.
 * `expect` and `actual` declarations have the same name and are located in the same package (have the same fully qualified name).
-* `expect` declarations never contain any implementation code.
+* `expect` declarations never contain any implementation code and are abstract by default.
+* In interfaces, functions in `expect` declarations cannot have bodies, but their `actual` counterparts can be non-abstract
+and have a body, allowing the inheritors not to implement a particular function.
```
koshachy, line 77: I bit complicated.
Does it work this way: It allows the inheritors not to implement....  ?

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,30 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
...
+
+  To indicate that common inheritors don't need to implement a function, mark it as `open`. All its `actual` implementations
+will be required to have a body.
```
koshachy, line 80:
```suggestion
will be required to have a body:
```

koshachy (approved):
Some minor suggestions.
Up to you.

docs/topics/mpp/mpp-connect-to-apis.md
```diff
@@ -72,7 +72,30 @@ internal actual fun writeLogMessage(message: String, logLevel: LogLevel) {
 The main rules regarding expected and actual declarations are:
 * An expected declaration is marked with the `expect` keyword; the actual declaration is marked with the `actual` keyword.
 * `expect` and `actual` declarations have the same name and are located in the same package (have the same fully qualified name).
-* `expect` declarations never contain any implementation code.
+* `expect` declarations never contain any implementation code and are abstract by default.
```
ilya-g, line 75: "abstract" has a particular meaning in Kotlin. Is it used in this meaning here?

danil-pavlov: It is, meaning 'the one that doesn't have an implementation'. Like in [Abstract classes](https://kotlinlang.org/docs/classes.html#abstract-classes) or [Interfaces](https://kotlinlang.org/docs/interfaces.html#properties-in-interfaces)

ilya-g: I suppose it would be confusing, because "abstract" usually means abstract modality. 
`expect` declarations can be explicitly declared `abstract` or not (e.g. `open`, `final`). Being expect doesn't imply being abstract.

danil-pavlov: Not sure that if it's not explicitly marked as `abstract` means it's not abstract

@dsavvinov, @koshachy What do you think? Should I change the wording here?

PR #2639, opened 2021-11-09
fix: typo - commonWebpackConfig

p0n3:
Should be `commonWebpackConfig` instead of `commonWbpackConfig`.


p7nov:
Thank you for the fix!

PR #2643, opened 2021-11-10
fix: move the JVM toolchain part to Targeting the JVM and fix headers this way

PetrakovichVictoria:


docs/topics/gradle.md
```diff
@@ -133,6 +133,158 @@ sourceSets {
...
+* `error` – the plugin will fail the build.
+* `ignore` – the plugin will skip the check and won't produce any messages.
+
+### Set custom JDK home
+
+> [Setting `jdkHome` option directly is deprecated](https://youtrack.jetbrains.com/issue/KT-46541).
+>
+{type="warning"}
+
+By default, Kotlin compile tasks use the current Gradle JDK. However, if you need, you can set the JDK home in the following ways:
```
PetrakovichVictoria, line 163: Is this ok as the description why to set custom JDK home?

docs/topics/gradle.md
```diff
@@ -133,6 +133,158 @@ sourceSets {
...
+the `compileJava` task has (or [inherits](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java-extension)) `targetCompatibility=15`.
+
+Control the behavior of this check by setting the `kotlin.jvm.target.validation.mode` property in the `build.gradle` file equal to:
+* `warning` – the default value; the Kotlin Gradle plugin will print a warning message.
+* `error` – the plugin will fail the build.
+* `ignore` – the plugin will skip the check and won't produce any messages.
+
+### Set custom JDK home
+
+> [Setting `jdkHome` option directly is deprecated](https://youtrack.jetbrains.com/issue/KT-46541).
```
p7nov, line 159: "The `jdkHome` compiler option is deprecated since Kotlin 1.*.*."

docs/topics/gradle.md
```diff
@@ -133,6 +133,158 @@ sourceSets {
...
+java {
+    toolchain {
+      languageVersion.set(JavaLanguageVersion.of(<MAJOR_JDK_VERSION>)) // "8" 
+    }
+}
+```
+
+To set any JDK (even local) for the specific task, use the Task DSL.
+
+### Setting JDK version with the Task DSL
```
p7nov, line 236: This subsection needs an intro describing the case - older Gradle without toolchains.
Otherwise readers that just scan headings with their eyes or Ctrl+F "JDK version" may get there and think that this is a general way.

docs/topics/gradle.md
```diff
@@ -133,6 +133,158 @@ sourceSets {
...
+
+The Task DSL allows setting any JDK version for any task implementing the `UsesKotlinJavaToolchain` interface.
+At the moment, these tasks are `KotlinCompile` and `KaptTask`.
+If you want Gradle to search for the major JDK version, replace the <MAJOR_JDK_VERSION> placeholder in your build script:
```
p7nov, line 240: Highlight the placeholder with code

p7nov (changes requested):
Generally OK, please see some comments

koshachy:
Hey folks,

Please, ensure that these changes don't overwrite changes from this PR — https://github.com/JetBrains/kotlin-web-site/pull/2654

docs/topics/gradle.md
PetrakovichVictoria: Well, here I wouldn't change the wording at least now. By the link, there is the ticket with the target version = 1.7 and the text telling that builds with jdkHome will fail soon. If we change this note to point to the concrete version of Kotlin, the version will be 1.5.30 and we'll lose the link.

PetrakovichVictoria:
I brought a little edit from 1.6.0 manually here.

docs/topics/gradle.md
PetrakovichVictoria: The edits are in this [commit](https://github.com/JetBrains/kotlin-web-site/pull/2643/commits/9ddff33d6ac6a98e7c0dd46a764d372966a59840).

docs/topics/gradle.md
p7nov: 1. The current wording is not precisely correct - the option itself is deprecated, not "setting it". 
2. This wording doesn't inform about versions in which it was OK to use jdkHome.
3. We describe the current state of things, not our plans. Many times we promised something to happen in the next release and it didn't. This ticket already shows such a case – see discussion in comments and version updates :)

p7nov:
I also removed the mention of Gradle 6.1. Otherwise, we'll have to update this each time we switch the minimum supported version.

p7nov (approved):

PR #2649, opened 2021-11-14
Fix JVM record code showing Kotlin property usage

huy-vuong:
The usage example currently shows the Kotlin property for a record being accessed at the class-level (as if it were static) instead of at the object-level. This change fixes that typo so that the Kotlin property is called on the object-level, since the properties would be at the object-level.


huy-vuong:
This is waiting to be reviewed by maintainers.

koshachy (approved):

koshachy:
Hey @huy-vuong 

I've slightly changed your PR :) Thanks for pointing this out!

PR #2650, opened 2021-11-14
Fix broken link to java.lang.Record documentation

huy-vuong:
The JVM records tutorial links to the Java documentation for `java.lang.Record`, but that link no longer works. This change updates the link to the current JDK 16 documentation.

(This change does not update the link to point to newer JDK versions' documentation, so that it keeps with the original intention of this page. JVM records came out in JDK 16.)


koshachy:
Hey @huy-vuong 
Thanks for your contribution!

PR #2654, opened 2021-11-16
Use configuration avoidance APIs

LouisCAD:
The `matching` method unfortunately causes all tasks to be "realized", which impacts the builds where these tasks don't need to be run for the given module.

This commit uses `withType()` along with `configureEach` to avoid this issue.

@melix Did I do this right?

### DIFF
--- a/docs/topics/gradle.md
+++ b/docs/topics/gradle.md
@@ -816,12 +816,9 @@ val service = project.extensions.getByType<JavaToolchainService>()
 val customLauncher = service.launcherFor {
     it.languageVersion.set(JavaLanguageVersion.of(<MAJOR_JDK_VERSION>)) // "8"
 }
-project.tasks
-    .withType<KotlinJvmCompile>()
-    .configureEach {
-        if (this !is UsesKotlinJavaToolchain) return@configureEach
-        kotlinJavaToolchain.toolchain.use(customLauncher)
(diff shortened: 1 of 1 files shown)


docs/topics/gradle.md
```diff
@@ -817,9 +817,10 @@ val customLauncher = service.launcherFor {
     it.languageVersion.set(JavaLanguageVersion.of(<MAJOR_JDK_VERSION>)) // "8"
 }
 project.tasks
-    .matching { it is UsesKotlinJavaToolchain && it.name == "compileKotlin" }
+    .withType<KotlinJvmCompile>()
```
melix, line 820: you can use `.withType(UsesKotlinJavaToolchain)`

LouisCAD: Oh that's right, I assumed it could not be of type `Task` because it was an interface, but `Task` is also an interface, and `UsesKotlinJavaToolchain` extends it.

Tapchicoma (approved):
Thank you for bringing attention to this, today I learned `matching { }` is actually triggers task creation :clap: 

Wonder why usage of this method is set to "OK, with issues" in Task configuration avoidance Gradle documentation :thinking:

koshachy:
Thanks a lot for your contribution!

PR #2657, opened 2021-11-16
chore: some edits in scope-functions.md

PetrakovichVictoria:


docs/topics/scope-functions.md
```diff
@@ -112,7 +112,7 @@ fun main() {
 
     // it
     str.let {
-        println("The receiver string's length is ${it.length}")
+        println("The argument string's length is ${it.length}")
```
PetrakovichVictoria, line 115: But have some doubts about this edit.

docs/topics/scope-functions.md
```diff
@@ -458,7 +458,7 @@ fun main() {
         val hexDigits = "A-Fa-f"
         val sign = "+-"
         
-        Regex("[$sign]?[$digits$hexDigits]+")
+        Regex("[$sign][$digits$hexDigits]+")
```
PetrakovichVictoria, line 461: With the ? sign "be" from "not-a-number" is included in the result, see https://kotlinlang.org/docs/scope-functions.html#run

Without ?, the result is what can be expected: https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS41LjIxIiwiY29kZSI6ImZ1biBtYWluKCkge1xuICAgICAgICB2YWwgaGV4TnVtYmVyUmVnZXggPSBydW4ge1xuICAgICAgICAgICAgdmFsIGRpZ2l0cyA9IFwiMC05XCJcbiAgICAgICAgICAgIHZhbCBoZXhEaWdpdHMgPSBcIkEtRmEtZlwiXG4gICAgICAgICAgICB2YWwgc2lnbiA9IFwiKy1cIlxuXG4gICAgICAgICAgICBSZWdleChcIlskc2lnbl1bJGRpZ2l0cyRoZXhEaWdpdHNdK1wiKVxuICAgICAgICB9XG5cbiAgICAgICAgZm9yIChtYXRjaCBpbiBoZXhOdW1iZXJSZWdleC5maW5kQWxsKFwiKzEyMzQgLUZGRkYgbm90LWEtbnVtYmVyXCIpKSB7XG4gICAgICAgICAgICBwcmludGxuKG1hdGNoLnZhbHVlKVxuICAgICAgICB9XG59IiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiJ9

docs/topics/scope-functions.md
p7nov: Well, it's still the receiver of `let`, definitely not its argument :)
If you find this contradictory, we can leave just `"The string's length is", but I don't see a problem here.

docs/topics/scope-functions.md
p7nov: * `be` is a valid hex number, so this works as expected now :) 
* If you remove `?`, you won't find numbers without a sign, which is incorrect. Try adding `10` in the string.

Although I totally agree that it's a tricky example, maybe we should replace the string we search with something simpler.

p7nov (changes requested):

docs/topics/scope-functions.md
PetrakovichVictoria: >Well, it's still the receiver of let, definitely not its argument :)
And we don't write that it's the argument of `let`, `it` is a lambda argument (the same as `this` is a lambda receiver above) and we take the length of `it`.
WDYT about it in this perspective?

docs/topics/scope-functions.md
PetrakovichVictoria: > be is a valid hex number, so this works as expected now :)
Yes, but it's  a bit confusing: you expect to see only two numbers as the result but see three including "be")

> If you remove ?, you won't find numbers without a sign, which is incorrect. 
Of course, I understand it)

We can replace "not-a-number" with "!!!%*&". WDYT?

docs/topics/scope-functions.md
p7nov: My thoughts are the same. 
- We can remove the word to avoid overloading readers with details. It's obvious what is _the string_ here.
- We can call it receiver because it stands before the dot and the function we talk about. 

Other options seem incorrect to me.

docs/topics/scope-functions.md
p7nov: Of course, that's what I meant. Or just take some letters that don't form a hex number. And add a simple number without a sign to show that the regex matches them as well.

docs/topics/scope-functions.md
PetrakovichVictoria: Removed the word "string".

docs/topics/scope-functions.md
PetrakovichVictoria: > And add a simple number without a sign to show that the regex matches them as well.

Added.

p7nov (approved):

PR #2660, opened 2021-11-16
Documentation updated

Lokik-5:
I think making property private for data class and allowing then access by public property using backing fields will print wrong output for componentN(), toString(), etc.

data class BatteryDemo(
    private val _level: Int,
    private val _temperature: Int,
    private val _voltage: Int,
) {
    val level get() = _level.toString()
    val temperature get() = _temperature.toString()
    val voltage get() = _voltage.toString()
}


udalov:
You can still write the code that way if it's needed for some reason though. This section is talking about the constraints that the language puts on the data classes. The compiler will report an error if one of these constraints is not satisfied, which is not true for the statement you propose to add.

Lokik-5:
Okay got it.
Thanks for your reply.

PR #2661, opened 2021-11-17
Adapt `Compatibility flags` section to new release cadence

martinbonnin:
Feature release now happen twice a year so 2 cycles are ~1 year

https://blog.jetbrains.com/kotlin/2020/10/new-release-cadence-for-kotlin-and-the-intellij-kotlin-plugin


danil-pavlov (approved):
Good catch! Thanks for your suggestion

PR #2662, opened 2021-11-19
fixed Kotlin DSL example for pre-installed Node.js

albertocavalcante:
The current example syntax is not working.

Got the right one from: https://youtrack.jetbrains.com/issue/KT-39106


p7nov:
Hey!

Thanks for spotting.

It seems that a shorter notation also works, please see my commits. I've also fixed similar examples for Yarn.

p7nov (approved):

PR #2665, opened 2021-11-19
Replace readln with readLine

leartgjoni:
fixes `Unresolved reference: readln` when following the guide


danil-pavlov:
Hi! Kotlin 1.6.0 [introduced](https://blog.jetbrains.com/kotlin/2021/11/kotlin-1-6-0-is-released/#standard-library) new functions for handling standard input: [readln()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/readln.html) and [readlnOrNull()](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/readln-or-null.html).

To avoid confusion, I'll add corresponding links to documentation shortly.

PR #2673, opened 2021-11-23
KTL-511: update samples

zoobestik:


static/js/page/index/jquery.jscroll.js
```diff
@@ -0,0 +1,103 @@
+function jScroll(){
+	this.id = arguments[0];
+	this.type = arguments[1] == 'h' ? 'h' : 'v';
+	this.width = typeof arguments[2] == 'number' ? arguments[2] : null;
+	this.height = typeof arguments[3] == 'number' ? arguments[3] : null;
+	this.$obj = null;
+	this.is_scroll = false;
+	this.start_pos = null;
+	this.init();
+}
```
nikpachoo, line 10: Is it https://jscroll.com/?
I would prefer to add it as a npm dependency.

krutilov (approved):

static/js/page/index/jquery.jscroll.js
zoobestik: No, it isn't. It is other script with same name. That's from npm no good for us.

nikpachoo (approved):

PR #2681, opened 2021-11-26
Corrected the broken link to the blog post.

Zork67:
Hello! At https://kotlinlang.org/docs/whatsnew16.html
There is a link that leads to a blog post. It is not valid at the moment. I've fixed that.


koshachy (approved):
Hey, thanks for pointing out that!

PR #2683, opened 2021-11-28
Add idiom example: ignore nullable mapping results

melentevpavel:


PetrakovichVictoria:
Hi @melentevpavel! Thank you for your pull request. 

I had thought deeply about this idiom and decided not to include it on this page. The main reason is that on this page we try to illustrate the basics of the language using simple examples. Your example is on the next level of learning the language because it is the specific library function and it's harder than the `firstOrNull()`, for example, which is understandable by its name.
 
We mention the `mapNotNull()` function among [other collection transformations](https://kotlinlang.org/docs/collection-transformations.html#map).

Anyway, thank you for your attention.

PR #2684, opened 2021-11-28
Change iterate over a range example code section

melentevpavel:
1. "if (x in 1..10) { ... }" isn't iteration example and shouldn't be placed at "iterate over a range" chapter.
2. I suggest one more consise iteration example.


danil-pavlov (approved):
Thanks for your suggestion!

PR #2685, opened 2021-11-29
Update native-command-line-compiler.md

Vova-SH:
fix platform name


p7nov:
Thank you for spotting this!

PR #2690, opened 2021-12-01
Add "Execute a code block if null" idiom

dorshinar:
I've added an idiom to get a value from a code block if another value is null

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -160,6 +160,10 @@ println(files?.size) // size is printed if files is not null
 val files = File("Test").listFiles()
 
 println(files?.size ?: "empty") // if files is null, this prints "empty"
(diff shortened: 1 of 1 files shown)


p7nov:
Hi! 

Thanks for this PR. This is a nice example.

However, we try to keep the idioms list shorter so that it doesn't scare readers when they first see it. 
There is the [If-not-null-else shorthand](https://kotlinlang.org/docs/idioms.html#if-not-null-else-shorthand) idiom. What do you think about adding your example as a second option in it? Something like "To calculate the fallback value in a code block, use run:" 

Regards, Pavel
Kotlin Technical Writer

dorshinar:
Thank you for the response! 
I've updated the PR

p7nov (approved):

p7nov:
Nice, thanks!

PR #2702, opened 2021-12-08
Use the term "area" instead of "square"

b0ri5:
I thought the width times the height of a rectangle was its area, and that a square was a rectangle with equal sides. And that the square of a number was it times itself.

But this could be a localization concept I am not aware of.


koshachy:
Oh, thanks for pointing out!
Thanks for your contribution :)

koshachy (approved):

PR #2703, opened 2021-12-09
Update Moshi support status

ZacSweers:
It is officially released now https://github.com/square/moshi/blob/master/CHANGELOG.md#version-1130


shadrina:
Thank you!

shadrina (approved):

PR #2706, opened 2021-12-10
KT-50087 update: add note about new readln() function

koshachy:


p7nov:
Let's sync with @danil-pavlov he was solving this problem on other pages recently.

p7nov (approved):

danil-pavlov:
Maybe we should add the same note to competitive-programming.md and native-get-started.md
Together with jvm-get-started.md these were the three topics where `readln()` was introduced

koshachy:
Sounds reasonable, I've updated the PR.

danil-pavlov (approved):

PR #2708, opened 2021-12-11
Fix typo on the opt-in compiler option

wlezzar:
The current documentation might have a typo in the compiler option name:

```
freeCompilerArgs += "-opt-in=org.mylibrary.OptInAnnotation"
```

The code above gives: `Invalid compiler option`.

### DIFF
--- a/docs/topics/opt-in-requirements.md
+++ b/docs/topics/opt-in-requirements.md
@@ -7,7 +7,7 @@
 {type="warning"}
 
 > `@RequireOptIn` and `@OptIn` annotations were introduced in 1.3.70 to replace previously used `@Experimental` and `@UseExperimental`;
-> at the same time, `-opt-in` compiler option replaced `-Xuse-experimental`.
+> at the same time, `-Xopt-in` compiler option replaced `-Xuse-experimental`.
 >
 {type="note"} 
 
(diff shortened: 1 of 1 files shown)


shadrina:
Hi! Which Kotlin version do you use? The `-opt-in` version works for Kotlin >= 1.6.0.

prof18:
Happened to me as well with Kotlin 1.5.21

koshachy:
Thanks for your feedback, guys. This really helps us, we appreciate it!

I see two problems here:
* We should provide a better explanation for such cases in the future
* We should find a way to make it clear to our users that Kotlin docs reflects the latest version of Kotlin

We'll discuss these points with the team and return with a solution.

wlezzar:
Oh sorry about that then!I didn't realize that the issue was that my kotlin version wasn't matching. I should have checked further before submitting the PR 👍🏼 .

Should I close it then?

koshachy:
> Should I close it then?

No, you don't need to.
Let's keep it as a communication channel :)

koshachy:
Hey folks, 

Some new information:
* I close this PR. We'll add a note in this one: https://github.com/JetBrains/kotlin-web-site/pull/2714
* We discussed and decided to put such notes for changes that users could face and there is no way to know what to do next/how to fix it (the IDE doesn't help, the compiler doesn't say anything about how to fix it, etc.)

Sorry for any inconvenience you may have with this.
Thanks for your participation, it helps us to make the documentation better.

PR #2709, opened 2021-12-13
Add missing X on opt-in compiler argument

prof18:
After copying and pasting the snippet to opt in to using an API in a module, I noticed (well not me, but the compiler :) ) that there was some missing X


prof18:
Closing, since I didn't saw #2708. It happened with Kotlin 1.5.21

PR #2714, opened 2021-12-15
update: Update the warning about opt-in requirements

shadrina:


shadrina:
There is another option to put this note above the first place where `-opt-in` is mentioned. @koshachy WDYT?

koshachy:
> There is another option to put this note above the first place where `-opt-in` is mentioned. @koshachy WDYT?

If I understood this correctly, you propose to also add the note after the first usage of the `-opt-in` option. Seems legit.

But this gave me the idea that we should mention this in the compatibility guide?

shadrina:
Adding it to the compatibility guide also seems legit. Can we do it in a separate PR? It will require technical review to verify the deprecation cycle.

koshachy (approved):

PR #2722, opened 2021-12-21
chore: clearer explanation about set operations on lists

p7nov:
https://youtrack.jetbrains.com/issue/KT-50011

### DIFF
--- a/docs/topics/set-operations.md
+++ b/docs/topics/set-operations.md
@@ -13,7 +13,6 @@ To find collection elements not present in another collection, use [`subtract()`
 Both these functions can be called in the infix form as well, for example, `a intersect b`.
 
 ```kotlin
-
 fun main() {
 //sampleStart
     val numbers = setOf("one", "two", "three")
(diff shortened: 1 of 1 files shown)


docs/topics/set-operations.md
```diff
@@ -29,5 +28,18 @@ fun main() {
...
+These operations – `union`, `intersect`, and `subtract` – apply to `List` as well.
+However, their result is _always_ a `Set`, even on lists. In this result, all the duplicate elements are merged into one 
+and the index access is not available.
+
+```kotlin
+fun main() {
+//sampleStart
+    val list1 = listOf(1, 1, 2 ,3, 5, 8, -1)
+    val list2 = listOf(1, 1, 2, 2 ,3, 5)
+    println(list1 intersect list2) // result on two lists is a Set:
```
koshachy, line 40: TBH, I don't understand why there is a colon sign after "Set"

docs/topics/set-operations.md
```diff
@@ -29,5 +28,18 @@ fun main() {
 ```
 
-Note that set operations are supported by `List` as well.
-However, the result of set operations on lists is still a `Set`, so all the duplicate elements are removed.
+These operations – `union`, `intersect`, and `subtract` – apply to `List` as well.
```
koshachy, line 31: WDYT about You can also apply all these set operations to `List`?
Or you can also apply union, intersect, and substarct operations to `List`?

Just want to make this sentence easier to read.

koshachy (commented):
I love the way you want to improve it, but I think we could find better wording.
Discussable.

koshachy (approved):

PR #2730, opened 2021-12-27
fix: Use correct syntax for cinterops in build.gradle

davidmhewitt:
When trying to follow this part of the tutorial, I was getting the following error after applying the `compilations.main` section to my `build.gradle.kts`:

```
build.gradle.kts:20:22: Unresolved reference: main
```

After looking up some code samples, I found this as a working solution.


danil-pavlov:
Thanks a lot for pointing out this error! I've extended your PR a bit and merged it in [#2735](https://github.com/JetBrains/kotlin-web-site/pull/2735)

PR #2731, opened 2021-12-28
Update type-safe-builders.md

DotDivee:
Fix inconsistency involving the use of colons to define function return types.


shadrina:
Thank you!

PR #2734, opened 2021-12-30
move `binaries` to be in `js {}`

aSemy:
error: 
image

works: 
image


p7nov:
You're right, thanks for this fix!

PR #2737, opened 2021-12-30
KSP documentation: syncing with the original repository

shadrina:
These are the latest changes from [google/ksp](https://github.com/google/ksp) (all except one little change that will be [here](https://github.com/JetBrains/kotlin-web-site/pull/2716) later). I made two commits per updated file: one with the original version and another with my modifications. I suppose we do not need a technical review here.


docs/topics/ksp/ksp-kmp.md
```diff
@@ -0,0 +1,63 @@
+[//]: # (title: KSP with Kotlin Multiplatform)
```
koshachy, line 1: I propose to rename the file to ksp-multiplatform

docs/topics/ksp/ksp-cmdline.md
```diff
@@ -0,0 +1,62 @@
+[//]: # (title: Running KSP from command line)
```
koshachy, line 1: Also, there is no need to short the name (SEO purposes).
ksp-command-line looks fine to me

koshachy (approved):
To move a bit faster, I propose to integrate this into our docs as is.

PR #2741, opened 2022-01-10
fix: indentation in safe.md

aecant:
Hello, I noticed the indentation in the "safe" feature tab is a little off, it should be better like this


p7nov:
You're right, thanks! And the next `val` should also have such an indentation.

PR #2743, opened 2022-01-11
Update until function example in ranges.md

mhmdcs:
The notation can be a little bit confusing to new programmers in the comment for the until function example, especially for new programmers reading the doc with no prior experience in mathematical square brackets and parenthesis inclusion/exclusion rules.


p7nov (approved):

p7nov:
thanks!

PR #2745, opened 2022-01-11
Update zip() transformation function description

mhmdcs:
Collections and arrays are different concepts in Kotlin, writing `collection (array)` gives the reader the implication that collections and arrays are one and the same.
Updated it to `collection (or array)` for better reflection.


p7nov:
Thank you!

PR #2750, opened 2022-01-13
Add return type to the Documentation Comment section examples

Dantaro:
The documentation example explicitly talks about how to document return types but the code examples fail to include any return type when one should exist.  Just hoping to clean that up and avoid any confusion


ilya-g (approved):

koshachy (approved):

koshachy:
Thanks!

PR #2751, opened 2022-01-15
chore: better property getters description

mhmdcs:
removed repeated `from the initializer` from the getters and setters description and added a comment to the getter custom accessor example to demonstrate the aforementioned optional type.


koshachy:
Thanks for your PR!

PR #2755, opened 2022-01-18
fix: use the second person

StarHamster:
Used the second person – “you” instead of "we" when not meaning the Kotlin team have done or are doing something.


p7nov (approved):

p7nov:
Thanks for this update!

PR #2763, opened 2022-01-30
Extension of Kotlin Idioms

n81ur3:
Added example of multiple declarations in a single source file to Kotlin Idioms.


p7nov:
Hey!

This is a valid addition, but Kotlin docs never declared that it's impossible to write multiple declarations in the same file. Neither did we see our users having problems with that. So we won't add this to the docs, but will take it into account if we see such problems in the future. Thanks!

PR #2768, opened 2022-02-03
update: add warning about non-working property

PetrakovichVictoria:


PetrakovichVictoria:
Screenshot 2022-02-03 at 23 35 57

docs/topics/gradle.md
```diff
@@ -890,7 +890,17 @@ Each of the options in the following list overrides the ones that came before it
   org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx1500m -Xms=500m
   ```
 
-* You can add the`kotlin.daemon.jvmargs` property in the `gradle.properties` file:
+  > Gradle ignores these properties if all the following conditions are satisfied:
+  > * Gradle uses JDK 1.9 or higher.
+  > * The version of Gradle is between 7.0 and 7.1.1 inclusively.
+  > * Gradle compiles Kotlin DSL scripts.
```
p7nov, line 896:
```suggestion
  > * The project's build script is written in Kotlin (`build.gradle.kts`).
```
The current wording is always true. Does Gradle compile Kotlin DSL scripts? Yes, it does :)

p7nov (approved):
One minor comment

docs/topics/gradle.md
PetrakovichVictoria: 😁🤔 And what if it's about a user's own DSL? @Tapchicoma could you help us, please? Is the @p7nov's wording more correct?

Tapchicoma: IMHO initial wording is ok. Gradle could also compile `*.kt` files inside `buildSrc` module via own Kotlin compiler that may lead to this  error.

PetrakovichVictoria: Ok, maybe we should also add about this condition
>Gradle could also compile *.kt files inside buildSrc module via own Kotlin compiler that may lead to this error.
?

Will it be 
"Gradle is compiling Kotlin DSL scripts OR Gradle is compiling *.kt files inside buildSrc module via own Kotlin compiler"?

But anyway Gradle is compiling them somehow, so seems like "Gradle is compiling Kotlin DSL scripts" is enough?
@Tapchicoma

docs/topics/gradle.md
```diff
@@ -890,7 +890,17 @@ Each of the options in the following list overrides the ones that came before it
   org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx1500m -Xms=500m
   ```
 
-* You can add the`kotlin.daemon.jvmargs` property in the `gradle.properties` file:
+  > Gradle ignores these properties if all the following conditions are satisfied:
+  > * Gradle uses JDK 1.9 or higher.
```
PetrakovichVictoria, line 894:
```suggestion
  > * Gradle is using JDK 1.9 or higher.
```

docs/topics/gradle.md
PetrakovichVictoria, line 896:
```suggestion
  > * Gradle is compiling Kotlin DSL scripts.
```

Tapchicoma: > so seems like "Gradle is compiling Kotlin DSL scripts" is enough?

IMHO this sentence is enough

PR #2774, opened 2022-02-08
Add KMM book as a tutorial resource

nrobi144:


koshachy:
Sorry for the late reply.
We were in the revamping process of Multiplatform and Multiplatform mobile pages so we couldn't process your PR.
It seems outdated now because our current structure doesn't suppose a separate page with tutorials and books.
Sorry for the inconvenience caused, and thank you for your contribution.

PR #2781, opened 2022-02-09
fix: fix arguments for org.gradle.jvmargs=-Dkotlin.daemon.jvm.options

PetrakovichVictoria:


PetrakovichVictoria:
@Tapchicoma wait, I'll add more about the rules of - and ,.

PetrakovichVictoria:
@Tapchicoma it's ready for review now.

docs/topics/gradle.md
```diff
@@ -877,9 +877,16 @@ Each of the options in the following list overrides the ones that came before it
 * If the Gradle daemon's JVM arguments have the `kotlin.daemon.jvm.options` system property – use it in the `gradle.properties` file:
 
  ```properties
-  org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx1500m -Xms=500m
+  org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx1500m,Xms=500m
   ```
 
+  > In this case, enumerate arguments according to the following rules:
```
p7nov, line 883: Three rules in a row; maybe use a bullet list?

docs/topics/gradle.md
```diff
@@ -877,9 +877,16 @@ Each of the options in the following list overrides the ones that came before it
...
+  > Use the minus sign `-` before the argument `Xmx` and don't use it before all other arguments.
```
p7nov, line 884: Does it depend on the args order? What if Xmx doesn't come first?

docs/topics/gradle.md
```diff
@@ -877,9 +877,16 @@ Each of the options in the following list overrides the ones that came before it
...
+  > Separate all arguments with the comma ','.
```
p7nov, line 885:
```suggestion
  > Separate arguments with commas (`,`).
```
shorter is better

docs/topics/gradle.md
```diff
@@ -877,9 +877,16 @@ Each of the options in the following list overrides the ones that came before it
...
+  > Don't use spaces between arguments, it will lead to the usage of the arguments after a space for the Gradle daemon, not for the Kotlin daemon.
```
p7nov, line 886: Isn't this a part of the second rule?
"Separate arguments with commas (`,`) _without spaces_. Arguments that come after a space will be used for .."

p7nov (commented):
Some questions from my side

docs/topics/gradle.md
p7nov: Also, I'd suggest writing this simpler and shorter. Smth like:
"When passing the arguments, follow these rules:"

docs/topics/gradle.md
PetrakovichVictoria: No, it doesn't.

p7nov (approved):
Nice!

nikpachoo (approved):

PR #2782, opened 2022-02-09
Add example of typealias to "fun interfaces vs. type aliases" section

mcpiroman:
It may not be evident to the reader that the type aliases brought by this section talk about aliasing a functional type.


mcpiroman:
.

koshachy (approved):

PR #2791, opened 2022-02-14
Fix the reference to java Annotation

mataevs:
java.lang.Annotation -> java.lang.annotation.Annotation


p7nov:
Thanks for spotting this!

PR #2806, opened 2022-02-18
Added KMM examples:

falcon4ever:
- Bottom Navigation sample
- MVI sample


koshachy:
Hey, thanks for your contribution!

We've looked through your samples and found that it uses the same techniques and tools already used in other examples. 
Such intersections could confuse our readers because they probably couldn't understand the difference between samples.

Also, it is hard for us to keep this list of samples up to date, so we're trying to merge only unique samples, with new libraries or approaches under the hood.
I hope you understand our point. Have a nice Kotlin!

PR #2813, opened 2022-02-21
fix: match the function value parameter

ykws:
the following reference `file` and more clear.

```kotlin
    is FileReadError -> { println("Error while reading file ${e.file}") }
```


p7nov (approved):

p7nov:
Good improvement, thank you!

PR #2814, opened 2022-02-22
add Xi'An Kotlin User Group

shengyou:


meilalina (commented):
Looks good to me! Thank you @shengyou cc @nikpachoo

PR #2816, opened 2022-02-23
Remove harmful instruction for `Boolean` properties

turansky:
[Related discussion](https://kotlinlang.slack.com/archives/C0B8L3U69/p1645523369441939)

### DIFF
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -78,50 +78,6 @@ external interface CustomComponentState : State {
 }
 ```
 
-## Make boolean properties nullable in external interfaces
-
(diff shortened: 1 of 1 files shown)


vonox7:
Probably also remove inspection in the IDE when this is not considered harmful

SebastianAigner:
Hey folks. Thanks for bringing this to our attention – you are of course right. However, we have found that this is still an important part of troubleshooting that people run into every once in a while. In #2824, I've moved the text to a separate section and softened the wording to hopefully make it clearer that this isn't a _necessary_ requirement, but something that may help avoid headaches in certain situations.

We will also be disabling the IDE inspection to avoid misleading people.

I'm closing this here, but I absolutely welcome your feedback on PR #2824 that tackles this!

PR #2820, opened 2022-02-27
Added missing arrowhead

imashnake0:
Added missing arrowhead for the `MutableList` extends `MutableCollection` line. Fixed arrows to `MutableCollection` (from `MutableIterable`), `Collection`, and `Set`.


p7nov (approved):
Thanks for this improvement!

PR #2824, opened 2022-03-02
Soften wording on IR compiler migration guide re nullable Booleans

SebastianAigner:
As per #2816

### DIFF
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -186,7 +186,7 @@ external interface SomeExternal {
 }
 
 fun main() {
-    val empty = js("{}") as SomeExternal
+    val empty: SomeExternal = js("{}") as SomeExternal
     println(empty.visible.toString()) // Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
 }
 ```
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -186,7 +186,7 @@ external interface SomeExternal {
 }
 
 fun main() {
-    val empty: SomeExternal = js("{}") as SomeExternal
+    val empty: SomeExternal = js("{}")
     println(empty.visible.toString()) // Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
 }
 ```
--- a/docs/topics/js/js-ir-migration.md
+++ b/docs/topics/js/js-ir-migration.md
@@ -172,35 +172,22 @@ kotlin {
 }
 ```
 
-## Additional style hints recommended when working with the Kotlin/JS IR compiler
+## Additional troubleshooting tips when working with the Kotlin/JS IR compiler
 
-These hints may help you when working with projects using the Kotlin/JS IR compiler. They're not strictly required for a migration, but may be worth keeping in mind.
+These hints may help you when troubleshooting problems in your projects using the Kotlin/JS IR compiler.
 
 ### Make boolean properties nullable in external interfaces
 
-**Issue**: JavaScript treats the `null` or undefined value of a boolean variable as `false`. So, Boolean properties can be used
-in expressions without being defined. If there is a chance that you receive Boolean properties that may be nullable (for example when your code is  called from JavaScript code you have no control over), you may want to avoid this behaviour in Kotlin:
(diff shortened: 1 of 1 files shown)


docs/topics/js/js-ir-migration.md
```diff
@@ -214,4 +170,52 @@ kotlin {
...
+
+```kotlin
+external interface ComponentProps: Props {
+   var isInitialized: Boolean
+   var visible: Boolean
+}
+```
+
+```kotlin
+val props = js("{}") as ComponentProps
```
turansky, line 192: `val props: ComponentProps = js("({})")`

without redundant cast

docs/topics/js/js-ir-migration.md
```diff
@@ -214,4 +170,52 @@ kotlin {
...
+props.isInitialized = true
+// visible is not initialized - OK in JS – means it's false
+```
+
+If you try to use such a property in a function overridden in Kotlin (for example, a React `button`), you'll get a `ClassCastException`:
+
+```kotlin
+button {
+   attrs {
```
turansky, line 201: No more `attrs`?

docs/topics/js/js-ir-migration.md
```diff
@@ -214,4 +170,52 @@ kotlin {
...
+       autoFocus = props.visible // ClassCastException here
```
turansky, line 202: `ClassCastException` isn't expected here:
* `autoFocus` property in nullable
* `null` works as `false`

turansky (changes requested):

SebastianAigner:
Further softened the wording. As per discussion with @turansky, turns out that the error surface is smaller than initially anticipated – in all cases (besides calling `toString`) `null` and `undefined` booleans behave correctly in Kotlin/JS, so I've made the warning specific to calling toString.

@turansky @bashor PTAL if time allows :)

docs/topics/js/js-ir-migration.md
```diff
@@ -214,4 +170,39 @@ kotlin {
...
+
+**Issue**: when you call `toString` on a `Boolean` from an external interface, you're getting an error like `Uncaught TypeError: Cannot read properties of undefined (reading 'toString')`. JavaScript treats the `null` or `undefined` values of a boolean variable as `false`. If you rely on calling `toString` on a `Boolean` that may be `null` or `undefined` (for example when your code is called from JavaScript code you have no control over), be aware of this:
+
+```kotlin
+external interface SomeExternal {
+    var visible: Boolean
+}
+
+fun main() {
+    val empty: SomeExternal = js("{}") as SomeExternal
```
turansky, line 189: `as SomeExternal` is redundant

turansky (changes requested):

docs/topics/js/js-ir-migration.md
SebastianAigner: ...finger must've slipped when copying this snippet from my sample project, d'oh 😱

turansky (approved):

docs/topics/js/js-ir-migration.md
```diff
@@ -214,4 +170,39 @@ kotlin {
...
 }
+```
+
+## Additional troubleshooting tips when working with the Kotlin/JS IR compiler
+
+These hints may help you when troubleshooting problems in your projects using the Kotlin/JS IR compiler.
+
+### Make boolean properties nullable in external interfaces
+
+**Issue**: when you call `toString` on a `Boolean` from an external interface, you're getting an error like `Uncaught TypeError: Cannot read properties of undefined (reading 'toString')`. JavaScript treats the `null` or `undefined` values of a boolean variable as `false`. If you rely on calling `toString` on a `Boolean` that may be `null` or `undefined` (for example when your code is called from JavaScript code you have no control over), be aware of this:
```
bashor, line 181: it, also, could be fail in casts to `Boolean` and `Any` (note not-null!) either explicit and implicit. Implicit ones could be generated by compiler, e.g. when refine generic type in override  (bridges), like:
https://pl.kotl.in/-NnNLbtOQ
```kotlin
external interface Foo {
    val bar: Boolean
}

interface A<T> {
    fun foo(a: T) = a
}

class B : A<Boolean> {
    override fun foo(a: Boolean) = a
}


fun main() {
    try {
        val o: Foo = js("{}")
	println("1")
        val bool = o.bar// as Boolean
        println("2")
        val a = B() as A<Boolean>
        println("3")
        println(a.foo(bool))
        println("done")
    } catch (e: Throwable) {
        println("fail")
    }
}
```

turansky: `Boolean` as type parameter doesn't look like typical case :)

turansky:
Merge?

PR #2832, opened 2022-03-10
update: Add -opt-in to compiler reference

shadrina:


docs/topics/compiler-reference.md
```diff
@@ -216,6 +216,10 @@ see [this](https://www.davidbcalhoun.com/2014/what-is-amd-commonjs-and-umd/) art
 
 Don't automatically include the default Kotlin/JS stdlib into the compilation dependencies.
 
+### -opt-in _marker_
+
+Enable usages of API that requires opt-in with an opt-in requirement marker with the given fully qualified name.
```
mglukhikh, line 221: May be drop second "opt-in"? Just "that requires opt-in with a requirement marker..."

p7nov: Agree with Misha.
I would also:
- add a link to opt-in doc page somewhere in this sentence.
- use "annotation" instead of "marker" because it's a commonly known term.

shadrina: Thanks! Changed the wording and added a link.

p7nov (approved):

mglukhikh (approved):

PR #2839, opened 2022-03-15
update: hmpp is now enabled by default

danil-pavlov:


docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -0,0 +1,108 @@
...
+
+```properties
+kotlin.mpp.hierarchicalStructureSupport=false
+```
+
+As for the `kotlin.mpp.enableCompatibilityMetadataVariant` option that enables compatibility of libraries published with
+the hierarchical project structure and non-hierarchical projects, you should disable it separately:
+
+```properties
+kotlin.mpp.enableCompatibilityMetadataVariant=false
```
antohaby, line 107: That might sound a little bit confusing. 
Yes, it is true that the user may have following set of properties:

```
kotlin.mpp.hierarchicalStructureSupport=false
kotlin.mpp.enableCompatibilityMetadataVariant=true
```

with meaning that they still want to compile `commonMain` source set into legacy metadata artifact. Along with new klib variant. But disable hierarchical support for some reason. However by default `enableCompatibilityMetadataVariant` is false already, so there is no logical reason to explicitly opt-out.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -0,0 +1,108 @@
...
+
+The Kotlin toolchain will automatically figure out the API available in the consumer source set while checking the
+unsafe usages, like using an API meant for the JVM in JS code.
+
+* Libraries published with the hierarchical project structure are compatible only with projects that have hierarchical
+  project structure. To enable compatibility with non-hierarchical projects, add the following to
+  the `gradle.properties` file in your library project:
+
+   ```properties
+   kotlin.mpp.enableCompatibilityMetadataVariant=true
```
antohaby, line 75: General remark about `enableCompatibilityMetadataVariant`, it is up to you how to deal with it.

Only source code from `commonMain` source set  will be compiled with legacy metadata compiler.

For example in project with such structure: `commonMain -> nativeMain -> [ linuxX64, linuxArm64 ]`
If `enableCompatibilityMetadataVariant` is enabled, then only code in `commonMain` will be compiled with legacy metadata variant.
And both `commonMain` and `nativeMain` will be compiled into klib. 
And the situation goes tricky if platform-specific code will be used in `commonMain`. For example function `getpid` from posix. In such cases, the compilation of `commonMain` source set to legacy format will fail. But if posix's `getpid` will be used in `nativeMain`, everything would compile fine :) 

So that is just an aspect of `enableCompatibilityMetadataVariant` behavior. It is up to you how to work with it.

docs/topics/multiplatform/multiplatform-share-on-platforms.md
```diff
@@ -268,18 +262,12 @@ it and call `runBlocking` from a source set that is shared between the JVM and n
 
 ### Use native libraries in the hierarchical structure
 
-You can use platform-dependent libraries like Foundation, UIKit, and POSIX in source sets shared among several native 
+You can use platform-dependent libraries like `Foundation`, `UIKit`, and `posix` in source sets shared among several native
```
antohaby, line 265: NIT:
I believe `POSIX` should be still written with capital letters. That is their official style.

docs/topics/multiplatform/multiplatform-hierarchy.md
danil-pavlov: I decided to add a warning on using platform-specific code in `commonMain` in this case

danil-pavlov: I decided to add a warning on using platform-specific code in `commonMain` in this case

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -0,0 +1,108 @@
...
+## For multiplatform project authors
+
+With the new hierarchical project structure support, you can share code among some, but not
+all [targets](multiplatform-dsl-reference.md#targets) in a multiplatform project.
+
+You can also use platform-dependent libraries, such as `UIKit`, and `posix` in source sets shared among several native
+targets. One popular case is having access to iOS-specific dependencies like `Foundation` when sharing code across all
+iOS targets. New structure helps you share more native code without being limited by platform-specific dependencies.
+
+By using the hierarchical structure along with platform-dependent libraries in shared source sets, you can eliminate the
```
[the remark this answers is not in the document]
danil-pavlov: Fixed the description to a default option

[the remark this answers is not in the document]
danil-pavlov: Fixed the description to a default option

docs/topics/multiplatform/multiplatform-hierarchy.md
danil-pavlov: Fixed the description to a default option

PR #2886, opened 2022-04-06
fix: Recommend kotlinx.html 0.7.5 for Kotlin 1.6.20

jakobkmar:


PetrakovichVictoria (approved):
Hi @jakobkmar, thank you for your fast reaction on the new kotlinx.html release!

PR #2893, opened 2022-04-11
fix: link to CocoapodsExtension.kt

Tiagoperes:
Fixes broken link to `CocoapodsExtension.kt`.


danil-pavlov (approved):
Thanks for noticing!

PR #2898, opened 2022-04-13
Use lower-case relation name for the data class

pavelgordon:
This tutorial works fine with select/find queries, but gives "relation "messages" does not exist" error during persisting the messages. This is due to the relation name being case-sensitive e.g. "MESSAGES" vs "messages"
`findMessages` uses implicit query  `@Query("select * from messages")`("messages") and therefore works.
`db.save(message)` uses the generated query from `SimpleJdbcRepository`("MESSAGES") and therefore fails.
This tutorial https://dba.stackexchange.com/questions/192897/postgres-relation-does-not-exist-error

To be transparent, I tested this tutorial on Postgres database running locally, and not on the in-memory H2.


koshachy (approved):
Hey @pavelgordon!
Thanks for pointing this out

PR #2899, opened 2022-04-13
Update idioms.md

Naidile-P-N:
Add a short description for the keyword lazy


koshachy:
I'm not quite sure that this comment should be here.
OTOH, I don't see any reason why not.

koshachy:
I'm not quite sure that this comment should be here.
OTOH, I don't see any reason why not.

koshachy (approved):

PR #2911, opened 2022-04-21
KTL-721 Community Overvew bottom block

krutilov:
https://youtrack.jetbrains.com/issue/KTL-721


nikpachoo (approved):

nikpachoo (changes requested):
Let's remove the images as the designer proposed.

PR #2914, opened 2022-04-23
Update constructing-collections.md

DJsushi:
It's not a constructor, it's a function.


DJsushi:
Sorry for the late thanking, but thank you! This is even better.

PR #2938, opened 2022-05-13
chore: excluding the dollar sign when copying code snippets

danil-pavlov:
* KT-45467


koshachy:
It complicates the mark up, while the final benefit is really unclear.
Maybe we could just omit the dollar sign?

koshachy (approved):

PR #2949, opened 2022-05-16
fix: paragraph nesting in delegated-properties.md

udalov:
"Translation rules for delegated properties" and "Providing a delegate" are not subsections of "Property delegate requirements"


koshachy (approved):
Thanks for fixing it!

PR #2960, opened 2022-05-24
fix: fix filenames mentioned in doc

ywnkm:


danil-pavlov:
Thank you for pointing out this issue! We decided to add Kotlin samples on this page instead (#2987)

PR #2963, opened 2022-05-26
Update maven.md

AlexBeggs:
Fix grammar


koshachy (approved):
Thank you for pointing this out!

PR #2964, opened 2022-05-27
Update multiplatform-mobile-plugin-releases.md

terrakok:
Should be published together with Kotlin 1.7.0 release


koshachy:
@terrakok oh, sorry, it seems that I broke the PR a bit 😨

PR #2974, opened 2022-06-01
Update in interfaces.md file

saurabh-jadhav-theporter:
Inside resolving interface conflicts,
I think `bar()` method is abstract by default in `A` interface.

So, we need to remove `not` word from the documentation.

Correct me If I am wrong, comments appreciated!


laxmikantrevdikar (approved):
nice catch

koshachy:
Thanks for your PR.
This sentence is correct:

> bar() is not marked as abstract in A, because this is the default for interfaces if the function has no body

We don't need to mark `bar()` as abstract, because it's abstract by default.

PR #2983, opened 2022-06-03
Compatibility guide for Kotlin 1.7

ilya-g:


docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
+****[//]: # (title: Compatibility guide for Kotlin 1.7)
```
danil-pavlov, line 1:
```suggestion
[//]: # (title: Compatibility guide for Kotlin 1.7)
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+>
+> **Incompatible change type**: source
+>
+> **Short summary**:
+>
+> **Deprecation cycle**:
+>
+> - 1.5.20: warning
+> - 1.7.0: report an error
+-->
```
danil-pavlov, line 42: Do you want to keep this template?

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Make safe call result always nullable
+
+> **Issue**: [KT-46860](https://youtrack.jetbrains.com/issue/KT-46860)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.7 will consider the type of safe call result always nullable, even when the receiver of the safe call is non-nullable.
```
danil-pavlov, line 52:
```suggestion
> **Short summary**: Kotlin 1.7 will consider the type of safe call result always nullable, even when the receiver of the safe call is non-nullable
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+>
+> **Deprecation cycle**:
+>
+> - \<1.3: report a warning on unnecessary safe call on non-nullable receivers
```
danil-pavlov, line 56:
```suggestion
> - \<1.3: report a warning on an unnecessary safe call on non-nullable receivers
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+> - 1.6.20: warn additionally that the result of unnecessary safe call will change its type in the next version
```
danil-pavlov, line 57:
```suggestion
> - 1.6.20: warn additionally that the result of an unnecessary safe call will change its type in the next version
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+> - 1.7.0: change the type of safe call result to nullable,
```
danil-pavlov, line 58:
```suggestion
> - 1.7.0: change the type of safe call result to nullable,
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+### Prohibit the delegation of super calls to an abstract superclass member
+
+> **Issues**: [KT-45508](https://youtrack.jetbrains.com/issue/KT-45508), [KT-49017](https://youtrack.jetbrains.com/issue/KT-49017), [KT-38078](https://youtrack.jetbrains.com/issue/KT-38078)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+> 
+> **Short summary**: Kotlin will report a compile error when an explicit or implicit super call is delegated 
+> to an _abstract_ member of the superclass, even if there’s a default implementation in a super interface.
```
danil-pavlov, line 70:
```suggestion
> to an _abstract_ member of the superclass, even if there's a default implementation in a super interface
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+>
+> **Deprecation cycle**:
+>
+> - 1.5.20: introduce a warning when non-abstract classes that do not override all abstract members are used
+> - 1.7.0: report an error if a super call if fact accesses an abstract member from a super class
```
danil-pavlov, line 75:
```suggestion
> - 1.7.0: report an error if a super call, in fact, accesses an abstract member from a superclass
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+### Prohibit exposing non-public types through public properties declared in a non-public primary constructor
+
+> **Issue**: [KT-28078](https://youtrack.jetbrains.com/issue/KT-28078)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin will prevent declaring public properties having non-public types in a private primary constructor.
+> Accessing such properties from another package could lead to an `IllegalAccessError`.
```
danil-pavlov, line 89:
```suggestion
> Accessing such properties from another package could lead to an `IllegalAccessError`
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+>
+> **Deprecation cycle**:
+>
+> - 1.3.20: report a warning on a public property declared in non-public constructor has non-public type
```
danil-pavlov, line 93:
```suggestion
> - 1.3.20: report a warning on a public property declared in a non-public constructor that has a non-public type
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.6 no longer allows calling user functions named `suspend` that have the single argument of a functional type passed as a trailing lambda
+>
+> **Deprecation cycle**:
+>
+> - 1.3.0: introduce a warning on such function calls
+> - 1.6.0: raise this warning to an error
+> - 1.7.0: introduce changes to the language grammar, so that `suspend` before `{` is parsed as a keyword
```
danil-pavlov, line 222:
```suggestion
> - 1.7.0: introduce changes to the language grammar so that `suspend` before `{` is parsed as a keyword
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+> **Incompatible change type**: source
+>
+> **Short summary**: declarations from the `kotlin.dom` and `kotlin.browser` packages are moved to the corresponding `kotlinx.*` packages to prepare for extracting them from stdlib
+>
+> **Deprecation cycle**:
+>
+> - 1.4.0: introduce the replacement API in `kotlinx.dom` and `kotlinx.browser` packages
+> - 1.4.0: deprecate the API in `kotlin.dom` and `kotlin.browser` packages and propose the new API above as a replacement
+> - 1.6.0: raise the deprecation level to an error
+> - \>= 1.8: remove the deprecated functions from stdlib
```
danil-pavlov, line 275:
```suggestion
> - \>= 1.8.0: remove the deprecated functions from stdlib; move the API in kotlinx.* packages to a separate library
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+> - \>= 1.8: move the API in kotlinx.* packages to a separate library
```
danil-pavlov, line 276:
```suggestion

```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Deprecate KotlinGradleSubplugin class
+
+> **Issue**: [KT-48830](https://youtrack.jetbrains.com/issue/KT-48830)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: the class `KotlinGradleSubplugin` will be removed in Kotlin 1.7. Use `KotlinCompilerPluginSupportPlugin` instead.
```
danil-pavlov, line 306:
```suggestion
> **Short summary**: the class `KotlinGradleSubplugin` will be removed in Kotlin 1.7. Use `KotlinCompilerPluginSupportPlugin` instead
```

danil-pavlov (changes requested):

docs/topics/compatibility-guide-17.md
ilya-g: Yes, it would be very handy for further updates and next version compatibility guides

docs/topics/compatibility-guide-17.md
ilya-g: As always, this trailing space is needed to have a soft break before the next line

docs/topics/compatibility-guide-17.md
ilya-g: I believer it should be fine to use >= 1.8 to indicate less certainty in the version that in >= 1.8.0

docs/topics/compatibility-guide-17.md
ilya-g: These are two independent actions of migration plan, they should not be merged together.

docs/topics/compatibility-guide-17.md
danil-pavlov: Do you use our documentation plugin? Its local preview shows that these trailing spaces don't make a line break, but in fact leave no space between the comma and the flag name
If it actually works correct on staging, I don't mind these extra spaces

docs/topics/compatibility-guide-17.md
danil-pavlov: Please check, trying a more detailed explanation

docs/topics/compatibility-guide-17.md
danil-pavlov: It's just for the sake of consistency. All other items in the listed deprecation cycles have a full release number
And I don't think it changes the meaning, but it's up to you

docs/topics/compatibility-guide-17.md
danil-pavlov: Yeah, I get that they are different. It’s just that every deprecation cycle lists different release periods. Since these two actions happen in the same period, some time after 1.8, we can use one item. As a compromise, we can introduce the the second level for the list.

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Deprecate KotlinGradleSubplugin class
+
+> **Issue**: [KT-48830](https://youtrack.jetbrains.com/issue/KT-48830)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: the class `KotlinGradleSubplugin` will be removed in Kotlin 1.7. Use `KotlinCompilerPluginSupportPlugin` instead
```
PetrakovichVictoria, line 306:
```suggestion
> **Short summary**: the class `KotlinGradleSubplugin` will be removed in Kotlin 1.7. Use the `KotlinCompilerPluginSupportPlugin` instead
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Deprecate KotlinGradleSubplugin class
+
+> **Issue**: [KT-48830](https://youtrack.jetbrains.com/issue/KT-48830)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: the class `KotlinGradleSubplugin` will be removed in Kotlin 1.7. Use the `KotlinCompilerPluginSupportPlugin` instead
```
PetrakovichVictoria, line 306:
```suggestion
> **Short summary**: remove the class `KotlinGradleSubplugin`. Use the `KotlinCompilerPluginSupportPlugin` instead
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,467 @@
...
+### Deprecate kotlin.compiler.execution.strategy system property
+
+> **Issue**: [KT-51830](https://youtrack.jetbrains.com/issue/KT-51830)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: deprecate the `kotlin.compiler.execution.strategy` system property used to choose a compiler execution strategy. 
+> The Gradle property `kotlin.compiler.execution.strategy` or the compile task property `compilerExecutionStrategy` can be used instead
```
PetrakovichVictoria, line 384:
```suggestion
> `use the Gradle property `kotlin.compiler.execution.strategy` or the compile task property `compilerExecutionStrategy` instead
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,467 @@
...
+> **Short summary**: deprecate the `kapt.use.worker.api` property that allowed to run kapt via Gradle Workers API (default: true)
+>
+> **Deprecation cycle**:
+>
+> - 1.6.20: raise the deprecation level to a warning
+> - \>= 1.8.0: remove this property
+
+### Remove kotlin.experimental.coroutines Gradle DSL option and kotlin.coroutines Gradle property
+
+> **Issue**: [KT-50494/](https://youtrack.jetbrains.com/issue/KT-50494)
```
PetrakovichVictoria, line 381:
```suggestion
> **Issue**: [KT-50494](https://youtrack.jetbrains.com/issue/KT-50494)
```

docs/topics/compatibility-guide-17.md
ilya-g: It can be seen not as a fixed version, but as a some unknown version with that constraint. Since the version is not yet clear, they may get their plan clarified individually. 
I don't see anything wrong in having several list items with the same version if it is done on purpose.

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+>
+> **Short summary**: Kotlin 1.4–1.6 mistakenly neglected some type constraints during type inference due to an incorrect optimization.
+> It could allow writing unsound code causing `ClassCastException` at runtime.
+> Kotlin 1.7 takes these constraints into account thus prohibiting the unsound code
+>
+> **Deprecation cycle**:
+>
+> - 1.5.20: report a warning if a type mismatch would happen should all the type inference constraints have been taken into account 
+> - 1.7.0: take all the constraints into account thus raising this warning to an error,  
+> `-XXLanguage:-ProperTypeInferenceConstraintsProcessing` can be used to temporarily revert to the pre-1.7 behavior
```
ilya-g, line 257: Added these two issues, please review them as well

docs/topics/compatibility-guide-17.md
ilya-g: Then I'd use >= 1.8 in other places for consistency.

ilya-g:
When merging this PR could we squash the commits in the way so the commits of different authors remained separate? It would help in future to find who can clarify this or that issue.

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+> **Incompatible change type**: source
+>
+> **Short summary**: remove the `kotlin2js` and `kotlin-dce-plugin` plugins
+>
+> **Deprecation cycle:**
+>
+> - 1.4.0: raise the deprecation level to a warning
+> - 1.7.0: remove the plugins
+
+### Remove KotlinGradleSubplugin class
```
ilya-g, line 472: Duplicates the first issue in this section

docs/topics/compatibility-guide-17.md
ilya-g: Hmm, it is supposed to work this way in markdown. What is the way then to insert a line break in webhelp md dialect?

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+### Prohibit smart cast on base class property if the base class is from another module
+
+> **Issue**: [KT-52629](https://youtrack.jetbrains.com/issue/KT-52629)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.7 will no longer allow smart casts on properties declared in another module 
+> that were erroneously allowed before
```
danil-pavlov, line 233:
```suggestion
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+>
+> **Short summary**: Kotlin 1.6 no longer allows calling user functions named `suspend` that have the single argument of a functional type passed as a trailing lambda
+>
+> **Deprecation cycle**:
+>
+> - 1.3.0: introduce a warning on such function calls
+> - 1.6.0: raise this warning to an error
+> - 1.7.0: introduce changes to the language grammar so that `suspend` before `{` is parsed as a keyword
+
+### Prohibit smart cast on base class property if the base class is from another module
```
danil-pavlov, line 224:
```suggestion
### Prohibit smart cast on a base class property if the base class is from another module
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.7 will no longer allow smart casts on properties declared in another module 
+> that were erroneously allowed before
+>
+> **Deprecation cycle**:
+>
+> - 1.6.0: report a warning on a smart cast of a property declared in the superclass that is in another module
```
danil-pavlov, line 237:
```suggestion
> - 1.6.0: report a warning on a smart cast on a property declared in the superclass of another module
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+
+> **Issue**: [KT-52668](https://youtrack.jetbrains.com/issue/KT-52668)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.4–1.6 mistakenly neglected some type constraints during type inference due to an incorrect optimization.
+> It could allow writing unsound code causing `ClassCastException` at runtime.
+> Kotlin 1.7 takes these constraints into account thus prohibiting the unsound code
```
danil-pavlov, line 251:
```suggestion
> Kotlin 1.7 takes these constraints into account, thus prohibiting the unsound code
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+
+### Do not neglect meaningful constraints during type inference
+
+> **Issue**: [KT-52668](https://youtrack.jetbrains.com/issue/KT-52668)
+>
+> **Component**: Core language
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.4–1.6 mistakenly neglected some type constraints during type inference due to an incorrect optimization.
```
danil-pavlov, line 249:
```suggestion
> **Short summary**: Kotlin 1.4—1.6 neglected some type constraints during type inference due to an incorrect optimization.
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+> It could allow writing unsound code causing `ClassCastException` at runtime.
+> Kotlin 1.7 takes these constraints into account thus prohibiting the unsound code
+>
+> **Deprecation cycle**:
+>
+> - 1.5.20: report a warning if a type mismatch would happen should all the type inference constraints have been taken into account 
+> - 1.7.0: take all the constraints into account thus raising this warning to an error,
```
danil-pavlov, line 256:
```suggestion
> - 1.7.0: take all the constraints into account, thus raising this warning to an error,  
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin 1.4–1.6 mistakenly neglected some type constraints during type inference due to an incorrect optimization.
+> It could allow writing unsound code causing `ClassCastException` at runtime.
+> Kotlin 1.7 takes these constraints into account thus prohibiting the unsound code
+>
+> **Deprecation cycle**:
+>
+> - 1.5.20: report a warning if a type mismatch would happen should all the type inference constraints have been taken into account
```
danil-pavlov, line 255:
```suggestion
> - 1.5.20: report a warning when a type mismatch should happen if all the type inference constraints were taken into account
```

danil-pavlov (approved):

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+
+### Remove noStdlib compiler option
+
+> **Issue**: [KT-49011](https://youtrack.jetbrains.com/issue/KT-49011)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: remove the `noStdlib` compiler option. Gradle always resolves the Kotlin standard library and adds it to the compiler classpath
```
Tapchicoma, line 450: > Gradle always resolves the Kotlin standard library and adds it to the compiler classpath

This statement is partially true. By default KGP adds stdlib as a normal dependency, but user could disable it via following gradle property:
```
kotlin.stdlib.default.dependency=false
```

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+
+### Remove kotlin2js and kotlin-dce-plugin plugins
+
+> **Issue**: [KT-48276](https://youtrack.jetbrains.com/issue/KT-48276)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: remove the `kotlin2js` and `kotlin-dce-plugin` plugins
```
Tapchicoma, line 465: Probably worth to add what is the proposed replacement?

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,502 @@
...
+> **Issue**: [KT-32805](https://youtrack.jetbrains.com/issue/KT-32805)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: Kotlin compile tasks no longer inherit the Gradle `AbstractCompile` task and that's why 
+> the `sourceCompatibility` and `targetCompatibility` inputs are no longer available in Kotlin users' scripts. 
+> The `SourceTask.stableSources` input is no longer available. The `sourceFilesExtensions` input was removed. 
+> The deprecated `Gradle destinationDir: File` output was replaced with the `destinationDirectory: DirectoryProperty` output
```
Tapchicoma, line 498: Also worth to mention `getClasspath()` and `setClasspath()` methods deprecation:
> All compile tasks are now using `libraries` input for list of compilation required libraries. `KotlinCompile` still has deprecated `setClasspath(FileCollection)` and `getClasspath(): FileCollection` methods.

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: Indeed, I'm sorry. Fixed

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: True. Added

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: Nevertheless, seems like we can leave it as is. Wrote in DM

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: When were these methods deprecated?

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: Added the same as to What's new

docs/topics/compatibility-guide-17.md
Tapchicoma: In KT-32805 :slightly_smiling_face:

I've wrote about it in my summary comment in this issue.

docs/topics/compatibility-guide-17.md
ilya-g: What's wrong with using en-dash for a range? https://www.thepunctuationguide.com/en-dash.html

docs/topics/compatibility-guide-17.md
ilya-g: @Tapchicoma isn't there the single deprecated `classpath: FileCollection` property instead?
> KotlinCompile still has deprecated setClasspath(FileCollection) and getClasspath(): FileCollection methods

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,489 @@
...
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: remove the `noStdlib` compiler option. The Gradle plugin uses the `kotlin.stdlib.default.dependency=true` property to control the Kotlin standard library presence
+>
+> **Deprecation cycle:**
+>
+> - 1.5.0: raise the deprecation level to a warning
```
ilya-g, line 454: Note: usually deprecation level is _raised_ when it was a warning and becomes an error. When the warning appears first time, we say that we introduce a warning, report a warning, and so on.

docs/topics/compatibility-guide-17.md
ilya-g: The intent here is to show that it was not in all cases of properties in another module, but only in those that were mistakenly allowed

docs/topics/compatibility-guide-17.md
```diff
@@ -0,0 +1,489 @@
...
+
+### Remove noStdlib compiler option
+
+> **Issue**: [KT-49011](https://youtrack.jetbrains.com/issue/KT-49011)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: remove the `noStdlib` compiler option. The Gradle plugin uses the `kotlin.stdlib.default.dependency=true` property to control the Kotlin standard library presence
```
PetrakovichVictoria, line 450:
```suggestion
> **Short summary**: remove the `noStdlib` compiler option. The Gradle plugin uses the `kotlin.stdlib.default.dependency=true` property to control whether the Kotlin standard library is present

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: > still has deprecated
doesn't sound as they are deprecared in this release, it looks like they were deprecated some time ago

PetrakovichVictoria: Added in this commit https://github.com/JetBrains/kotlin-web-site/pull/2983/commits/e05244ba5e7b5e27b53e4d2d74b1d5d56f8406ed

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: Ok, I'll fix in my Tools features

PetrakovichVictoria: Actually, in the previos compatibility guide we used the following wording:
> - 1.5.0: raise the deprecation level to a warning

> - 1.5.0: raise the deprecation level to an error

Maybe we should keep in mind your note for future compatibility guides.

docs/topics/compatibility-guide-17.md
danil-pavlov: Then we need to specify what kind of cases were incorrectly allowed, because now it doesn't make much sense — "Kotlin will no longer allow smarcasts... that were erroneously allowed before"

docs/topics/compatibility-guide-17.md
ilya-g: reworded differently

docs/topics/compatibility-guide-17.md
ilya-g: Still, what's about property `classpath` versus getter and setter methods? I though we provided the deprecated property for Kotlin Gradle DSL users.

ilya-g:
If no more changes are expected, I'm going to squash sequential commits keeping the authorship and rebase on master

docs/topics/compatibility-guide-17.md
PetrakovichVictoria: In What's new we wrote:
> The `KotlinCompile` task still has the deprecated Kotlin property `classpath`, which will be removed in future releases.

@Tapchicoma, should I write it instead of 
> The `setClasspath(FileCollection)` and `getClasspath(): FileCollection` methods are deprecated

?

Tapchicoma: yeah, @ilya-g is right. Should be `classpath` property instead of `setClasspath` and `getClasspath`

PetrakovichVictoria: Fixed in the separate PR https://github.com/JetBrains/kotlin-web-site/pull/3005

PR #2990, opened 2022-06-06
Update native-app-with-c-and-libcurl.md

SvyatoslavScherbina:


SvyatoslavScherbina:
The removed part was incorrect (see also https://youtrack.jetbrains.com/issue/KT-52522/Native-cinterop-fatal-error-h-file-not-found).

PR #2992, opened 2022-06-06
update: update compiler execution strategies

PetrakovichVictoria:


docs/topics/gradle.md
```diff
@@ -975,16 +975,16 @@ _Kotlin compiler execution strategy_ defines where the Kotlin compiler is execut
...
+| Out of process | In a separate process for each call  | No                      | -                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
Tapchicoma, line 982: Probably worth to explain as well:

> Same as `in-process`, but instead of using Gradle workers - each compilation will run in a separate Java process. The slowest of the three execution strategies

PetrakovichVictoria: > each compilation will start and run in a separate Java process

We already have it in the second column:
> In a separate process for each call

Is it really worth mentioning here that it is
> instead of using Gradle workers
?

This is the inner kitchen of it, in the forth column for the "In process" we mention workers in terms of JIT optimization and as one of the reasons why this strategy is slow. Not because of a user really needs to know what's under the hood.

PetrakovichVictoria: And I wanted to consult both of you: won't we frighten users telling that these strategies are slow? Won't it make them using only the default strategy? cc @ALikhachev

PetrakovichVictoria: Ok, seems like we can write that they are slow, and it's all a user's responsibility to use slower strategies knowing this fact about them, yeah?

docs/topics/gradle.md
```diff
@@ -975,16 +975,16 @@ _Kotlin compiler execution strategy_ defines where the Kotlin compiler is execut
...
+| Out of process | In a separate process for each call  | No                      | The slowest execution strategy.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
Tapchicoma, line 982: Probably better to replace `for each call` with `for each compilation` to make it easier to understand.

docs/topics/gradle.md
```diff
@@ -975,16 +975,16 @@ _Kotlin compiler execution strategy_ defines where the Kotlin compiler is execut
...
+| In process     | Inside the Gradle daemon process     | No                      | May share the heap with the Gradle daemon. Note that the in process execution is *slower* than the Daemon execution strategy. This is so because Gradle spawns multiple [workers](https://docs.gradle.org/current/userguide/worker_api.html) with the Kotlin compiler, and they could be not yet [JIT-optimized](https://en.wikipedia.org/wiki/Just-in-time_compilation#:~:text=In%20computing%2C%20just%2Din%2D,time)%20rather%20than%20before%20execution.) when the compilation runs. Also, the Kotlin daemon has additional caches. |
```
Tapchicoma, line 981: Replace:

> This is so because Gradle spawns multiple [workers](https://docs.gradle.org/current/userguide/worker_api.html) with the Kotlin compiler, and they could be not yet [JIT-optimized](https://en.wikipedia.org/wiki/Just-in-time_compilation#:~:text=In%20computing%2C%20just%2Din%2D,time)%20rather%20than%20before%20execution.) when the compilation runs. Also, the Kotlin daemon has additional caches.

with:

> In this mode plugin on each compilation inside Gradle worker running with no-isolation mode creates a separate classloader to execute Kotlin compiler.

docs/topics/gradle.md
Tapchicoma, line 982: Imho, still better to explain the difference:

> The slowest execution strategy. Similar to 'in-process', but additionally creates a separate Java process within a Gradle worker to execute a Kotlin compiler.

docs/topics/gradle.md
```diff
@@ -975,16 +975,16 @@ _Kotlin compiler execution strategy_ defines where the Kotlin compiler is execut
...
+| Daemon         | Inside its own daemon process        | Yes                     | *The default strategy*. Can be shared between different Gradle daemons.                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
Tapchicoma, line 980: Probably worth to add in the end:

> and multiple parallel compilations

docs/topics/gradle.md
PetrakovichVictoria: Ok, I'll add.

docs/topics/gradle.md
PetrakovichVictoria: I'll add.

docs/topics/gradle.md
PetrakovichVictoria: Seems like I can resolve this?

docs/topics/gradle.md
PetrakovichVictoria: Discussed in DM

docs/topics/gradle.md
```diff
@@ -975,16 +975,16 @@ _Kotlin compiler execution strategy_ defines where the Kotlin compiler is execut
...
+|----------------|--------------------------------------------|-------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Daemon         | Inside its own daemon process              | Yes                     | *The default and the fastest strategy*. Can be shared between different Gradle daemons and multiple parallel compilations.                                                                                                                                     |
+| In process     | Inside the Gradle daemon process           | No                      | May share the heap with the Gradle daemon. The "In process" execution strategy is *slower* than the "Daemon" execution strategy. Each [worker](https://docs.gradle.org/current/userguide/worker_api.html) creates a separate classloader for each compilation. |
```
Tapchicoma, line 981: Probably better to clarify:
> a separate classloader

to 

> a separate Kotlin compiler classloader

Tapchicoma (approved):

PR #3003, opened 2022-06-09
Fix capitalization of Xcode

tiwoc:
Fixes a typo: Xcode is spelled with a lower-case c (see [Apple Developer](https://developer.apple.com/xcode/)).

This is the only instance of this typo in this repo.


koshachy (approved):
Thanks!

PR #3007, opened 2022-06-10
Update idioms.md

rk012:
Add idiom for smart casting a nullable type.


dzharkov:
Hi @rk012!
I really like that early-checking-then-return approach, but this exactly kind of idiom looks rather controversial to me. 
The first thing is that `?:` operator due to its expression's nature looks unnatural in the statement position.
On the other hand, its counterpart `if (value == null) return` looks much clearer especially when read out loud and at the same time it's not too wordy.

koshachy:
I close this PR.
If you got any questions, don't hesitate to contact me or just re-open this PR.

PR #3013, opened 2022-06-14
Fix spelling of Javadoc

rillig:
Oracle's documentation spells the tool name as Javadoc, not JavaDoc.


koshachy (approved):
That's true, Roland :)
Thanks for pointing this out.

PR #2991, opened 2022-06-06
add: add about -jdk-release option

PetrakovichVictoria:

PR #3018, opened 2022-06-15
fix: -xjdk-release guaranterely works only for OpenJDK

PetrakovichVictoria:


docs/topics/compiler-reference.md
```diff
@@ -159,7 +159,7 @@ Use a custom JDK home directory to include into the classpath if it differs from
 ### -Xjdk-release=version
 
 Specify the target version of the generated JVM bytecode. Limit the API of the JDK in the classpath to the specified Java version. 
-Automatically sets [`-jvm-target version`](#jvm-target-version). Works only for [OpenJDK distributions](https://openjdk.java.net/projects/jdk/18/).
+Automatically sets [`-jvm-target version`](#jvm-target-version). The effect of use is [guaranteed](https://youtrack.jetbrains.com/issue/KT-29974) only for [OpenJDK distributions](https://openjdk.java.net/projects/jdk/18/).
```
max-kammerer, line 162: I think it's best to avoid usage of any specific JDK distributions and just write that it's not guranteed to work on all JDK disributions like suggested in https://github.com/JetBrains/kotlin-web-site/pull/2991#discussion_r894456634

PetrakovichVictoria: Oops, indeed, by some reason I had the page opened without the udalov's comment, sorry.

docs/topics/compiler-reference.md
```diff
@@ -159,9 +159,13 @@ Use a custom JDK home directory to include into the classpath if it differs from
 ### -Xjdk-release=version
 
 Specify the target version of the generated JVM bytecode. Limit the API of the JDK in the classpath to the specified Java version. 
-Automatically sets [`-jvm-target version`](#jvm-target-version). Works only for [OpenJDK distributions](https://openjdk.java.net/projects/jdk/18/).
+Automatically sets [`-jvm-target version`](#jvm-target-version).
 Possible values are `1.8`, `9`, `10`, ..., `18`. The default value is `%defaultJvmTargetVersion%`.
 
+> This option is not [guaranteed](https://youtrack.jetbrains.com/issue/KT-29974) to be effective for each JDK distribution.
```
koshachy, line 165: I guess **not guaranteed** should be marked as a link, not **guaranteed**.

Could we provide a link to a particular comment in this issue? It's not obvious what statement in the issues explains about this guarantee.

PetrakovichVictoria: > I guess not guaranteed should be marked as a link, not guaranteed.

Indeed. 

> Could we provide a link to a particular comment in this issue? It's not obvious what statement in the issue explains about this guarantee.

I guess no, because it's in the [ticket's description](https://youtrack.jetbrains.com/issue/KT-29974):
> To load the version information on different declarations (akin to Kotlin's @SinceKotlin), we need to load and parse the ct.sym file from the OpenJDK distribution. Sadly, this is not a part of the spec, so this file is absent in other JDK distributions, in which case we should report an error/warning saying this option is only supported when compiling against OpenJDK.

docs/topics/whatsnew17.md
```diff
@@ -210,7 +210,7 @@ target bytecode version and limits the API of the JDK in the classpath to the sp
 example, `kotlinc -Xjdk-release=1.8` won't allow referencing `java.lang.Module` even if the JDK in the dependencies is
 version 9 or higher.
 
-> This option works only for OpenJDK distributions.
+> This option is not [guaranteed](https://youtrack.jetbrains.com/issue/KT-29974) to be effective for each JDK distribution.
```
PetrakovichVictoria, line 213:
```suggestion
> This option is [not guaranteed](https://youtrack.jetbrains.com/issue/KT-29974) to be effective for each JDK distribution.
```

koshachy (approved):

docs/topics/compiler-reference.md
koshachy: TBH, this sentence does not give me enough information to make it clear, when it applies and when it doesn't apply.

PetrakovichVictoria: I agree, but now there's the incorrect info on the site, so let's mege this and try to invent something more correct if we wish.

Also, this discussion started in another PR, the comment is [here](https://github.com/JetBrains/kotlin-web-site/pull/2991#discussion_r894456634). I can't invent right now how to make it less vague.

PR #3031, opened 2022-06-20
Fix invalid C string conversion references

anton-novikau:
In kotlin `1.7.0` the extension function `ByteArray. stringFromUtf8()` referenced in
the `C interop` tutorial is not available anymore. The invalid references are replaced
with `ByteArray.decodeToString()`.


koshachy (approved):

koshachy:
Hey, thanks for your suggestions!
We really appreciate that :)

PR #3033, opened 2022-06-21
KTL-764 kotlin.Result extension function sample provides kotlin.Collections samples

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-764/kotlinResult-extension-function-sample-provides-kotlinCollections-samples

It is just a quick fix, and then we will review all redirects to find and remove redundant.


ilya-g:
Looking at the redirects list, there seem to be many other redirects preventing navigation to some API pages.
I'd insist on clearing redirects from API documentation thoroughly. Let's discuss in DM how to do it.

PR #3034, opened 2022-06-21
Rephrase `final` requirement for inline classes

SebastianAigner:
Saying "must be `final`" could be misinterpreted as "needs `final` modifier in order to work". Since `final` is the default, I propose rephrasing it to show that finality is _always_ the case.


koshachy (approved):
Thanks, @SebastianAigner!

PR #3035, opened 2022-06-22
update: bump the minimal Gradle versions

koshachy:
update minimal Gradle and Android Gradle plugin versions
add microformat markup


danil-pavlov (approved):
Good idea!

PR #3036, opened 2022-06-23
Explicitly state lack of support for bytecode < Java 8

SebastianAigner:
> 'compileJava' task (current target is 16) and 'compileKotlin' task (current target is 1.6) jvm target compatibility should be set to the same Java version.
> w: JVM target 1.6 is deprecated and will be removed in a future release. Please migrate to JVM target 1.8 or above

Historically, Kotlin had supported compiling all the way down to JVM 1.6 targets, and you can still find references to this on [StackOverflow](https://stackoverflow.com/questions/48036758/is-the-jdk-required-for-kotlin).

Since this target is deprecated, it makes sense to state the requirement of targeting Java 8 compatible code explicitly.

### DIFF
--- a/docs/topics/faq.md
+++ b/docs/topics/faq.md
@@ -96,7 +96,7 @@ When targeting native, Kotlin will produce platform-specific code (via LLVM).
 Kotlin lets you choose the version of JVM for execution. By default, the Kotlin/JVM compiler produces Java 8 compatible bytecode.
 If you want to make use of optimizations available in newer versions of Java, you can explicitly specify the target Java
 version from 9 to 18. Note that in this case the resulting bytecode might not run on lower versions.
-Starting with [Kotlin 1.5](https://kotlinlang.org/docs/whatsnew15.html#new-default-jvm-target-1-8), the compiler does not support producing bytecode compatible with Java versions below 8.
+Starting with [Kotlin 1.5](whatsnew15.md#new-default-jvm-target-1-8), the compiler does not support producing bytecode compatible with Java versions below 8.
 
 ### Is Kotlin hard?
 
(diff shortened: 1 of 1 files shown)


SebastianAigner:
(I've also submitted an edit to the [StackOverflow answer](https://stackoverflow.com/a/48036888/13770600) which is currently waiting for review.)

koshachy:
Hey @SebastianAigner 
I'm not quite sure that we should write about it explicitly, because the previous sentence does the job.

Our documentation reflects the latest version of the language. We usually use "Starting with 1.x.x" constructions in What's new docs, or if the change requires any migration steps from users (or if the feature changes the behavior completely).

So I'd like to leave things as they are. WDYT?

SebastianAigner:
I don't think the precious sentence covers this. It says "if you want to make use of optimisations in newer Java versions". However, it doesn't mention anything about compatibility with older versions.

It took me about half an hour of googling before finding the actual resource that stated that JVM 1.6 is deprecated, so I think it would be useful to mention this here again. We can try to find a different phrasing, but I feel this is still useful to people who come in with the historical knowledge that 1.6 was once supported.

koshachy:
Ok, I see. So let's merge this and then think about how to improve our documentation.

koshachy (approved):

docs/topics/faq.md
```diff
@@ -95,7 +95,8 @@ When targeting native, Kotlin will produce platform-specific code (via LLVM).
 
 Kotlin lets you choose the version of JVM for execution. By default, the Kotlin/JVM compiler produces Java 8 compatible bytecode.
 If you want to make use of optimizations available in newer versions of Java, you can explicitly specify the target Java
-version from 9 to 18. Note that in this case the resulting bytecode might not run on lower versions. 
+version from 9 to 18. Note that in this case the resulting bytecode might not run on lower versions.
+Starting with [Kotlin 1.5](https://kotlinlang.org/docs/whatsnew15.html#new-default-jvm-target-1-8), the compiler does not support producing bytecode compatible with Java versions below 8.
```
koshachy, line 99:
```suggestion
Starting with [Kotlin 1.5](whatsnew15.md#new-default-jvm-target-1-8), the compiler does not support producing bytecode compatible with Java versions below 8.
```

PR #3041, opened 2022-06-28
update: update information about Gradle configuration cache

koshachy:
[KT-53001](https://youtrack.jetbrains.com/issue/KT-53001) Update the information on the Gradle Configuration Cache

### DIFF
--- a/docs/topics/gradle.md
+++ b/docs/topics/gradle.md
@@ -747,12 +747,13 @@ If you use [kapt](kapt.md), note that kapt annotation processing tasks are not c
 ## Gradle configuration cache support
 
 > The configuration cache is available in Gradle 6.5 and later as an experimental feature.
+> It is available only on JVM and JS platforms.
 > You can check the [Gradle releases page](https://gradle.org/releases/) to see whether it has been promoted to stable.
 >
 {type="note"}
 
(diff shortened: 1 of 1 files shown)


koshachy:
Just a quick fix for the documentation

antohaby (approved):

docs/topics/gradle.md
```diff
@@ -747,12 +747,13 @@ If you use [kapt](kapt.md), note that kapt annotation processing tasks are not c
 ## Gradle configuration cache support
 
 > The configuration cache is available in Gradle 6.5 and later as an experimental feature.
+> It is available only on JVM and JS platforms.
```
Tapchicoma, line 750: Probably better to say which Gradle plugins support it:
- 'org.jetbrains.kotlin.jvm'
- 'org.jetbrains.kotlin.js'
- 'org.jetbrains.kotlin.android'

Using Platfrom approach suggests MPP/JS is supported, while most probably it may be not

koshachy: Oh, I see!
Thanks for clarification!

danil-pavlov (approved):

Tapchicoma (approved):

PR #3042, opened 2022-06-28
Update of destination directory configuration sample

wild-lynx:
According to the replacement of the deprecated API made in 1.7.0. See https://kotlinlang.org/docs/whatsnew17.html#changes-in-compile-tasks for details.


koshachy (approved):
LGTM!
Thanks!

PR #3047, opened 2022-06-30
update: add link to the FIR template for K2

koshachy:


danil-pavlov (approved):
@koshachy No, I'd leave it like that for the sake of consistency. we don't use notes for similar feedback links in this document

PR #3052, opened 2022-07-07
[K/N] Add a section about -Xexport-kdoc to native-objc-interop.md

sbogolepov:


docs/topics/native/native-objc-interop.md
```diff
@@ -409,6 +409,47 @@ this library would disable these compiler checks.
 See [Interoperability with C](native-c-interop.md) for an example case where the library uses some plain C features,
 such as unsafe pointers, structs, and so on.
 
+## Experimental features
```
SvyatoslavScherbina, line 412: Do we really need to group experimental features under the dedicated section?
We already have a partially experimental feature here: https://github.com/JetBrains/kotlin-web-site/blob/b9dd48d8ca233b72b92e13984cf6171dd1bd1781/docs/topics/native/native-objc-interop.md#suspending-functions
Generally, it might make sense to simply add stability status mentions to the feature descriptions instead.

docs/topics/native/native-objc-interop.md
```diff
@@ -409,6 +409,47 @@ this library would disable these compiler checks.
...
+```objective-c
+/**
+ * Prints the sum of the arguments.
+ * Properly handles the case when the sum doesn't fit in 32-bit integer.
+ */
++ (void)printSumA:(int32_t)a b:(int32_t)b __attribute__((swift_name("printSum(a:b:)")));
+```
+
+Known limitations:
+* Dependency documentation is not exported unless it is compiled with `-Xexport-kdoc` itself.
```
SvyatoslavScherbina, line 449: Should we add a caution like this:
> The feature is experimental, so libraries compiled with this flag might be incompatible with other compiler versions.

?

docs/topics/native/native-objc-interop.md
```diff
@@ -409,6 +409,47 @@ this library would disable these compiler checks.
...
+* Commentaries are mostly exported "as is" and many KDoc features (for example, `@property`) are not supported.
+* The feature is not properly tested, so it might break generated header.
```
SvyatoslavScherbina, line 451: Sounds too strong. Why do we need this statement here at all?

docs/topics/native/native-objc-interop.md
sbogolepov: An explicit list of opt-ins makes it easier to distinguish a set of features that require some action from the user (passing an additional compiler flag in this case). Otherwise, it will make reading of the document harder because reader will have to constantly make "context switch" between opt-in and opt-out features.

docs/topics/native/native-objc-interop.md
sbogolepov: Yep, we should.

docs/topics/native/native-objc-interop.md
sbogolepov: Yeah, let's drop it. Experimental "label" should be enough.

docs/topics/native/native-objc-interop.md
koshachy: I don't think that it works in that way. When users come to the docs, they usually don't want to know which are experimental or not.
Otherwise, we should move suspend functions text to the Experimental section. I'd recommend eliminating the Experimental features section and adding a typical Experimental note (same as for suspend).

docs/topics/native/native-objc-interop.md
```diff
@@ -409,6 +409,47 @@ this library would disable these compiler checks.
 See [Interoperability with C](native-c-interop.md) for an example case where the library uses some plain C features,
 such as unsafe pointers, structs, and so on.
 
+## Experimental features
+
+### Export KDoc to generated Objective-C headers
+
+By default, [KDocs](../kotlin-doc.md) are not translated into corresponding commentaries when generating an Objective-C header.
```
koshachy:
```suggestion
By default, [KDoc](kotlin-doc.md) comments are not translated into corresponding comments when generating an Objective-C header.
```

docs/topics/native/native-objc-interop.md
```diff
@@ -409,6 +409,47 @@ this library would disable these compiler checks.
...
+ * Properly handles the case when the sum doesn't fit in 32-bit integer.
+ */
+fun printSum(a: Int, b: Int) = println(a.toLong() + b)
+```
+will produce an Objective-C declaration without a commentary:
+```objective-c
++ (void)printSumA:(int32_t)a b:(int32_t)b __attribute__((swift_name("printSum(a:b:)")));
+```
+
+KDoc export can be enabled by adding the following lines to `build.gradle.kts`:
```
koshachy:
```suggestion
To enable KDoc comments export, add the following to `build.gradle.kts`:
```

koshachy (approved):
I've slightly updated this section and eliminated the Experimental heading

PR #3058, opened 2022-07-12
update: range expression in comment

kasisoft:


koshachy:
But I think that I understand your concern: would it be clearer if there will be `i >=1 && && i <= 4`?

PR #3065, opened 2022-07-18
Remove links to outdated Kotlin/Native samples

SvyatoslavScherbina:


koshachy (approved):
LGTM.

@danil-pavlov should we add some other links instead of eliminated in the "What's new" section? WDYT?

danil-pavlov:
> should we add some other links instead of eliminated in the "What's new" section? WDYT?

@koshachy We don't have other examples for the c-libcurl tutorial. We could add a link to the C-interop page, but we already have it inside the article. so I think we can leave it as is

danil-pavlov (approved):

PR #3072, opened 2022-07-20
Replace non-Java-Identifier const by constant

IrgendNJulian:
const is in Java a keyword and therefore not a valid identifier.
So the example Java Code 
```java
int const = Obj.CONST;
int max = ExampleKt.MAX;
int version = C.VERSION;
```
about using const properties in Java wouldn't compile.
I propose renaming the variable const to constant.

### DIFF
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -200,7 +200,7 @@ In Java:
 
 ```java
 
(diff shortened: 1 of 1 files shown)


danil-pavlov (changes requested):
Hi! Thanks for your contribution. I propose changing the variable name for this constant in the Kotlin code above as well to make it consistent with the new variable name for Java

danil-pavlov (approved):

PR #3073, opened 2022-07-20
update: multiplatform libs structure

danil-pavlov:
* first part of the multiplatform libs section revamp


docs/topics/multiplatform/multiplatform-library.md
```diff
@@ -27,7 +27,24 @@ Download and install the [latest version of IntelliJ IDEA](https://www.jetbrains
...
+
+   * Specify dependencies between modules:
+   
+     * Multiplatform and Android modules
+     * Multiplatform and iOS modules
+     * JVM modules
+
+       ![Add module dependencies](multiplatform-project-3.png){width=700}
+
+9. Click **Finish**.
```
koshachy, line 47:
```suggestion
8. Click **Finish**.
```

docs/topics/multiplatform/multiplatform-library.md
```diff
@@ -27,7 +27,24 @@ Download and install the [latest version of IntelliJ IDEA](https://www.jetbrains
 
 4. Select the Gradle DSL – Kotlin or Groovy.
 5. Specify the [JDK](https://www.jetbrains.com/help/idea/sdk.html#jdk), which is required for developing Kotlin projects.
-5. Click **Next**, and then click **Finish**.
+6. Click **Next**.
+7. You can finish creating the project by clicking **Finish** on the next screen or configure it if necessary:
```
koshachy, line 31: Should it be a separate step?

If so, we could write

"7. Configure the project if necessary:"

koshachy: Usually, tutorials provide some exact steps what to do.
Here I see options, and there is no info about their difference (or what reader should select to move on)

koshachy (changes requested):

docs/topics/multiplatform/multiplatform-library.md
danil-pavlov: That's not actually a step for this tutorial's project, but the team asked to add a description of this wizard window, because the UX here is very confusing 
I changed the text to make it separate from the tutorial

koshachy (approved):

PR #3078, opened 2022-07-25
Fix typo in sealed classes `when` example

scolsen:
The final branch was missing an `is` prior to the matched case.


koshachy (approved):
Nice catch, thank you! ;)
We appreciate your contribution to the Kotlin documentation.

PR #3107, opened 2022-08-13
Updating example for Anonymous classes

iamraajkanchan:
Adding a more useful example to understand anonymous classes for enum


danil-pavlov:
Hi! Thank you for your example, but after consulting with our development team, we decided that the example we currently have is sufficient for this particular case

PR #3128, opened 2022-08-26
KTL-817 Fix search builder: an error occurred during docs indexation

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-817/Fix-search-builder-an-error-occurred-and-new-pages-are-skipped

### DIFF
--- a/src/search.py
+++ b/src/search.py
@@ -114,7 +114,7 @@ def get_valuable_content(page_path, content: Iterator[Tag]) -> List[str]:
             valuable_content.append(child.text)
         elif child.name in ['ul', 'ol', 'blockquote', 'div', 'section', 'dl']:
             valuable_content += get_valuable_content(page_path, child.children)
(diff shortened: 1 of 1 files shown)


src/search.py
```diff
@@ -188,7 +188,7 @@ def get_webhelp_page_index_objects(content: Tag, url: str, page_path: str, title
             chapter_title_node = chapter.select_one('[data-toc]')
             if chapter_title_node:
                 chapter_title = chapter_title_node.extract().text
-                chapter_title_anchor = chapter_title_node.attrs["data-toc"].split('#')[1]
```
nikpachoo, line 191: Anchors in right ToC had these format:
`data-toc="multiplatform#kotlin-multiplatform-use-cases"`
And now they look like:
`data-toc="kotlin-multiplatform-use-cases"`
So, splitting by # is obsolete.

PR #3133, opened 2022-08-29
specifically call out the handling of .toString()

elliottswan-toast:
I found this behavior confusing and worth specifically calling out

### DIFF
--- a/docs/topics/null-safety.md
+++ b/docs/topics/null-safety.md
@@ -61,6 +61,11 @@ But if you want to access the same property on `b`, that would not be safe, and 
 val l = b.length // error: variable 'b' can be null
 ```
 
+An exception is `.toString()` [which returns the string "null" if called on a null object](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to-string.html)
+```kotlin
+null.toString() // returns "null"
+```
+
 But you still need to access that property, right? There are a few ways to do so.
 
 ## Checking for `null` in conditions


koshachy:
Hey there and thank you for your contribution!
I don't think that this "exception" is useful in this context (and it's not an exception actually).
I'd rather decline your suggestion.

elliottswan-toast:
I considered it an exception because every other attempt to call a function or access a property on a nullable object will result in a compile time error if it is not made a null-safe call.  By not enforcing this and allowing `.toString()` to be called on a null object it's possible to miss a bug at compile time.  I'm open to other ways to document this, but I think it should be included on the page about null safety, since this behavior of `.toString()` seems to be related to null safety.

SebastianAigner:
So the problem with this explanation is that it's not `toString` that's magical, it's that `toString` happens to be defined on a receiver of type `Any?` – a nullable receiver (https://kotlinlang.org/docs/extensions.html#nullable-receiver). In fact, the example on the nullable receiver page explains exactly how `toString` works 😃 

So, I think we may want to add a sentence to the page on nullability that mentions Nullable receivers and links the corresponding docs page.

elliottswan-toast:
> So the problem with this explanation is that it's not `toString` that's magical, it's that `toString` happens to be defined on a receiver of type `Any?` – a nullable receiver (https://kotlinlang.org/docs/extensions.html#nullable-receiver). In fact, the example on the nullable receiver page explains exactly how `toString` works 😃
> 
> So, I think we may want to add a sentence to the page on nullability that mentions Nullable receivers and links the corresponding docs page.

Yeah I think that would make sense.  The docs on `toString()` are great, but as a seasoned Java dev I'm not likely to read those until this happens to me.  The docs on null safety are IMHO the best launching off point to further details on the differences between Java and Kotlin from a nullability perspective.  

Also, are there many other classic Java functions built into the language that are defined on nullable receivers?

SebastianAigner:
> Also, are there many other classic Java functions built into the language that are defined on nullable receivers?

I wouldn't call them "classic Java functions" – they're usually helpers to make working with nullable objects a little bit smoother, such as [`isNullOrBlank`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-blank.html). I'd argue these are _rather_ self-explanatory in comparison, though, just based on their name.

elliottswan-toast:
> So, I think we may want to add a sentence to the page on nullability that mentions Nullable receivers and links the corresponding docs page.

I think this is a good approach, I made the edits

elliottswan-toast:
@SebastianAigner can you take another look?

SebastianAigner:
The reference seems fine to me, but I think there’s some minor typos, and I would ask @kosachy  to check whether this is all in accordance with our writing style :)

SebastianAigner:
@koshachy , since that ping was broken somehow.

elliottswan-toast:
@koshachy do you have any further thoughts on this?

koshachy:
Sorry for the late reply,
We'll come back to your PR in a few days. Were a bit busy with the 1.7.20 release.

docs/topics/null-safety.md
```diff
@@ -143,6 +143,17 @@ is `null`, the assignment is skipped and the expression on the right is not eval
 person?.department?.head = managersPool.getManager()
 ```
 
+## Nullable receiver
+
+Functions could be extended to use a [nullable receiver](extensions.md#nullable-receiver).
```
SebastianAigner, line 148: “Functions could be extended” is a confusing phrasing. How about “Extension functions can be defined on a nullable receiver”?

docs/topics/null-safety.md
```diff
@@ -143,6 +143,17 @@ is `null`, the assignment is skipped and the expression on the right is not eval
...
+This way you can check for the null condition and have a default defined in one place.
```
SebastianAigner, line 149: “Check for the null condition” sounds odd to me. Maybe something like “Specify behaviour for potentially null values without the need to use null-checking logic at each call-site” or something like that. Otherwise, the explanation focuses too much on “what you can do in the body”, rather than the actual effect this has (changing the calling convention)

docs/topics/null-safety.md
```diff
@@ -143,6 +143,17 @@ is `null`, the assignment is skipped and the expression on the right is not eval
 person?.department?.head = managersPool.getManager()
 ```
 
+## Nullable receiver
+
+Functions could be extended to use a [nullable receiver](extensions.md#nullable-receiver).
+This way you can check for the null condition and have a default defined in one place. 
+
+For example, the [`toString()` function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to-string.html) can be called with a nullable receiver and it returns the "null" string value (not `null`):
```
SebastianAigner, line 151: “Can be called with a nullable receiver” sounds like it’s a choice the user makes. “The toString function is defined on a nullable receiver” is probably more appropriate.
Also maybe something like
“…for null values, it returns the String ‘null’” to make clear when that behaviour happens.

docs/topics/null-safety.md
```diff
@@ -143,6 +143,17 @@ is `null`, the assignment is skipped and the expression on the right is not eval
...
+
+```kotlin
+null.toString() // returns "null"
```
SebastianAigner, line 154: Let’s maybe add an example that introduces an explicit nullable type like “String?” here. Free-standing null-literals are kind of icky (and we don’t want to invite people to break their brain by doing things like `null + null`, which is senseless code anyway 😉 (https://youtu.be/wwplVknTza4)

SebastianAigner (changes requested):

docs/topics/null-safety.md
elliottswan-toast: that is more precise, yeah [dee3b18](https://github.com/JetBrains/kotlin-web-site/pull/3133/commits/dee3b18d426b6f05ae9851974e57c3c4b955e3e6)

docs/topics/null-safety.md
elliottswan-toast: [dee3b18](https://github.com/JetBrains/kotlin-web-site/pull/3133/commits/dee3b18d426b6f05ae9851974e57c3c4b955e3e6)

docs/topics/null-safety.md
elliottswan-toast: I tried to simplify the bug I created by not understanding this and boiled it down to an example, let me know what you think: [dee3b18](https://github.com/JetBrains/kotlin-web-site/pull/3133/commits/dee3b18d426b6f05ae9851974e57c3c4b955e3e6) and [8f1a862](https://github.com/JetBrains/kotlin-web-site/pull/3133/commits/8f1a8629e2dda5c52c48dccd489ca972ce1f2113)

SebastianAigner (approved):

SebastianAigner:
I've made a bunch of changes to your text that make the PR fit better with the rest of the docs, @elliottswan-toast – most notably, I've tried to simplify the examples and go from a "this is how you do it wrong" to "this is how you should do it" type explanation. Feel free to have another look. From my end, this is ready to merge now.

koshachy (approved):

elliottswan-toast:
@SebastianAigner looks good to me, one of ya'll can merge it

SebastianAigner:
Thank you for the contribution and working with us to make sure the change is the best it potentially can be, @elliottswan-toast ! Hoping this will alleviate some confusion for people in the future! 🎉

PR #3139, opened 2022-09-01
chore: remove unnecessary icon of the home page

nikpachoo:
Hey @koshachy,
let's remove this icon from the "command line compiler" block. 
Screenshot 2022-09-01 at 13 42 31


koshachy:
True, that looks awful.
Thanks for pointing out this.

koshachy (approved):

PR #3142, opened 2022-09-02
Fix typo in distribution directory name

Sejsel:
Checked in a new project to see what the real name is.


koshachy (approved):
Nice catch, thank you!

PR #3150, opened 2022-09-06
chore: update the Join the community section

koshachy:
Added some bells and whistles


koshachy:
@danil-pavlov don't pay attention to the Qodana checks

PR #3152, opened 2022-09-06
fix: KTL-821: Home page has little 'esc'

zoobestik:


nikpachoo (approved):
Ouch, that's my fault, I didn't notice when I deleted the assets.

krutilov (approved):

PR #3156, opened 2022-09-11
Update serializationVersion to 1.4.0

grian32:


nikpachoo (approved):
Nice catch, thank you!

PR #3158, opened 2022-09-12
Update basic-types.md

Ijaiswalshivam:
Added some words


koshachy:
Thanks for your contribution!
However, it doesn't make any significant improvement of the docs: we don't mark examples in this way.
So I'd rather decline your PR.
Don't hesitate to reach out to me if you have questions.

PR #3160, opened 2022-09-12
KTL-802: empty pages

zoobestik:
* **fix**: Add redirect for [dropped](https://github.com/JetBrains/kotlin-web-site/pull/2913/files#diff-c81630bb0ada7c9f9a6c68dec3a6abb7dfc9d77f31e22b868663d3b95f8fdaba) urls
* **fix**: replace index instead append data

### DIFF
--- a/src/search.py
+++ b/src/search.py
@@ -3,7 +3,7 @@ from typing import Dict, List, Iterator
 
 from algoliasearch import algoliasearch
 from algoliasearch.index import Index
-from bs4 import BeautifulSoup, Tag
+from bs4 import Tag
 from googleapiclient.discovery import build, Resource
 from oauth2client.service_account import ServiceAccountCredentials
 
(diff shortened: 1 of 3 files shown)


src/search.py
```diff
@@ -337,7 +337,9 @@ def wh(*args):
     wh_index = get_wh_index()
 
     if wh_index:
+        print("Clearing WH index in " + wh_index.index_name + " index")
+        wh_index.clear_index()
         print("Submitting WH index objects to " + wh_index.index_name + " index")
-        wh_index.add_objects(wh_index_objects)
+        wh_index.save_objects(wh_index_objects)
```
nikpachoo, line 343: I'd recommending to use [replace_all_objects](https://www.algolia.com/doc/api-reference/api-methods/replace-all-objects/) instead of "clear_index" and "save_objects". This method replaces all records in an index without any downtime.
Since we run it regularly, avoiding downtime for the search index is better.
As I see, it requires an updated version of algoliasearch, but hopefully, that won't be a problem

nikpachoo (changes requested):

src/search.py
zoobestik: I thought about that, but it is not preset in our current API version:
https://github.com/algolia/algoliasearch-client-python/blob/1.12.0/algoliasearch/index.py

I\m not sure it's not waste of time to upgrade it without refactoring. What do you think?

zoobestik: Looks good with 1.20 - [TC ktl-802-empty-pages](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinSites_Builds_KotlinlangOrg_BuildBetaSerachIndex?branch=ktl-802-empty-pages&mode=builds)

nikpachoo (approved):

PR #3165, opened 2022-09-14
fix: non-null to non-nullable types

PetrakovichVictoria:


PetrakovichVictoria:
We agreed to investigate non-null values, params, and so on.

koshachy:
@PetrakovichVictoria @danil-pavlov Should we proceed with this one?

PetrakovichVictoria:
Yes, we agreed with @danil-pavlov that I'll ask Matt about some more details. I haven't done it yet, but it's in my plan

PetrakovichVictoria:
Sorry for the prolonged inaction in this PR.
We discussed this with Matt once more, and we both have an impression plus it's also so when googling examples, that:

- "non-null" is more appropriate for things that have been finally defined (e.g. values)
- non-nullable would imply that they cannot ever be null based on their type, etc

However, in our docs, we have some quotes from IDEA tips. In IDEA tips, I suppose, it's a bit incorrect – I mean using of "non-null type". But I would leave them as they are to save consistency between tips and their quotation.

docs/topics/jvm/java-interop.md
```diff
@@ -340,7 +340,7 @@ The `@UnderMigration` annotation (provided in a separate artifact `kotlin-annota
 maintainers to define the migration status for the nullability type qualifiers.
 
 The status value in `@UnderMigration(status = ...)` specifies how the compiler treats inappropriate usages of the
-annotated types in Kotlin (e.g. using a `@MyNullable`-annotated type value as non-null):
+annotated types in Kotlin (e.g. using a `@MyNullable`-annotated type value as non-nullable):
```
PetrakovichVictoria, line 343:
```suggestion
annotated types in Kotlin (e.g. using a `@MyNullable`-annotated type value as non-null):
```

PR #3168, opened 2022-09-15
update: how to enable IR for kapt

PetrakovichVictoria:


docs/topics/kapt.md
```diff
@@ -76,6 +76,15 @@ If you use annotation processors for your `androidTest` or `test` sources, the r
 `kaptAndroidTest` and `kaptTest`. Note that `kaptAndroidTest` and `kaptTest` extends `kapt`, so you can just provide the
 `kapt` dependency and it will be available both for production sources and tests.
 
+> To use all the newest Kotlin features with kapt, for example, [repeatable annotations](annotations.md#repeatable-annotations), 
+> add the following option to your `gradle.properties`:
```
danil-pavlov, line 80: I'd suggest explaining why this particular option is used, like 'To use the newest Kotlin features in kapt, for example..., enable the support for the IR backend(link) with following option'

PetrakovichVictoria: Good idea, just one remark: I. suppose the preposition should be "with" kapt, not "in" kapt.

docs/topics/kapt.md
```diff
@@ -15,75 +15,76 @@ Please read below about how to apply the *kapt* plugin to your Gradle/Maven buil
...
 
-```groovy
-apply plugin: 'kotlin-kapt'
-```
+   ```groovy
+   apply plugin: 'kotlin-kapt'
+   ```
 
-Then add the respective dependencies using the `kapt` configuration in your `dependencies` block:
+   Applying Kotlin plugins with `apply` in the Kotlin Gradle DSL is not recommended – [see why](#using-the-gradle-kotlin-dsl).
```
danil-pavlov, line 48: The link is broken

PetrakovichVictoria: Oops

danil-pavlov (approved):

PR #3171, opened 2022-09-16
Update extensions.md when member method private.

He-Pin:
When then member method is private, then the extension method with same method will win.


PetrakovichVictoria:
Hi @He-Pin! Thank you for your suggestion. However, I suppose that it's quite obvious that the private method is not accessible; and this page is not about modifiers.

PR #3177, opened 2022-09-22
Update equality.md

debChowdhury:
Added this line "NaN is considered greater than any other element including POSITIVE_INFINITY". It is already given in this link https://kotlinlang.org/docs/numbers.html#floating-point-numbers-comparison


danil-pavlov (approved):

PR #3184, opened 2022-09-28
fix: use the right case on table name when adding database support

rlespinasse:
Based on the code repository, the table name should be in uppercase in order for H2 to found the table on the H2 database.


PetrakovichVictoria:
Hi @rlespinasse! Thank you for your suggestion!

PR #3187, opened 2022-09-29
update: add about usage of freeCompilerArgs in KGP

PetrakovichVictoria:


docs/topics/gradle.md
```diff
@@ -927,12 +927,12 @@ Here is a complete list of options for Gradle tasks:
...
-| `freeCompilerArgs` | A list of additional compiler arguments |  | [] |
+| Name | Description                                                                                                                                         | Possible values |Default value |
+|------|-----------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|--------------|
+| `allWarningsAsErrors` | Report an error if there are any warnings                                                                                                           |  | false |
+| `suppressWarnings` | Don't generate warnings                                                                                                                             |  | false |
+| `verbose` | Enable verbose logging output. Works only when the [Gradle debug log level enabled](https://docs.gradle.org/current/userguide/logging.html)         |  | false |
+| `freeCompilerArgs` | A list of additional compiler arguments. You can use experimental `-X` arguments here too. See an [example](#example-of-additional-arguments-usage) |  | [] |
```
Tapchicoma, line 935: Is it possible somehow to mention that `freeCompilerArgs` usage is discouraged. If user is missing some option in Gradle DSL - new Kotlin issue should be opened?

docs/topics/gradle.md
```diff
@@ -966,6 +966,55 @@ Here is a complete list of options for Gradle tasks:
 | `target` | Generate JS files for specific ECMA version | "v5" | "v5" |
 | `typedArrays` | Translate primitive arrays to JS typed arrays |  | true |
 
+#### Example of additional arguments usage
+
+Use the attribute `freeCompilerArgs` to supply additional (including experimental) compiler arguments. You can add a single 
+argument to this attribute or a list of arguments divided by a comma:
```
Tapchicoma, line 972: This option has type `List<String>`, so "a list of arguments divided by a comma" doesn't make sense

docs/topics/gradle.md
```diff
@@ -966,6 +966,55 @@ Here is a complete list of options for Gradle tasks:
...
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
```
Tapchicoma, line 978:
```suggestion
import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
```

docs/topics/gradle.md
PetrakovichVictoria: Sure

docs/topics/gradle.md
```diff
@@ -966,6 +966,55 @@ Here is a complete list of options for Gradle tasks:
...
+// ...
+
+val compileKotlin: KotlinCompile by tasks
+
+// Single experimental argument
+compileKotlin.kotlinOptions.freeCompilerArgs += "-Xexport-kdoc"
+// Single additional argument, can be a key-value pair
+compileKotlin.kotlinOptions.freeCompilerArgs += "-opt-in=org.mylibrary.OptInAnnotation"
+// List of arguments
+compileKotlin.kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
```
Tapchicoma, line 988: replace `[..]` to `listOf(...)`

docs/topics/gradle.md
PetrakovichVictoria: Are other such fixes correct? Check, please. And what is about other classes from the `tasks` package? Where they moved to `dsl` too?

docs/topics/gradle.md
PetrakovichVictoria: Please. check now.

docs/topics/gradle.md
```diff
@@ -927,12 +927,17 @@ Here is a complete list of options for Gradle tasks:
...
+|------|-----------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|--------------|
+| `allWarningsAsErrors` | Report an error if there are any warnings                                                                                                           |  | false |
+| `suppressWarnings` | Don't generate warnings                                                                                                                             |  | false |
+| `verbose` | Enable verbose logging output. Works only when the [Gradle debug log level enabled](https://docs.gradle.org/current/userguide/logging.html)         |  | false |
+| `freeCompilerArgs` | A list of additional compiler arguments. You can use experimental `-X` arguments here too. See an [example](#example-of-additional-arguments-usage) |  | [] |
+
+> We are  going to remove the attribute `freeCompilerArgs` in future releases. If you miss some option in the Kotlin Gradle DSL,
```
Tapchicoma, line 937: > remove

This option will stay in some other form, but will produce build warning on any usage. Better to replace it with "to discourage `freeCompilerArgs` usage in future releases"?

docs/topics/gradle.md
Tapchicoma: > Check, please.

Done

> And what is about other classes from the tasks package? Where they moved to dsl too?

Not yet. Also `dsl.KotlinCompile` is deprecated in Kotlin 1.8.0 release :sweat_smile:

docs/topics/gradle.md
PetrakovichVictoria: What about "We are going to deprecate..."? We should tell the reason why we discourage something. And if we say that something will deprecate – it's obvious that it's better not to use it.

docs/topics/gradle.md
```diff
@@ -869,10 +869,10 @@ To configure a single task, use its name. Examples:
 <tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
 // ...
 
-val compileKotlin: KotlinCompile by tasks
+val compileKotlin: KotlinCompile by dsl
```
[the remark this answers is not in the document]
PetrakovichVictoria: How does it work?😃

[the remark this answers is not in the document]
Tapchicoma: via delegation and generated by Gradle Kotlin accessors/DSL

docs/topics/gradle.md
```diff
@@ -966,6 +971,55 @@ Here is a complete list of options for Gradle tasks:
...
+argument to this attribute or a list of arguments:
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
+// ...
+
+val compileKotlin: KotlinCompile by dsl
```
Tapchicoma, line 986: Here `by tasks` as well :slightly_smiling_face:

Tapchicoma (approved):
Small issue to fix and then LGTM to merge!

docs/topics/gradle.md
Tapchicoma: `freeCompilerArgs` has following issues:
- values that could be passed are not discoverable and user has to find them by searching in the internet
- values does not have proper deprecation cycle notifying user in advance that it is deprecated
- passing options with paths could break build cache via using absolute paths. Generally some values could use Gradle inputs specific normalization
- some passed values should not be considered as task inputs as they are not affect compilation output
- not nice build scan report for task inputs comparison between builds

docs/topics/gradle.md
PetrakovichVictoria, line 986:
```suggestion
val compileKotlin: KotlinCompile by tasks
```

danil-pavlov (commented):
A couple of comments and suggestions:
1. It looks like this example is related to the JS attributes. I'd move it under the common attributes section
2. Since the usage of this option is discouraged, but we still add a new example section, I suggest making it collapsible
3. In the example section, spacing in code blocks on the Groovy tab is too short

PetrakovichVictoria:
@danil-pavlov, thank you. I've fixed the first and the third items. Spacing is somehow reduced to 2 spaces when I copy the code from the Groovy script where it has 4 spaces🤷‍♀️

Regarding the second: this new piece of docs is long-awaited and desired by many developers. If we want to collapse this section, then I would at least rename it to "Example of additional arguments usage via freeCompilerArgs" so people can find this item on the page via a page search. So, I renamed)

docs/topics/whatsnew1620.md
```diff
@@ -668,7 +668,7 @@ If you are publishing a `klib` and want to use only relative paths of source fil
 <tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile::class).configureEach {
+tasks.withType(org.jetbrains.kotlin.gradle.dsl.KotlinCompile::class).configureEach {
```
koshachy, line 671: If it was a mistake — we should fix it.
But if this is how it was used in Kotlin 1.6.20 — we shouldn't fix it.

Tapchicoma: Change is also fine for Kotlin 1.6.20

PetrakovichVictoria:
@danil-pavlov Let's reconsider the collapsible state of the section. It revealed that it's impossible to make a link to a header that contains such technical information.

`#### Example of additional arguments usage via freeCompilerArgs {initial-collapse-state="collapsed"}`

And I would like users to be able to share this link.

The PR with the new Gradle section structure is here.

danil-pavlov:
@PetrakovichVictoria I checked, links to collapsible blocks work, see [an example](https://kotlinlang.org/docs/multiplatform-mobile-create-first-app.html#run-on-a-different-android-simulated-device)

PetrakovichVictoria:
@danil-pavlov thank you, great! Dunno why it didn't work for me🙈

PetrakovichVictoria:
I thought that the link doesn't work on the staging because the header has collapsible attributes. The thing is that in IDEA the link works if to remove these collapsible attributes from the header and leave upper letters in the link. But on the staging, it works if to make upper case letters lower case letters.

PR #3188, opened 2022-09-29
update: add about usage of the Lombok plugin with JPS

PetrakovichVictoria:


docs/topics/lombok.md
```diff
@@ -169,3 +169,21 @@ The Lombok compiler plugin works correctly with [kapt](kapt.md) if annotation pr
...
+-Xplugin=$KOTLIN_HOME/lib/lombok-compiler-plugin.jar
+```
+
+Replace `<PATH_TO_CONFIG_FILE>` with a path to your `lombok.config` file:
+
+```bash
+# The plugin option format is: "-P plugin:<plugin id>:<key>=<value>". 
+# Options can be repeated.
+
+-P plugin:org.jetbrains.kotlin.lombok:config:=<PATH_TO_CONFIG_FILE>
```
demiurg906, line 188: Redundant extra `:` before `=`

demiurg906 (changes requested):

docs/topics/lombok.md
PetrakovichVictoria, line 188:
```suggestion
-P plugin:org.jetbrains.kotlin.lombok:config=<PATH_TO_CONFIG_FILE>
```

PetrakovichVictoria: Indeed, I haven't noticed, thank you.

demiurg906 (approved):

danil-pavlov (commented):
Just to clarify, are these two alternative options to provide Lombok compiler jar (xplugin and lomboc.config) or 2 steps required for completing this task?

demiurg906:
@danil-pavlov First argument (`-Xplugin`) is required to register lombok plugin in compiler. The second one (`-P`) is optional for the case when user actually uses `lombok.config` file and wants to koltin compiler to understand it

danil-pavlov (changes requested):
I think the text should be clear that the second one being optional, as of now, it's an imperative

danil-pavlov (approved):

PR #3195, opened 2022-09-30
fix: sequences are not containers

PetrakovichVictoria:


docs/kr.tree
```diff
@@ -276,6 +275,7 @@
             <toc-element id="set-operations.md" toc-title="Set-specific operations"/>
             <toc-element id="map-operations.md" toc-title="Map-specific operations"/>
         </toc-element>
+        <toc-element id="sequences.md"/>
```
PetrakovichVictoria, line 278: Is it right to move this page out of collections?

ilya-g: No, I don't see it as an improvement. Sequences are very close to collections semantically.

ilya-g (changes requested):

docs/kr.tree
PetrakovichVictoria: Ok, fixed. Check the text, please.

ilya-g (approved):

danil-pavlov (approved):

PR #3196, opened 2022-09-30
Misspell in object-declarations.md

HaliksaR:
data object?


SebastianAigner:
Yes.

Data object declarations is an Experimental feature. It may be dropped or changed at any time. Opt-in is required with the kotlinOptions.languageVersion = "1.9" compiler option.

PR #3197, opened 2022-09-30
Added some more features of Kotlin and Link

AbhiiVops:
Added some more features of Kotlin and a dedicated link to learn android Development using Kotlin from scratch


koshachy (commented):
Sorry, I couldn't accept these changes.
* There is no need to update the link to another course.
* Google / JetBrains created all content for this page, and we decided not to change this page in the nearest future.

PR #3200, opened 2022-10-04
Made some paragraphs sound better to read

mangoisbest:


docs/topics/functions.md
```diff
@@ -127,37 +121,32 @@ reformat(
 )
 ```
 
-You can skip all the ones with default values:
+You can disregard all of the ones with default values:
```
sarahhaggarty, line 124:
```suggestion
You can skip all of the ones with default values:
```

docs/topics/functions.md
```diff
@@ -127,37 +121,32 @@ reformat(
...
 
 ```kotlin
 reformat("This is a long String!")
 ```
 
-You are also able to skip specific arguments with default values, rather than omitting them all. However, after the first
-skipped argument, you must name all subsequent arguments:
+You can also skip specific arguments with default values rather than omitting all of them. After the first skipped argument, however, you must name all subsequent arguments:
```
sarahhaggarty, line 130:
```suggestion
You can also skip specific arguments with default values, rather than omitting them all. However, after the first skipped argument, you must name all subsequent arguments:
```
It is clearer for the reader if **however** is at the start of the sentence, to make it clear that this is an exception.

docs/topics/functions.md
```diff
@@ -127,37 +121,32 @@ reformat(
...
 
 ```kotlin
 reformat("This is a short String!", upperCaseFirstLetter = false, wordSeparator = '_')
 ```
 
-You can pass a [variable number of arguments (`vararg`)](#variable-number-of-arguments-varargs) with names using the
-`spread` operator:
+Using the'spread' operator, you can pass a [variable number of arguments ('vararg')](#variable-number-of-arguments-varargs) with names:
```
sarahhaggarty, line 136:
```suggestion
You can pass a [variable number of arguments (`vararg`)](#variable-number-of-arguments-varargs) with names using the `spread` operator:
```
We prefer to enclose operator names and variables in backticks to format them as inline code.

docs/topics/functions.md
```diff
@@ -127,37 +121,32 @@ reformat(
...
 fun foo(vararg strings: String) { /*...*/ }
 
 foo(strings = *arrayOf("a", "b", "c"))
 ```
 
-> On the JVM: You can't use the named argument syntax when calling Java functions because Java bytecode does not
-> always preserve the names of function parameters.
->
-{type="note"}
+When calling Java functions on the JVM, you cannot use the named argument syntax because Java bytecode does not always preserve the names of function parameters.
```
sarahhaggarty, line 144:
```suggestion
> When calling Java functions on the JVM, you can't use the named argument syntax because Java bytecode does not always preserve the names of function parameters.
{type="note"}
```
It's important that the reader notices this, so we should keep it as a note.

docs/topics/functions.md
```diff
@@ -127,37 +121,32 @@ reformat(
...
 
 ### Unit-returning functions
 
-If a function does not return a useful value, its return type is `Unit`. `Unit` is a type with only one value - `Unit`.
-This value does not have to be returned explicitly:
+If a function returns an unusable value, its return type is 'Unit.' 'Unit' is a type that has only one value.
+This value does not have to be explicitly returned:
```
sarahhaggarty, line 149:
```suggestion
If a function returns an unusable value, its return type is `Unit`. `Unit` is a type that has only one value - `Unit`.
This value doesn't have to be returned explicitly:
```
We use backticks to format inline code, rather than single quotes.
@koshachy, what do you think about **unusable**?

sarahhaggarty (changes requested):
Hi @mangoisbest,

Thanks for taking the time to make a contribution to our documentation! 😃 Have you read our [style guide](https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit?usp=sharing)? This explains what requirements our documents have to meet to be uploaded on our site. It's also a great starting point if you're new to technical writing.

Here is a summary of the feedback I've given on your changes:
* When writing sentences explaining how to do something, we prefer to start with the goal, and then describe how to achieve it.
* Use backticks instead of single quotes for inline code formatting.
* Use active voice, rather than passive voice.
* Conditional statements (if) should ideally be used as early as possible in a sentence so that readers can skip ahead if it doesn't apply.

Please let me know if you have any questions!

(P.S. Cool name! 🥭 )

koshachy:
@mangoisbest are you ready to proceed with this PR?

koshachy:
Hey @sarahhaggarty 
You could close this PR and move some useful changes in the separate one.

sarahhaggarty:
Incorporated useful suggestions in #3264.

PR #3204, opened 2022-10-05
Added for loop to get items and indices together.

ShoaibKakal:


PetrakovichVictoria:
Hi @ShoaibKakal! Thank you for your suggestion. However, we have this example on the other page https://kotlinlang.org/docs/control-flow.html#for-loops (at the end of the section). I suppose that the Idioms page is already quite big.

PR #3232, opened 2022-10-20
Change a block code for better consistancy

KavrinDEV:
Property delegate requirements -> resourceDelegate()

Throughout this part, in code examples, "Resource" class has been used for properties type. the function name is resourceDelegate() too which adds up to the confusion. also the function is an example of using anonymous objects instead of classes, which have been used in the previous example, so I think it's more readable and intuitive if the code examples be consistent.


danil-pavlov (approved):

PR #3238, opened 2022-10-24
Remove old stdlib artifacts for JDK 7 and 8

sarahhaggarty:
As part of KT-51907.

### DIFF
--- a/docs/topics/maven.md
+++ b/docs/topics/maven.md
@@ -27,11 +27,9 @@ To use the standard library in your project, add the following dependency in the
 </dependencies>
 ```
 
-If you're targeting JDK 7 or JDK 8, you can use extended versions of the Kotlin standard library. They contain
-additional extension functions for APIs added in new JDK versions. Instead of `kotlin-stdlib`, use `kotlin-stdlib-jdk7`
-or `kotlin-stdlib-jdk8`, depending on your JDK version. 
-
->For Kotlin versions older than 1.2, use `kotlin-stdlib-jre7` and `kotlin-stdlib-jre8`.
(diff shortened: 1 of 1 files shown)


docs/topics/maven.md
```diff
@@ -27,11 +27,10 @@ To use the standard library in your project, add the following dependency in the
...
 ```
 
-If you're targeting JDK 7 or JDK 8, you can use extended versions of the Kotlin standard library. They contain
-additional extension functions for APIs added in new JDK versions. Instead of `kotlin-stdlib`, use `kotlin-stdlib-jdk7`
-or `kotlin-stdlib-jdk8`, depending on your JDK version. 
-
->For Kotlin versions older than 1.2, use `kotlin-stdlib-jre7` and `kotlin-stdlib-jre8`.
+> For Kotlin versions older than:
+> * 1.8, if you're targeting JDK 7 or 8, use `kotlin-stdlib-jdk7`
+     or `kotlin-stdlib-jdk8`.
```
PetrakovichVictoria, line 32: Maybe to unite this line with the previous and add "respectively"?

`> * 1.8, if you're targeting JDK 7 or 8, use `kotlin-stdlib-jdk7` or `kotlin-stdlib-jdk8` respectively.`

PetrakovichVictoria (commented):
That is almost good; let's discuss a minor point.

docs/topics/maven.md
sarahhaggarty: I incorporated your suggestion but also decided to change the format. Would you mind checking it again?

PetrakovichVictoria (approved):

PR #3240, opened 2022-10-24
fix: default implementation of MutableList is ArrayList

PetrakovichVictoria:
https://youtrack.jetbrains.com/issue/KT-51602/Web-feedback-from-Overview-httpskotlinlangorgdocscollections-overviewhtml


ilya-g (changes requested):
In that case, makes sense to fix the similar problem for mutable set and mutable map.

PR #3249, opened 2022-11-04
fix: Fix glide link in ksp-overview

jporsay:


koshachy (approved):
Thanks for pointing this out!

PR #3253, opened 2022-11-07
fix: fixed dead link

bemoty:
The subheading to Tony Hoare's apologies and retractions was recently removed. (see https://en.wikipedia.org/w/index.php?title=Tony_Hoare&diff=1114844581&oldid=1108470526)


koshachy (approved):
Thanks for fixing this!

PR #3257, opened 2022-11-09
chore: update the Gradle prop for K/N parallel builds

koshachy:


docs/topics/native/native-improving-compilation-time.md
```diff
@@ -29,7 +29,8 @@ Here are some recommendations for configuring Gradle for better compilation perf
 
 * **Increase the [Gradle heap size](https://docs.gradle.org/current/userguide/performance.html#adjust_the_daemons_heap_size)**.
   Add `org.gradle.jvmargs=-Xmx3g` to `gradle.properties`. If you use [parallel builds](https://docs.gradle.org/current/userguide/performance.html#parallel_execution),
-  you might need to make the heap even larger or choose the right number of threads with `org.gradle.parallel.threads`.
+  you might need to increase the number of workers with the `org.gradle.workers.max` property or the `--max-workers` command-line option.
```
ALikhachev, line 32: I'd preserve the old wording,
```suggestion
  you might need to choose the right number of workers with the `org.gradle.workers.max` property or the `--max-workers` command-line option.
```
In most cases if you need to change the value, you will set it to lower value than the default one.

koshachy: Oh, I see, thanks!

ALikhachev (approved):

docs/topics/native/native-improving-compilation-time.md
PetrakovichVictoria: Doesn't it sound ambiguous now? How can a user know what is *the right number of workers*?
Maybe to remove the word "right"?

PetrakovichVictoria: And I'm not sure, but doesn't this option `org.gradle.workers.max` works only when org.gradle.parallel=true (https://docs.gradle.org/current/userguide/build_environment.html)? Is it worth saying?

koshachy: The wording is pretty same as it was, I just changed the property name.

We could change it to "lower", but I'm afraid that it won't be more clear: we cannot know the device configuration in advance.

PetrakovichVictoria: > The wording is pretty same as it was

I understand; maybe we should improve it in the future.

koshachy: > And I'm not sure, but doesn't this option org.gradle.workers.max works only when org.gradle.parallel=true (https://docs.gradle.org/current/userguide/build_environment.html)? Is it worth saying?

The sentence begins with mentioning parallel building 🤔

koshachy: Or did you mean to write the property explicitly? 🤔

PetrakovichVictoria: >Or did you mean to write the property explicitly? 🤔

Yes. But from the other side, they can understand it by the link.

PetrakovichVictoria (approved):

koshachy:
I'll merge it as it is.
But, as @PetrakovichVictoria rightly noted, it is necessary to remove the ambiguity of terms.

PR #3258, opened 2022-11-09
update: update info on Kotlin for Eclipse

koshachy:
https://youtrack.jetbrains.com/issue/KT-54740


yalishevant (approved):

danil-pavlov (approved):
Just an idea to include Eclipse into the page description as well

PR #3263, opened 2022-11-12
Corrected the build filename, improved clarity.

julianharty:
Following along with this tutorial I was confused which file to edit and whether to add a new plugin section.


danil-pavlov (approved):
Thank you, Julian! Approved

PR #3274, opened 2022-11-17
Corrects text for h1 header in tutorial documentation

derekleerock:


koshachy (approved):
What a catch! Thanks for pointing this out!

PR #3281, opened 2022-11-21
feat: new header filter

danil-pavlov:
* [KT-54001](https://youtrack.jetbrains.com/issue/KT-54001/Kotlin-Native-support-header-exclusion-in-cinterop-def-files)

### DIFF
--- a/docs/topics/native/native-c-interop.md
+++ b/docs/topics/native/native-c-interop.md
@@ -98,21 +98,30 @@ in case of direct dependencies.
 
 #### Filter headers by globs
 
-It is possible to filter headers by globs. The `headerFilter` property value
-from the `.def` file is treated as a space-separated list of globs. If the
-included header matches any of the globs, then the declarations from this header
-are included into the bindings.
+It is possible to filter headers by globs using filter properties from the `.def` file.
+They are treated as a space-separated list of globs.
 
-The globs are applied to the header paths relative to the appropriate include
-path elements, e.g. `time.h` or `curl/curl.h`. So if the library is usually
-included with `#include <SomeLibrary/Header.h>`, then it would probably be
-correct to filter headers with
+* To include declarations from headers, use the `headerFilter` property. If the included header matches any of the globs,
+  the declarations are included in the bindings.
 
-```c
-headerFilter = SomeLibrary/**
-```
+  The globs are applied to the header paths relative to the appropriate include path elements,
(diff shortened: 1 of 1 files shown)


docs/topics/native/native-c-interop.md
```diff
@@ -98,21 +98,30 @@ in case of direct dependencies.
...
+
+  The globs are applied to the header paths relative to the appropriate include path elements,
+  for example, `time.h` or `curl/curl.h`. So if the library is usually included with `#include <SomeLibrary/Header.h>`,
+  it would probably be correct to filter headers with the following filter:
+    
+  ```c
+  headerFilter = SomeLibrary/**
+  ```
+    
+  If `headerFilter` is not specified, then all headers are included.
```
sbogolepov, line 115: Let's mention that it is recommended to explicitly specify `headerFilter` with the most precise glob possible, so that the generated interop library contains only the necessary declarations. It should help avoid various problems when upgrading Kotlin or environment (e.g. Xcode).

sbogolepov (approved):

docs/topics/native/native-c-interop.md
```diff
@@ -98,21 +98,37 @@ in case of direct dependencies.
...
+  and specify the glob as precisely as possible. In this case, the generated library contains only the necessary
+  declarations. It can help avoid various issues when upgrading Kotlin or tools in your development environment.
+
+* To exclude specific headers, use the `excludeFilter` property.
+  
+  It can be helpful to remove redundant or problematic headers and optimize compilation,
+  as declarations from the specified headers are not included into the bindings.
+
+  ```properties
+  excludeFilter = SomeLibrary/time.h
```
koshachy, line 125: Shouldn't we also use glob here instead of "time.h"? 🤔

koshachy (approved):
One minor comment, 
overall is OK!

docs/topics/native/native-c-interop.md
sbogolepov: This one is a good example as it shows the main use case of `excludeFilter` which is to remove some problematic header.

koshachy: Ok, I see.
I'm a bit confused because the section's name is `Filter headers by **globs**`

danil-pavlov: I thought about it, but the `excludeFilter` can be useful while filtering by globs with headerFilter, so I think it makes sense to keep them together

PR #3292, opened 2022-11-30
Update Arrays page

sarahhaggarty:
This PR updates the Arrays page with additional content.

It will resolve the following YouTrack tickets:
- KT-55214
- KT-46054
- KT-46629
- KT-49167
- KT-48481
- KT-50043
- KT-52564
- KT-56228
- KT-57718
- KT-57473
- KT-58318
- KT-58810
- KWHF-18
- KWHF-86
- KWHF-765
- KWHF-934

### DIFF
--- a/docs/topics/arrays.md
+++ b/docs/topics/arrays.md
@@ -1,62 +1,384 @@
 [//]: # (title: Arrays)
 
-Arrays in Kotlin are represented by the `Array` class. It has `get()` and `set()` functions that turn into `[]` by operator overloading conventions,
-and the `size` property, along with other useful member functions:
+An array is a data structure that holds a fixed number of values of the same type or its subtypes. 
+The most common type of array in Kotlin is the object-type array, represented by the [`Array`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-array/) class.
 
-```kotlin
-class Array<T> private constructor() {
-    val size: Int
-    operator fun get(index: Int): T
-    operator fun set(index: Int, value: T): Unit
+> If you use primitives in an object-type array, this has a performance impact because your primitives are [boxed](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html)
+> into objects. To avoid boxing overhead, use [primitive-type arrays](#primitive-type-arrays) instead.
+>
+{type="note"}
 
-    operator fun iterator(): Iterator<T>
-    // ...
-}
-```
+## When to use arrays
 
-To create an array, use the function `arrayOf()` and pass the item values to it, so that `arrayOf(1, 2, 3)` creates an array `[1, 2, 3]`.
-Alternatively, the `arrayOfNulls()` function can be used to create an array of a given size filled with `null` elements.
+Use arrays in Kotlin when you have specialized low-level requirements that you need to meet. For example, if you have 
+performance requirements beyond what is needed for regular applications, or you need to build custom data structures. If
+you don't have these sorts of restrictions, use [collections](collections-overview.md) instead.
 
-Another option is to use the `Array` constructor that takes the array size and the function that returns values
-of array elements given its index:
+Collections have the following benefits compared to arrays:
+* Collections can be read-only, which gives you more control and allows you to write robust code that has a clear intent.
+* It is easy to add or remove elements from collections. In comparison, arrays are fixed in size. The only way to 
+add or remove elements from an array is to create a new array each time, which is very inefficient:
+
+  ```kotlin
+  fun main() {
+  //sampleStart
+      var riversArray = arrayOf("Nile", "Amazon", "Yangtze")
+
+      // Using the += assignment operation creates a new riversArray,
+      // copies over the original elements and adds "Mississippi"
+      riversArray += "Mississippi"
+      println(riversArray.joinToString())
+      // Nile, Amazon, Yangtze, Mississippi
+  //sampleEnd
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-rivers-array-kotlin"}
+
+* You can use the equality operator (`==`) to check if collections are structurally equal. You can't use this operator
+for arrays. Instead, you have to use a special function, which you can read more about in [Compare arrays](#compare-arrays).
+
+For more information about collections, see [Collections overview](collections-overview.md).
+
+## Create arrays
+
+To create arrays in Kotlin, you can use:
+* functions, such as [`arrayOf()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of.html), [`arrayOfNulls()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of-nulls.html#kotlin$arrayOfNulls(kotlin.Int)) 
+or [`emptyArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/empty-array.html).
+* the `Array` constructor.
+
+This example uses the [`arrayOf()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of.html) function 
+and passes item values to it:
 
 ```kotlin
 fun main() {
 //sampleStart
-    // Creates an Array<String> with values ["0", "1", "4", "9", "16"]
-    val asc = Array(5) { i -> (i * i).toString() }
-    asc.forEach { println(it) }
+    // Creates an array with values [1, 2, 3]
+    val simpleArray = arrayOf(1, 2, 3)
+    println(simpleArray.joinToString())
+    // 1, 2, 3
 //sampleEnd
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-simple-array-kotlin"}
 
-The `[]` operation stands for calls to member functions `get()` and `set()`.
-
-Arrays in Kotlin are _invariant_. This means that Kotlin does not let us assign an `Array<String>`
-to an `Array<Any>`, which prevents a possible runtime failure (but you can use `Array<out Any>`,
-see [Type Projections](generics.md#type-projections)).
-
-## Primitive type arrays
-
-Kotlin also has classes that represent arrays of primitive types without boxing overhead: `ByteArray`,
-`ShortArray`, `IntArray`, and so on. These classes have no inheritance relation to the `Array` class, but they
-have the same set of methods and properties. Each of them also has a corresponding factory function:
+This example uses the [`arrayOfNulls()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of-nulls.html#kotlin$arrayOfNulls(kotlin.Int))
+function to create an array of a given size filled with `null` elements:
 
 ```kotlin
-val x: IntArray = intArrayOf(1, 2, 3)
-x[0] = x[1] + x[2]
+fun main() {
+//sampleStart
+    // Creates an array with values [null, null, null]
+    val nullArray: Array<Int?> = arrayOfNulls(3)
+    println(nullArray.joinToString())
+    // null, null, null
+//sampleEnd
+}
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-null-array-kotlin"}
+
+This example uses the [`emptyArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/empty-array.html) function to 
+create an empty array :
 
 ```kotlin
-// Array of int of size 5 with values [0, 0, 0, 0, 0]
-val arr = IntArray(5)
-
-// Example of initializing the values in the array with a constant
-// Array of int of size 5 with values [42, 42, 42, 42, 42]
-val arr = IntArray(5) { 42 }
-
-// Example of initializing the values in the array using a lambda
-// Array of int of size 5 with values [0, 1, 2, 3, 4] (values initialized to their index value)
-var arr = IntArray(5) { it * 1 } 
+    var exampleArray = emptyArray<String>()
 ```
+
+> You can specify the type of the empty array on the left-hand or right-hand side of the assignment due to Kotlin's type
+> inference.
+>
+> For example:
+> ```Kotlin
+> var exampleArray = emptyArray<String>()
+> 
+> var exampleArray: Array<String> = emptyArray()
+>```
+>
+{type="note"}
+
+The `Array` constructor takes the array size and a function that returns values for array elements given its index:
+
+```kotlin
+fun main() {
+//sampleStart
+    // Creates an Array<Int> that initializes with zeros [0, 0, 0]
+    val initArray = Array<Int>(3) { 0 }
+    println(initArray.joinToString())
+    // 0, 0, 0
+
+    // Creates an Array<String> with values ["0", "1", "4", "9", "16"]
+    val asc = Array(5) { i -> (i * i).toString() }
+    asc.forEach { print(it) }
+    // 014916
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-array-constructor-kotlin"}
+
+> Like in most programming languages, indices start from 0 in Kotlin.
+>
+{type="note"}
+
+### Nested arrays
+
+Arrays can be nested within each other to create multidimensional arrays:
+
+```kotlin
+fun main() {
+//sampleStart
+    // Creates a two-dimensional array
+    val twoDArray = Array(2) { Array<Int>(2) { 0 } }
+    println(twoDArray.contentDeepToString())
+    // [[0, 0], [0, 0]]
+
+    // Creates a three-dimensional array
+    val threeDArray = Array(3) { Array(3) { Array<Int>(3) { 0 } } }
+    println(threeDArray.contentDeepToString())
+    // [[[0, 0, 0], [0, 0, 0], [0, 0, 0]], [[0, 0, 0], [0, 0, 0], [0, 0, 0]], [[0, 0, 0], [0, 0, 0], [0, 0, 0]]]
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-multidimensional-array-kotlin"}
+
+> Nested arrays don't have to be the same type or the same size.
+>
+{type="note"}
+
+## Access and modify elements
+
+Arrays are always mutable. To access and modify elements in an array, use the [indexed access operator](operator-overloading.md#indexed-access-operator)`[]`:
+
+```kotlin
+fun main() {
+//sampleStart
+    val simpleArray = arrayOf(1, 2, 3)
+    val twoDArray = Array(2) { Array<Int>(2) { 0 } }
+
+    // Accesses the element and modifies it
+    simpleArray[0] = 10
+    twoDArray[0][0] = 2
+
+    // Prints the modified element
+    println(simpleArray[0].toString()) // 10
+    println(twoDArray[0][0].toString()) // 2
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-access-array-kotlin"}
+
+Arrays in Kotlin are _invariant_. This means that Kotlin doesn't allow you to assign an `Array<String>`
+to an `Array<Any>` to prevent a possible runtime failure. Instead, you can use `Array<out Any>`. For more information,
+see [Type Projections](generics.md#type-projections).
+
+## Work with arrays
+
+In Kotlin, you can work with arrays by using them to pass a variable number of arguments to a function or perform operations
+on the arrays themselves. For example, comparing arrays, transforming their contents or converting them to collections.
+
+### Pass variable number of arguments to a function
+
+In Kotlin, you can pass a variable number of arguments to a function via the [`vararg`](functions.md#variable-number-of-arguments-varargs)
+parameter. This is useful when you don't know the number of arguments in advance, like when formatting a message or
+creating an SQL query.
+
+To pass an array containing a variable number of arguments to a function, use the _spread_ operator
+(`*`). The spread operator passes each element of the array as individual arguments to your chosen function:
+
+```kotlin
+fun main() {
+    val lettersArray = arrayOf("c", "d")
+    printAllStrings("a", "b", *lettersArray)
+    // abcd
+}
+
+fun printAllStrings(vararg strings: String) {
+    for (string in strings) {
+        print(string)
+    }
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-vararg-array-kotlin"}
+
+For more information, see [Variable number of arguments (varargs)](functions.md#variable-number-of-arguments-varargs).
+
+### Compare arrays
+
+To compare whether two arrays have the same elements in the same order, use the [`.contentEquals()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-equals.html)
+and [`.contentDeepEquals()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-deep-equals.html) 
+functions:
+
+```kotlin
+fun main() {
+//sampleStart
+    val simpleArray = arrayOf(1, 2, 3)
+    val anotherArray = arrayOf(1, 2, 3)
+
+    // Compares contents of arrays
+    println(simpleArray.contentEquals(anotherArray))
+    // true
+
+    // Using infix notation, compares contents of arrays after an element 
+    // is changed
+    simpleArray[0] = 10
+    println(simpleArray contentEquals anotherArray)
+    // false
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-compare-array-kotlin"}
+
+> Don't use equality (`==`) and inequality (`!=`) [operators](equality.md#structural-equality) to compare the contents 
+> of arrays. These operators check whether the assigned variables point to the same object.
+> 
+> To learn more about why arrays in Kotlin behave this way, see our [blog post](https://blog.jetbrains.com/kotlin/2015/09/feedback-request-limitations-on-data-classes/#Appendix.Comparingarrays).
+> 
+{type="warning"}
+
+### Transform arrays
+
+Kotlin has many useful functions to transform arrays. This document highlights a few but this isn't an 
+exhaustive list. For the full list of functions, see our [API reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-array/).
+
+#### Sum
+
+To return the sum of all elements in an array, use the [`.sum()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/sum.html)
+function:
+
+```Kotlin
+fun main() {
+//sampleStart
+    val sumArray = arrayOf(1, 2, 3)
+
+    // Sums array elements
+    println(sumArray.sum())
+    // 6
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-sum-array-kotlin"}
+
+> The `.sum()` function can only be used with arrays of [numeric data types](numbers.md), such as `Int`.
+>
+{type="note"}
+
+#### Shuffle
+
+To randomly shuffle the elements in an array, use the [`.shuffle()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/shuffle.html)
+function:
+
+```Kotlin
+fun main() {
+//sampleStart
+    val simpleArray = arrayOf(1, 2, 3)
+
+    // Shuffles elements [3, 2, 1]
+    simpleArray.shuffle()
+    println(simpleArray.joinToString())
+
+    // Shuffles elements again [2, 3, 1]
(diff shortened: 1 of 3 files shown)


docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+To pass an array to the `vararg` parameter of a function, use the _spread_ operator (`*`).
 
 ```kotlin
-val x: IntArray = intArrayOf(1, 2, 3)
-x[0] = x[1] + x[2]
+val a = arrayOf(1, 2, 3)
+val list = asList(-1, 0, *a, 4)
 ```
 
+For more information, see [Variable number of args (varargs)](functions.md#variable-number-of-arguments-varargs).
```
SebastianAigner, line 263: nit: the section is called "Variable number of arguments (varargs)﻿"

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+
+## Convert to collection
+
+To convert an array to a [collection](collections-overview.md) (`Set`, `List`, or `Map`), use the [`toSet()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-set.html), [`toList()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-list.html), [`toMap()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-map.html) functions respectively.
```
SebastianAigner, line 267: `toList` is by far the most common use case, should be mentioned first. Also, `toMap` is likely unintuitive enough to warrant further explanation / its own short paragraph. Specifically, arbitrary arrays can't be converted to maps, only arrays of pairs / `Array<Pair<K,V>>`.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
 ```kotlin
-val x: IntArray = intArrayOf(1, 2, 3)
-x[0] = x[1] + x[2]
+val a = arrayOf(1, 2, 3)
+val list = asList(-1, 0, *a, 4)
 ```
 
+For more information, see [Variable number of args (varargs)](functions.md#variable-number-of-arguments-varargs).
+
+## Convert to collection
```
SebastianAigner, line 265: Again, check if these headlines are consistent with the way we're usually writing headlines. These phrasings strike me as odd (and even for our very short titles, a little _too_ concise)

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
-val arr = IntArray(5) { 42 }
+## What's next?
 
-// Example of initializing the values in the array using a lambda
-// Array of int of size 5 with values [0, 1, 2, 3, 4] (values initialized to their index value)
-var arr = IntArray(5) { it * 1 } 
-```
\ No newline at end of file
+* Learn about other [basic types](basic-types.md).
+* Learn about [collections](collections-overview.md).
```
SebastianAigner, line 293: Once again, big emphasis on the point that developers likely should've gone straight to the collections API anyway.

docs/topics/collections-overview.md
```diff
@@ -6,9 +6,9 @@ of items (possibly zero) that are significant to the problem being solved and ar
 Collections are a common concept for most programming languages, so if you're familiar with, for example, Java or Python 
 collections, you can skip this introduction and proceed to the detailed sections. 
 
-A collection usually contains a number of objects (this number may also be zero) of the same type. Objects in a collection
-are called _elements_ or _items_. For example, all the students in a department form a collection that can be used to
-calculate their average age. 
+A collection usually contains a number of objects (this number may also be zero) of the same type. However if you need them to,
+collections can contain objects of different types. Objects in a collection are called _elements_ or _items_. For example,
```
SebastianAigner, line 10: This sentence about objects of different types is making me somewhat uneasy. Is this supposed to be a didactic simplification? I need some background on this, because it feels like it should read something more like "of the same type (and its subtypes)" or similar, but I'm curious if this phrasing aims at a particular goal.

docs/topics/collections-overview.md
```diff
@@ -6,9 +6,9 @@ of items (possibly zero) that are significant to the problem being solved and ar
...
+all the students in a department form a collection that can be used to calculate their average age.
```
SebastianAigner, line 11: I'd split this in two:
```suggestion
all the students in a department form a collection. This collection can then be used to calculate their average age, for example.
```

docs/topics/collections-overview.md
```diff
@@ -30,6 +30,12 @@ collections of any type.
 The collection interfaces and related functions are located in the `kotlin.collections` package. Let's get an overview 
 of its contents.
 
+> Arrays are not a type of collection. 
+> Arrays have a fixed size and their elements **must** all be of the same type.
```
SebastianAigner, line 34: See one of my first comments, not sure where this part is coming from

docs/topics/collections-overview.md
```diff
@@ -287,3 +293,28 @@ The default implementation of `MutableMap` – [`LinkedHashMap`](https://kotlinl
...
+    println(deque.last()) // 4
+
+    deque.removeFirst()
+    deque.removeLast()
+    println(deque) // [1, 2, 3]
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.4"}
+
+For more information about arrays, see [Arrays](arrays.md).
```
SebastianAigner, line 320: This section should not reference the arrays page. Unfortunately, `ArrayDeque` is a "false friend": It's a `Deque` collection, the fact that it is backed by an array is an implementation detail and not relevant to using this collection type. (Similar to an `ArrayList`, it's array-esque nature is, from a usage standpoint, of no concern)

docs/topics/collections-overview.md
```diff
@@ -287,3 +293,28 @@ The default implementation of `MutableMap` – [`LinkedHashMap`](https://kotlinl
 preserves the order of elements insertion when iterating the map.
 In turn, an alternative implementation – [`HashMap`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-hash-map/index.html) – 
 says nothing about the elements order.
+
+### ArrayDeque
+
+[`ArrayDeque<T`>](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-array-deque/) is an implementation of a double-ended queue, which allows you to add or remove elements both at the beginning or end of the queue.
```
SebastianAigner, line 299: For purposes of SEO, it might be worth mentioning here that as such, `ArrayDeque` also fills the role of both a Stack and Queue data structure in Kotlin. (keyword here being `Stack` – it's one of the basic data structures we don't have in Kotlin, but that can be conveniently emulated using an ArrayDeque)

docs/topics/arrays.md
```diff
@@ -1,28 +1,47 @@
...
+Arrays in Kotlin are represented by the `Array` class. Arrays have a fixed size and their elements **must** all be of the same type. If you need to be able to change the size or have elements of different types, use [collections](collections-overview.md) instead.
 
-```kotlin
-class Array<T> private constructor() {
-    val size: Int
-    operator fun get(index: Int): T
-    operator fun set(index: Int, value: T): Unit
+## Create arrays
+
+To create an array, use the function [`arrayOf()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of.html) and pass the item values to it, so that `arrayOf(1, 2, 3)` creates an array `[1, 2, 3]`. For example:
```
SebastianAigner, line 7: ESL disclaimer obviously, but "so that" seems like an way of putting it here

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+    println(simpleArray.contentEquals(anotherArray))
+
+    simpleArray[0] = 10
+    println(simpleArray.contentEquals(anotherArray))
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="compare-array-kotlin"}
+
+> Don't use equality (`==`) and inequality (`!=`) [operators](equality.md#structural-equality) to compare the contents of arrays. These operators
```
SebastianAigner, line 159: Another big drawback of using Arrays, might be included in the discussion in the beginning (see my first comment)

SebastianAigner (changes requested):
Some content and structure changes are required here – Arrays are a traditionally tricky subject in Kotlin from a philosophical standpoint, but for more on that, see my detailed comments.

docs/topics/arrays.md
```diff
@@ -1,28 +1,47 @@
 [//]: # (title: Arrays)
```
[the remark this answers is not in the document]
sarahhaggarty: I've updated the page with a note at the beginning trying to reflect Roman's words. Instead of highlighting the negatives of arrays I've highlighted some benefits of collections that arrays don't have. Is this worded strongly enough in your view?

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
 ```
```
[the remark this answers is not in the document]
sarahhaggarty: It's a good point. I've updated the content so that all samples are runnable except for the `emptyArray()` because it could confuse readers when nothing is printed.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="multidimensional-array-kotlin"}
+
+> Nested arrays don't have to be the same type or the same size.
+>
+{type="note"}
+
+## Access and modify elements
```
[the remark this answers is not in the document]
sarahhaggarty: I've moved "Primitive arrays" out from "Create arrays" and to the bottom of the document so this content is higher up. I tried moving the "Empty arrays" and "Multidimensional arrays" content inside collapsible deflists to try and save space but the formatting of code blocks and notes looked terrible so I decided against that in the end.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+    val unzipArray = pairArray.unzip()
+    println(unzipArray)
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="unzip-array-kotlin"}
+
+## Concatenate arrays
+
+Arrays of type `String` can be concatenated using the `+` operator.
```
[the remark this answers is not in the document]
sarahhaggarty: I've added a comment in the code block to explain that a new array is being created.

docs/topics/arrays.md
sarahhaggarty: Can you explain a bit more what part you are finding odd? Would "Convert arrays to collection" sound better to you? Or is it actually that you prefer titles with noun phrases? E.g. Array to collection conversion. 🤔 My preference is to use bare infinitives when it relates to a task. Like Google recommends in their [developer's documentation style guide](https://developers.google.com/style/headings). I checked this with @koshachy  just in case and this also matches with his thinking.

docs/topics/arrays.md
```diff
@@ -1,28 +1,47 @@
...
 
-Another option is to use the `Array` constructor that takes the array size and the function that returns values
-of array elements given its index:
+Alternatively, the [`arrayOfNulls()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of-nulls.html#kotlin$arrayOfNulls(kotlin.Int)) function can be used to create an array of a given size filled with `null` elements.
 
 ```kotlin
 fun main() {
 //sampleStart
+    val nullArray = arrayOfNulls(3)
+    println(nullArray.joinToString(" "))
```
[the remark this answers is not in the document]
sarahhaggarty: I actually hadn't noticed that. Thanks for pointing it out! I've added this to all examples.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+
+To create an empty array, use [`emptyArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/empty-array.html):
+
+```kotlin
+var exampleArray = emptyArray<String>()
+```
+
+### Primitive type arrays
+
+Kotlin also has classes that represent arrays of primitive types without boxing overhead:
```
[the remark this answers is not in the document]
sarahhaggarty: Added to the top of the page where we try to persuade the reader to consider using collections instead.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+
+| | | | |
```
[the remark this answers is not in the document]
sarahhaggarty: Yikes. That certainly isn't very pretty. You're right that it is something new I am suggesting to try and save space on the page. It looks like this on the website. Are you not a fan?
Screenshot 2023-01-11 at 17 07 25

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="multidimensional-array-kotlin"}
+
+> Nested arrays don't have to be the same type or the same size.
+>
+{type="note"}
+
+## Access and modify elements
+
+Square brackets `[]` are transformed into `get()` or `set()` function calls, like in Java. For example, `x[a,b]` instead of `x.get(a,b)`:
```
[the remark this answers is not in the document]
sarahhaggarty: I've removed this statement and rewritten this part based on your feedback.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+
+### Empty arrays
+
+To create an empty array, use [`emptyArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/empty-array.html):
+
+```kotlin
+var exampleArray = emptyArray<String>()
+```
+
+### Primitive type arrays
```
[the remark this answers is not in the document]
sarahhaggarty: I've moved it to the bottom of the page as its own section.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+    val simpleArray = arrayOf("a", "b", "c")
+    val anotherArray = arrayOf("d", "e", "f")
+
+    println((simpleArray + anotherArray).joinToString())
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="concatenate-array-kotlin"}
+
+## Pass array as vararg
```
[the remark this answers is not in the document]
sarahhaggarty: I agree this header isn't nice. I've updated it to "Use arrays with variable number of arguments".

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
 
-Kotlin also has classes that represent arrays of primitive types without boxing overhead: `ByteArray`,
-`ShortArray`, `IntArray`, and so on. These classes have no inheritance relation to the `Array` class, but they
-have the same set of methods and properties. Each of them also has a corresponding factory function:
+To pass an array to the `vararg` parameter of a function, use the _spread_ operator (`*`).
```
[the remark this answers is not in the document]
sarahhaggarty: I've updated the content based on your feedback. Can you let me know what you think? I'll fix OT2 in a separate PR.

docs/topics/collections-overview.md
sarahhaggarty: This was due to a knowledge gap at the time of writing. (See earlier comment about `Any` type). I've removed it now.

docs/topics/arrays.md
```diff
@@ -31,32 +60,293 @@ fun main() {
...
+
+In some cases it is useful to be able to pass a variable number of arguments to a function without having to define the number of arguments in advance.
+In Kotlin you can use the `vararg` parameter for this. If you want to pass your variable number of arguments to a function as the contents of an array, use the _spread_ operator (`*`).
+The spread operator passes each element of your array as individual arguments to your chosen function.
+
+```kotlin
+fun main() {
+    val lettersArray = arrayOf("c", "d")
+    val lettersList = listOf("e", "f")
+    printAllStrings("a", "b", *lettersArray, *lettersList.toTypedArray())
```
SebastianAigner, line 281: If you want to keep the `lettersList` in here, it would be required to explain briefly what the `toTypedArray()` function does. Actually, `toTypedArray()` seems like a valuable addition to this page (Something along the lines of "Converting collections to arrays")
Also required to document: the difference between `toTypedArray` and `toIntArray` / the other primitives. Those can be related back to the sections between primitive arrays and object-arrays as well.

docs/topics/arrays.md
```diff
@@ -1,28 +1,57 @@
...
+>   * They have no boxing overhead
+>
+{type="note"}
 
-```kotlin
-class Array<T> private constructor() {
-    val size: Int
-    operator fun get(index: Int): T
-    operator fun set(index: Int, value: T): Unit
+Arrays in Kotlin are represented by the `Array` class. Arrays are always mutable and have a fixed size.
```
SebastianAigner, line 11: I actually don't think we mention anywhere on the page that when using the `Array` class, you're always storing _objects_. Again, this is what potentially influences boxing, and what you can use to contrast with the primitive arrays further down the page.

docs/topics/arrays.md
```diff
@@ -31,32 +60,293 @@ fun main() {
...
+}
+```
+
+For more information, see [Variable number of arguments (varargs)](functions.md#variable-number-of-arguments-varargs).
+
+## Convert to collection
+
+To convert an array to a [collection](collections-overview.md) (`List`, `Set` or `Map`), use the [`toList()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-list.html), [`toSet()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-set.html), [`toMap()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-map.html) functions respectively.
+
+> Only an array of [`Pair`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/) can be converted to a map: `Array<Pair<K,V>>`.
```
SebastianAigner, line 297: You might want to consider actually having a separate subsection for converting to Maps, to have a place where there'se a locality between this note and the related code snippet.

docs/topics/arrays.md
```diff
@@ -31,32 +60,293 @@ fun main() {
...
+//sampleStart
+    val simpleArray = arrayOf("a", "b", "c", "c")
+    
+    // Convert to Set ["a", "b", "c"]
+    println(simpleArray.toSet())
+    
+    // Convert to List ["a", "b", "c", "c"]
+    println(simpleArray.toList())
+
+    val pairArray = arrayOf(Pair(1,1), Pair(2,2), Pair(3,3))
```
SebastianAigner, line 313: Please use an example that is more expressive than just pairs of numbers. I am a fan of something like prizes `gold/silver/bronze -> 500/200/100`, or calories `apple/banana/kiwi -> 300/ 250/400`.
Bonus: add a second `kiwi` to your input array to illustrate that map keys are unique and will be overwritten.

docs/topics/collections-overview.md
```diff
@@ -8,7 +8,7 @@ collections, you can skip this introduction and proceed to the detailed sections
 
 A collection usually contains a number of objects (this number may also be zero) of the same type. Objects in a collection
```
SebastianAigner, line 9: Again, if we want to be accurate, change this to state "of the same type (and its subtypes)".

SebastianAigner (changes requested):
Already looking better! I've added additional comments on your applied changes, as well as additional points that were uncovered by taking a closer look at the current state of the document.

Also, at least for my PRs: Please don't resolve conversations opened by me. I need to validate that the changes you make reflect my original intention, and that I didn't accidentally miscommunicate or that something fell by the wayside. Once the subject of discussion from a conversation has been addressed, I'll make sure to close them for you!

AlexanderPrendota:
Bad samples
Snippets:
docs/topics/arrays.md

```kotlin
fun main() {
//sampleStart
    // Creates an array with values [null, null, null]
    val nullArray = arrayOfNulls(3)
    println(nullArray.joinToString())
//sampleEnd
}
```

(3:20, 3:32) **ERROR** Not enough information to infer type variable T
    
------

AlexanderPrendota:
Bad samples
Snippets:
docs/topics/arrays.md

```kotlin
fun main() {
//sampleStart
    // Creates an array with values [null, null, null]
    val nullArray = arrayOfNulls(3)
    println(nullArray.joinToString())
//sampleEnd
}
```

(3:20, 3:32) **ERROR** Not enough information to infer type variable T
    
------

docs/topics/arrays.md
```diff
@@ -1,28 +1,47 @@
...
+{type="note"}
 
-Another option is to use the `Array` constructor that takes the array size and the function that returns values
-of array elements given its index:
+Alternatively, the [`arrayOfNulls()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of-nulls.html#kotlin$arrayOfNulls(kotlin.Int)) function can be used to create an array of a given size filled with `null` elements.
 
 ```kotlin
 fun main() {
 //sampleStart
+    val nullArray = arrayOfNulls(3)
```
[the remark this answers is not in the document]
sarahhaggarty: Thanks for the question @SebastianAigner! We have the Verify samples check which should fail if a sample generates a compiler error. We found that this check was incorrectly reporting  a pass even though it wasn't running successfully. Now this is fixed and any samples containing errors will be flagged. (You can see the latest PR comments for an example of it in action).

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+> To convert primitive arrays to typed arrays, use [`toTypedArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-typed-array.html).
+>
+{type="note"}
+
+### Multidimensional arrays
+
+Arrays can be nested within each other to create multidimensional arrays:
+
+```kotlin
+fun main() {
```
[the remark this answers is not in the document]
sarahhaggarty: I added an example in the Access and modify elements section instead of this code  block.

docs/topics/arrays.md
```diff
@@ -31,32 +50,244 @@ fun main() {
...
+    println(simpleArray.joinToString())
+
+    simpleArray.shuffle()
+    println(simpleArray.joinToString())
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="shuffle-array-kotlin"}
+
+### Zip()
```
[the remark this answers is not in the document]
sarahhaggarty: This section was added to give an example of how to use `shuffle()` as this was requested in [KT-48481](https://youtrack.jetbrains.com/issue/KT-48481). The examples of `zip()` and `sum()` were just additional examples I added so it wasn't just one `shuffle()`. If you don't think it's useful, I could always remove this section completely?

docs/topics/arrays.md
```diff
@@ -177,6 +182,8 @@ fun main() {
 > Don't use equality (`==`) and inequality (`!=`) [operators](equality.md#structural-equality) to compare the contents of arrays. These operators
 > check whether the assigned variables point to the same object.
 > 
+> For an explanation why we have this behavior for arrays, see our [blog post](https://blog.jetbrains.com/kotlin/2015/09/feedback-request-limitations-on-data-classes/#Appendix.Comparingarrays).
```
SebastianAigner, line 185: I'd suggest replacing "we have" with "Why Arrays in Kotlin behave this way"

docs/topics/arrays.md
```diff
@@ -320,28 +294,34 @@ fun main() {
 ### Convert to Map
 
 To convert an array to a `map`, use the [`toMap()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-map.html) function.
-Only an array of [`Pair`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/) can be converted to a map: `Array<Pair<K,V>>`.
-The first value of a `Pair` instance becomes a key and the second becomes a value.
+Only an array of [`Pair<K,V>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/) can be converted to a map.
+The first value of a `Pair` instance becomes a key and the second becomes a value. In the example below, we use the [infix notation](functions.md#infix-notation) to call the [`to()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to.html) function to create tuples of `Pair`.
```
SebastianAigner, line 298: remove the parentheses from `to()`, since it's pretty much exclusively called as infix

docs/topics/arrays.md
```diff
@@ -320,28 +294,34 @@ fun main() {
...
     // overwrites the first
     println(pairArray.toMap())
-    
+
 //sampleEnd
 }
 ```
 
-## Primitive type arrays
+> To convert primitive type arrays to object type arrays, use [`toTypedArray()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-typed-array.html).
```
SebastianAigner, line 315: Why is this warning now above the introduction of primitive type arrays?

SebastianAigner (changes requested):

docs/topics/arrays.md
sarahhaggarty: Human error. Fixed now!

SebastianAigner (approved):
Looks great now! 🎉🎉🎉

SebastianAigner:
Bump!

koshachy (changes requested):
Huge work!
I'm so sorry for so late review, but I think it will help make our docs better.

Besides comments I still have a doubt about the document structure.
I tried to create a new one quickly, but we need to think over it:

-- When to use arrays
-- Create arrays
--- via functions
---- regular
---- empty
---- multidimensional
--- using constructor
-- access and modify elements
--- operator[]
--- getters and setters
--- pass variable (isn't it about accessing array elements?)
-- operations with arrays
--- compare
--- transform
--- concatenate
--- convert array to collection
-- primitive-type arrays

Feel free to contact me if you have any questions or would like to discuss the topic further.

docs/topics/arrays.md
```diff
@@ -1,62 +1,447 @@
...
+To create an array, use the [`arrayOf()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/array-of.html) function 
+and pass the item values to it:
+
+```kotlin
+fun main() {
+//sampleStart
+    // Creates an array with values [1, 2, 3]
+    val simpleArray = arrayOf(1, 2, 3)
+    println(simpleArray.joinToString())
+    // 1, 2, 3
```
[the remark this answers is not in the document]
sarahhaggarty: This was an idea from @SebastianAigner which I liked. It means that the reader doesn't have to run the snippet to know what the result will be. It's mainly for users who might not have a great internet connection so runnable samples aren't very quick. We actually have this style already in the Kotlin tour 😶

docs/topics/arrays.md
```diff
@@ -1,62 +1,447 @@
...
+
+    // Accesses the element and modifies it
+    simpleArray[0] = 10
+    twoDArray[0][0] = 2
+
+    // Prints the modified element
+    println(simpleArray[0].toString()) // 10
+    println(twoDArray[0][0].toString()) // 2
+//sampleEnd
+}
```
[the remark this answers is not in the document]
sarahhaggarty: Since `[]` is translated to `.get()` and `.set()` functions, is there a benefit from covering these functions?

docs/topics/arrays.md
```diff
@@ -1,62 +1,447 @@
...
+    // Creates an array with values [1, 2, 3]
+    val simpleArray = arrayOf(1, 2, 3)
+    println(simpleArray.joinToString())
+    // 1, 2, 3
+//sampleEnd
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="arrays-simple-array-kotlin"}
+
+> When creating arrays, you can use a [trailing comma](coding-conventions.md#trailing-commas).
```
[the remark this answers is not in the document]
sarahhaggarty: After thinking about it, there's not a benefit in the case of arrays. So I'll take this note out.

docs/topics/arrays.md
```diff
@@ -1,62 +1,447 @@
...
+{type="note"}
+
+## When to use arrays
+
+Use arrays in Kotlin when you have specialized low-level requirements that you need to meet. For example, if you have 
+performance requirements beyond what is needed for regular applications, or you need to build custom data structures. If
+you don't have these sorts of restrictions, we strongly recommend that you consider using [collections](collections-overview.md)
+instead.
+
+Collections have many benefits compared to arrays. Whereas arrays are always mutable, collections can be read-only. This
```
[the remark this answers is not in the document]
SebastianAigner: @koshachy beyond that, collections scale automatically, and provide a much more versatile API than Arrays. This note is inspired by Roman's post https://twitter.com/relizarov/status/1313475595892912134?s=20

docs/topics/arrays.md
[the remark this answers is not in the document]
SebastianAigner: I would not consider using `.get` and `.set` for arrays idiomatic Kotlin, so rather than give users the "choice" here, I'd nudge them towards using indexed access throughout the array docs.

sarahhaggarty:
Thanks so much for your review @koshachy!
I have updated the PR based on your comments. Please can you review again paying particular attention to:
* Overall docs structure
* Restructured "When to use arrays" section about collection benefits
* Restructured "Create arrays" section with empty arrays now included
* Revamped table for primitive-type arrays (I tried lists before but found it took up too much space)
* New intro to "Pass variable number of arguments to a function" section

You can find the latest commit in TC preview.

koshachy (approved):
Great job!
I hope it will help our reader to understand Arrays in Kotlin better!

PR #3296, opened 2022-12-01
Update events in November

DevPicon:
Compose Multiplatform talk


MKrishtal:
Hi Armando,
The title is supposed to be the event title 'Compose Camp 2022'
and the subject is your talk: 'Compose Multiplatform Overview'
Please update. Thanks.

PR #3303, opened 2022-12-05
update: cocoapods setup for m chips and different tools

danil-pavlov:


docs/topics/native/native-cocoapods.md
```diff
@@ -14,16 +14,117 @@ as well as [a Kotlin Gradle project and an Xcode project](native-cocoapods-xcode
...
+{type="note"}
+
+<tabs>
+<tab title="Homebrew">
+
+1. Install [Homebrew](https://brew.sh/) in case you don't have yet.
+2. Install Ruby. You can choose a specific version:
+
+    ```bash
+    brew install ruby@3.0
```
IliaSolovei, line 30: My intuition tells me that instructions for installing CocoaPods with new instance of ruby on m1 and Intel should be roughly the same.
I don't have Intel mac near me, so I can't test this intuition, but macOS doesn't work differently on these chips, so the installation may be different only for the case when we are trying to install CocoaPods with system ruby
With M1 mac it is impossible to properly install the CocoaPods without changing the system ruby (somehow it works on one of my virtual machines, but currently I don't know why and I can't propose such a way to users)

docs/topics/native/native-cocoapods.md
```diff
@@ -14,16 +14,117 @@ as well as [a Kotlin Gradle project and an Xcode project](native-cocoapods-xcode
...
+    ```
+
+4. Install CocoaPods:
+
+    ```bash
+    sudo gem install cocoapods
+    ```
+
+</tab>
+<tab title="Bundler">
```
IliaSolovei, line 78: I'm not sure that we should promote using Bundler right now. After installation of CocoaPods via Bundler, it won't be possible to use `pod` from CocoaPods gradle plugin because it doesn't support execution of pods via `bundle exec pod install` our plugin just don't know anything about Bundler and his commands.

docs/topics/native/native-cocoapods.md
```diff
@@ -14,16 +14,117 @@ as well as [a Kotlin Gradle project and an Xcode project](native-cocoapods-xcode
...
+   bundle config set --local path 'vendor/bundle'
+   ```
+   
+4. Run the installation command:
+
+    ```bash
+    bundle install
+    ```
+   
+5. Install CocoaPods:
```
IliaSolovei, line 105: In this step, CocoaPods are already installed via `bundle install` command, so we are executing installation of concrete pod (e.g. shared pod generated with CocoaPods gradle plugin)

IliaSolovei (changes requested):
Finished the first run

IliaSolovei (approved):
Good job! Now all instructions look correct!

docs/topics/native/native-cocoapods.md
```diff
@@ -14,17 +14,106 @@ as well as [a Kotlin Gradle project and an Xcode project](native-cocoapods-xcode
...
+
+   ```bash
+   echo 'export PATH="/opt/homebrew/opt/ruby/bin:$PATH"' >> /.zshrc
+   ```
+
+4. Make sure your export was added:
+
+    ```bash
+    source .zshrc
+    ```
```
koshachy, line 100: Will it show something? Output? How to ensure?

koshachy (approved):

PR #3305, opened 2022-12-05
Change data object required languageVersion to 1.9

romainbsl:


koshachy (approved):
Nice catch! 
Things change, and we didn't notice that this feature has a new target version!
Thanks!

PR #3324, opened 2022-12-15
Correct Get started with KMM link in Multiplatform overview

sarahhaggarty:
Discussed with @danil-pavlov.
Currently this link points to the general Get started with Kotlin Multiplatform page when it should point to the Get started with Kotlin Multiplatform Mobile page.


koshachy (approved):
Nice catch!

PR #3325, opened 2022-12-16
Grammatical improvements of `ksp-why-ksp.md`

YektaDev:


sarahhaggarty:
Hi @YektaDev, thanks for your suggestions. Unfortunately most of your suggestions aren't applicable. Metaprogramming and prebuilt are correct spellings and don't need hyphens. If you'd like to read more about when to use a hyphen, I'd recommend Grammarly's blog [on this topic](https://www.grammarly.com/blog/hyphen/).

PR #3330, opened 2022-12-19
Add a note regarding polymorphism

sdousti:
Explain how polymorphism can be a better design choice over instance checking.


sdousti:
Closing due to a discussion here: https://stackoverflow.com/a/74857026/17189500
In summary, it depends on the use case. One cannot say in general whether polymorphism is advantageous over instance checking.

PR #3336, opened 2022-12-20
update: jvmToolchain(x) instead of languageVersion.set(JavaLanguageVe…

PetrakovichVictoria:
…rsion.of(x))

### DIFF
--- a/docs/topics/gradle/gradle-configure-project.md
+++ b/docs/topics/gradle/gradle-configure-project.md
@@ -179,9 +179,7 @@ Use the following code to set a toolchain. Replace the placeholder `<MAJOR_JDK_V
 
 ```kotlin
 kotlin {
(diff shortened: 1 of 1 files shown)


docs/topics/gradle/gradle-configure-project.md
```diff
@@ -179,9 +179,7 @@ Use the following code to set a toolchain. Replace the placeholder `<MAJOR_JDK_V
 
 ```kotlin
 kotlin {
-    jvmToolchain {
```
Tapchicoma, line 182: I think we should show both ways to configure - with such shortcut or without it. Without it still could be useful for some cases.

Tapchicoma (changes requested):

Tapchicoma (approved):

PR #3337, opened 2022-12-22
fix: add missed import to Groovy script

PetrakovichVictoria:


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -47,6 +47,11 @@ compileKotlin.kotlinOptions.suppressWarnings = true
 <tab title="Groovy" group-key="groovy">
 
 ```groovy
```
Tapchicoma, line 50:
```suggestion
import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -47,6 +47,11 @@ compileKotlin.kotlinOptions.suppressWarnings = true
...
+// ...
+
+val compileKotlin: KotlinCompile by tasks
```
PetrakovichVictoria, line 53: And here it stays "by tasks", right?

PR #3340, opened 2022-12-22
add: new section Kotlin compiler fallback strategy

PetrakovichVictoria:


docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -283,6 +284,61 @@ tasks.withType(KotlinCompile)
...
+
+The Kotlin compiler's fallback strategy is to run a compilation outside a Kotlin daemon if the daemon somehow fails. 
+If the Gradle daemon is on, the compiler uses the ["In process" strategy](#defining-kotlin-compiler-execution-strategy). 
+If the Gradle daemon is off, the compiler uses the "Out of process" strategy.
+
+When this fallback happens, you have the following warning lines in your Gradle's build output:
+
+```
+Could not perform incremental compilation: Could not connect to Kotlin compile daemon
+Could not connect to kotlin daemon. Using fallback strategy.
```
ALikhachev, line 297: Since 1.8.0 the message is slightly changed:
```suggestion
Failed to compile with Kotlin daemon: java.lang.RuntimeException: Could not connect to Kotlin compile daemon
[exception stacktrace]
Using fallback strategy: Compile without Kotlin daemon
Try ./gradlew --stop if this issue persists.
```

ALikhachev: In the case of disabled fallback strategy the message would look like:
```
Failed to compile with Kotlin daemon. Fallback strategy (compiling without Kotlin daemon) is turned off. Try ./gradlew --stop if this issue persists.
```

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -283,6 +284,61 @@ tasks.withType(KotlinCompile)
...
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+val compileKotlin: KotlinCompile by tasks
+
+compileKotlin {
+    useDaemonFallbackStrategy = false
+}
```
ALikhachev, line 322: Such syntax with assigning won't work in Kotlin DSL until https://github.com/gradle/gradle/issues/9268 is resolved. Gradle is actively working on it now.

Also the example promotes non-lazy way to configure tasks (non-conformant with [Task Configuration Avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html)).
```suggestion
tasks {
    compileKotlin {
        useDaemonFallbackStrategy.set(false)
    }   
}
```

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -283,6 +284,61 @@ tasks.withType(KotlinCompile)
...
+<tab title="Groovy" group-key="groovy">
+
+```groovy
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+val compileKotlin: KotlinCompile by tasks
+
+compileKotlin {
+    useDaemonFallbackStrategy = false
+}
```
ALikhachev, line 335: It's also Kotlin DSL 😅 
I've suggested a Groovy variant also respecting the Task Configuration Avoidance feature
```suggestion
tasks.named("compileKotlin").configure {
    useDaemonFallbackStrategy = false
}
```

ALikhachev (changes requested):

docs/topics/gradle/gradle-compilation-and-caches.md
PetrakovichVictoria: Thanks! Seems like this message 
>In the case of disabled fallback strategy the message would look like:

is not needed in docs, right?

ALikhachev: I think yes, it's not necessary here

ALikhachev (approved):

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -283,6 +284,57 @@ tasks.withType(KotlinCompile)
...
+```
+Failed to compile with Kotlin daemon: java.lang.RuntimeException: Could not connect to Kotlin compile daemon
+[exception stacktrace]
+Using fallback strategy: Compile without Kotlin daemon
+Try ./gradlew --stop if this issue persists.
+```
+
+However, a silent fallback to another strategy can consume a lot of system resources or lead to non-deterministic builds, 
+read more about this in this [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-48843/Add-ability-to-disable-Kotlin-daemon-fallback-strategy).
+To avoid this, there is a Gradle property `kotlin.daemon.useFallbackStrategy`, whose default value is `true`.
```
PetrakovichVictoria, line 304: @sarahhaggarty Could you please help me with this sentence? Does it sound weird with this beginning?

PetrakovichVictoria: Thanks!

PR #3345, opened 2022-12-26
fix: fix Groovy syntax

PetrakovichVictoria:


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -78,19 +78,26 @@ It is also possible to configure all of the Kotlin compilation tasks in the proj
...
 </tab>
 <tab title="Groovy" group-key="groovy">
 
 ```groovy
-tasks.withType(org.jetbrains.kotlin.gradle.dsl.KotlinCompile).configureEach {
-    kotlinOptions { /*...*/ }
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
+// ...
+
+tasks.named('compileKotlin', KotlinCompilationTask) {
```
PetrakovichVictoria, line 96: I'm not sure about `named('compileKotlin')`

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -93,19 +93,26 @@ It is also possible to configure all of the Kotlin compilation tasks in the proj
 <tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
-tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask<*>>().configureEach {
-    compilerOptions { /*...*/ }
-}
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
+// ...
+
+val compileKotlin: KotlinCompilationTask<*> by tasks
```
Tapchicoma, line 99: Better to use here the same approach as for Groovy:
```
val compileKotlin = tasks.named<KotlinCompilationTask<*>>("compileKotlin")
```

Tapchicoma: `by tasks` eagerly realizes task

docs/topics/gradle/gradle-compiler-options.md
Tapchicoma: `tasks.named('compileKotlin', KotlinCompilationTask)` should work

PetrakovichVictoria:
I merged because I've fixed all @Tapchicoma's remarks. Seems like there is no need to waste his time in rechecking.

PR #3354, opened 2022-12-28
update: add about Removal of the old JVM backend

PetrakovichVictoria:


docs/topics/compatibility-guide-18.md
```diff
@@ -492,6 +492,23 @@ perspective
...
+> **Incompatible change type**: behavioral
+>
+> **Short summary**: Starting from 1.5.0, Kotlin uses the [IR-based backend](https://blog.jetbrains.com/kotlin/2021/02/the-jvm-backend-is-in-beta-let-s-make-it-stable-together/)
+> for the Kotlin/JVM compiler by default. In Kotlin 1.8.0, the old backend is deleted.
+>
+> **Deprecation cycle**:
+>
+> - < 1.5: by default, the old JVM backend is used
+> - \>= 1.5: by default, the IR-based backend is used
+> - 1.8.0: remove the old backend, the `-Xuse-old-backend` compiler option, and the `useOldBackend` Gradle option
```
ilya-g, line 510: I believe this is not a language incompatible change, more likely a tooling one. Moreover, [we do not promise](https://kotlinlang.org/docs/compiler-reference.html#x) that `-X` options are stable, so I'm not even sure that it has place in the compatibility guide.
All known incompatible changes appeared in the result of switching to the IR backend should have been covered in previous compatibility guides.

ilya-g (changes requested):

docs/topics/compatibility-guide-18.md
PetrakovichVictoria: I almost agree, but what about the `useOldBackend` Gradle option? It's also removed; it was not covered in any guide as deprecated/removed.

ilya-g: This indeed can be placed in the tooling section together with the other Gradle plugin API changes.

PetrakovichVictoria:
@koshachy please follow up

ilya-g (approved):

PetrakovichVictoria:
@udalov JFYI

PR #3368, opened 2023-01-09
Add if-else-if example to Conditions and loops page

sarahhaggarty:
Add an if-else-if example to the [Conditions and loops](https://kotlinlang.org/docs/control-flow.html) page to resolve [KWHF-789.](https://youtrack.jetbrains.com/issue/KWHF-789)

### DIFF
--- a/docs/topics/control-flow.md
+++ b/docs/topics/control-flow.md
@@ -6,20 +6,34 @@ In Kotlin, `if` is an expression: it returns a value.
 Therefore, there is no ternary operator (`condition ? then : else`) because ordinary `if` works fine in this role.
 
 ```kotlin
-var max = a 
-if (a < b) max = b
+fun main() {
+  val a = 2
+  val b = 3
 
-// With else 
-var max: Int
-if (a > b) {
+  //sampleStart
+  var max = a
+  if (a < b) max = b
+
+  // With else
+  if (a > b) {
     max = a
-} else {
+  } else {
     max = b
+  }
+
+  // As expression
+  max = if (a > b) a else b
+
+  // You can also use `else if` in expressions:
+  val maxLimit = 1
+  val maxOrLimit = if (maxLimit > a) maxLimit else if (a > b) a else b
+
+  //sampleEnd
+  println("max is $max")
+  println("maxOrLimit is $maxOrLimit")
 }
- 
-// As expression 
-val max = if (a > b) a else b
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="if-else-if-kotlin"}
 
 Branches of an `if` expression can be blocks. In this case, the last expression is the value of a block:
 


sarahhaggarty:
Reviewed and approved by Anton Arhipov. @PetrakovichVictoria, could you please review?

docs/topics/control-flow.md
```diff
@@ -6,30 +6,57 @@ In Kotlin, `if` is an expression: it returns a value.
 Therefore, there is no ternary operator (`condition ? then : else`) because ordinary `if` works fine in this role.
 
 ```kotlin
-var max = a 
-if (a < b) max = b
+val tax: Int
+val value = 50
```
PetrakovichVictoria, line 10: Let's rename "value" to a more meaningful name. For example, income? It's not good to call variables "value"/"variable" and so on; it makes code hard readable.

docs/topics/control-flow.md
```diff
@@ -6,30 +6,57 @@ In Kotlin, `if` is an expression: it returns a value.
...
+
+// With else if
+val tax: Int
+val value = 50
+
+if (value < 10) {
+  tax = 1
+} else if (value < 100) {
+  tax = 10
+} else if (value < 1000) {
```
PetrakovichVictoria, line 35: I would remove this branch; I think it's enough to show one "else if" branch. Otherwise, this new example is very wordy and this additional branch doesn't add any valuable sense.

docs/topics/control-flow.md
```diff
@@ -6,30 +6,57 @@ In Kotlin, `if` is an expression: it returns a value.
...
-val max = if (a > b) {
-    print("Choose a")
-    a
+val tax = if (value < 10) {
+    print("Tax is 1%")
+    1
+} else if (value < 100) {
+    print("Tax is 10%")
+    10
+} else if (value < 1000) {
```
PetrakovichVictoria, line 54: The same as in the comment above.

PetrakovichVictoria (changes requested):
Some minor things to improve.

Actually, I liked the brevity of the previous variant, but I see that you wanted to show the usage of an "else if" branch, and you need something more meaningful than "a" and "b" to do it correctly and with some sense. Of course, calling variables "a" is also not according to Martin's postulates in his "Clean Code"; however, I suppose that sometimes not adding additional sense to some easy things is good and doesn't distract a reader from a topic we want to explain.

docs/topics/control-flow.md
sarahhaggarty: Would "productValue" be acceptable?

PetrakovichVictoria: Let's pause this review for a bit: let's discuss with the team our approach to writing examples.

docs/topics/control-flow.md
```diff
@@ -6,30 +6,52 @@ In Kotlin, `if` is an expression: it returns a value.
...
+
+if (value < 10) {
+    tax = 1
 } else {
-    max = b
+    tax = 10
 }
  
 // As expression 
-val max = if (a > b) a else b
```
PetrakovichVictoria, line 21: I would propose returning everything as it was and just adding a couple of lines like this:

> // You can also use `else if` branches in expressions:
> val maxOrLimit = if (maxLimit > a) maxLimit else if (a > b) a else b

We can discuss our approach to examples on our next sync.

sarahhaggarty: @PetrakovichVictoria, I've made changes as we discussed over Slack. Is this good to go?

docs/topics/control-flow.md
```diff
@@ -6,20 +6,34 @@ In Kotlin, `if` is an expression: it returns a value.
...
     max = a
-} else {
+  } else {
     max = b
+  }
+
+  // As expression
+  max = if (a > b) a else b
+
+  // You can also use else if branches in expressions:
```
PetrakovichVictoria, line 27:
```suggestion
  // You can also use `else if` in expressions:
```

PetrakovichVictoria: I wouldn't call it a branch; it's just an ordinary "if" in the "else" branch.

PetrakovichVictoria:
Except for this minor [comment](https://github.com/JetBrains/kotlin-web-site/pull/3368#discussion_r1101621905), everything else LGTM! Thank you for this work!👍

docs/topics/control-flow.md
PetrakovichVictoria: (And I'm sorry, it was my proposal about "else if branches", but after that research (I gave links in Slack), I realized that there are no "else if" branches as in Kotlin, as in Java. Sometimes people call them so but I think we shouldn't to be accurate)

PetrakovichVictoria (approved):

PR #3370, opened 2023-01-09
Update kspVersion to 1.0.8

Jon889:
There is no KSP version 1.0.7, only 1.0.8, for Kotlin 1.8.0 https://github.com/google/ksp/releases


koshachy (approved):
Thanks for pointing out!

PR #3371, opened 2023-01-10
Remove Dokka generation content from KDoc and Dokka page

sarahhaggarty:
As we introduce Dokka's documentation, this PR is to update the KDoc and Dokka page to remove any duplicating content and focus scope on KDoc.

### DIFF
--- a/docs/topics/kotlin-doc.md
+++ b/docs/topics/kotlin-doc.md
@@ -1,15 +1,13 @@
-[//]: # (title: Document Kotlin code: KDoc and Dokka)
+[//]: # (title: Document Kotlin code: KDoc)
 
-The language used to document Kotlin code (the equivalent of Java's Javadoc) is called **KDoc**. In its essence, KDoc
+The language used to document Kotlin code (the equivalent of Java's Javadoc) is called **KDoc**. In essence, KDoc
 combines Javadoc's syntax for block tags (extended to support Kotlin's specific constructs) and Markdown for
 inline markup.
 
-## Generate the documentation
-
-Kotlin's documentation generation tool is called [Dokka](https://github.com/Kotlin/dokka). See the
-[Dokka README](https://github.com/Kotlin/dokka/blob/master/README.md) for usage instructions.
-
-Dokka has plugins for Gradle, Maven, and Ant, so you can integrate documentation generation into your build process.
+> Kotlin's documentation engine: Dokka, understands KDoc and can be used to generate documentation in various formats.
+> For more information, read our [Dokka documentation](dokka-introduction.md).
+>
+{type="note"}
 
 ## KDoc syntax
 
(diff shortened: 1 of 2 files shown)


docs/topics/kotlin-doc.md
```diff
@@ -137,36 +135,10 @@ Names in links are resolved using the same rules as if the name was used inside
 In particular, this means that if you have imported a name into the current file, you don't need to fully qualify it
 when you use it in a KDoc comment.
 
-Note that KDoc does not have any syntax for resolving overloaded members in links. Since the Kotlin documentation generation
+Note that KDoc does not have any syntax for resolving overloaded members in links. Since Kotlin's documentation generation
 tool puts the documentation for all overloads of a function on the same page, identifying a specific overloaded function
 is not required for the link to work.
 
-## Module and package documentation
```
IgnatBeresnev, line 144: Oof, I don't think this part is actually covered in Dokka's documentation, it might instead link to this section, actually. And I'm not 100% sure it should migrate there

The primary concern is that this format might already be used by other tools and thus relied upon. I'm afraid we'd be rug pulling if Dokka became its owner and started changing things for its own use.

The safest option would be to re-write this section to be tool-independent, i.e just describe what it could be used for and what the format is, and then mention at the end that Dokka can understand it. 

But since there's no predefined name/path for such files (unlike with Java's `package-info.java`), I don't know how useful it'll be and whether we should keep it here

I'll ask for a second opinion and come back to you

docs/topics/kotlin-doc.md
```diff
@@ -1,15 +1,13 @@
-[//]: # (title: Document Kotlin code: KDoc and Dokka)
+[//]: # (title: Document Kotlin code: KDoc)
 
 The language used to document Kotlin code (the equivalent of Java's Javadoc) is called **KDoc**. In its essence, KDoc
 combines Javadoc's syntax for block tags (extended to support Kotlin's specific constructs) and Markdown for
 inline markup.
```
IgnatBeresnev, line 5: Just thinking out loud: I wonder if we're past the point of needing to compare Kotlin with Java. It totally made sense in the beginning as people were primarily migrating from Java, but how many users/beginners are there now that are not familiar with Java and Javadoc?

IgnatBeresnev:
We discussed that this could be a good chance to revisit the whole page and see if it needs any additional changes. My thoughts and observations, do tell me if I need to create separate tickets for any of them:

___

I can't find the issue now, but a user came to Dokka's issues asking about the `@property` tag. They were just starting out with Kotlin and presumably never coded in Java.

The user wanted to annotate class properties, they saw [the example](https://kotlinlang.org/docs/kotlin-doc.html#kdoc-syntax) and did this:

```kotlin
/**
 * @property myProperty description
 * for my property
 */
class MyClass {

    val myProperty = "value"
}
```

Whereas it's not against the spec, the more common approach would be

```kotlin
class MyClass {

    /**
     * description
     * for my property
     */
    val myProperty = "value"
}
```

^ this is obvious and comes natural to those with some Java background as it's the only way to annotate fields, but it's not clear for beginners without Java experience.

When I told them the same thing as I outlined above, the user said they had documented a lot of code this way already and probably wouldn't change it.

I'd add an example of documenting standalone properties to the example, as well as examples for some other popular tags.

___

Currently, none of the tags have examples, and for beginners it might not be obvious how they can be used, for instance see https://github.com/Kotlin/dokka/issues/2595

I'd be happy to help come up with small and universal examples
___

> @see identifier﻿
> Adds a link to the specified **class or method** to the See also block of the documentation.

I don't think it's limited to classes or methods only. It works with properties, enum entries, interfaces, typealiases. IntelliJ IDEA also resolves packages and parameters. Should it be changed to `element` to avoid possible confusion?

docs/topics/kotlin-doc.md
IgnatBeresnev: We've discussed it with the team and came to the conclusion that it's unlikely that anyone other than Dokka is using it, and it's unlikely that this will be part of the KDoc specification itself in its current state.

So we've agreed to move it to Dokka's documentation. I can't think of a place where we can put it though, since it's runner-independent, so it looks like it's going to be a separate page altogether

@sarahhaggarty do you have any objections with this? I can move it by the time we deploy documentation live, but you can do it yourself if you have the time

sarahhaggarty: @IgnatBeresnev, no objections from me. Do you know if we have an example of this I could preview somewhere?

docs/topics/kotlin-doc.md
sarahhaggarty: I actually asked this question recently and apparently the majority of users are still Java developers migrating to Kotlin. So I think for now we should keep it as it is. Perhaps if the audience type shifts in future we can reconsider.

sarahhaggarty:
> We discussed that this could be a good chance to revisit the whole page and see if it needs any additional changes. My thoughts and observations, do tell me if I need to create separate tickets for any of them:

@IgnatBeresnev, thanks for taking the time to complete such a thorough review of the page! I've captured your feedback in [KT-55998](https://youtrack.jetbrains.com/issue/KT-55998/Update-KDoc-page). There are some other KDoc related tickets that I'd like to combine with this to revamp the page. I'll do the page revamp in a separate PR as it will need a lot of review from different people. I will definitely ask you about some good examples to add as part of that 😄

docs/topics/kotlin-doc.md
IgnatBeresnev: Thanks for sharing!

IgnatBeresnev (approved):
Looking good!

IgnatBeresnev:
> There are some other KDoc related tickets that I'd like to combine with this to revamp the page

That's awesome to hear, thanks! I see there are some linked issues with changes that I wanted to propose, but hesitated (like adding examples for code blocks)

danil-pavlov (approved):

PR #3373, opened 2023-01-12
fix: fix small typo from "need" to "want"

Elio-Muculo:
fix small typo on [Properties Page](https://kotlinlang.org/docs/properties.html#backing-fields) that lead to misunderstanding


koshachy (approved):
Thanks!

PR #3379, opened 2023-01-16
fix: correct a file path

PetrakovichVictoria:


PetrakovichVictoria:
Approved by @antonarhipov in DM.

PetrakovichVictoria:
Fixing this just in case we use these materials again.

PR #3388, opened 2023-01-21
fix fold button display

Skwodo:
Fixed unfolded button being offset with the folded one

Before:
image

After: 
image


zoobestik (approved):

zoobestik:
Thank you!

PR #3389, opened 2023-01-23
Update kotlinx html dependency

hfhbd:


koshachy (approved):
Thanks for pointing out!

PR #3397, opened 2023-01-25
Use PascalCase for type alias in documentation

vadimsemenov:


koshachy (approved):
True, this one should use PascalCase.
Thanks!

PR #3403, opened 2023-01-29
Update extensions.md

debChowdhury:
Added definition of virtual functions. It may be good to find info about virtual functions here in this page.


sarahhaggarty:
Hi @debChowdhury! Firstly, sorry for taking so long to respond to your PR. I've reviewed your suggestion and I think it's better to remove any mention of "virtual" because we don't mention it in any of our other materials. It can also be easily confused with the C++ definition of "virtual functions" which is not the same as Kotlin's. Instead of suggesting changes for you to apply to this PR, I've updated the paragraph to provide a simpler explanation as part of another PR: #3706. So I'll close this PR. Thanks for sharing your thoughts with us. If you'd like to contribute again, please request my review with your next PR and I'll take a look post-haste!

PR #3404, opened 2023-01-29
Update data-classes.md

debChowdhury:
Given explanation for why 2 objects of the same data class having different body properties but same constructor properties are equal.


sarahhaggarty:
Hi @debChowdhury! Again, sorry for taking so long to respond to your PR. I've reviewed your suggestion and I agree that I think this can be explained better. Instead of suggesting changes for you to apply to this PR, I've updated the paragraph to provide a simpler explanation as part of another PR: #3706. So I'll close this PR. Thanks for sharing your thoughts with us!

PR #3412, opened 2023-01-31
Multiplatform Library Guide Erroneous Code Fix

mackimaow:
There is some erroneous code in the multiplatform-library guide that is replaced.
Also, some clarification to vague parts of the guide is added.


danil-pavlov (approved):
Thanks a lot for your contribution! It helped to fix a lot of mistakes

PR #3430, opened 2023-02-16
Add experimental Enum.entries and update examples

sarahhaggarty:
Explain the new `entries` property for Enum classes. This page needs restructuring but I plan to do this in a separate PR to save time on review.

### DIFF
--- a/docs/topics/enum-classes.md
+++ b/docs/topics/enum-classes.md
@@ -87,11 +87,23 @@ EnumClass.valueOf(value: String): EnumClass
 EnumClass.values(): Array<EnumClass>
 ```
 
+Below is an example of these methods in action:
+
+```kotlin
+enum class RGB { RED, GREEN, BLUE }
+
+fun main() {
+    for (color in RGB.values()) println(color.toString()) // prints RED, GREEN, BLUE
+    println("The first color is: " + RGB.valueOf("RED")) // prints "The first color is: RED"
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="rgb-enums-kotlin"}
+
 The `valueOf()` method throws an `IllegalArgumentException` if the specified name does
 not match any of the enum constants defined in the class.
 
 You can access the constants in an enum class in a generic way using
-the `enumValues<T>()` and `enumValueOf<T>()` functions:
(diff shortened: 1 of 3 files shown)


docs/topics/enum-classes.md
```diff
@@ -87,11 +87,23 @@ EnumClass.valueOf(value: String): EnumClass
...
 ```
 
+Below is an example of these methods in action:
+
+```kotlin
+enum class RGB { RED, GREEN, BLUE }
+
+fun main() {
+    for (color in RGB.values()) println(color.toString()) // prints RED, GREEN, BLUE
+    println("The first color is: " + RGB.valueOf("RED")) // prints "The first color is: RED"
```
zarechenskiy, line 97: Minor: `println("The first color is: ${RGB.valueOf("RED")}")`

docs/topics/enum-classes.md
```diff
@@ -103,9 +115,41 @@ inline fun <reified T : Enum<T>> printAllValues() {
...
+enum class RGB { RED, GREEN, BLUE }
+
+@OptIn(ExperimentalStdlibApi::class)
+fun main() {
+    for (color in RGB.entries) println(color.toString())
+    // prints RED, GREEN, BLUE
+}
+```
+
+> The `entries` property is [Experimental](components-stability.md#stability-levels-explained). To use it you need
```
zarechenskiy, line 133: Hmm, there is a problem that this compatibility guide is a bit incorrect and confuses technologies such as Kotlin/JVM and certain language features which should have their own definitions of compatibility levels. We are thinking about how to redesign it to include correct information about language features. I'd propose leaving the text as is, but remove the link to this guide, there should be enough explanation in `ExperimentalStdlibApi` annotation

zarechenskiy (changes requested):
Looks good, only minor notes, thank you!

docs/topics/enum-classes.md
```diff
@@ -103,9 +115,41 @@ inline fun <reified T : Enum<T>> printAllValues() {
 printAllValues<RGB>() // prints RED, GREEN, BLUE
 ```
 
-Every enum constant has properties for obtaining its name and position (starting with 0) in the enum class declaration:
+If you don't want your enum constants returned in an array, you can use the `entries` property instead. The
```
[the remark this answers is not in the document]
sarahhaggarty: Thanks @zarechenskiy! Can you check if the updated text is OK now?

zarechenskiy (approved):
👍

PR #3434, opened 2023-02-21
Fix link to LINQ (old link no longer update)

ningpp:
old link no longer update


sarahhaggarty (approved):
Thanks for your contribution @ningpp!

PR #3437, opened 2023-02-23
KTL-1010 StdLib documentation: Move build site pages to the TeamCity DSL

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1010/StdLib-documentation-Move-build-site-pages-to-the-TeamCity-DSL


.teamcity/builds/kotlinlang/SiteProject.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import builds.kotlinlang.buidTypes.*
+import builds.kotlinlang.templates.DockerImageBuilder
+import builds.kotlinlang.vcs.GrammarGenerator
+import builds.kotlinlang.vcs.KotlinSpec
+import builds.kotlinlang.vcs.WebHelp
+import jetbrains.buildServer.configs.kotlin.Project
+
+object SiteProject : Project({
+  name = "Build kotlinlang.org"
```
zoobestik, line 11: Build is neccessery? I still think its better have structure based by project more then function 🙄 like that: 

 - kotlinlang.org
    - build
    - deploy

But it's unnecessary, we can discuss about somewhere else 🙄

.teamcity/builds/kotlinlang/buidTypes/BuildSearchIndex.kt
```diff
@@ -0,0 +1,81 @@
...
+import jetbrains.buildServer.configs.kotlin.AbsoluteId
+import jetbrains.buildServer.configs.kotlin.BuildType
+import jetbrains.buildServer.configs.kotlin.FailureAction
+import jetbrains.buildServer.configs.kotlin.buildFeatures.dockerSupport
+import jetbrains.buildServer.configs.kotlin.buildSteps.ScriptBuildStep
+import jetbrains.buildServer.configs.kotlin.buildSteps.script
+import jetbrains.buildServer.configs.kotlin.triggers.schedule
+import vcsRoots.KotlinLangOrg
+
+const val SearchAppId = "7961PKYRXV"
```
zoobestik, line 12: It would be nice move it higher and use for all builds.

.teamcity/builds/kotlinlang/buidTypes/BuildSitePages.kt
```diff
@@ -0,0 +1,208 @@
...
+
+  steps {
+    script {
+      name = "Build html pages"
+      scriptContent = """
+                #!/bin/bash
+                
+                set -x
+                
+                cat $kotlinWebsiteSetup
```
zoobestik, line 45: O_o

.teamcity/builds/kotlinlang/buidTypes/BuildSitePages.kt
```diff
@@ -0,0 +1,208 @@
...
+                echo " ##teamcity[buildStatus text='{build.status.text} ${'$'}{url}'] "
+            """.trimIndent()
+    }
+  }
+
+  features {
+    /*notifications {
+      notifierSettings = slackNotifier {
+        connection = "PROJECT_EXT_486"
+        sendTo = "#kotlin-web-site-e2e-tests"
```
zoobestik, line 112: Not so secret, but it's better to remove it, what do you think?

src/search.py
```diff
@@ -70,10 +70,6 @@ def get_page_views_statistic() -> Dict[str, int]:
     return page_views
 
 
-def get_client():
```
zoobestik, line 73: What happened here? :)

zoobestik (approved):

src/search.py
nikpachoo: it is obsolete one, we don't use it anywhere

.teamcity/builds/kotlinlang/buidTypes/BuildSitePages.kt
nikpachoo: It just works as a variable because it is a Kotlin file, no magic here :)

.teamcity/builds/kotlinlang/SiteProject.kt
nikpachoo: Yeah, I agree with your proposal. 
I've fixed the name, but the structure itself in DSL right now is a bit different. I suggest to return to aligning the structure when the first deploy is created.

PR #3442, opened 2023-02-24
fix: gradle languageVersion fix for .kts

PascalHonegger:
I tried to follow the current documentation and noticed that it didn't compile when using the provided example code in my .kts Gradle project. As far as I can tell, the `it` was simply copy-pasted from groovy and doesn't belong there, which is why I removed it.


PetrakovichVictoria:
@PascalHonegger you're right; thank you!

PR #3444, opened 2023-02-25

ltttttttttttt:
fix path


danil-pavlov:
Thank you! Fixed in #3412

PR #3460, opened 2023-03-05
Change title for "Sealed Classes" page

SebastianAigner:
The page talks about both sealed classes and sealed interfaces. This PR changes the title to reflect this.


sarahhaggarty (changes requested):
Thanks for your suggestion @SebastianAigner. Can you update the `kr.tree` as well? Let me know if you need any assistance.

sarahhaggarty (approved):

PR #3461, opened 2023-03-05
Expand section on data objects

SebastianAigner:
This PR expands the data objects section with information about the `equals` and `hashCode` semantics for `data objects` in preparation for the Kotlin 1.8.20 release.


docs/topics/object-declarations.md
```diff
@@ -193,32 +193,91 @@ fun main() {
...
+    println(MyDataObject) // MyDataObject
 }
 ```
 
-[Sealed class hierarchies](sealed-classes.md) are a particularly good fit for `data object` declarations, since they allow you to maintain symmetry with any data classes you might have defined alongside the object:
+The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
+In most cases, you will only have a single instance of your data object at runtime (after all, a `data object` declares a singleton).
+However, in the edge case where another object of the same type is generated at runtime (e.g. due to reflection or when using a serialization library), this ensures that the objects are treated as being equal. 
+
+> Make sure to only compare `data objects` structurally (i.e. using the `==` operator), never by reference (the `===` operator).
```
strangepleasures, line 218: I'd expand this a little, e.g. "To avoid the pitfalls of having more than one instance of a data object class, make sure to only compare `data objects` structurally (i.e. using the `==` operator), never by reference (the `===` operator), similarly to data classes."

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,91 @@ fun main() {
...
 
 fun main() {
-    println(MyObject) // MyObject
+    println(MyDataObject) // MyDataObject
 }
 ```
 
-[Sealed class hierarchies](sealed-classes.md) are a particularly good fit for `data object` declarations, since they allow you to maintain symmetry with any data classes you might have defined alongside the object:
+The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
+In most cases, you will only have a single instance of your data object at runtime (after all, a `data object` declares a singleton).
```
strangepleasures, line 215: "The `hashCode()` function is implemented to be consistent with `equals()`, so that all instances of a certain data object class will have the same hash code."

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,91 @@ fun main() {
...
-sealed class ReadResult {
-    data class Number(val value: Int): ReadResult()
-    data class Text(val value: String): ReadResult()
-    data object EndOfFile: ReadResult()
+import java.lang.reflect.Constructor
+
+data object MySingleton
+
+fun main() {
+    val fromSerializationLibrary = createInstanceViaReflection()
```
strangepleasures, line 227: Via reflection or serialization? These are different things, even though a serialization library may use reflection. I'd maybe give the variable a humorist name, such as `evilTwin` or `dopplelganger`, to make it clear that it's an exceptional thing to occur.

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,91 @@ fun main() {
...
 fun main() {
-    println(MyObject) // MyObject
+    println(MyDataObject) // MyDataObject
 }
 ```
 
-[Sealed class hierarchies](sealed-classes.md) are a particularly good fit for `data object` declarations, since they allow you to maintain symmetry with any data classes you might have defined alongside the object:
+The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
+In most cases, you will only have a single instance of your data object at runtime (after all, a `data object` declares a singleton).
+However, in the edge case where another object of the same type is generated at runtime (e.g. due to reflection or when using a serialization library), this ensures that the objects are treated as being equal.
```
strangepleasures, line 216: It's probably worth mentioning that Kotlin reflection API doesn't permit instantiation of object classes, so you're safe here. It can be only achieved via platform reflection, e.g. `java.lang.reflction.*` API.  The same should be true for serialization libraries as well.

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,91 @@ fun main() {
...
+}
+
+fun createInstanceViaReflection(): MySingleton {
+    // Creates a new MySingleton instance "by force" (i.e. reflection)
+    // Don't do this yourself!
+    return (MySingleton.javaClass.declaredConstructors[0].apply { isAccessible = true } as Constructor<MySingleton>).newInstance()
+}
+```
+
+The automatically generated `hashCode` implementation accompanies the semantics of the `equals()` function.
```
strangepleasures, line 246: "..., so that all instances of a certain data object class will have the same hash code."

strangepleasures (changes requested):

docs/topics/object-declarations.md
SebastianAigner: I've expanded the warning. I'm not sure about the "data object class" term, so take a look and see if the new phrasing works for you :)

docs/topics/object-declarations.md
SebastianAigner: I've expanded the paragraph below the code sample. Do you want the `hashCode` function mentioned here, instead?

docs/topics/object-declarations.md
SebastianAigner: changed to `evilTwin`, good idea!

strangepleasures (approved):

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
 
 fun main() {
-    println(ReadResult.Number(1)) // Number(value=1)
-    println(ReadResult.Text("Foo")) // Text(value=Foo)
-    println(ReadResult.EndOfFile) // EndOfFile
+  printReadResult(EndOfFile) // EOF
 }
 ```
 
+
```
danil-pavlov, line 279:
```suggestion

```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+
```
danil-pavlov, line 280:
```suggestion

```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+
```
danil-pavlov, line 281:
```suggestion

```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+    is Text -> println("Txt(${r.text}")
+    is EndOfFile -> println("EOF")
+  }
 }
 
 fun main() {
-    println(ReadResult.Number(1)) // Number(value=1)
-    println(ReadResult.Text("Foo")) // Text(value=Foo)
-    println(ReadResult.EndOfFile) // EndOfFile
+  printReadResult(EndOfFile) // EOF
```
danil-pavlov, line 275:
```suggestion
    printReadResult(EndOfFile) // EOF
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+`data object` declarations are a particularly good fit for sealed hierarchies, like [sealed classes or sealed interfaces](sealed-classes.md), since they allow you to maintain symmetry with any data classes you might have defined alongside the object:
+
+```kotlin
+sealed interface ReadResult
+data class Number(val number: Int) : ReadResult
+data class Text(val text: String) : ReadResult
+data object EndOfFile : ReadResult
+
+fun printReadResult(r: ReadResult) {
+  when(r) {
```
danil-pavlov, line 267:
```suggestion
    when(r) {
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+    is Number -> println("Num(${r.number}")
+    is Text -> println("Txt(${r.text}")
```
danil-pavlov, line 269:
```suggestion
        is Text -> println("Txt(${r.text}")
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+    is EndOfFile -> println("EOF")
```
danil-pavlov, line 270:
```suggestion
        is EndOfFile -> println("EOF")
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+  }
```
danil-pavlov, line 271:
```suggestion
    }
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+The generated `hashCode()` function has behavior that is consistent with the `equals()` function, so that all runtime instances of a `data object` have the same hash code.
+
+#### No copy and componentN functions
+
+While `data object` and `data class` declarations are often used together and have some similarities, there are some functions that are not generated for a `data object`:
+
+- No `copy()` function: Because a `data object` declaration is intended to be used as singleton objects, no `copy()` function is generated. The singleton pattern restricts the instantiation of a class to one, single instance, which would be violated by allowing copies of the instance to be created.
+- No `componentN()` function: unlike a `data class`, a `data object` does not have any data properties. Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
+
+#### Use cases for data objects
```
danil-pavlov, line 256: Since it's just one use case for now, maybe we can use "Sealed hierarchies" header?

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+}
+```
+
+The generated `hashCode()` function has behavior that is consistent with the `equals()` function, so that all runtime instances of a `data object` have the same hash code.
+
+#### No copy and componentN functions
+
+While `data object` and `data class` declarations are often used together and have some similarities, there are some functions that are not generated for a `data object`:
+
+- No `copy()` function: Because a `data object` declaration is intended to be used as singleton objects, no `copy()` function is generated. The singleton pattern restricts the instantiation of a class to one, single instance, which would be violated by allowing copies of the instance to be created.
```
danil-pavlov, line 253:
```suggestion
* No `copy()` function. Because a `data object` declaration is intended to be used as singleton objects, no `copy()` function is generated. The singleton pattern restricts the instantiation of a class to a single instance, which would be violated by allowing copies of the instance to be created.
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+- No `componentN()` function: unlike a `data class`, a `data object` does not have any data properties. Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
```
danil-pavlov, line 254:
```suggestion
* No `componentN()` function. Unlike a `data class`, a `data object` does not have any data properties. Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+    // Kotlin reflection does not permit the instantiation of data objects.
+    // This creates a new MySingleton instance "by force" (i.e. Java platform reflection)
+    // Don't do this yourself!
+    return (MySingleton.javaClass.declaredConstructors[0].apply { isAccessible = true } as Constructor<MySingleton>).newInstance()
+}
+```
+
+The generated `hashCode()` function has behavior that is consistent with the `equals()` function, so that all runtime instances of a `data object` have the same hash code.
+
+#### No copy and componentN functions
```
danil-pavlov, line 249: I'd suggest "Differences between data objects and data classes", seems more descriptive. WDYT?

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
 }
 ```
 
-Just like [data classes](data-classes.md), you can mark your `object` declaration with the `data` modifier to get a nicely formatted string representation without having to manually provide an implementation for its `toString` function:
+Just like [data classes](data-classes.md), you can mark an `object` declaration with the `data` modifier. This instructs the compiler to generate a number of functions for your object:
 
+* `toString()` returning the name of the data object
+* `equals()`/`hashCode()` pair
+
+Note that you can't provide a custom `equals` or `hashCode` implementation for a `data object`.
```
danil-pavlov, line 201:
```suggestion
  > You can't provide a custom `equals` or `hashCode` implementation for a `data object`.
  >
  {type="note}
```

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
 fun main() {
-    println(MyObject) // MyObject
+    println(MyDataObject) // MyDataObject
 }
 ```
 
-[Sealed class hierarchies](sealed-classes.md) are a particularly good fit for `data object` declarations, since they allow you to maintain symmetry with any data classes you might have defined alongside the object:
+The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
+In most cases, you will only have a single instance of your data object at runtime (after all, a `data object` declares a singleton).
+However, in the edge case where another object of the same type is generated at runtime (e.g. via platform reflection through `java.lang.reflect`, or by using a JVM serialization library that uses this API under the hood), this ensures that the objects are treated as being equal.
```
danil-pavlov, line 216:
```suggestion
However, in the edge case where another object of the same type is generated at runtime (for example, by using platform reflection with `java.lang.reflect` or a JVM serialization library that uses this API under the hood), this ensures that the objects are treated as being equal.
```

danil-pavlov (changes requested):
Thank you! Huge work
please consider some header suggestions

docs/topics/object-declarations.md
```diff
@@ -193,32 +193,92 @@ fun main() {
...
+
+> Make sure to only compare `data objects` structurally (i.e. using the `==` operator), never by reference (the `===` operator). This helps you avoid pitfalls when more than once instance of a data object exists at runtime.
```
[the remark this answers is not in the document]
SebastianAigner: I've added back the word "using" because otherwise I don't think grammatically these work (you're using the operator to achieve the comparison). Other than that, agreed, that's a better phrasing!

docs/topics/object-declarations.md
SebastianAigner: While we don't do a full comparison in the paragraph, I think "differences" is also fine!

docs/topics/object-declarations.md
SebastianAigner: Seems two of these newline changes were conflicting after applying them in order. If you still need changes here, do take a look :)

danil-pavlov (approved):

PR #3464, opened 2023-03-08
feat: podspec files available in the new dsl for building final artifacts

danil-pavlov:


daugeldauge (commented):
I think it should be mentioned somewhere that cocoapods plugin need to be applied to make it work.

PR #3465, opened 2023-03-08
update: add example of setting a languageVersion

PetrakovichVictoria:


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -191,6 +191,31 @@ tasks.named('compileKotlin', KotlinCompilationTask) {
 
 Also, see [Types for compiler options](#types-for-compiler-options).
 
+#### Example of setting a languageVersion
```
Tapchicoma, line 194: Should we also provide an example how to set `apiVersion`? :thinking:

PetrakovichVictoria: I think that is not necessary; the main idea is to have code snippets with the languageVersion setting and to give a link to this example in What's new.

Tapchicoma (approved):

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -199,17 +199,27 @@ To set a language version, use the following syntax:
...
 </tab>
 <tab title="Groovy" group-key="groovy">
 
 ```groovy
-tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask.class).configureEach {
-    compilerOptions.languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
+tasks
+    .withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask.class)
+    .configureEach {
+        compilerOptions
```
Tapchicoma, line 220: Nit: a little strange formatting here. I would propose to do it following:
```groovy
compilerOptions.languageVersion =
    org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
```

Tapchicoma (approved):

danil-pavlov (approved):
One note, I'd move the sentence "Also, see Types for compiler options." below the example, It's just that the structure Attributes → Example of attribute → See also looks more natural to me

docs/topics/gradle/gradle-compiler-options.md
PetrakovichVictoria: Should I do almost the same for the Groovy script?
```
.configureEach {
        compilerOptions.languageVersion
            .set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9)
    }
```

PetrakovichVictoria:
@danil-pavlov, aren't you afraid that people who don't read the example about languageVersion will not see this "see also"?

danil-pavlov:
> @danil-pavlov, aren't you afraid that people who don't read the example about languageVersion will not see this "see also"?

I just think "Also, see [Types for compiler options](#types-for-compiler-options)." is less important reference information that usually goes at the very bottom of the section, and the example with code snippet should be higher.

PetrakovichVictoria:
> > @danil-pavlov, aren't you afraid that people who don't read the example about languageVersion will not see this "see also"?
> 
> I just think "Also, see [Types for compiler options](#types-for-compiler-options)." is less important reference information that usually goes at the very bottom of the section, and the example with code snippet should be higher.

@danil-pavlov, ok, I think that both of us are right; I changed as you proposed.

PR #3467, opened 2023-03-09
feat: memory manager statistics

danil-pavlov:


docs/topics/native/native-memory-manager.md
```diff
@@ -54,6 +54,15 @@ build script:
 
 ## Memory consumption
 
+### Check for memory leaks
+
+To access the memory manager metrics, call `kotlin.native.internal.GC.lastGCInfo()`. It returns statistics for the last
+run of the garbage collector.
+
+The statistics can be useful for debugging memory leaks or checking if there are any leaks when running tests.
```
kunyavskiy, line 62: I would suggest clarifying, that leacks can happen only because of storing some objects into globals.

Also I think having an example would be nice. For example, use something like this

```
import kotlin.native.internal.*
import kotlin.test.*

class A

val global = mutableListOf<A>()

@OptIn(kotlin.ExperimentalStdlibApi::class)
fun getUsage() : Long {
	GC.collect()
	return GC.lastGCInfo!!.memoryUsageAfter["heap"]!!.totalObjectsSizeBytes
}

fun run() {
	global.add(A())
	// test would fail, if remove next line
	global.clear()
}

@Test
fun test() {
	val before = getUsage()
	// separate function is used to be sure, that all temporaries are cleaned
	run()
	val after = getUsage()
	assertEquals(before, after)
}
```

danil-pavlov: Great suggestion! made a small fix on the class name and the opt-in annotation

docs/topics/native/native-memory-manager.md
```diff
@@ -54,9 +54,49 @@ build script:
...
+* Debugging memory leaks when using global variables
+* Checking if there are any leaks when running tests
+
+```kotlin
+import kotlin.native.internal.*
+import kotlin.test.*
+
+class Resource
+
+val global = mutableListOf<A>()
```
kunyavskiy, line 71: probably it should be Resource here, not A

danil-pavlov: Good catch! missed that one

kunyavskiy (approved):

docs/topics/native/native-memory-manager.md
```diff
@@ -54,9 +54,49 @@ build script:
 
 ## Memory consumption
```
sarahhaggarty, line 55: I think it would be good to have at least one sentence here to introduce the subject matter so that there isn't a sub-heading immediately after a heading. Maybe something like: "With the memory manager it's possible to monitor memory consumption so that you can check for memory leaks or detect other memory problems."

sarahhaggarty (approved):
Just one suggestion!

PR #3476, opened 2023-03-16
feat: import obj-c-extensions as class members

danil-pavlov:


docs/topics/native/native-objc-interop.md
```diff
@@ -208,10 +208,16 @@ Learn more about the [`async`/`await` mechanism in Swift](https://docs.swift.org
 
 ### Extensions and category members
 
-Members of Objective-C categories and Swift extensions are imported to Kotlin
+Members of Objective-C categories and Swift extensions are generally imported to Kotlin
 as extensions. That's why these declarations can't be overridden in Kotlin.
 And the extension initializers aren't available as Kotlin constructors.
 
+> Currently, there are two exceptions in UIKit. Starting with Kotlin 1.8.20, category members that are declared
```
sbogolepov, line 215: NSView is from AppKit and UIView is from UIKit.

docs/topics/native/native-objc-interop.md
```diff
@@ -521,12 +527,12 @@ After that the Objective-C header will contain a corresponding comment:
 Known limitations:
 * Dependency documentation is not exported unless it is compiled with `-Xexport-kdoc` itself. The feature is experimental, 
 so libraries compiled with this flag might be incompatible with other compiler versions.
-* KDoc comments are mostly exported "as is" , many KDoc features (for example, `@property`) are not supported.
+* KDoc comments are mostly exported "as is", many KDoc features (for example, `@property`) are not supported.
```
sarahhaggarty, line 530:
```suggestion
* KDoc comments are mostly exported "as is". Many KDoc features (for example, `@property`) are not supported.
```

docs/topics/native/native-objc-interop.md
```diff
@@ -78,8 +78,8 @@ These classes and interfaces are placed into a package [specified in build confi
 The names of Kotlin classes and interfaces are prefixed when imported to Objective-C.
 The prefix is derived from the framework name.
 
-Objective-C does not support packages in a framework. Thus, the Kotlin compiler renames Kotlin classes which have the
-same name but different package in the same framework. This algorithm is not stable yet and can change between Kotlin
+Objective-C does not support packages in a framework. Thus, the Kotlin compiler renames Kotlin classes, which have the
+same name but different package, in the same framework. This algorithm is not stable yet and can change between Kotlin
 releases. As a workaround, you can rename the conflicting Kotlin classes in the framework.
```
sarahhaggarty, line 83:
```suggestion
Objective-C does not support packages in a framework. If the Kotlin compiler finds Kotlin classes in the same framework which have the same name but different packages, it renames them. To work around this, you can rename the conflicting Kotlin classes in the framework.
```

docs/topics/native/native-objc-interop.md
```diff
@@ -170,7 +170,7 @@ Swift has only checked errors. So if Swift or Objective-C code calls a Kotlin me
 which throws an exception to be handled, then the Kotlin method should be marked
 with a `@Throws` annotation specifying a list of "expected" exception classes.
 
-When compiling to Objective-C/Swift framework, non-`suspend` functions having or inheriting
+When compiling to the Objective-C/Swift framework, non-`suspend` functions having or inheriting
```
sarahhaggarty, line 173:
```suggestion
When compiling to the Objective-C/Swift framework, non-`suspend` functions that have or inherit the
```

docs/topics/native/native-objc-interop.md
```diff
@@ -208,11 +208,17 @@ Learn more about the [`async`/`await` mechanism in Swift](https://docs.swift.org
...
 ### Extensions and category members
 
-Members of Objective-C categories and Swift extensions are imported to Kotlin
+Members of Objective-C categories and Swift extensions are generally imported to Kotlin
 as extensions. That's why these declarations can't be overridden in Kotlin.
 And the extension initializers aren't available as Kotlin constructors.
 
-Kotlin extensions to "regular" Kotlin classes are imported to Swift and Objective-C as extensions and category members respectively.
+> Currently, there are two exceptions. Starting with Kotlin 1.8.20, category members that are declared
+> in the same headers as the NSView class (from the AppKit framework) and UIView classes (from the UIKit framework) are
```
sarahhaggarty, line 216: Since you say there are two exceptions, should it be: "headers as the NSView class ... **or** UIView classes"?

docs/topics/native/native-objc-interop.md
```diff
@@ -208,11 +208,17 @@ Learn more about the [`async`/`await` mechanism in Swift](https://docs.swift.org
...
+> imported as members of these classes. It means that you can override methods subclassing from NSView and UIView.
```
sarahhaggarty, line 217:
```suggestion
> imported as members of these classes. This means that you can override methods when subclassing from NSView and UIView.
```

sarahhaggarty (commented):
Looking good @danil-pavlov! Just one question and a few minor suggestions. Should be super quick to resolve.

docs/topics/native/native-objc-interop.md
danil-pavlov: The initial idea was to say "you can override methods that subclass from NSView and UIView." but maybe your variant works too

docs/topics/native/native-objc-interop.md
danil-pavlov: I agree, let't use 'or' in all cases

PR #3477, opened 2023-03-17
Fix explanation of KDoc links to elements

DRSchlaubi:
You cannot actually use `[]()`for elements, but only for external links, for elements you use `[][]`.

Not sure whether that's intended or not, but that's how the IntelliJ Plugin interprets it, so I figured it's intended


sarahhaggarty:
Thanks for letting us know about this @DRSchlaubi! I've incorporated this as well as some other changes as part of #3510

PR #3490, opened 2023-03-22
Clarify floating-point numbers comparison

sarahhaggarty:
Ticket: [KT-57282](https://youtrack.jetbrains.com/issue/KT-57282/Numbers-running-code-gives-different-result-than-stated-in-the-docs)

This PR clarifies the difference in behavior of statically typed/not statically typed operands as floating point numbers when completing comparisons.


PetrakovichVictoria (approved):
LGTM. One question: did Ilya also check the text? Just in case

PR #3491, opened 2023-03-22
feat: setting up hierarchy in multiplatform projects

danil-pavlov:


docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
-projects, [update to the latest release](releases.md#update-to-a-new-release). If you need to keep using an earlier version
-than 1.6.20, you can still enable this feature manually. For this, add the following to your `gradle.properties`:
-
-```none
-kotlin.mpp.enableGranularSourceSetsMetadata=true
-kotlin.native.enableDependencyPropagation=false
-```
+Multiplatform projects have support for hierarchical structure.
+This means you can form a hierarchy of source sets for sharing the common code among some, but not all,
+[supported targets](multiplatform-dsl-reference.md#targets).
```
dsavvinov, line 5: After this sentence, I immediately have a question: why would I want to share my code across some, but not all targets? 

The next few sentences kinda discuss it, but don't put it clearly enough. Let's try to clarify that.

Messaging:
Why do you need intermediate source-sets?

* If you're a library author, you might provide a specialized API for some, but not all targets (e.g. specialized API for K/N targets, but not for K/JVM targets). For that, you'll need to introduce an intermediate source set

* If you're a consumer of a library, you might want to use that specialized API, for that you'll have to introduce a similar intermediate source set

* Additionally, some libraries are not even available on some platforms. Specifically, Native Platform libraries are obviously available only in source sets that compile for K/Native only

* Kotlin Toolchain will ensure that each source set has access only to the API that is available on all targets to which this source set compiles. You're safe from using Windows-only API and then compiling it to MacOS, getting linkage errors or even undefined behaviour at runtime!

---

Note that the case "Sharing code in similar targets" mentioned in the original text is omitted. This is because IMO it's quite vague. Like I can theoretically come up with some case like "I use this utility function only in Native source-sets, thus I'll place it in allNative", but frankly, no one does that AFAIK. If you can write code in common, you write it in common and don't mess with hierarchies. 

That's a somewhat wild swing from the initial messaging of this article, so I'd like to hear your opinions on what do you think about adding those messages here. cc @KaterinaPetrova

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
+> For the Kotlin Gradle build scripts, opt-in is required with `@OptIn(ExperimentalKotlinGradlePluginApi::class)`.
+>
+{type="warning"}
+
+Starting with Kotlin 1.8.20, you can set up source set hierarchy in your multiplatform projects with default target hierarchy.
+It's a template for all possible targets and their shared source sets hardcoded in the Kotlin Gradle plugin.
+
+When you declare the targets to which your project compiles,
+the plugin picks necessary source sets from the template and creates them in your project based on the specified targets.
+See the full hierarchy template:
```
dsavvinov, line 33: I think the full template interrupts the flow of reading heavily here. We just barely explained the basics, and most likely, it's not very clear what's going on without seeing an example, as we're avoiding strict terms and definitions and using "intuitive" wording like "the plugin picks necessary source sets" (which ones are necessary?), "based on the specific targets" (how exactly targets affect it?). 

Two options:
- Introduce a short example right here (you can then just reference it from the section with code sample, like: "This code will create the structure from the example above")
- Push full template somewhere below, and just give a hint to a reader that if they are confused at this point, it's OK and just keep reading, like: "Let's see how it works on the example"

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
 
-You can also use platform-dependent libraries, such as `UIKit` and `POSIX`, in source sets shared among several native
-targets. One popular case is having access to iOS-specific dependencies like `Foundation` when sharing code across all
-iOS targets. The new structure helps you share more native code without being limited by platform-specific dependencies.
+### Set up your project
 
-By using the hierarchical structure along with platform-dependent libraries in shared source sets, you can eliminate the
-need to use workarounds to get IDE support for sharing source sets among several native targets, for
-example `iosArm64` and `iosX64`:
+To set up a hierarchy, call `targetHierarchi.default()` in the `kotlin` block of your `build.gradle(.kts)` and list all the targets you need.
```
atyrin, line 44: typo `targetHierarchi`

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
+@OptIn(ExperimentalKotlinGradlePluginApi::class)
+
 kotlin {
-    val iOSTarget: (String, KotlinNativeTarget.() -> Unit) -> KotlinNativeTarget =
-        if (System.getenv("SDK_NAME")?.startsWith("iphoneos") == true)
-            ::iosArm64
-        else
-            ::iosX64
+    // Enable the default target hierarchy:
+    targetHierarchi.default()
```
atyrin, line 52: the same typo `targetHierarchi`

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
+
+        val nativeMain by getting {
+            dependsOn(jvmAndNativeMain)
+        }
+
+        val jvmMain by getting {
+            dependsOn(jvmAndNativeMain)
+        }
+    }
+}
```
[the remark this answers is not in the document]
atyrin: 🤓 If you want it pretty hard, it is possible with
```
group("native"){
    withoutCompilations { it.target == watchosX64() }
}
```

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
-| No                                          | No                                          | ✅                                                                               |
+## Manual configuration
 
-## How to opt-out
+To create the hierarchical structure manually, introduce an intermediate source set that holds the shared code for
+several
+targets and create a structure of the source sets including the intermediate one.
 
-To disable hierarchical structure support, set the following option to `false` in your `gradle.properties`:
+For example, if you want to share code among native Linux, Windows, and macOS targets – `linuxX64M`, `mingwX64`, and
```
atyrin, line 249: typo in `linuxX64M`

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
+
+The resulting hierarchical structure will look like this:
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+kotlin {
+    sourceSets {
+        val commonMain by sourceSets.getting
```
atyrin, line 157: This code is already inside `sourceSets{}`. So, in `sourceSets.getting` call `sourceSets.` part might be omitted.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,322 @@
...
+### Create your own hierarchy
+
+If necessary, you can further configure the resulting hierarchy manually [using the dependsOn relation](#manual-configuration).
+To do that, apply the `by getting` construction for the source sets created with `targetHierarchy.default()`.
+
+Consider this example of a project with a source set shared between the `jvm` and `native` targets only:
+
+```kotlin
+@OptIn(ExperimentalKotlinGradlePluginApi::class)
+
```
atyrin, line 88: Is there an empty line between annotation and `kotlin{}` ? Seems unnecessary.

docs/topics/multiplatform/multiplatform-hierarchy.md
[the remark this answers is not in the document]
atyrin: 🤓 If you want it pretty hard, it is possible with
```
group("native"){
    withoutCompilations { it.target == watchosX64() }
}
```

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,21 +1,25 @@
...
+[supported targets](multiplatform-dsl-reference.md#targets). Introducing intermediate source sets has some important advantages:
+
+* If you're a library author, you might want to provide a specialized API for some, but not all targets. For example,
+  an intermediate source set for Kotlin/Native targets, but not for Kotlin/JVM may come in handy.
+* You might want to use platform-dependent libraries in your project. Introducing an intermediate source set allows
+  you to utilize that specialized API among several native targets. For example,
+  having access to iOS-specific dependencies like Foundation when sharing code across all iOS targets.
+* Some libraries aren't available for some platforms. Specifically, native libraries are only available for source sets
+  that compile to Kotlin/Native. Intermediate source set will solve this issue.
+* The Kotlin toolchain ensures that each source set has access only to the API that is available for all targets to which
```
dsavvinov, line 14: Very minor: I think this point isn't the _reason_ why you want to have an intermediate source set, it's just a nice feature we provide to you **if** you had to use an intemdetiate source set (by any of the reasons above). So I'd just remove it from the bullet point list

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -153,11 +156,13 @@ The resulting hierarchical structure will look like this:
 
 ```kotlin
 kotlin {
+    ios()
```
dsavvinov, line 159: As far as I can see, this section wants to show a manual configuration equivalent to the usage of `ios`, so you can't use `ios` here (it's kinda self-referential)

Use `iosArm64()` and `iosX64()` instead

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -171,6 +176,8 @@ kotlin {
 
 ```groovy
 kotlin {
+    ios()
```
dsavvinov, line 179: Ditto

docs/topics/whatsnew1620.md
```diff
@@ -481,7 +481,7 @@ By default, libraries published with the hierarchical project structure are comp
 #### Better code-sharing in your project
 
 Without hierarchical structure support, there is no straightforward way to share code across _some_ but not _all_ [Kotlin targets](multiplatform-dsl-reference.md#targets).
-One popular example is sharing code across all iOS targets and having access to iOS-specific [dependencies](multiplatform-share-on-platforms.md#use-native-libraries-in-the-hierarchical-structure), like `Foundation`.
+One popular example is sharing code across all iOS targets and having access to iOS-specific [dependencies](multiplatform-share-on-platforms.md#conn), like Foundation.
```
dsavvinov: NB: link looks weird

docs/topics/multiplatform/multiplatform-hierarchy.md
danil-pavlov: I guess we have a confusion here. The section is about target shortcuts actually. We introduce an example with the ios() shortcut first, then show a diagram, and then explain the resulting hierarchy. So I'm not sure what's the best way to write the code snippet for the resulting hierarchy in this case

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
-kotlin.mpp.enableGranularSourceSetsMetadata=true
-kotlin.native.enableDependencyPropagation=false
-```
+Multiplatform projects support hierarchical structures.
+This means you can form a hierarchy of intermediate source sets for sharing the common code among some, but not all,
+[supported targets](multiplatform-dsl-reference.md#targets). Introducing intermediate source sets has some important advantages:
+
+* If you're a library author, you might want to provide a specialized API for some, but not all targets. For example,
+  an intermediate source set for Kotlin/Native targets, but not for Kotlin/JVM may come in handy.
+* You might want to use platform-dependent libraries in your project. Introducing an intermediate source set allows
```
koshachy, line 9: In that case, you can use an intermediate source set to utilize that specialized (specific?) API among several native target. For example, to have access to iOS-specific dependencies, such as Foundation, when sharing code across all iOS targets.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+> In this example, `apple` and `native` source sets compile only to `iosArm64` and `iosSimulatorArm64` targets.
+> Therefore, despite their names, they have access to the full iOS API.
+> This might be counter-intuitive for source sets like `native`, as you may expect that only APIs available on all
+> native targets are accessible in this source set. This behavior may change in the future.
+>
+{type="note"}
+
+### Adjust resulting hierarchy
+
+If necessary, you can further configure the resulting hierarchy manually [using the `dependsOn` relation](#manual-configuration).
```
koshachy, line 74: It's not clear what actually "If necessary" means and how user can understand that it is necessary for their project.
Could we omit this?

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+    }
+}
 ```
-{initial-collapse-state="collapsed"}
 
-Instead of doing this, you can create a hierarchical structure
-with [target shortcuts](multiplatform-share-on-platforms.md#use-target-shortcuts)
-available for typical multi-target scenarios, or you can manually declare and connect the source sets. For example, you
-can create two iOS targets and a shared source set with the `ios()` shortcut:
+Removing `dependsOn` relations automatically created by `targetHierarchy.default()` could be cumbersome.
```
koshachy, line 107: It could be cumbersome (taking time) to remove "dependsOn", that was automatically created by the `...()) call.

Or any other way, just to make the sentence more easy to read.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+If it's necessary, use entirely [manual configuration](#manual-configuration) instead of calling default hierarchy.
```
koshachy, line 108: I still don't understand how a user understands if it's necessary or not.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+
+> We're currently working on API for creating your own target hierarchies. It should be useful for projects
+> which hierarchy configurations differ a lot from the default template.
```
koshachy, line 111: I think "a lot" can be omitted.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+>
+> At the moment, this API is not ready yet. But if you're eager to try it,
+> look into the `targetHierarchy.custom { ... }` block and the declaration of `targetHierarchy.default()` as an example.
+> Keep in mind that this API is still in development. It might be under-tested, and can change dramatically in further releases.
```
koshachy, line 115: "dramatically" can be omitted.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
+{type="tip"}
+
+#### See the full hierarchy template {initial-collapse-state="collapsed"}
+
+When you declare the targets to which your project compiles,
+the plugin picks shared source sets based on the specified targets from the template and creates them in your project.
+
+![Default target hierarchy](full-template-hierarchy.svg)
+
+> Here we describe the production part of the project only, omitting the suffix `Main`
```
koshachy, line 126: This example shows the production part of the project, omitting the `Main` suffix.
...

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -1,111 +1,340 @@
...
-| No                                          | Yes                                         | Library can't be used in a shared native source set                             |
-| No                                          | No                                          | ✅                                                                               |
+## Manual configuration
 
-## How to opt-out
+You can manually introduce an intermediate source in the source set structure.
+It will hold the shared code for several targets.
 
-To disable hierarchical structure support, set the following option to `false` in your `gradle.properties`:
+For example, if you want to share code among native Linux, Windows, and macOS targets – `linuxX64`, `mingwX64`, and
```
koshachy, line 259: Maybe it's better to use a "you + verb" instead of dash.

koshachy (approved):
Approved.
Huge work!
I'd recommend fixing all the wording issues and reviewing the intro part after the upcoming release.
Thanks!

PR #3493, opened 2023-03-23
Mention about the escape sequence `\$` since it's allowed in the current version

yuantj:
In [Concepts/Types/Basic types/Strings](https://kotlinlang.org/docs/strings.html), it mentions that:

> To insert the dollar sign $ in a raw string **(which doesn't support backslash escaping)** before any symbol ...

However, `\$` is allowed in the current version, so the document should be updated as well. Moreover, `\$` is much clearer and has much better readability than `${'$'}` mentioned in the current document.

### DIFF
--- a/docs/topics/strings.md
+++ b/docs/topics/strings.md
@@ -123,12 +123,11 @@ fun main() {
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 You can use templates both in raw and escaped strings.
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @yuantj,
In your suggestion you've changed the string from a raw string to an escaped string, so now the  `\$` will work. If you try it with a raw string (delimited by `"""`) you'll find that this doesn't work. Hope this helps!

PR #3501, opened 2023-03-26
Update generics.md

wahab-owolabi:


sarahhaggarty:
Hi @wahab-owolabi
Thanks for your suggestion. I checked the code sample and it is correct as it is.
If we didn't have the compile-time error when assigning the reference `strs` to `objs` then when an integer is added to `obj` it would also be accessible via `strs` which then triggers a ClassCastException error. I hope this explanation helps! As a result, I'm closing this PR.

PR #3505, opened 2023-03-26
Remove kapt.use.jvm.ir=true

PetrakovichVictoria:
[KT-56314](https://youtrack.jetbrains.com/issue/KT-56314) [Docs][JVM] Enable JVM IR for KAPT stub generation by default

### DIFF
--- a/docs/topics/kapt.md
+++ b/docs/topics/kapt.md
@@ -70,14 +70,6 @@ Follow these steps:
    `kaptAndroidTest` and `kaptTest`. Note that `kaptAndroidTest` and `kaptTest` extends `kapt`, so you can just provide the
    `kapt` dependency and it will be available both for production sources and tests.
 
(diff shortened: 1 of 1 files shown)


docs/topics/kapt.md
```diff
@@ -70,14 +70,6 @@ Follow these steps:
    `kaptAndroidTest` and `kaptTest`. Note that `kaptAndroidTest` and `kaptTest` extends `kapt`, so you can just provide the
    `kapt` dependency and it will be available both for production sources and tests.
 
-4. To use the newest Kotlin features with kapt, for example, [repeatable annotations](annotations.md#repeatable-annotations),
```
PetrakovichVictoria, line 73: @udalov, WDYT, should we still mention separately that a user can use the newest Kotlin features with kapt, for example, repeatable annotations?

udalov: No, I don't think it's necessary

PR #3524, opened 2023-04-07
fix: gradle compilerOptions for .kts

PascalHonegger:
I believe this is more in-line with how Gradle tasks should be configured, also the existing example doesn't compile without modification (e.g. one could also add a `.get()` within line 150, but that seemed less ideal)


PascalHonegger:
As far as I can tell this change would still be relevant, anything blocking this @Tapchicoma ?

PetrakovichVictoria:
Hello folks. @PascalHonegger, thank you for your contribution. 

@Ziette7, there's no need to quote another opinion without any additional information – we see all the comments. If you agree with the initial contributor, you can put a reaction (emoji) under a post you like.

Unfortunately, we were a little bit busy and indeed forgot about this PR, sorry for that. To not forget about PRs, we enabled this Stale-review mechanism that today reminded us about this PR. We'll return to this PR as soon as we have time. We'll try to do this ASAP.

docs/topics/opt-in-requirements.md
```diff
@@ -145,9 +145,10 @@ If you build your module with Gradle, you can add arguments like this:
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 // ...
 
-val compileKotlin = tasks.named<KotlinCompilationTask<*>>("compileKotlin")
+tasks.named<KotlinCompilationTask<*>>("compileKotlin").configure {
+    compileKotlin.compilerOptions.freeCompilerArgs.add("-opt-in=org.mylibrary.OptInAnnotation")
```
Tapchicoma, line 149: Hm, in 1.9.0 release there is a separate option for `optIn`:
`compilerOptions.optIn.add("..")`

@PetrakovichVictoria should we update this PR and wait for `1.9.0` final release or merge it as is and then, after release, update? :thinking:

PetrakovichVictoria:
The problem with the proposed changes is that they are not compilable too:
Screenshot 2023-06-06 at 17 13 05

But ok, just `compileKotlin` should be removed.

Also, we have usage of the same construction in two other places.

For me, it's easier to merge @PascalHonegger's change to my branch and apply the needed change to two more places there.

docs/topics/opt-in-requirements.md
PetrakovichVictoria, line 149:
```suggestion
    compilerOptions.freeCompilerArgs.add("-opt-in=org.mylibrary.OptInAnnotation")
```

PetrakovichVictoria (approved):

PR #3374, opened 2023-01-12
feat: the Dokka docs

IgnatBeresnev:

PR #3526, opened 2023-04-07
feature: add api-guidelines docs

PetrakovichVictoria:
Probably, this PR should be used instead of #3525
Done by example from #3374

### DIFF
--- a/docs/kr.tree
+++ b/docs/kr.tree
@@ -186,6 +186,9 @@
                 <toc-element toc-title="Collections" id="java-to-kotlin-collections-guide.md"/>
                 <toc-element toc-title="Nullability" id="java-to-kotlin-nullability-guide.md"/>
             </toc-element>
+            <toc-element toc-title="Library creators' guidelines">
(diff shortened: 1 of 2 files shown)


docs/kr.tree
```diff
@@ -186,6 +186,9 @@
                 <toc-element toc-title="Collections" id="java-to-kotlin-collections-guide.md"/>
                 <toc-element toc-title="Nullability" id="java-to-kotlin-nullability-guide.md"/>
             </toc-element>
+            <toc-element toc-title="Library creators' guidelines">
+                <include from="kl" target="api-guidelines" origin="api-guidelines"/>
```
PetrakovichVictoria, line 190: IDK what "kl" means)

IgnatBeresnev (approved):
:+1: :+1:

docs/kr.tree
koshachy: It could be renamed to have another prefix

koshachy (approved):

PR #3529, opened 2023-04-11
Update multiplatform-mobile-plugin-releases.md

terrakok:


koshachy (approved):

danil-pavlov:
I think the version should be updated in the Components stability table as well

koshachy:
Why? Did we change the status from Beta to Stable?

PR #3539, opened 2023-04-18
Update kotlinx.serialization api reference

nikpachoo:


IgnatBeresnev:
We've discussed this already, but cc @sandwwraith just in case - the API reference for serialization will be built from master once it gets merged

PR #3543, opened 2023-04-21
Update user-groups.yml

lailaaaed03:
Adding Morocco Kotlin User Group to the list with its informations !


data/user-groups.yml
```diff
@@ -1250,3 +1250,9 @@
     position:
       lat: 9.2034963
       lng: 12.4953899
+ - name: Morocco KUG
```
nikpachoo, line 1253: Thanks for the contribution!
Slight inaccuracy in the format:
```suggestion
  - name: Morocco KUG
```

nikpachoo (changes requested):

PR #3552, opened 2023-04-27
Fix misleading idiom if-not-null-else shorthand.

sdoerner:
The return statement in the version before this change will not return the value to the run statement but to the enclosing function.

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -159,11 +159,13 @@ println(files?.size) // size is printed if files is not null
 ```kotlin
 val files = File("Test").listFiles()
 
+// For easy fallback values:
 println(files?.size ?: "empty") // if files is null, this prints "empty"
 
-// To calculate the fallback value in a code block, use `run`
+// To calculate a more complicated fallback value in a code block, use `run`
 val filesSize = files?.size ?: run { 
-    return someSize 
+    val someSize = getSomeSize()
+    someSize * 2
 }
 println(filesSize)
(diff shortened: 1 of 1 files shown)


sdoerner:
Hi,
@sarahhaggarty would you be able to share any progress or ETA on the review for this change?

Thank you
Sebastian

docs/topics/idioms.md
```diff
@@ -159,11 +159,13 @@ println(files?.size) // size is printed if files is not null
 ```kotlin
 val files = File("Test").listFiles()
 
+// For easy fallback values:
```
SebastianAigner, line 162: I don't think "easy" is the right word to use. I'd rather we use "simple" here.

SebastianAigner, line 162:
```suggestion
// For simple fallback values:
```

SebastianAigner (changes requested):
Minor wording, apart from that, this is indeed much clearer than what we had so far.

docs/topics/idioms.md
sdoerner: Accepted your suggestion.

SebastianAigner (approved):

SebastianAigner:
Thanks a bunch for your contribution! 🎉

sdoerner:
Just to clarify the process for closing this out:
I don't seem to have permissions to merge this (despite "approval not required"). So I assume you will do that?

Thanks

PR #3557, opened 2023-04-29
Update server-overview.md

AndriiMaliuta:
Added Quarkus to frameworks list


koshachy (approved):
Thanks for your contribution!
I've slightly modified the list of frameworks and added a note.

PR #3559, opened 2023-05-01
update: update Kotlin/JS docs

katia-energizer:
Updated Kotlin/JS docs as agreed with Egor Tolstoy and Nastya Kapanina.


docs/topics/js/js-project-setup.md
```diff
@@ -1,4 +1,4 @@
-[//]: # (title: Set up a Kotlin/JS project)
+[//]: # (title: Understand a Kotlin/JS project)
```
koshachy, line 1: This page isn't about Understanding the project.
It shows how to set up JS-related Gradle props. I'd like you to rollback the changes (or start a discussion on why it should be named "Understand").

koshachy (changes requested):

docs/topics/js/js-project-setup.md
katia-energizer: We can keep `Set up`.

koshachy (approved):
Nice job.

Don't forget to delete pics that are used in tutorials (multiplatform-full-stack-app and multiplatform-library).

docs/topics/js/js-project-setup.md
katia-energizer: Done ✅

danil-pavlov (approved):

PR #3565, opened 2023-05-05
KTL-1094 kotlinlang: there is not Kotlin version is Docs header

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1094/kotlinlang-there-is-not-Kotlin-version-is-Docs-header


koshachy:
Thanks for fixing the issue!

PR #3577, opened 2023-05-16
chore: add a note about Kotlin and Java source files

koshachy:
[KT-58565](https://youtrack.jetbrains.com/issue/KT-58565) Write a notice in Kotlin docs that java files shouldn't be in src/main/kotlin

### DIFF
--- a/docs/topics/gradle/gradle-configure-project.md
+++ b/docs/topics/gradle/gradle-configure-project.md
@@ -85,7 +85,9 @@ The `version` should be literal in this block, and it cannot be applied from ano
 
 ### Kotlin and Java sources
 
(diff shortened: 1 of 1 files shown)


docs/topics/gradle/gradle-configure-project.md
```diff
@@ -95,6 +97,12 @@ project
             - java
 ```
 
+> Do not store Java `.java` and Kotlin `.kt` files in the same `src/*/kotlin` directory, as the `.java` files will not be compiled.
```
Tapchicoma, line 100:
```suggestion
> Do not store Java `.java`` files in the `src/*/kotlin` directory, as the `.java` files will not be compiled.
```

Tapchicoma (approved):

PR #3585, opened 2023-05-18
Reflection: Fix property reference type names

OliverO2:


OliverO2:
If the fix is not as obvious as it appeared to me: Is there anything I can do in order to help?

sarahhaggarty:
Hi @OliverO2! Sorry for the delayed response. What you have proposed is more precise however the current text is still correct as `KProperty0` is a subtype of `KProperty`. If we include `KMutableProperty0<Int>` then for completeness we should also include `KMutableProperty1<Int>` and `KMutableProperty2<Int>` but this is a bit overkill. So after some consideration, I'm going to leave the text as it is and close this PR. But thanks for your suggestion and please consider contributing to Kotlin Docs again!

OliverO2:
Hi @sarahhaggarty, thanks for explaining your considerations. Unfortunately, I disagree, having been misled more than once by the wordings in question:

Assume I wanted to pass `::x` as a parameter to another method by following the doc.

First, it says:
> The expression ::x evaluates to a KProperty<Int> type property object.

Then, after setting the stage for a `KProperty<Int>` type:
> You can read its value using get()

This not entirely incorrect, as it applies to `::x`, but misleading as it does not apply to `KProperty<Int>`: Trying to achieve the above, I'd now declare a parameter of type `KProperty<Int>`. Then I'll get stuck as there is no method `get()` in `KProperty<Int>`. (However, there's a `getter` property which would further lead me down the wrong track. It always took me some time to discover the correct subclass required to continue.)

For `KMutableProperty<Int>`, the documentation is plainly wrong, explicitly saying

> the KMutableProperty<Int> type which has a set() method

It does not.

Please reconsider.

sarahhaggarty:
Hi @OliverO2, thanks for taking the time to share this additional information with me! Now I can see the trap that our current text creates. I will re-open the PR.

sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @OliverO2!

PR #3593, opened 2023-05-25
Clarified `this` reference nullability in case extension receiver is null

Mikhail2048:


Mikhail2048:
Hello @sarahhaggarty! 
When do you plan to review this PR?)

Mikhail2048:
Any progress erviewing this PR?) @sarahhaggarty

sarahhaggarty:
Hi @Mikhail2048, sorry for the delayed response. I just got back from a long vacation. I'll be reviewing this shortly. Thanks for your patience!

docs/topics/extensions.md
```diff
@@ -115,14 +115,15 @@ fun main() {
 ## Nullable receiver
 
 Note that extensions can be defined with a nullable receiver type. These extensions can be called on an object variable
-even if its value is null, and they can check for `this == null` inside the body.
+even if it's value is null. If the receiver is `null`, then reference to `this` inside the extension function will be `null`
+as well. So you will be forced to check for `this == null` inside the body.
```
sarahhaggarty, line 119:
```suggestion
even if its value is null. Since the receiver type can be `null`, then `this` can also be `null`. So when defining an extension with a nullable receiver type, it's a good idea to perform a `this == null` check inside the function body to avoid compiler errors. 
```

docs/topics/extensions.md
```diff
@@ -115,14 +115,15 @@ fun main() {
...
 
-This way, you can call `toString()` in Kotlin without checking for `null`, as the check happens inside the extension function:
+Because of nullable receiver, you can call `toString()` in Kotlin without checking for `null`, as the check happens inside the extension function:
 
 ```kotlin
 fun Any?.toString(): String {
-    if (this == null) return "null"
-    // after the null check, 'this' is autocast to a non-null type, so the toString() below
+    if (this == null) return "null" // This extension is defined on nullable Any, so `this` reference might be null
```
sarahhaggarty, line 125:
```suggestion
    if (this == null) return "null"
```

Mikhail2048: Can you please explain why you remove this comment?) I do not fully understand why we cannot be very explicit here. In my opinion, we should be very clear on why we are doing this null check. It is generally better to over communicate than under communicate.

docs/topics/extensions.md
```diff
@@ -115,14 +115,15 @@ fun main() {
 ## Nullable receiver
 
 Note that extensions can be defined with a nullable receiver type. These extensions can be called on an object variable
-even if its value is null, and they can check for `this == null` inside the body.
+even if it's value is null. If the receiver is `null`, then reference to `this` inside the extension function will be `null`
+as well. So you will be forced to check for `this == null` inside the body.
 
-This way, you can call `toString()` in Kotlin without checking for `null`, as the check happens inside the extension function:
+Because of nullable receiver, you can call `toString()` in Kotlin without checking for `null`, as the check happens inside the extension function:
```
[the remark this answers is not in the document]
Mikhail2048: I think it generally ok, we can leave it this way, if you want :)

docs/topics/extensions.md
Mikhail2048: I really appreciate your suggestion, but I think we can and should rethink it a bit :) 

Sentence `Since the receiver type can be null, then this can also be null` is a bit obscure. I have specifically written, that if the receiver is `null`, that `this` **_will be null_**. Suggested change is:

> can also be null

So the only difference between these two formulations is that initial version have explicitly mentioned when `this` will be `null`. IMO, it is generally better to be explicit here :)

If specify when `this` might be `null` clearly, the last suggested sentence would lose its meaning. The programmer will understand in what scenarios `this` reference will be `null` and will react accordingly. So, IMO, checking for null is not a good idea, but the necessity, which occurs under very specific circumstances :)

Mikhail2048: I really appreciate your suggestion, but I think we can and should rethink it a bit :) 

Sentence `Since the receiver type can be null, then this can also be null` is a bit obscure. I have specifically written, that if the receiver is `null`, that `this` **_will be null_**. Suggested change is:

> can also be null

So the only difference between these two formulations is that initial version have explicitly mentioned when `this` will be `null`. IMO, it is generally better to be explicit here :)

If specify when `this` might be `null` clearly, the last suggested sentence would lose its meaning. The programmer will understand in what scenarios `this` reference will be `null` and will react accordingly. So, IMO, checking for null is not a 'good idea', but the necessity, which occurs under very specific circumstances :)

Mikhail2048 (commented):
I appreciate our work, thanks. I have left my comments to your comments as well. 
In general, I think it is better to clearly and explicitly state the conditions in the documentation. 
I am the Kotlin developer as well, and the doc on kotlin web site is the source of truth in all aspects. It is better, IMO, for doc to be explicit, clear and maybe sometimes to over communicate, overstate, overemphasize things, because others rely on this doc heavily. It is the bible, that gets quotes on SO, Github e.t.c.

docs/topics/extensions.md
sarahhaggarty: Hi @Mikhail2048, thanks for sharing your thoughts. I've updated the suggestion to make it more explicit. The main motivation for change here from my side is to use present tense so it is in line with our style guide. 

I've also changed "it's a good idea" to the stronger "we recommend".

docs/topics/extensions.md
sarahhaggarty: I removed this comment to avoid duplication of content as the need for the null check is explained in the first paragraph. Also, the purpose of the code block is to demonstrate that `toString()` already contains a null check.

Mikhail2048:
Hey @sarahhaggarty, now I think it is far more explicit and straightforward :) 
I will apply your changes, thanks for suggestions

sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution and your patience @Mikhail2048 !

PR #3597, opened 2023-05-31
explain that dropped functions will be gone

robstoll:


robstoll:
what do I need to do that this gets reviewed?

robstoll:
@ilya-g done: https://youtrack.jetbrains.com/issue/KT-60405/Kotlin-does-not-provide-polyfils-for-dropped-deprecated-things-in-case-someone-uses-old-apiVersion

PR #3600, opened 2023-06-02
Swap description of SourceDirectory layout explanation

Jon889:
> They create a confusing `SourceDirectories` layout, as `src/androidTest/java` has a `UnitTest` and `src/androidTest/kotlin` has an `InstrumentedTest`.

I think the description (above) of the confusing SourceDirectories layouts in what's new in Kotlin 1.8 is the wrong way round. I'm not entirely sure though as I'm primarily an IOS dev, and have only been using KMP, never Android.

But from what the tables that are below that line say, that line doesn't make sense to me, as `src/androidTest/java` would be for the AndroidSourceSet `androidTest` (corresponding to the KotlinSourceSet `androidAndroidTest`) which is instrumentation tests. And `src/androidTest/kotlin` is the KotlinSourceSet which corresponds to the AndroidSourceSet `test` which are unit tests.

Apologies if I've misunderstood!


sellmair (approved):

danil-pavlov (approved):
Good catch, thank you!

PR #3605, opened 2023-06-05
Update jvm-spring-boot-add-db-support.md

sirajalam049:
Error code fixed.


antonarhipov (commented):
This change opens the possibility for SQL injection. No SQL statements should be executed as a concatenated string. Prepared statements are the way to go.

PR #3609, opened 2023-06-07
feat: dependencies between pods

danil-pavlov:


docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,29 @@ kotlin {
         }
     }
 }
+```
+
+### Dependencies between Pods
```
daugeldauge, line 290: I think it's better not to use the plain "dependency" term. Because:
1) That's not really correct technically because dependency is added not between pods themselves but between their C-interops
2) That sounds quite simple, but the case for which it is needed is complicated. I think users can misinterpret that
Due to these reasons, the functions in the DSL are not called "dependsOn" and "dependencies"

I'd suggest smth like "Share Kotlin C-interop between dependant pods". But I don't know if we're ok of using C-interop term because it's quite technical.

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,29 @@ kotlin {
...
+
+You can use the cinterop binding generated for another Pod when building a binding for the new Pod.
```
daugeldauge, line 292: I think the case for which it's required should be described here. Because we don't want users to use this DSL "just in case".

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,29 @@ kotlin {
...
+The dependent Pod should be declared before setting up the dependency. The Kotlin CocoaPods Gradle plugin supports two ways
+of adding dependencies between Pods:
+
+* `useInteropBindingFrom()` function specifies the name of the existing Pod that is used as dependency.
+* `interopBindingDependencies` contains a list of all dependencies to other Pods.
```
daugeldauge, line 297: I guess we can omit `interopBindingDependencies` from the docs. It was added just for consistency. Most of the users should be fine with `useInteropBindingFrom`

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,29 @@ kotlin {
...
+
+```kotlin
+pod("WebImage") {
+    version = "~> 1.0.0"
```
daugeldauge, line 301: I'd suggest to stop promoting `~>` versions. We support them, but the Gradle users are far more used to the fixed versions.

docs/topics/native/native-cocoapods-libraries.md
danil-pavlov: I think it's OK. We're using "cinterop" a lot in the context of ios dependencies

docs/topics/native/native-cocoapods-libraries.md
danil-pavlov: Thank you! Please check the new intro

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,25 @@ kotlin {
         }
     }
 }
+```
+
+### Share Kotlin cinterop between dependent Pods
+
+If you're adding multiple dependencies on Pods using the `pod()` function, you might encounter issues in Xcode
```
daugeldauge, line 292: It doesn't matter whether a build is from Xcode or not. Most probably, users will encounter it inside our IDE (AS/Intellij) first.

I would add some specifics here, like
```
// pod("WebImage") cinterop
fun image(): WebImage

// pod("Info") cinterop
fun printImageInfo(image: WebImage)

// user code
printImageInfo(image())
```

If a user doesn't set-up the correct dependencies between interops, then the code would be red because `WebImage` type would come from the different interops (and thus packages). If interop-dependencies are set up correctly, then the code would be green.

danil-pavlov: Thank you! Removed mentions of Xcode and replaced the code sample

daugeldauge (approved):

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
         }
     }
 }
-```
\ No newline at end of file
+```
+
+### Sharing Kotlin cinterop between dependent Pods
```
koshachy, line 290: Maybe just "Share" and not "sharing"?

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+
+If you're adding multiple dependencies on Pods using the `pod()` function, you might encounter issues in
```
koshachy, line 292: Do we need present continuous here?
If you add ...

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+case there are dependencies between APIs of your Pods.
```
koshachy, line 293: in case of?
or better to add another if?
or

If you add multiple dependencies on Pods using the `pod()` function, you might face issues when there are dependencies between the APIs of your Pods.

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+
+To make the code compile in this case, use the `useInteropBindingFrom()` function.
```
koshachy, line 295: in such cases?

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+It utilizes the cinterop binding generated for another Pod when building a binding for the new Pod.
```
koshachy, line 296: while building?

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+
+The dependent Pod should be declared before setting up the dependency:
```
koshachy, line 298: It's important to declare the dependent Pod before setting up the dependency:

Does this work?

koshachy: You should declare the dependent Pod before setting up the dependency:

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+fun loadImage(): WebImage
+
+// The cinterop of pod("Info"):
+fun printImageInfo(image: WebImage)
+
+// Your code:
+printImageInfo(loadImage())
+```
+
+If you didn't set up the correct dependencies between cinterops in this case,
```
koshachy, line 311: If you haven't configured the correct dependencies...

IDK, maybe it sounds a bit better?

docs/topics/native/native-cocoapods-libraries.md
```diff
@@ -285,4 +285,28 @@ kotlin {
...
+the code would be invalid because the `WebImage` type would come from different cinterops and, thus, packages.
```
koshachy, line 312: ...would be sourced from different cinterop files and, consequently, different packages.

koshachy (changes requested):
I'd recommend rewriting the "Sharing Kotlin cinterop between dependent Pods" section.
See my comments. Feel free to rephrase.

koshachy (approved):
LGTM!

PR #3618, opened 2023-06-13
update: multiplatform compatibility table layout

danil-pavlov:


koshachy (approved):
Thanks, now it looks better!

PR #3630, opened 2023-06-21
feat: disable sources publication

danil-pavlov:


docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -156,6 +156,48 @@ kotlin {
...
+you can configure and disable sources publication with the `withSourcesJar()` API:
+
+* To disable sources publication for all the targets:
+
+  ```kotlin
+  kotlin {
+      withSourcesJar(publish = false)
+  
+      jvm()
+      ios()
```
atyrin, line 175: Maybe for future us, we should no longer use shortcuts for targets like `ios()`. We are going to deprecate them in 1.9.20. But it up to you.

atyrin (approved):

docs/topics/multiplatform/multiplatform-publish-lib.md
danil-pavlov: Thank you! I agree it makes sense not to use them in new texts, plus we can align these three examples

koshachy (approved):

PR #3633, opened 2023-06-22
feat: object deinit on the main thread

danil-pavlov:


docs/topics/native/native-ios-integration.md
```diff
@@ -48,9 +48,14 @@ The resulting output:
 ```text
 init on <_NSMainThread: 0x600003bc0000>{number = 1, name = main}
 shared.SwiftExample
-deinit on <NSThread: 0x600003b9b900>{number = 7, name = (null)}
+deinit on <_NSMainThread: 0x600003bc0000>{number = 1, name = main}
 ```
```
koshachy, line 53: The list should show the two different ways of doing it, shouldn't it?
Now it seems that the second way is just an addition to the first one...

docs/topics/native/native-ios-integration.md
```diff
@@ -48,9 +48,14 @@ The resulting output:
...
 
+* If Swift/Objective-C objects are passed to Kotlin on a thread other than main or if the main dispatch queue isn't processed,
```
koshachy, line 54: Maybe this bullet should be a list?
Something like:

Kotlin calls deinit on the S/O objects on a specific GC thread, if:
* S/O objects are passed to Kotlin on a thread other than the main
* Main dispatch queue isn't processed (by the GC?)

If you want to call deinit on .... explicitly, set ` ` in your `gradle.properties`. It calls deinit, even if these objects are passed to Kotlin on the main thread.

docs/topics/native/native-ios-integration.md
```diff
@@ -48,9 +48,14 @@ The resulting output:
...
+  deinit on the Swift/Objective-C objects is called on a special GC thread.
```
koshachy, line 55: Also, WDYT if we use "deinitialization" word instead of "deinit"?

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/deinitialization/

koshachy (changes requested):
Please, take a look at the comments.

docs/topics/native/native-ios-integration.md
danil-pavlov: No, this is actually a list of exceptions from the main rule about the main queue, like minor notes. so the second item is indeed an addition

koshachy (approved):
LGTM!

PR #3588, opened 2023-05-23
feat: add rangeUntil operator

sarahhaggarty:
This PR updates our Ranges page with the `..<` operator. It also adds this operator the list of keywords and operators.

PR #3637, opened 2023-06-26
KT-53643: Update coding style conventions wrt rangeUntil operator

zarechenskiy:
I'm slightly hesitant about including part "Prior to Kotlin 1.9, it was recommended to use the `until` function". In the end, decided not to add it as new projects will use only `..<` and old projects will continue to use `until` as they used to.


sarahhaggarty:
I'm including these changes as part of #3588 so I'll close this PR.

PR #3639, opened 2023-06-26
Rephrase definition of single-expression function

vegarsti:
The current definition says "When a function returns a single expression", which I don't think is correct: As long as the return type is not `Void`, a function always returns a single expression. And it sounds as if the function returns all the expressions inside it.

This is an opinionated nitpick, though, so feel free to close this!


docs/topics/functions.md
```diff
@@ -177,7 +177,7 @@ fun printHello(name: String?) { ... }
 
 ### Single-expression functions
 
-When a function returns a single expression, the curly braces can be omitted and the body is specified after a `=` symbol:
+When a function contains a single `return` expression, the curly braces can be omitted and the body is specified after a `=` symbol:
```
sarahhaggarty, line 180:
```suggestion
When the function body consists of a single expression, the curly braces can be omitted and the body specified after an `=` symbol:
```

sarahhaggarty (changes requested):
I agree that using "returns" is confusing. What do you think about this version that I suggest instead?

sarahhaggarty (approved):

sarahhaggarty:
Thanks for sharing your thoughts with us @vegarsti !

docs/topics/functions.md
Cesar556: - []()@_********_

PR #3644, opened 2023-06-28
feat: secondary constructors in inline classes

danil-pavlov:


docs/topics/inline-classes.md
```diff
@@ -40,29 +36,39 @@ usages (similar to how content of [inline functions](inline-functions.md) is inl
 ## Members
 
 Inline classes support some functionality of regular classes. In particular, they are allowed to declare properties and 
-functions, and have the `init` block:
+functions, have the `init` block and [secondary constructors](classes.md#secondary-constructors):
 
 ```kotlin
 @JvmInline
-value class Name(val s: String) {
+value class Person(private val firstName: String) {
```
zarechenskiy, line 43: Probably, `firstName` -> `fullName`

zarechenskiy (approved):

docs/topics/inline-classes.md
```diff
@@ -21,10 +21,6 @@ To declare an inline class for the JVM backend, use the `value` modifier along w
 value class Password(private val s: String)
 ```
 
-> The `inline` modifier for inline classes is deprecated.
-> 
-{type="warning"}
-
 An inline class must have a single property initialized in the primary constructor. At runtime, instances of the inline
```
zarechenskiy, line 24: Also, it looks like we have to call them "inline values class"

docs/topics/inline-classes.md
```diff
@@ -40,29 +36,39 @@ usages (similar to how content of [inline functions](inline-functions.md) is inl
 ## Members
 
 Inline classes support some functionality of regular classes. In particular, they are allowed to declare properties and 
-functions, and have the `init` block:
+functions, have the `init` block and [secondary constructors](classes.md#secondary-constructors):
```
sarahhaggarty, line 39:
```suggestion
functions, have an `init` block and [secondary constructors](classes.md#secondary-constructors):
```

sarahhaggarty (approved):
Looks great! Just one tiny suggestion.

PR #3645, opened 2023-06-29
changed some .md files ( doc files )

enyoos:
When using the MutableList ( with mutableListOf() method ), the author used a val, which is redudant and counter-intuitive.
Mutable list are mutable, but using the val keyword makes it immutable, which defeat its purpose.

### DIFF
--- a/docs/topics/collection-write.md
+++ b/docs/topics/collection-write.md
@@ -12,7 +12,7 @@ To add a single element to a list or a set, use the [`add()`](https://kotlinlang
 
 fun main() {
 //sampleStart
-    val numbers = mutableListOf(1, 2, 3, 4)
+    var numbers = mutableListOf(1, 2, 3, 4)
     numbers.add(5)
     println(numbers)
 //sampleEnd
@@ -32,7 +32,7 @@ Other elements of the argument collection will follow it, shifting the receiver 
 
 fun main() {
 //sampleStart
-    val numbers = mutableListOf(1, 2, 5, 6)
+    var numbers = mutableListOf(1, 2, 5, 6)
     numbers.addAll(arrayOf(7, 8))
     println(numbers)
     numbers.addAll(2, setOf(3, 4))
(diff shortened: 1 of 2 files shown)


sarahhaggarty:
Hi @enyoos,
Thanks for sharing your suggestion with us! Mutable collections don't have to be assigned to `var` for write operations to succeed. There is a difference between **the collection** being mutable and **the reference** to the mutable collection being mutable. As per our [coding conventions](https://kotlinlang.org/docs/coding-conventions.html#immutability), we always recommend to declare immutable local variables as much as possible. This makes your code safer as it protects against unintentionally modifying a reference, which could then affect other parts of your code that were relying on that reference.

Checking our documentation, I think we can explain this better so I've updated the text in our Collections overview as part of another PR I am working on (#3292). For now, I will close this PR but please let me know if my explanation isn't clear or if you have any questions! Thanks again for sharing this with us.

enyoos:
tyvm for the feedback, the explanation is very clear.
________________________________
De : Sarah Haggarty ***@***.***>
Envoyé : lundi 17 juillet 2023 08:29
À : JetBrains/kotlin-web-site ***@***.***>
Cc : envyoos ***@***.***>; Mention ***@***.***>
Objet : Re: [JetBrains/kotlin-web-site] changed some .md files ( doc files ) (PR #3645)


Hi @enyoos<https://github.com/enyoos>,
Thanks for sharing your suggestion with us! Mutable collections don't have to be assigned to var for write operations to succeed. There is a difference between the collection being mutable and the reference to the mutable collection being mutable. As per our coding conventions<https://kotlinlang.org/docs/coding-conventions.html#immutability>, we always recommend to declare immutable local variables as much as possible. This makes your code safer as it protects against unintentionally modifying a reference, which could then affect other parts of your code that were relying on that reference.

Checking our documentation, I think we can explain this better so I've updated the text in our Collections overview as part of another PR I am working on (#3292<https://github.com/JetBrains/kotlin-web-site/pull/3292>). For now, I will close this PR but please let me know if my explanation isn't clear or if you have any questions! Thanks again for sharing this with us.

PR #3650, opened 2023-07-03
ktl-1115 feat: add next.js front page

krutilov:


nikpachoo (changes requested):
Let's change the target branch. It shouldn’t be the master branch.

PR #3654, opened 2023-07-05
Add SealedX to list of KSP supported libraries

skydoves:
Add [SealedX](https://github.com/skydoves/sealedx) to list of KSP-supported libraries.


koshachy:
Thanks for your contribution!

PR #3680, opened 2023-07-14
Fix basePackage value for "Build a full-stack web app with Kotlin Multiplatform" tutorial

kuroski:
I am currently going through the tutorial for ["Build a full-stack web app with Kotlin Multiplatform"](https://kotlinlang.org/docs/multiplatform-full-stack-app.html) but found an issue when I reached the ["Serve HTML and JavaScript files from Ktor"](https://kotlinlang.org/docs/multiplatform-full-stack-app.html#serve-html-and-javascript-files-from-ktor) part.

In my case, I was facing a 404 error when running the app for `shoppinglist.js` file.

image

After some trials and errors, I finally found that the issue was that the `basePackage` property from `Ktor` should actually be empty

```gradle
staticResources("/", "")
```

image

### DIFF
--- a/docs/topics/multiplatform/multiplatform-full-stack-app.md
+++ b/docs/topics/multiplatform/multiplatform-full-stack-app.md
@@ -480,7 +480,7 @@ root directory.
             ContentType.Text.Html
         )
     }
-    staticResources("/", "static")
+    staticResources("/", "")
     route(ShoppingListItem.path) {
         // ...
     }


danil-pavlov:
Daniel, thank you for the report, and I'm sorry it took so long. We discuused this issue with developers and decided to take a different approach to solve this issue. Please see #3764

PR #3690, opened 2023-07-19
Deprecate declaring multiple targets of the same type

antohaby:
Hi!
I want to add deprecation information about declaring multiple targets in Kotlin Multiplatform Projects.

YT issue about deprecation: https://youtrack.jetbrains.com/issue/KT-59316

I also would like to ask your advice on how to structure this information best and consistently.

### DIFF
--- a/docs/topics/multiplatform/multiplatform-compatibility-guide.md
+++ b/docs/topics/multiplatform/multiplatform-compatibility-guide.md
@@ -496,9 +496,7 @@ In Kotlin 1.9.0, a deprecation warning is introduced when the `android` name is 
 
 **What's changed?**
 
-Declaring multiple targets of the same kind in one project is not recommended and will be forbidden in future releases.
-There are multiple ways of achieving same results with other Kotlin Gradle Plugin mechanisms. As part of Kotlin
-Multiplatform stabilization we are trying to emphasize the essential and reduce ambiguous APIs. 
+We discourage declaring multiple targets of the same kind in a single Gradle project.
 
 Example:
 ```kotlin
--- a/docs/topics/multiplatform/multiplatform-compatibility-guide.md
+++ b/docs/topics/multiplatform/multiplatform-compatibility-guide.md
@@ -502,54 +502,186 @@ Example:
 ```kotlin
 kotlin {
     jvm()
-    jvm("jvm6") // is not recommended and will produce deprecation warning
+    jvm("jvmKtor")
+    jvm("jvmOkio") // is not recommended and will produce deprecation warning
 }
 ```
 
 **What's the best practice now?**
 
-There are two ways of how to avoid multiple target declaration: different compilations and different gradle projects.
+There are two ways of how to avoid multiple target declaration:
 
-If you need to compile some sources with different parameters such as dependencies or compiler options for example
-functional tests, different JDK, debug or release artifacts. Then instead of declaring multiple targets you can
-create multiple compilations in one target.
+1. Creating different compilations
 
-Functional test build configuration sample 
-```kotlin
-kotlin {
-    jvm {
-        // Creates new compilation and 'jvmFunctionalTest' source set associated with it
-        val functionalTest by compilations.creating
+    If you need to compile some sources with different parameters such as dependencies or compiler options for example
+    for functional tests. Then instead of declaring multiple targets you can create multiple compilations in one target.
+ 
+   <tabs group="build-script">
+   <tab title="Before" group-key="before">
 
-        // functionalTest should see public and internal declarations from 'main' compilation now.
-        // it is possible to associate with 'test' compilation as well
-        functionalTest.associateWith(compilations.getByName("main"))
-    }
-}
-```
+    ```kotlin
+   // Project ':myLibrary:integrationTests'
+   kotlin {
+      jvm("jvmKtor") // !!!
+      jvm("jvmOkio") // !!!
+   
+       sourceSets {
+           // 'commonTest' source set contains shared integration tests of 'myLibrary' between ktor and okio 
+           val commonTest by getting {
+               dependencies {
+                   implementation(projects.myLibrary) 
+               }
+           }
+           // 'jvmKtorTest' source set contains ktor-specific integration tests
+           val jvmKtorTest by getting {
+               dependencies {
+                   implementation(libs.ktor)
+                   implementation(projects.myLibrary.ktor)
+               }
+           }
+   
+           // 'jvmOkioTest' source set contains okio-specific integration tests
+           val jvmOkioTest by getting {
+               dependencies {
+                   implementation(libs.okio)
+                   implementation(projects.myLibrary.okio)
+               }
+           }
+       }
+   }
+    ```
 
-Different projects can be configured like this:
-```kotlin
-// :lib-jvm
-kotlin {
-    jvm()
-}
-
-// :lib-jvm6
-kotlin {
-    jvm {
-        // JVM specific configurations
-    }
-    
-    sourceSets {
-        jvmMain {
+   </tab>
+   <tab title="Now" group-key="now">
+   
+    ```kotlin
+   // Project ':myLibrary:integrationTests'
+   kotlin {
+      jvm {
+         val ktorIntegrationTest by compilations.creating
+         val okioIntegrationTest by compilations.creating
+   
+         // Associate with test compilation to see tests from commonTest
+         ktorIntegrationTest.associateWith(compilations.getByName("test"))
+         okioIntegrationTest.associateWith(compilations.getByName("test"))
+   
+         // Create a Test Run to for each integration test
+         testRuns {
+            create("ktorIntegration") { setExecutionSourceFrom(ktorIntegrationTest) }
+            create("okioIntegration") { setExecutionSourceFrom(okioIntegrationTest) }
+         }
+      }
+   
+      sourceSets {
+         val commonTest by getting {
             dependencies {
-                api(project(":lib-jvm"))
+               implementation(kotlin("test"))
+               implementation(projects.myLibrary)
             }
-        }
-    }
+         }
+   
+         val jvmKtorIntegrationTest by getting {
+            dependencies {
+               implementation(kotlin("test"))
+               implementation(libs.ktor)
+               implementation(projects.myLibrary.ktor)
+            }
+         }
+   
+         val jvmOkioIntegrationTest by getting {
+            dependencies {
+               implementation(kotlin("test"))
+               implementation(libs.okio)
+               implementation(projects.myLibrary.okio)
+            }
+         }
+      }
+   }
+    ```
+
+   </tab>
+   </tabs>
+
+
+
+2. Extracting new target into a separate Gradle project
+
+   <tabs group="build-script">
+   <tab title="Before" group-key="before">
+
+   ```kotlin
+   // Project ':myUILibrary'
+   kotlin {
+       jvm("jvmCompose") // !!!
+       jvm("jvmSwing")   // !!!
+   
+       sourceSets {
+           val commonMain by getting
+           val commonJvmMain by sourceSets.creating {
+               dependsOn(commonMain)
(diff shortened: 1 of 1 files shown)


antohaby:
This is currently in a draft state. We might adjust it with more real use cases. 
I will search for some community projects where multiple targets are declared and I'll try to understand why they do that and how to fix it in the recommended way. If I find some interesting and repeating cases I will consider adding them to this doc.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
 **When do the changes take effect?**
 
 In Kotlin 1.9.0, a deprecation warning is introduced when the `android` name is used in Kotlin Multiplatform projects.
+
+<anchor name="declaring-multiple-targets"></anchor>
+**What's changed?**
```
dsavvinov, line 495: Heading is missing

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
...
+
+Declaring multiple targets of the same kind in one project is not recommended and will be forbidden in future releases.
+There are multiple ways of achieving same results with other Kotlin Gradle Plugin mechanisms. As part of Kotlin
```
dsavvinov, line 498: I'd drop those two sentences, we're expanding on this in details below anyways

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
...
+                api(project(":lib-jvm"))
+            }
+        }
+    }
+}
+```
+
+**When do the changes take effect?**
+
+In Kotlin 1.9.20, a deprecation warning is introduced when multiple targets of the same type is used in Kotlin Multiplatform projects.
```
dsavvinov, line 556: Normally, versions where next steps of deprecation cycle take palce are expected to be described here. E.g.:

- In Kotlin 1.9.20, deprecation warning is reported
- Kotlin 2.0, error will be reported in such cases, causing your build to fail

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
...
+        val functionalTest by compilations.creating
+
+        // functionalTest should see public and internal declarations from 'main' compilation now.
+        // it is possible to associate with 'test' compilation as well
+        functionalTest.associateWith(compilations.getByName("main"))
+    }
+}
+```
+
+Different projects can be configured like this:
```
dsavvinov, line 531: Currently it reads as continutiation of the previous idea/case, but it's in fact the second case (and second approach to addressing this deprecation). Let's separate them properly and add a bit of a structure along the way -- I'd suggest using numbered sections, like:

```
1. Creating custom compilations

...

2. Extracting new target into a separate Gradle project

...
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
...
+kotlin {
+    jvm {
+        // Creates new compilation and 'jvmFunctionalTest' source set associated with it
+        val functionalTest by compilations.creating
+
+        // functionalTest should see public and internal declarations from 'main' compilation now.
+        // it is possible to associate with 'test' compilation as well
+        functionalTest.associateWith(compilations.getByName("main"))
+    }
+}
```
dsavvinov, line 528: I'd suggest using tables with "before"/"after" because currently it's not clear for which multi-target configuration this code snippet is suggested.

Also, with separate target people might've had:
- separate compiler options
- separate dependencies
- separate sources

This is not the exhaustive list, but I think it covers most of the top-level requests (feel free to add something). Please show how to migrate them to the "separate compilation"-approach.

dsavvinov (changes requested):

docs/topics/multiplatform/multiplatform-compatibility-guide.md
antohaby: But I think something should be placed under "what's changed" section.
I'll try to rephrase it to something like:

> As part of Kotlin Multiplatform stabilization we are trying to emphasize the essential and reduce ambiguous APIs. Therefore we discourage declaring multiple targets of the same kind in a single Gradle project.

The first sentence can also be dropped as it is just a "soft" introduction to what exactly we changed. But it can blur the reader's eye.

dsavvinov: > The first sentence can also be dropped

Yes, I'd agree

antohaby: Updated

docs/topics/multiplatform/multiplatform-compatibility-guide.md
antohaby: Updated

docs/topics/multiplatform/multiplatform-compatibility-guide.md
antohaby: Separated and I made more detailed samples. Can you check them?

docs/topics/multiplatform/multiplatform-compatibility-guide.md
antohaby: Reworked with tables. Tried to add detailed samples. I'm open for more suggetsions

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,67 @@ projects.
...
+```
+
+Different projects can be configured like this:
+```kotlin
```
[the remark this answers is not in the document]
antohaby: I've made a sample with UI library where you can have 'compose' and 'swing'. With some shared code between them.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+    If you need to compile some sources with different parameters such as dependencies or compiler options for example
+    for functional tests. Then instead of declaring multiple targets you can create multiple compilations in one target.
+ 
+   <tabs group="build-script">
+   <tab title="Before" group-key="before">
+
+    ```kotlin
+   // Project ':myLibrary:integrationTests'
+   kotlin {
+      jvm("jvmKtor") // !!!
```
dsavvinov, line 525: I'd recommend to avoid using JVM-shared source sets without the specific the need. I.e., let's add `linuxX64` or something here.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+   
+    ```kotlin
+   // Project ':myLibrary:integrationTests'
+   kotlin {
+      jvm {
+         val ktorIntegrationTest by compilations.creating
+         val okioIntegrationTest by compilations.creating
+   
+         // Associate with test compilation to see tests from commonTest
+         ktorIntegrationTest.associateWith(compilations.getByName("test"))
```
dsavvinov, line 565: Is `associateWith` considered non-experimental for KMP? I thought it's not, and only `associateWith` in JVM is.

dsavvinov, line 565: Why is it associated with `test`? Does it even work in IDE? 

I think the direct counterpart for the previous snippet would be to:

- associate new compilations with `main`
- connect link their default source sets with `dependsOn` to the `commonTest`

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+2. Extracting new target into a separate Gradle project
+
+   <tabs group="build-script">
+   <tab title="Before" group-key="before">
+
+   ```kotlin
+   // Project ':myUILibrary'
+   kotlin {
+       jvm("jvmCompose") // !!!
+       jvm("jvmSwing")   // !!!
```
dsavvinov, line 616: Ditto about having `commonMain` as JVM-shared source set

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+   }
+   ```
+
+   </tab>
+   <tab title="Now" group-key="now">
+
+```kotlin
+// Project ':myUILibrary
+kotlin { 
+   jvm()
```
dsavvinov, line 648: I'd also avoid promoting single-target projects

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+               }
+           }
+       }
+   }
+   ```
+
+   </tab>
+   <tab title="Now" group-key="now">
+
+```kotlin
```
dsavvinov, line 645: I'm a bit conernced that this example is neither concise, nor complete. I think we have to make a firm decision: either we're providing complete examples, or we don't provide them at all and just assume that the users that had multiple same targets are advanced enough to handle this refactoring on their own.

I'm leaning towards the latter. Mostly because it's so hard to give a proper minimal migration example -- as it involves multiple files in the project, the only solution I see is an actual GitHub project with "migration commit", but IMO it's too much for this case (but we can expand it, if we see demand/questions)

dsavvinov (changes requested):
Thanks for providing examples, and sorry for suggesting to drop this work, but I'm really concerned that they are not "complete", but still quite verbose. Let's discuss the idea of dropping them completley?

docs/topics/multiplatform/multiplatform-compatibility-guide.md
martinbonnin: Agree with @dsavvinov here. Also, it would be nice to include a paragraph about when to choose separate compilations over separate Gradle projects. Adding a separate Gradle project is usually more involved. Unless there's something fundamental it unlocks, I would even not mention it.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+         val okioIntegrationTest by compilations.creating
+   
+         // Associate with test compilation to see tests from commonTest
+         ktorIntegrationTest.associateWith(compilations.getByName("test"))
+         okioIntegrationTest.associateWith(compilations.getByName("test"))
+   
+         // Create a Test Run to for each integration test
+         testRuns {
+            create("ktorIntegration") { setExecutionSourceFrom(ktorIntegrationTest) }
+            create("okioIntegration") { setExecutionSourceFrom(okioIntegrationTest) }
```
martinbonnin, line 571: Can I do this for native targets? The below doesn't work:

```kotlin
kotlin {
  macosArm64 {
    testRuns {
      // Unresolved reference
      setExecutionSourceFrom(compilation)
    }
  }
}
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +490,200 @@ projects.
...
+**What's changed?**
+
+We discourage declaring multiple targets of the same kind in a single Gradle project.
+
+Example:
+```kotlin
+kotlin {
+    jvm()
+    jvm("jvmKtor")
+    jvm("jvmOkio") // is not recommended and will produce deprecation warning
```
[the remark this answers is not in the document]
antohaby: right, we can remake it to just:

```
jvm()
jvm("jvmKtor") // is not recommended...
```

dsavvinov:
Changes in this request will be merged as part of the https://github.com/JetBrains/kotlin-web-site/pull/3789 . I'm closing this one.

PR #3692, opened 2023-07-19
update: js code and sample projects

danil-pavlov:


pahill:
Something I'm missing is where we add the base-64 npm module as a dependency?

pahill:
I'm also not seeing where the module is being used - it's not existing code.
I believe we also needed to make dependency updates?

danil-pavlov:
The thing is that this project wasn't created for this tutorial (and vice versa), they a bit different. I got confused because the sample project uses npm dependency for creating JS tests, but the tutorial doesn't have this part

I've added instructions on how to add an npm dependency to jsMain. The module itself is used to implement JS-specific base64 encoding in separate `Base64Npm.kt` file

pahill (approved):
LGTM! Thank you for making the changes.

PR #3710, opened 2023-07-26
Small update to Type checks and casts page

sarahhaggarty:
This PR updates the Type checks and casts page to make it clear earlier in the page that generics documentation is available on another page. Other changes have been added, like an introduction and other modifications, to tidy up the page a bit.

This PR resolves KT-57270.

### DIFF
--- a/docs/topics/typecasts.md
+++ b/docs/topics/typecasts.md
@@ -1,5 +1,12 @@
 [//]: # (title: Type checks and casts)
 
+In Kotlin, you can perform type checks to check the type of an object at runtime. Type casts convert objects to a 
+different type.
+
+> To learn specifically about **generics** type checks and casts, for example `List<T>`, `Map<K,V>`, see [Generics type checks and casts](generics.md#generics-type-checks-and-casts).
+>
+{type="tip"}
+
 ## is and !is operators
 
 Use the `is` operator or its negated form `!is` to perform a runtime check that identifies whether an object conforms to a given type:
@@ -9,7 +16,7 @@ if (obj is String) {
     print(obj.length)
 }
 
-if (obj !is String) { // same as !(obj is String)
+if (obj !is String) { // Same as !(obj is String)
     print("Not a String")
 } else {
     print(obj.length)
@@ -37,7 +44,7 @@ if (x !is String) return
 print(x.length) // x is automatically cast to String
 ```
 
-or if it is on the right-hand side of `&&` or `||` and the proper check (regular or negative) is on the left-hand side:
+Or if it is on the right-hand side of `&&` or `||` and the proper check (regular or negative) is on the left-hand side:
 
 ```kotlin
 // x is automatically cast to String on the right-hand side of `||`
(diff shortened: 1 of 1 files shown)


docs/topics/typecasts.md
```diff
@@ -60,17 +67,47 @@ when (x) {
...
+        <td>
+            Never, because the variable can be modified at any time by other code.
+        </td>
+    </tr>
+</table>
 
 ## "Unsafe" cast operator
 
-Usually, the cast operator throws an exception if the cast isn't possible. And so, it's called *unsafe*.
+Usually, the cast operator throws an exception if the cast isn't possible. Thus, it's called *unsafe*.
```
danil-pavlov, line 110:
```suggestion
Usually, the cast operator throws an exception if the cast isn't possible. Thus, it's called _unsafe_.
```

docs/topics/typecasts.md
```diff
@@ -1,5 +1,12 @@
 [//]: # (title: Type checks and casts)
 
+In Kotlin, you can perform type checks to check what type an object is at runtime. Type casts allow you to convert objects
```
danil-pavlov, line 3:
```suggestion
In Kotlin, you can perform type checks to check the type of an object at runtime. Type casts convert objects
```

docs/topics/typecasts.md
```diff
@@ -60,17 +67,47 @@ when (x) {
...
 ```
 
-Note that smart casts work only when the compiler can guarantee that the variable won't change between the check and the usage.
-More specifically, smart casts can be used under the following conditions:
-
-* `val` local variables - always, with the exception of [local delegated properties](delegated-properties.md).
-* `val` properties - if the property is private or internal or if the check is performed in the same [module](visibility-modifiers.md#modules) where the property is declared. Smart casts cannot be used on open properties or properties that have custom getters.
-* `var` local variables - if the variable is not modified between the check and the usage, is not captured in a lambda that modifies it, and is not a local delegated property.
-* `var` properties - never, because the variable can be modified at any time by other code.
+Note that smart casts work only when the compiler can guarantee that the variable won't change between the check and its usage.
```
danil-pavlov, line 70: I think an obligatory requirement like that should be an actual note (warning) or have a separate header

danil-pavlov (commented):
Nice update! I'm just wondering why you opted for an xml table instead of Markdown? We usually try to use md syntax because it's more readable except for special cases when it's not enough
Is it because of the necessary table header in md tables? You can still add something like Element | Condition at the top of the table

sarahhaggarty:
> Nice update! I'm just wondering why you opted for an xml table instead of Markdown? We usually try to use md syntax because it's more readable except for special cases when it's not enough Is it because of the necessary table header in md tables? You can still add something like Element | Condition at the top of the table

Thanks @danil-pavlov ! Yes, it's because I wanted to use a table without a header, which is only possible via XML right now. Originally this was just a bulleted list but I think it is easier to read in a table format. I couldn't find header names that I liked unfortunately. Even "element" doesn't seem relevant enough to me for the first column 🤔 Is it OK to stick with the XML or could you help me find better header names?

danil-pavlov:
I like the table idea and I can't come up with anything better than "Variables and properties", so let's keep the XML table as is

danil-pavlov (approved):

PR #3717, opened 2023-08-01
Update EAP and Classes pages

sarahhaggarty:
This PR adds a note to EAP pages explaining that if the Kotlin EAP plugin can't find the latest Kotlin version, users need to check that they are on the latest IDE version. It also clarifies the definition of a primary constructor as the current description is misleading. In addition, there is some formatting fixes for data class member functions.

This PR resolves:
- [KT-57525](https://youtrack.jetbrains.com/issue/KT-57525/Primary-constructor-definition)
- [KT-57864](https://youtrack.jetbrains.com/issue/KT-57864/Web-feedback-from-Install-the-Kotlin-EAP-Plugin-user-cannot-update)

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -16,8 +16,8 @@ class Empty
 
 ## Constructors
 
-A class in Kotlin can have a _primary constructor_ and one or more _secondary constructors_. The primary constructor is a
-part of the class header, and it goes after the class name and optional type parameters.
+A class in Kotlin has a _primary constructor_ and possibly one or more _secondary constructors_. The primary constructor
+is declared in the class header, and it goes after the class name and optional type parameters.
 
 ```kotlin
 class Person constructor(firstName: String) { /*...*/ }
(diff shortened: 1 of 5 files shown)


docs/topics/classes.md
```diff
@@ -29,8 +29,8 @@ If the primary constructor does not have any annotations or visibility modifiers
 class Person(firstName: String) { /*...*/ }
 ```
 
-The primary constructor cannot contain any code. Initialization code can be placed in _initializer blocks_ prefixed with
-the `init` keyword.
+The primary constructor initializes a class instance and its properties. Initialization code can be placed in _initializer
```
koshachy, line 32: Why have we decided to omit "cannot containing any code"?
Is it insignificant?

koshachy (approved):
LGTM

docs/topics/classes.md
sarahhaggarty: This sentence was causing confusion for some readers. They thought initializer blocks were code in the primary constructor even though the primary constructor is only defined in the class header. So I thought it would be cleaner to remove the sentence entirely.

koshachy: Well, maybe we should write explicitly about "primary constructor is only defined in the class header"?
What do you think?

koshachy: Anyway, feel free to merge the PR :)

sarahhaggarty: Thanks for your suggestion! I updated this paragraph with a couple more sentences to make it clear that the primary constructor is declared in the class header which can't contain any runnable code. I also explain that to run some code as part of object creation, users can use initialization blocks. Hopefully everything is covered now :)

PR #3724, opened 2023-08-03
code formatting in kotlin tour functions

olzhy:


sarahhaggarty:
Thanks for catching this @olzhy! Since this is such a small change, I've added it as part of #3717 so I'll close this PR now. Thanks again!

PR #3727, opened 2023-08-03

freesme:
When using IDEA (versions before 2023.1.4) to create a project following a tutorial, the default setting for the js compiler is 'BOTH', which may lead to errors. Need to manually change it to 'IR'.
image

### DIFF
--- a/docs/topics/multiplatform/multiplatform-library.md
+++ b/docs/topics/multiplatform/multiplatform-library.md
@@ -152,6 +152,19 @@ JS-specific functionality:
        fun encode(s: String): String
    }
    ```
+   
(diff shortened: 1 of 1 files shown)


koshachy:
Hello,
Thanks for your participation in improving the Kotlin documentation, we appreciate it!
In the section "[Set up the environment](https://kotlinlang.org/docs/multiplatform-library.html#set-up-the-environment)" they said that you should install the latest version of the IDEA.
I'd rather close this PR instead of adding this text to the documentation. 
We'll think about creating a note about this and add this to the tutorial. I close this PR.

Don't hesitate to open an issue in our [issue tracker](https://kotl.in/issue) or ask questions here.

cc @danil-pavlov

PR #3730, opened 2023-08-07
Add definitely non-nullable types documentation

sarahhaggarty:
This PR adds documentation for generic types that are definitely non-nullable. The content is primarily added to the [Generics](https://kotlinlang.org/docs/generics.html) page. A link has been added to the [Null safety](https://kotlinlang.org/docs/null-safety.html) page, and the [Java to Kotlin nullability guide](https://kotlinlang.org/docs/java-to-kotlin-nullability-guide.html) also has a small update.

This PR resolves [KT-54422](https://youtrack.jetbrains.com/issue/KT-54422/Add-documentation-on-website-for-Definitely-non-nullable-types-feature).

### DIFF
--- a/docs/topics/generics.md
+++ b/docs/topics/generics.md
@@ -322,6 +322,41 @@ fun <T> copyWhenGreater(list: List<T>, threshold: T): List<String>
 The passed type must satisfy all conditions of the `where` clause simultaneously. In the above example, the `T` type
 must implement _both_ `CharSequence` and `Comparable`.
 
+## Definitely non-nullable types
+
+To make interoperability with generic Java classes and interfaces easier, Kotlin supports declaring a generic type parameter
+as **definitely non-nullable**. 
+
+To declare a generic type `T` as definitely non-nullable, declare the type with `& Any`. For example: `T & Any`.
+
+A definitely non-nullable type must have a nullable [upper bound](#upper-bounds).
+
+The most common use case for declaring definitely non-nullable types is when you want to override a Java method that 
+contains `@NotNull` as an argument. For example, consider the `load()` method:
+
+```java
+import org.jetbrains.annotations.*;
+
+public interface Game<T> {
+    public T save(T x) {}
+    @NotNull
+    public T load(@NotNull T x) {}
+}
+```
+
+To override the `load()` method in Kotlin successfully, you need `T1` to be declared as definitely non-nullable:
+
+```kotlin
+interface ArcadeGame<T1> : Game<T1> {
+    override fun save(x: T1): T1
+    // T1 is definitely non-nullable
+    override fun load(x: T1 & Any): T1 & Any
+}
+```
+
+When working only with Kotlin, it's unlikely that you will need to declare definitely non-nullable types explicitly because 
+Kotlin's type inference takes care of this for you.
+
 ## Type erasure
 
 The type safety checks that Kotlin performs for generic declaration usages are done at compile time.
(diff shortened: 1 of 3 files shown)


dzharkov (approved):

docs/topics/generics.md
```diff
@@ -322,6 +322,73 @@ fun <T> copyWhenGreater(list: List<T>, threshold: T): List<String>
...
+To make interoperability with generic Java classes and interfaces easier, Kotlin supports declaring a generic type parameter
+as **definitely non-nullable**. 
+
+To declare a generic type `T` as definitely non-nullable, declare the type with `& Any`. For example: `T & Any`. 
+
+> The syntax represents an intersection between type `T` and the non-nullable type `Any`.
+> 
+{type="note"}
+
+A definitely non-nullable type must have a nullable [upper bound](#upper-bounds). In addition, `Any` must resolve to `kotlin.Any`.
```
sandwwraith, line 336: > In addition, `Any` must resolve to `kotlin.Any`

IMO, this sentence may create unnecessary confusion. In the rest of the page, it's just `Any` without `kotlin.Any`. It may leave questions like 'what else Any can resolve to?' without providing a solid answer or a link to that answer.

docs/topics/generics.md
```diff
@@ -322,6 +322,73 @@ fun <T> copyWhenGreater(list: List<T>, threshold: T): List<String>
...
 must implement _both_ `CharSequence` and `Comparable`.
 
+## Definitely non-nullable types
+
+To make interoperability with generic Java classes and interfaces easier, Kotlin supports declaring a generic type parameter
+as **definitely non-nullable**. 
+
+To declare a generic type `T` as definitely non-nullable, declare the type with `& Any`. For example: `T & Any`. 
+
+> The syntax represents an intersection between type `T` and the non-nullable type `Any`.
```
sandwwraith, line 332: I'm also not sure if this is needed. While this sentence is correct, Kotlin in general does not provide any kind of intersection types (besides this feature) and does not mention them anywhere in the documentation, pretending they are non-existent. Such `&` syntax is correct only for `& Any` case.

 Also, 'intersection type' is a kind of advanced topic in programming languages that may confuse inexperienced users.

docs/topics/generics.md
```diff
@@ -322,6 +322,73 @@ fun <T> copyWhenGreater(list: List<T>, threshold: T): List<String>
...
+    public T bar(@NotNull T x) {}
+}
+```
+
+To override the `bar()` method in Kotlin successfully, you need `T` (and its subtypes) to be declared as definitely non-nullable:
+
+```kotlin
+interface B<T1> : A<T1> {
+    override fun foo(x: T1): T1
+    //Subtype T1 is definitely non-nullable
```
sandwwraith, line 356: I'm not sure 'subtype' is a correct term here, as it is just a generic substitution T1 -> T

docs/topics/generics.md
sarahhaggarty: Removed!

docs/topics/generics.md
sarahhaggarty: Fixed!

sandwwraith (approved):

sandwwraith:
Great job!

danil-pavlov (commented):
Great job! My only note is that we try not to use `foo` and `bar` in code samples anymore. The same goes for one-letter names of interfaces and classes. So maybe it's possible to make more real-life examples on both pages

sarahhaggarty:
> Great job! My only note is that we try not to use `foo` and `bar` in code samples anymore. The same goes for one-letter names of interfaces and classes. So maybe it's possible to make more real-life examples on both pages

Thanks @danil-pavlov ! Good point. I have corrected the code samples to be more real-life. Please can you review again? 🙏

PR #3742, opened 2023-08-17
Update Gradle compiler option API and language versions

sarahhaggarty:
This PR updates the table for possible API and language versions for the Gradle compiler options.

### DIFF
--- a/docs/topics/gradle/gradle-compiler-options.md
+++ b/docs/topics/gradle/gradle-compiler-options.md
@@ -237,8 +237,8 @@ tasks.named('compileKotlin', KotlinCompilationTask) {
 
 | Name | Description | Possible values |Default value |
 |------|-------------|-----------------|--------------|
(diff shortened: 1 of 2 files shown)


ALikhachev (approved):

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -254,7 +254,7 @@ tasks
         compilerOptions
             .languageVersion
             .set(
-              org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
+              org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
```
koshachy, line 257: What if we use a variable here? Will it help us with updating the v.list file instead of updating the example?

I mean if it's not the only place where we have this value, we can use a var.

koshachy (approved):
LGTM,

See the comment

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: Great idea! At the moment this is the only instance of it but by making a it a variable it's much less likely that we will forget to update it in future. Updated!

PR #3754, opened 2023-08-26
improve documentation on no-arg-plugin.md

ygdrasil-io:
- Provide a sample annotation to illustrate how the plugin should be used.
- Explain how to utilize the plugin using Kotlin DSL in Gradle, as an alternative for Groovy.


koshachy (approved):
Hello,
Thank you for your interest in contributing to the Kotlin docs!
We appreciate it!
I've slightly updated the PR, without any significant changes.
Thanks again!

PR #3758, opened 2023-08-27
Added note that kotlin serialization plugin does not support v1.9.x

Maelstromeous:
It seems in this case depending upon `%kotlinVersion%` is causing a problem, which as a new Kotlin adopter going through for the first time the "Hello world" guide, having to debug the fact this package doesn't exist for Kotlin `>=1.9.0` is not a great first impression.


danil-pavlov:
`kotlinx.serialization` does support the latest Kotlin versions, inlcluding 1.9. See its [installation instructions](https://github.com/Kotlin/kotlinx.serialization)

PR #3761, opened 2023-08-28
update: warnings for the new and the legacy memory manager pages

danil-pavlov:
* [KT-59547](https://youtrack.jetbrains.com/issue/KT-59547) First result for "Kotlin concurrency" on Google leads to "legacy memory manager" page
* [KT-58582](https://youtrack.jetbrains.com/issue/KT-58582/Web-feedback-from-Kotlin-Native-memory-management-https-kotlinlang.org-docs-native-memory-manager.html)


koshachy (approved):
LGTM

PR #3769, opened 2023-09-05
KT-61711: Add documentation for via scope function

vlas-ilya:
Add documentation for [KT-61711](https://youtrack.jetbrains.com/issue/KT-61711/Add-scope-function-via-for-calling-functions-that-need-to-use-the-same-argument)


sarahhaggarty:
Hi @vlas-ilya,
I see that you are submitting a proposal for a new scope function. Until this has been assessed by our developers as part of [KT-61711](https://youtrack.jetbrains.com/issue/KT-61711/Add-scope-function-via-for-calling-functions-that-need-to-use-the-same-argument) and (if they agree) implemented, we can't accept any changes to Kotlin docs. Our documentation reflects what is possible with the latest Stable Kotlin release. So I'm going to close this PR for now. Good luck with your proposal!

PR #3776, opened 2023-09-07
KTL-1136 Main page revamp: Kotlin usage highlights

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1136/Main-page-revamp-Kotlin-usage-highlights


pages/index.tsx
```diff
@@ -1,11 +1,141 @@
...
+import MultiplatformPreviewImage from '../public/images/index/multiplatform-preview.svg';
+
+import { InfoBlock } from '../blocks/main/info-block/info-block';
+import { DividerLine } from '../blocks/main/divider-line/divider-line';
+
+import Link from 'next/link';
+
+import styles from './index.module.css';
+
+const kotlinUsageHighlightsCases = [
```
krutilov, line 30: I'm now not sure about keeping all the data on this page instead of putting it into the components. Seems like this page is getting too big

krutilov (approved):

pages/index.tsx
nikpachoo: The better option is to put it in a YAML file, but I don't know how to refer to the images. Probably by name? Indeed, I can match an image with the company by name, and this contract can be mentioned in Readme.

zoobestik (approved):

PR #3788, opened 2023-09-13
feat: using platform-specific apis

danil-pavlov:


docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
+different platforms. Use expected and actual functions and properties for simple cases.
 
-fun logDebug(message: String) = writeLogMessage(message, LogLevel.DEBUG)
-fun logWarn(message: String) = writeLogMessage(message, LogLevel.WARN)
-fun logError(message: String) = writeLogMessage(message, LogLevel.ERROR)
-```
+## Interfaces in common code
+
+If platform-specific logic is too big and complex, you can define an interface to represent it in the common code and
+then provide different implementations in platform source sets.
```
sarahhaggarty, line 93:
```suggestion
then providing different implementations in platform source sets.
```

docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
 ```
 
-</tab>
-</tabs>
+On Android, you create an instance of `AndroidPlatform` and pass it to the `application()` function, while on iOS, you
+similarly create and pass an instance of `IOSPlatform`. These entry points don't need to be the entry points of your
+apps, but this is where you can call the specific functionality of your shared module.
+
+Providing correct implementations with expected and actual functions or directly through entry points works well for
+simple scenarios. However, if you use DI in your project, we recommend you use it in such cases as well for consistency.
```
sarahhaggarty, line 213: Since DI is discussed in the next section, should we remove this paragraph to avoid duplication of content?

docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
+![Using dependency injection framework](expect-di-framework.png){width=500}
+
+This approach only works if you put implementations in platform modules. It isn't very scalable because your Kotlin
+Multiplatform module can't be self-sufficient − you should implement the common interface in a different module.
+
+<!-- If you're interested in expanding this functionality to a shared module, please vote for this issue [TODO create or find an issue] and describe your use case. -->
+
+## What's next?
+
+For more examples and information on the expect/actual mechanism, see [Expected and actual declarations](multiplatform-expect-actual.md).
```
sarahhaggarty, line 276:
```suggestion
For more examples and information on expected and actual declarations, see [Expected and actual declarations](multiplatform-expect-actual.md).
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+   common code but should not include implementation. Instead, the platform-specific code provides this implementation.
+3. In each platform-specific source set, declare the same construct in the same package and mark it with the `actual`
+   keyword. It's your _actual declaration_. Typically, it contains an implementation using platform-specific libraries.
+
+During compilation for a specific target, the compiler tries to match each actual declaration it finds with the
+corresponding _expected_ declaration in common code. The compiler ensures that:
+
+* Every expected declaration in the common source set has a matching actual declaration in every platform-specific
+  source set.
+* Expected declarations do not contain any implementation.
```
sarahhaggarty, line 33:
```suggestion
* Expected declarations don't contain any implementation.
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+actual class Identity {
+    actual val userName: String = getlogin()?.toKString() ?: "None"
+    actual val processID: Long = getpid().toLong()
+}
+```
+
+You might have seen this approach in demonstration materials before. However, using classes in simple cases where
+interfaces would be sufficient is _not recommended_.
+
+With interfaces, you do not limit your design to one implementation per target platform. Also, it's much easier to
```
sarahhaggarty, line 284:
```suggestion
With interfaces, you don't limit your design to one implementation per target platform. Also, it's much easier to
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+}
+
+expect class MyDate {
+    fun getYear(): Int
+    fun getMonth(): Month
+    fun getDayOfMonth(): Int
+}
+```
+
+Within a JVM module, the `java.time.Month` enum could be used to implement the first expected declaration and
```
sarahhaggarty, line 411:
```suggestion
Within a JVM module, the `java.time.Month` enum can be used to implement the first expected declaration and
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+Instead, you can use [type aliases](type-aliases.md) to connect the expected declarations and the platform-specific
+types:
+
+```kotlin
+actual typealias Month = java.time.Month
+actual typealias MyDate = java.time.LocalDate
+```
+
+The `typealias` declaration should be defined in the same package as the expected declaration, while the referred class
+should be defined elsewhere.
```
sarahhaggarty, line 424: Can we remove "should" and make this whole paragraph imperative?

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+
+<!-- See [Using platform-specific APIs](multiplatform-connect-to-apis.md#actualizing-an-interface-or-a-class-with-an-existing-platform-class-using-typealiases)
+for an Android-specific example of this pattern. -->
+
+### Widened visibility in actual declarations
+
+You can make actual implementations more visible than the corresponding expected declaration. This is useful if you
+don't want to expose your API as public for common clients.
+
+Currently, the Kotlin compiler issues an error in case of visibility changes. You can suppress this error by
```
sarahhaggarty, line 438:
```suggestion
Currently, the Kotlin compiler issues an error in the case of visibility changes. You can suppress this error by
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+// In the jvmMain source set:
+actual enum class Department { IT, HR, Sales, Legal }
+```
+
+```kotlin
+// In the nativeMain source set:
+actual enum class Department { IT, HR, Sales, Marketing }
+```
+
+However, matching `Department` in common code can never be exhaustive now. Therefore, the compiler requires you
```
sarahhaggarty, line 485: I don't understand this sentence. Does it mean that all possible cases need to be covered?

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,549 @@
...
+
+actual typealias XmlSerializable = XmlRootElement
+```
+
+There is an additional consideration when using `expect` with annotation classes. Annotations are used to attach
+metadata to code and do not appear as types in signatures. It's not essential for an expected annotation to have an
+actual class on a platform where it's never required.
+
+Because of this, you only need to provide an `actual` declaration on platforms where the annotation is used. This
+behavior is not enabled by default, and it requires the type to be marked with [`OptionalExpectation`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-optional-expectation/).
```
sarahhaggarty, line 532:
```suggestion
behavior isn't enabled by default, and it requires the type to be marked with [`OptionalExpectation`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-optional-expectation/).
```

sarahhaggarty (changes requested):
Looking good @danil-pavlov!
I mostly have minor comments. There are two things I consistently changed:
* "expected/actual mechanism" -> "expected and actual declarations". I think this is clearer for the reader.
* The "correct" implementation -> The "right" or "relevant" implementation. "Correct" makes it sound like there are incorrect implementations. This could just be a nit from me as a native speaker though.

docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
+## Expected and actual functions and properties
 
-> Use expected and actual declarations only for Kotlin declarations that have platform-specific
-> dependencies. Implementing as much functionality as possible in the shared module is better, even if doing so takes
-> more time.
->
-> Don't overuse expected and actual declarations – in some cases, an [interface](interfaces.md) may be a better choice
-> because it is more flexible and easier to test.
+> The mechanism of expected and actual declarations is in [Beta](components-stability.md).
+> It is almost stable, but migration steps may be required in the future.
```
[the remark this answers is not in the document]
danil-pavlov: I guess we're not referring directly to the status name here. Anyway, this change would require the [guide](https://docs.google.com/document/d/1zmjZvQHsvkohMZCQTpv64gPDZEpUOQwYqQ-YH601oWc/edit#heading=h.ty9tig7yrvtm) update

docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
 ```kotlin
-// Common
+// In the common source set:
 expect fun randomUUID(): String
 ```
 
-In each platform-specific module (iOS and Android), provide the actual implementation for the function `randomUUID()`
-expected in the common module. Use the `actual` keyword to mark the actual implementation.
+In each platform-specific source set (iOS and Android), provide the actual implementation for the
+function `randomUUID()` expected in the common module. Use the `actual` keyword to mark these actual implementations.
```
sarahhaggarty, line 55:
```suggestion
`randomUUID()` function expected in the common module. Use the `actual` keyword to mark these actual implementations.
```

docs/topics/multiplatform/multiplatform-connect-to-apis.md
```diff
@@ -1,380 +1,276 @@
...
-    actual fun sendMessage(msg: String) {
-        webSocket?.send(msg)
-    }
+// In the iosMain source set:
+class IOSPlatform : Platform
+
+actual fun platform() = IOSPlatform()
+```
+
+WHen you call `platform()` in the common code, it can work with an object of the `Platform` type. When you run this common
```
sarahhaggarty, line 153:
```suggestion
When you call the `platform()` function in the common code, it can work with an object of the `Platform` type. When you run this common
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,553 @@
...
+To define expected and actual declarations, follow these rules:
+
+1. In the common source set, declare a standard Kotlin construct. This can be a function, property, class, interface,
+   enumeration, or annotation.
+2. Mark this construct with the `expect` keyword. It's your _expected declaration_. These declarations can be used in
+   common code but shouldn't include any implementation. Instead, the platform-specific code provides this implementation.
+3. In each platform-specific source set, declare the same construct in the same package and mark it with the `actual`
+   keyword. It's your _actual declaration_. Typically, it contains an implementation using platform-specific libraries.
+
+During compilation for a specific target, the compiler tries to match each actual declaration it finds with the
```
sarahhaggarty, line 28:
```suggestion
During compilation for a specific target, the compiler tries to match each _actual_ declaration it finds with the
```

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,553 @@
...
+
+  actual fun buildIdentity() = Identity(
+      getlogin()?.toKString() ?: "None",
+      getpid().toLong()
+  )
+  ```
+
+  Here, platform functions return platform-specific `Identity` instances.
+
+> Starting with Kotlin version 1.9.0, using `getlogin()` and `getpid()` requires the `@OptIn` annotation.
```
sarahhaggarty, line 115:
```suggestion
> Starting with Kotlin version 1.9.0, using `getlogin()` and `getpid()` functions requires the `@OptIn` annotation.
```

sarahhaggarty (commented):
Just a few more minor comments! There were also a couple left over from the last review.

docs/topics/multiplatform/multiplatform-connect-to-apis.md
sarahhaggarty: OK, then shouldn't it be "expect**ed**/actual mechanism"?

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -0,0 +1,553 @@
...
+// In the jvmMain source set:
+actual enum class Department { IT, HR, Sales, Legal }
+```
+
+```kotlin
+// In the nativeMain source set:
+actual enum class Department { IT, HR, Sales, Marketing }
+```
+
+However, matching `Department` in common code can never be exhaustive now. Therefore, the compiler requires you
```
sarahhaggarty, line 489:
```suggestion
However, it's now possible that these extra constants in platform source sets won't match with those in common code. Therefore, the compiler requires you
```

sarahhaggarty (approved):
Looking good! Just a couple suggestions left but LGTM!

docs/topics/multiplatform/multiplatform-connect-to-apis.md
danil-pavlov: Not really, the key words in the mechanism are "expect" and "actual"

sarahhaggarty: Then shouldn't "expect" and "actual" be inline code? Like we discussed in the past?

PR #3789, opened 2023-09-13
Update KMP Compatibility Guide for 1.9.20

dsavvinov:


docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +492,251 @@ projects.
...
+```kotlin
+kotlin {
+    jvm("jvmKtor")
+    jvm("jvmOkio") // is not recommended and will produce deprecation warning
+}
+```
+
+One popular case is having two related pieces of code together. For example, you might be compelled to use
+`jvm("jvmKtor")` and `jvm("jvmOkio")` in your `:shared` Gradle project for providing an implementation of networking
+code based on Ktor or OKIO:
```
martinbonnin, line 513: FWIW, the example might be better with `Ktor` & `OkHttp`. Okio is the low level IO library but doesn't deal with networking.

dsavvinov: Indeed, that was a brainlag from me (though about `OkHttp` but typed `Okio`). Thanks for pointing out, Martin! 👍 

cc @danil-pavlov could you fix that during the TW editing?

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +502,251 @@ projects.
...
+
+kotlin {
+    compilerOptions {
+        // ... copy compiler options of jvmKtorMain here ...
+    }
+}
+```
+
+While this approach requires more initial setup work, it doesn't use any low-level entities of Gradle and KGP, making
+easier to work with the resulting build and maintain it.
```
martinbonnin, line 623: I think where this falls short is tests? Suppose you had a common test suite that tested both `ktor` and `okhttp` clients. The code in the tests is all the same, it just expects a HTTP client:

```kotlin
// commonMain
interface HttpClient { ... }
expect fun HttpClient(): HttpClient

// commonTest
@Test 
fun testClient() {
  assertThat(HttpClient().get("https://...").data).isEqualsTo("Hello World")
}
```

With multiple JVM targets, it was only a matter of defining 2 actuals:

```kotlin
// ktorMain
actual fun HttpClient() = KtorHttpClient()

// okHttpMain
actual fun HttpClient() = OkHttpClient()
```

Now with different Gradle projects, how would one do this? `expect`/`actual` do not cross Gradle projects boundaries. And duplicating the test code doesn't feel great. Any thoughts?

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -504,227 +503,244 @@ projects.
...
 **When do the changes take effect?**
 
-In Kotlin 1.9.20, a deprecation warning is introduced when multiple targets of the same type is used in Kotlin Multiplatform projects.
-In Kotlin 2.0, error will be reported in such cases, causing your build to fail.
+* 1.9.20: introduce a deprecation warning when multiple targets of the same type are used in Kotlin Multiplatform projects
+* 2.0: report an error in such cases, causing the build to fail
 
 <anchor name="jvmWithJava-preset-deprecation"></anchor>
-## Deprecation of 'jvmWithJava'-preset
+## Deprecated the jvmWithJava preset
```
dsavvinov, line 635: Minor: weird wording, probably a typo

dsavvinov (commented):
Please take a loot at my comment about forms of verbs in deprecation timeline sections and about "similat target types". Everything else is a minor

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -329,8 +328,8 @@ kotlin {
         }
 
         // Option #1. Shorter and more readable, use it when possible:
-       val jvmMain by getting { // Usually, the name of the default source set 
-                                // is a simple concatenation of the target name and the compilation name
+        val jvmMain by getting { // Usually, the name of the default source set 
+            // is a simple concatenation of the target name and the compilation name
```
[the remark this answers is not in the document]
danil-pavlov: I does look better, thanks!

docs/topics/multiplatform/multiplatform-compatibility-guide.md
danil-pavlov: true

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -504,227 +503,244 @@ projects.
 In Kotlin 1.9.0, a deprecation warning is introduced when the `android` name is used in Kotlin Multiplatform projects.
 
 <anchor name="declaring-multiple-targets"></anchor>
-## Declaring multiple targets of the same kind
+## Declaring multiple targets of the same type
```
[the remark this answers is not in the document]
danil-pavlov: we can avoid terms here with 'similar', for example

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -243,9 +242,9 @@ If you're a library author and want to be extra safe, check that consumers can w
 
 **When do the changes take effect?**
 
-* In 1.8.20, usage of these properties provokes a warning
-* In 1.9.20, usage of these properties provokes an error
-* In 2.0, support for these properties will be removed, and Kotlin Gradle Plugin will ignore them 
+* 1.8.20: report a warning when these properties are used
```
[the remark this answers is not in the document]
danil-pavlov: We already used this approach in the Multiplatform compatibility guide when we have a list for deprecation cycles. Now it's consistent in all sections

Here we follow Kotlin compatibility guides that have described the cycle in this way since [Kotlin 1.4](https://kotlinlang.org/docs/compatibility-guide-14.html)

docs/topics/multiplatform/multiplatform-compatibility-guide.md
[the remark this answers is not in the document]
dsavvinov: Yes, "similar" works

docs/topics/multiplatform/multiplatform-compatibility-guide.md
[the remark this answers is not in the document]
dsavvinov: Well from the link the difference is quite clear. The linked page (and any of the language compatibility pages) doesn't give any single recommendation to users, like "Migrate your code by replacing...". And KMP Compatibility page does that all over the place.

Which one is the proper way is an open question, but it's clear that decision from Language Compatibility Guide can't be blindly retrofitted to KMP Compatibiltiy Guide. They're written in significantly different styles.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
dsavvinov: That's true. The brutally honest description for this deprecation would've been something like: "if you've encountered this deprecation, and you don't see an easy way to migrate your project, then reach us and tell more about your case". 

But that would be a bit unfriendly, so we tried our best to provide some intrsuctions which applicable at least in some cases. It is expected that these instructions miss some cases, and we'll be happy to learn if someone encounters such cases in real projects

martinbonnin: We actually have this exact `expect`/`actual` scenario [in this integration test](https://github.com/apollographql/apollo-kotlin/blob/main/tests/engine/src/commonTest/kotlin/engines.kt#L4). I've turned this problem upside down but still not sure how to convert. Any guidance would be super welcome.

martinbonnin: Just took another look at it. In the current state, it's working with a single JVM target and 2 compilations but shows this warning:

```
w: Kotlin Source Set 'ktorTest' can't depend on 'commonTest' as they are from different Source Set Trees.
Please remove this dependency edge.
```

I'm not sure how to remove the common source set without breaking the `expect`/`actual`

docs/topics/multiplatform/multiplatform-compatibility-guide.md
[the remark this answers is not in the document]
danil-pavlov: But these are different sections. We give recommendations to users in the **What's the best practice now?** section, and the **When do the changes take effect?** section just lists Kotlin versions that describe the deprecation cycle as stated in the header, which is exactly the same as in Kotlin compatibility guides 

added a sentence about the deprecation cycle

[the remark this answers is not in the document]
dsavvinov: Ok

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+            dependsOn(commonJvmMain)
+            dependencies {
+                // OkHttp dependencies
+            }
+        }
+    }
+}
+```
+
+The implementation comes with a non-trivial configuration complexity:
```
sarahhaggarty, line 564:
```suggestion
The implementation comes with non-trivial configuration complexity:
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+
+* You have to set up Gradle attributes on the `:shared` side and each consumer's side. Otherwise, Gradle can't
+  resolve dependencies in such projects because it's not clear whether the consumer should receive the Ktor-based
+  or the OkHttp-based implementation without additional information.
```
sarahhaggarty, line 568:
```suggestion
or the OkHttp-based implementation.
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+* You have to set up the `commonJvmMain` source set manually.
+* The configuration involves a handful of low-level Gradle and the Kotlin Gradle plugin abstractions and APIs.
```
sarahhaggarty, line 570:
```suggestion
* The configuration involves a handful of low-level Gradle and Kotlin Gradle plugin
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+
+**What's the best practice now?**
+
+The configuration gets complicated because Ktor-based and OkHttp-based implementations are
```
sarahhaggarty, line 574:
```suggestion
The configuration is complex because Ktor-based and OkHttp-based implementations are
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+_in the same Gradle project_. In many cases, it's possible to extract those parts into separate Gradle projects.
+Here's a general outline of this refactoring:
```
sarahhaggarty, line 576:
```suggestion
Here's a general outline of such a refactoring:
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+
+1. Replace two duplicating targets from the original project with a single target. If you had a shared source set between
```
sarahhaggarty, line 578:
```suggestion
1. Replace two duplicated targets from the original project with a single target. If you had a shared source set between
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+    ```
+
+3. Configure each new Gradle project:
+
+   * Most likely, you don't need to apply the `kotlin("multiplatform")` plugin, as these projects compile only to one
+     target. In this example, you can apply `kotlin("jvm")`.
+   * Move the content of original target-specific source sets to their respective projects, for example,
+     from `jvmKtorMain` to `ktor-impl/src`.
+   * Copy the configuration of source sets: dependencies, compiler options, and so on.
+   * Add a dependency from a new Gradle project to the original one.
```
sarahhaggarty, line 608:
```suggestion
* Add a dependency from the new Gradle project to the original one.
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+            // Copy compiler options of jvmKtorMain here
+        }
+    }
+    ```
+
+While this approach requires more work on the initial setup, it doesn't use any low-level entities of Gradle and
+the Kotlin Gradle plugin, making it easier to use and maintain the resulting build.
+
+> Unfortunately, we can't provide detailed migration steps for each case. If the instructions above don't work
+> for you, describe your case in this [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-59316).
```
sarahhaggarty, line 632:
```suggestion
> for you, describe your use case in this [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-59316).
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+>
+{type="tip"}
+
+**When do the changes take effect?**
+
+Here's the planned deprecation cycle:
+
+* 1.9.20: introduce a deprecation warning when similar multiple targets are used in Kotlin Multiplatform projects
```
sarahhaggarty, line 640:
```suggestion
* 1.9.20: introduce a deprecation warning when multiple similar targets are used in Kotlin Multiplatform projects
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+## Deprecated commonMain and commonTest with custom dependsOn
+
+**What's changed?**
+
+The `commonMain` and `commonTest` source sets usually represent the roots of the `main` and `test` source set hierarchies,
+respectively. However, it was possible to override that by manually configuring `dependsOn` relations of these source sets.
+
+Maintaining such configuration requires extra effort and knowledge about multiplatform build internals. Additionally,
+it decreases code readability and reusability of the code because you need to read the particular buildscript
+to be sure whether the `commonMain` is the root of the `main` source set hierarchy.
```
sarahhaggarty, line 702:
```suggestion
to be sure whether `commonMain` is the root of the `main` source set hierarchy.
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+
+Therefore, accessing `dependsOn` on `commonMain` and `commonTest` is now deprecated.
+
+**What's the best practice now?**
+
+Suppose you need to migrate to 1.9.20 the `customCommonMain` source set that uses `commonMain.dependsOn(customCommonMain)`.
+In most cases, `customCommonMain` participates in the same compilations as `commonMain`, so you can merge the
```
sarahhaggarty, line 709:
```suggestion
In most cases, `customCommonMain` participates in the same compilations as `commonMain`, so you can merge
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+`customCommonMain` into `commonMain`:
+
+1. Copy sources of `customCommonMain` into `commonMain`.
+2. Add all dependencies of `customCommonMain` to `commonMain`.
+3. Add all compiler option settings of `customCommonMain` to `commonMain`.
+
+In rare cases, `customCommonMain` might be participating in more compilations than `commonMain`.
+Such configuration requires additional low-level configuration of the buildscript. If you're not sure if that's your case,
```
sarahhaggarty, line 717:
```suggestion
Such a configuration requires additional low-level configuration of the build script. If you're not sure if that's your use case,
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+it most likely isn't.
+
+If it's actually your case, "swap" these two source sets. To do that, move the sources and settings of `customCommonMain`
```
sarahhaggarty, line 720:
```suggestion
If it is your use case, "swap" these two source sets by moving the sources and settings of `customCommonMain`
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+* \>=1.9.20: report an error when `dependsOn` is used in `commonMain` or `commonTest`
+
+<anchor name="target-presets-deprecation"></anchor>
+## Deprecated target presets API
+
+**What's changed?**
+
+In the very early development stages, Kotlin Multiplatform introduced an API for working with so-called _target presets_.
+Each target preset essentially represented a factory for Kotlin Multiplatform targets. This API turned out to be largely
+redundant, as DSL functions like `jvm()` or `iosSimulatorArm64()` cover the same cases while being much more straightforward
```
sarahhaggarty, line 737:
```suggestion
redundant, as DSL functions like `jvm()` or `iosSimulatorArm64()` cover the same use cases while being much more straightforward
```

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -490,3 +508,281 @@ projects.
...
+
+**When do the changes take effect?**
+
+Here's the planned deprecation cycle:
+
+* 1.9.20: report a warning on any usages of the presets-related API
+* 2.0: raise this warning to an error
+* \>2.0: remove the presets-related API from the public API of the Kotlin Gradle plugin; sources that still use it fail
+  with "unresolved reference" errors, and binaries (for example, Gradle plugins) might fail with linkage errors
+  unless recompiled against the modern versions of the Kotlin Gradle plugin
```
sarahhaggarty, line 788:
```suggestion
unless recompiled against the latest versions of the Kotlin Gradle plugin
```

PR #3799, opened 2023-09-20
Fix small bit of wording on CocoaPods setup page

zsmb13:


koshachy (approved):
Thanks for pointing this out!

PR #3802, opened 2023-09-24
fix kotlin tour functions (lambda expressions) documentation

alexey-ivanov-official:
fix copy-pasted naming in "kotlin tour > functions > lambda expressions > pass to another function > map()" example

the term "predicate" seems to be copy-pasted from a nearby example of filter() method. 
the suitable term for map() method param is "transformer" I guess

### DIFF
--- a/docs/topics/tour/kotlin-tour-functions.md
+++ b/docs/topics/tour/kotlin-tour-functions.md
@@ -383,7 +383,7 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-tour-lambda-map"}
 
-The `.map()` function accepts a lambda expression as a transformer:
+The `.map()` function accepts a lambda expression as a transform function:
 * `{ x -> x * 2 }` takes each element of the list and returns that element multiplied by 2.
 * `{ x -> x * 3 }` takes each element of the list and returns that element multiplied by 3.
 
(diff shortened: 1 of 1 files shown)


docs/topics/tour/kotlin-tour-functions.md
```diff
@@ -383,7 +383,7 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-tour-lambda-map"}
 
-The `.map()` function accepts a lambda expression as a predicate:
+The `.map()` function accepts a lambda expression as a transformer:
```
sarahhaggarty, line 386:
```suggestion
The `.map()` function accepts a lambda expression as a transform function:
```

sarahhaggarty (changes requested):
Thanks for pointing this our @alexey-ivanov-official! This is indeed incorrect. I would like to use slightly different text. Could you make the changes?

docs/topics/tour/kotlin-tour-functions.md
alexey-ivanov-official: Yes. That's better. Thank you!

sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @alexey-ivanov-official !

PR #3804, opened 2023-09-30
Add a new idiom in idioms.md

mahozad:
If you think this idiom is worth adding, it can probably be merged to the idiom before it as well.

It is especially useful for the below examples (only if the expression is not null; otherwise, doing nothing):
  - Creating an object of a class having a single-parameter constructor
  - Invoking a single-parameter **callback**
  - Invoking a single-parameter function

```kotlin
val path = System
    .getProperty("myDestination")
    ?.let(::Path)

fun convert(onProgress: (Float) -> Unit) {
    var progress: Float? = null
    // ...
    progress?.let(::println)
    progress?.let(onProgress)
    path?.parent?.toFile()?.let(Desktop.getDesktop()::open)
}
```

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -196,6 +196,14 @@ value?.let {
 }
 ```
 
+## Call function with value only if value is not null
+
+```kotlin
+val value = ...
+
+value?.let(::println)
+```
+
 ## Map nullable value if not null
 
 ```kotlin


SebastianAigner:
Hi @mahozad – while valid Kotlin code, this isn't actually something we recommend. For most basic cases, we encourage people to use simple `== null` / `!= null` checks, and relying on Kotlin's smart casting functionality:

```kotlin
if (value != null) println(value)
```

When you do use let, try to nudge people towards using the lambda syntax, potentially using implicitly named parameters, i.e.:

```kotlin
val path = System
    .getProperty("myDestination")
    ?.let { Path(it) }
```

mahozad:
OK, thanks for the tip.

koshachy:
Hey, 
Thanks everyone for participating.
I close this PR.
Don't hesitate to ask any questions or re-open the PR.

PR #3818, opened 2023-10-09
Remove nullable boolean "idiom"

SebastianAigner:
Open for discussion here @JetBrains/kotlin-technical-writing – I don't understand the original motivation for adding this as a Kotlin "idiom", seeing how this seems like an instance of the [tri-state boolean](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem) problem.

If there is a reason as to why we're highlighting this as an idiom, I'd appreciate a short explanation – otherwise, it's probably better if it is removed.

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -330,17 +330,6 @@ stream.buffered().reader().use { reader ->
 inline fun <reified T: Any> Gson.fromJson(json: JsonElement): T = this.fromJson(json, T::class.java)
 ```
 
-## Nullable Boolean
-
-```kotlin
-val b: Boolean? = ...
-if (b == true) {
-    ...
-} else {
-    // `b` is false or null
-}
-```
-
 ## Swap two variables
 
 ```kotlin


koshachy (approved):
Hey @SebastianAigner
Thanks for raising the issue!
This idiom was added long ago, and I can't find any reason why it should be in the documentation.

I agree with you that we should remove it.
Thanks again for pointing this out!
We appreciate it!

YSakhno:
Frankly speaking, I do not understand why this was removed. This idiom, as it was written, while very trivial, is not obvious. As in, it's hard to come up with it on your own — when someone unfamiliar with the "idom" is hard pressed to use nullable `Boolean` in an "if" (or otherwise convert nullable `Boolean` to a non-nullable one), they will most likely write `b!!` or `b ?: false`. To me, `if (b == true)` is more idiomatic than `if (b ?: false)` (and much, much more idiomatic than `if (b!!)`).

If your argument was "simply don't use nullable `Boolean`s", then why is it in the language at all? Sometimes you _have_ to use it because it is more convenient, or because such are the constraints of the underlying system.

P.S. I would actually _extend_ this idiom by including the case of defaulting `null` to `true`: `b != false` (although this might be less obvious to the reader than `b ?: true`).

P.P.S. Also see [Detekt's `NullableBooleanCheck`](https://detekt.dev/docs/rules/style#nullablebooleancheck). Your change will break their rules, because they are referencing the very much idiom you removed.

PR #3829, opened 2023-10-13
ktl-1123 feat: new main page why kotlin block

krutilov:
Known issue: Run button doesn't work yet


blocks/main/why-kotlin/playground.css
```diff
@@ -0,0 +1,87 @@
+.kotlin-code-examples-section {
```
zoobestik, line 1: Better if it will be as theme of component playground 🤔

blocks/main/why-kotlin/why-kotlin.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+                                <NavItem icon={<PlayIcon />} onClick={handleRunButton}>
+                                    Run
+                                </NavItem>
+                            </div>
+                        </div>
+
+                        <div className="tab-content kotlin-code-examples-section">
+                            <ContentSwitcher index={activeIndex}>
+                                <div className={styles.tab} key={0}>
```
zoobestik, line 117: Looks strange `key` without array 😈

zoobestik (approved):

blocks/main/why-kotlin/why-kotlin.tsx
krutilov: It's a tab component hack, to it knows that it's different content inside the tab 😂. I'll add samples to `codeExamplesList` and do a map here, good catch

PR #3832, opened 2023-10-16
Java Sources have to be created in src/main/java

jazzyF:
Kotlin Code does not recognize Java classes created under src/main/kotlin


udalov:
I've tried to create a Java class in an IntelliJ project under `src/main/kotlin` and everything works as expected. It is compiled successfully, and its usages in Kotlin are also compiled fine and resolved correctly in the IDE. So I don't think a change to the docs is needed. Rather, if Java files under `src/main/kotlin` are not recognized in your case, could you please create an issue at https://kotl.in/issue so that we could investigate it?

PR #3839, opened 2023-10-19
Suggest disabling sandboxing in the target with embedAndSign task

abdulowork:


danil-pavlov:
Multiplatform pages will be updated in a different repository, the rest in #3864

PR #3841, opened 2023-10-20
feat: full support for gradle configuration cache in kmp

danil-pavlov:
* [KT-54362](https://youtrack.jetbrains.com/issue/KT-54362)

### DIFF
--- a/docs/topics/gradle/gradle-compilation-and-caches.md
+++ b/docs/topics/gradle/gradle-compilation-and-caches.md
@@ -144,13 +144,6 @@ To disable caching for all Kotlin tasks, set the system property `kotlin.caching
 
 ## Gradle configuration cache support
 
-> The Gradle configuration cache is supported only by the following Gradle plugins:
->   * `org.jetbrains.kotlin.jvm`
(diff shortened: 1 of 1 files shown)


docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -146,13 +146,13 @@ To disable caching for all Kotlin tasks, set the system property `kotlin.caching
 
 > The Gradle configuration cache is supported only by the following Gradle plugins:
 >   * `org.jetbrains.kotlin.jvm`
->   * `org.jetbrains.kotlin.js`
+>   * `org.jetbrains.kotlin.multiplatform`
```
dsavvinov, line 149: I think at this point it's fair to say that all plugins support GCC, so there's no need to mention it at all. I suggest to remove this note. Of course, not all tasks/actions will be fully GCC-optimized, but that's a usual situation

dsavvinov (changes requested):

dsavvinov (approved):

PR #3843, opened 2023-10-20
Fix module name casing

matytyma:
Fixed module names casing from `NativeMain` and `NativeTest` to `nativeMain` and `nativeTest`


danil-pavlov:
Thank you for the report! We've fixed the issue in the latest 1.9.20 release

PR #3857, opened 2023-10-26
Update kotlin-tour-functions.md

HiroVodka:
Remove Default argument

The Default argument is used here is confusing because it is explained in the next heading


sarahhaggarty (approved):

sarahhaggarty:
Thanks for pointing this out @HiroVodka!

PR #3867, opened 2023-10-30
update: cinterop opt-in requirement

danil-pavlov:


docs/topics/native/mapping-function-pointers-from-c.md
```diff
@@ -1,5 +1,11 @@
 [//]: # (title: Mapping function pointers from C – tutorial)
 
+> The C libraries import is [Experimental](components-stability.md#stability-levels-explained).
+> All Kotlin declarations generated by the `cinterop` tool from C libraries,
```
SvyatoslavScherbina, line 4: Why the comma?

docs/topics/native/native-c-interop.md
```diff
@@ -1,5 +1,11 @@
 [//]: # (title: Interoperability with C)
 
+> The C libraries import is [Experimental](components-stability.md#stability-levels-explained).
```
SvyatoslavScherbina, line 3: Objective-C libraries import is also experimental. So maybe we should also include this header in https://kotlinlang.org/docs/native-objc-interop.html? And mention Objective-C explicitly.

docs/topics/native/native-c-interop.md
```diff
@@ -1,5 +1,11 @@
...
+> All Kotlin declarations generated by the `cinterop` tool from C libraries,
+> should have the `@ExperimentalForeignApi` annotation.
```
SvyatoslavScherbina, line 5: It makes sense to mention that platform libraries are not affected.

PR #3882, opened 2023-11-02
fix: include print statement in object expression example

oliverscherf-tomtom:
The `print(helloWorld)` line should be visible on the docs page, otherwise one could only guess why "Hello World" is printed when running the example code.

[See example](https://kotlinlang.org/docs/object-declarations.html#creating-anonymous-objects-from-scratch).


koshachy:
Hey @oliverscherf-tomtom 
Thanks for pointing out the issue. I definitely agree with you that this line shouldn't be hidden.

koshachy (approved):

PR #3898, opened 2023-11-15
Set CocoaPods headers through DSL

daugeldauge:
There's a designated DSL property for this, no need to configure the task manually.


danil-pavlov (approved):
Looks great, thank you!

PR #3900, opened 2023-11-16
Put emphasis on enum `entries` rather than `values()`

SebastianAigner:
`.entries` is a stable part of the enum API since 1.9.0, so the featured code snippets should use it. A reference to the soft-deprecated `values()` function can of course remain :)


qurbonzoda (approved):
LGTM

sarahhaggarty (approved):

PR #3907, opened 2023-11-19
Update : data class documentation

hardikm9850:
Explained the need for a minimum of one constructor parameter in the data class for proper generation of essential methods.


sarahhaggarty:
Hi @hardikm9850
Thanks for your suggestion! Your suggestion to add additional information about the relationship between primary constructor and generated methods actually duplicates what is explained in the earlier sentence:

> To ensure consistency and meaningful behavior of the generated code

However, I agree that the wording of the requirements could be stronger. Instead of taking up your time to make such minor changes I incorporated these as part of another PR on inline classes: #3919.

So for now I'll close this PR. Thanks for your contribution and I hope you'll contribute again in future!

PR #3909, opened 2023-11-20
update: mobile rename and unused images

danil-pavlov:


redirects/pages.yml
```diff
@@ -756,3 +756,23 @@
     - /docs/kmm-understand-project-structure.html
     - /docs/mpp-discover-project.html
   to: /docs/multiplatform-discover-project.html
+
+- from:
```
koshachy, line 760: @zoobestik shouldn't these be implemented via "accept-web-file-names" in kr.tree?

koshachy (approved):
Works for me.
Please, request the Kotlin web team review.

redirects/pages.yml
zoobestik: IMO, there is not a big difference between both ways.
If it's difficult to read `kr.tree`, it's okay use `redirects/*.yml`, up to you.

PR #3913, opened 2023-11-22
Update inline-classes.md

filograno:
This PR slightly refine the intro about the Inline classes.


sarahhaggarty:
Hi @filograno 
Thanks so much for your suggestion! You inspired me to take this even further. So after some discussion with our developer advocates, I updated the introductory sentence as part of #3919. As a result, I'll close this PR. I hope you'll contribute again in future!

PR #3917, opened 2023-11-23
fix: remove redundant text in nested block comment

LightQuanta:


danil-pavlov (approved):
Good catch, thank you!

PR #3920, opened 2023-11-27
update to boolean and character sizes in memory and about mangling

daniCsorbaJB:
minor updates to boolean and char docs about their respective sizes in JVM and binary incompatibility of changing type from unsigned to its signed counterpart

### DIFF
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -1,7 +1,10 @@
 [//]: # (title: Booleans)
 
 The type `Boolean` represents boolean objects that can have two values: `true` and `false`. 
-On the JVM, the primitive `boolean` type typically uses 8 bits.
+
+> On the JVM, the primitive `boolean` type typically uses 8 bits.
+>
+{type="note"}
 
 `Boolean` has a nullable counterpart `Boolean?` that also has the `null` value.
 
@@ -28,6 +31,6 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> **On the JVM**: nullable references to boolean objects are boxed similarly to [numbers](numbers.md#numbers-representation-on-the-jvm).
+> On the JVM, nullable references to boolean objects are boxed in Java classes, just like with [numbers](numbers.md#numbers-representation-on-the-jvm).
 >
 {type="note"}
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -2,7 +2,7 @@
 
 The type `Boolean` represents boolean objects that can have two values: `true` and `false`. 
 
-> On the JVM, the primitive `boolean` type typically uses 8 bits.
+> On the JVM, booleans stored as primitive type: `boolean`, typically use 8 bits.
 >
 {type="note"}
 
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -1,6 +1,7 @@
 [//]: # (title: Booleans)
 
-The type `Boolean` represents boolean objects that can have two values: `true` and `false`.
+The type `Boolean` represents boolean objects that can have two values: `true` and `false`. 
+On the JVM, the primitive `boolean` type typically uses 8 bits.
 
 `Boolean` has a nullable counterpart `Boolean?` that also has the `null` value.
 
@@ -27,6 +28,6 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> **On JVM**: nullable references to boolean objects are boxed similarly to [numbers](numbers.md#numbers-representation-on-the-jvm).
+> **On the JVM**: nullable references to boolean objects are boxed similarly to [numbers](numbers.md#numbers-representation-on-the-jvm).
 >
 {type="note"}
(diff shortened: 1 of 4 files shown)


docs/topics/unsigned-integer-types.md
```diff
@@ -2,16 +2,22 @@
...
-* `UByte`: an unsigned 8-bit integer, ranges from 0 to 255
-* `UShort`: an unsigned 16-bit integer, ranges from 0 to 65535
-* `UInt`: an unsigned 32-bit integer, ranges from 0 to 2^32 - 1
-* `ULong`: an unsigned 64-bit integer, ranges from 0 to 2^64 - 1
+| Type     | Size (bits) | Min value | Max value                                       |
+|----------|-------------|-----------|-------------------------------------------------|
+| `UByte`  | 8           | 0         | 255                                             |
+| `UShort` | 16          | 0         | 65,535                                          |
+| `UInt`   | 32          | 0         | 4,294,967,295 (2<sup>32</sup> - 1)              |
+| `ULong`  | 64          | 0         | 18,446,744,073,709,551,615 (2<sup>64</sup> - 1) |
```
sarahhaggarty, line 10:
```suggestion
| `UInt`   | 32          | 0         | 2<sup>32</sup> - 1 |
| `ULong`  | 64          | 0         | 2<sup>64</sup> - 1 |
```
I think it looks better without explicitly writing out these numbers.

docs/topics/unsigned-integer-types.md
```diff
@@ -2,16 +2,22 @@
...
+
 
 Unsigned types support most of the operations of their signed counterparts.
 
 > Unsigned numbers are implemented as [inline classes](inline-classes.md) with the single storage property of the corresponding
```
sarahhaggarty, line 15:
```suggestion
> Unsigned numbers are implemented as [inline classes](inline-classes.md) with a single storage property of the corresponding 
```

docs/topics/unsigned-integer-types.md
```diff
@@ -2,16 +2,22 @@
...
 Unsigned types support most of the operations of their signed counterparts.
 
 > Unsigned numbers are implemented as [inline classes](inline-classes.md) with the single storage property of the corresponding 
-> signed counterpart type of the same width. Nevertheless, changing type from unsigned type to signed counterpart (and vice versa) 
-> is a _binary incompatible_ change.
+> signed counterpart type of the same width. However, due to the compiler's mangling of signatures containing unsigned types, 
+> changing the type from unsigned to its signed counterpart or the other way around 
+> is a _binary incompatible_ change. 
+> 
+> For more information, see [Mangling](inline-classes.md#mangling).
```
sarahhaggarty, line 20: It's not clear to me how mangling makes conversion between signed and unsigned integer types binary incompatible. Can you explain it a bit more?

Also, it reads like mangling only occurs for unsigned types. Is this correct?

docs/topics/unsigned-integer-types.md
```diff
@@ -22,19 +28,19 @@ Unsigned types support most of the operations of their signed counterparts.
...
+* `UShortArray`: an array of unsigned shorts.
+* `UIntArray`: an array of unsigned ints.
+* `ULongArray`: an array of unsigned longs.
 
-Same as for signed integer arrays, they provide similar API to `Array` class without boxing overhead.
+Same as for signed integer arrays, they provide a similar API to the `Array` class without boxing overhead.
 
-When you use unsigned arrays, you'll get a warning that indicates that this feature is not stable yet.
-To remove the warning, opt-in the `@ExperimentalUnsignedTypes` annotation.
+When you use unsigned arrays, you will receive a warning that indicates that this feature is not stable yet.
```
sarahhaggarty, line 40:
```suggestion
When you use unsigned arrays, you receive a warning that indicates that this feature is not stable yet.
```
I see you're expanding the contraction but let's go one step further and get rid of the future tense entirely 😄

sarahhaggarty (changes requested):
Nice start! I have some suggestions and questions for you 🤔

docs/topics/strings.md
```diff
@@ -1,6 +1,11 @@
 [//]: # (title: Strings)
 
-Strings in Kotlin are represented by the type [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/).
+Strings in Kotlin are represented by the type [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/). 
+
+> On the JVM, a `String` in UTF-16 encoding uses approximately 2 bytes per character.
```
sarahhaggarty, line 5:
```suggestion
> On the JVM, a string in UTF-16 encoding uses approximately 2 bytes per character.
```
Same comment as I wrote for the Characters page.

docs/topics/booleans.md
```diff
@@ -1,6 +1,10 @@
 [//]: # (title: Booleans)
 
-The type `Boolean` represents boolean objects that can have two values: `true` and `false`.
+The type `Boolean` represents boolean objects that can have two values: `true` and `false`. 
+
+> On the JVM, the primitive `boolean` type typically uses 8 bits.
```
sarahhaggarty, line 5: I noticed that we don't mention primitives before this. Can you introduce the `boolean` type before this? Or do we know the size of `Boolean` objects that we can use instead?

sarahhaggarty (changes requested):
Almost there!

sarahhaggarty (approved):
LGTM! 🚀

PR #3935, opened 2023-12-08
KTL-1232: Set up workflow to update kotlin news

zoobestik:


scripts/latest-news/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "fast-xml-parser": "^4.3.2"
+  }
+}
```
nikpachoo, line 5: I doubt that we need a separate package.json. I recall we tried to use a separate package file and removed it.
But on the other hand, it speeds up the build time. @krutilov WDYT?

nikpachoo (changes requested):

PR #3945, opened 2023-12-15
fix: clarity in the tip for the specific source set

zamulla:
I was trying to understand what exactly I was supposed to do: don't use the name like it's used in the example, or don't use the name like it's not used in the example.

I think I get it now, so I tried to clarify this.

### DIFF
--- a/docs/topics/multiplatform/multiplatform-add-dependencies.md
+++ b/docs/topics/multiplatform/multiplatform-add-dependencies.md
@@ -205,7 +205,7 @@ kotlin {
 If you want to use a multiplatform library just for specific source sets, you can add it exclusively to them. The
 specified library declarations will then be available only in those source sets.
 
-> Don't use a platform-specific name in such cases, like SQLDelight `native-driver` in the example below. Find the exact name in the library's documentation.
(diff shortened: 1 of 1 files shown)


docs/topics/multiplatform/multiplatform-add-dependencies.md
```diff
@@ -205,7 +205,7 @@ kotlin {
 If you want to use a multiplatform library just for specific source sets, you can add it exclusively to them. The
 specified library declarations will then be available only in those source sets.
 
-> Don't use a platform-specific name in such cases, like SQLDelight `native-driver` in the example below. Find the exact name in the library's documentation.
+> Use a general library name in such cases, not a platform-specific one. Like with SQLDelight in the example below, use `native-driver`, not `native-driver-iosx64` or other iOS versions. Find the exact name in the library's documentation.
```
danil-pavlov, line 208:
```suggestion
> Use a common library name in such cases, not a platform-specific one. Like with SQLDelight in the example below, use `native-driver`, not `native-driver-iosx64`. Find the exact name in the library's documentation.
```

danil-pavlov (changes requested):

danil-pavlov (approved):

PR #3947, opened 2023-12-15
update: rewrite the outdated section on avoiding duplicate publications

zamulla:
The configuration offered is unnecessary since it solves a problem from back when Kotlin/Native couldn't handle all the cross-compilation.


docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -4,6 +4,10 @@ You can publish a multiplatform library to a local Maven repository with the [`m
 Specify the group, version, and the [repositories](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:repositories) 
 where the library should be published. The plugin creates publications automatically.
 
+> Here and further down the page all examples show the `shared/build.gradle.kts` file, not the root project configuration.
```
danil-pavlov, line 7: I think it would be better to repeat this in every section with a code sample  (note as a note, but as an action), for example:
> In the `shared/build.gradle.kts` file, specify the group, version..

docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -42,73 +46,53 @@ This `kotlinMultiplatform` publication includes metadata artifacts and reference
...
-## Avoid duplicate publications
+## Host requirements
 
-To avoid duplicate publications of modules that can be built on several platforms (like JVM and JS), 
-configure the publishing tasks for these modules to run conditionally.
+Except for Apple platform targets, Kotlin/Native supports cross-compilation, allowing any host to produce needed artifacts.
 
-You can detect the platform in the script, introduce a flag such as `isMainHost` and set it to `true` for the main target 
-platform. Alternatively, you can pass the flag from an external source, for example, from CI configuration. 
+So, to avoid any issues:
```
danil-pavlov, line 58:
```suggestion
To avoid any issues during publication:
```

docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -42,73 +46,53 @@ This `kotlinMultiplatform` publication includes metadata artifacts and reference
...
+* Publish only from an Apple host when your project targets Apple operating systems.
+* Never publish from multiple hosts to avoid duplicating publications at the repo.
```
danil-pavlov, line 60:
```suggestion
* Publish all artifacts from one host only to avoid duplicating publications in the repository.
```

docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -42,73 +46,53 @@ This `kotlinMultiplatform` publication includes metadata artifacts and reference
...
+  
+  Maven Central, for example, explicitly forbids duplicate publications and fails the process. <!-- TBD: add the actual error -->
+  
+### Problems solved since Kotlin 1.9.20 {initial-collapse-state="collapsed"}
```
danil-pavlov, line 64: I think this whole section could be rephrased to be more actionable, rather than a historic reference, like this:

### If you use Kotlin 1.9.0 and earlier

Before Kotlin 1.9.20, the Kotlin/Native compiler didn't support all the cross-compilation options. If you use these earlier versions, modules in multiplatform projects can be published more than once, resulting in duplicate publications in the repository.

...

To avoid that, assign a main host for each target and check for it like this:

...

We recommend upgrading to the latest version of Kotlin and using the current approach:..

---
Also, the indentation is off in the code block

Did you intentionally removed the Groovy option?

danil-pavlov (changes requested):

docs/topics/multiplatform/multiplatform-publish-lib.md
zamulla: Actionable is good indeed. Rewrote stuff. Also, "1.9.20" was fine to signify something _from before_; to rewrite as you suggested, I asked for a more precise version number.

I did remove the Groovy option because it seemed unnecessary as an example of what not to do, but in this revision, it's back again.

danil-pavlov (approved):
Perfect, thank you!

PR #3950, opened 2023-12-18
KT-64374 rewriting section to increase clarity

daniCsorbaJB:
updating docs to clarify default behavior of data class methods with class body properties

### DIFF
--- a/docs/topics/data-classes.md
+++ b/docs/topics/data-classes.md
@@ -1,7 +1,6 @@
 [//]: # (title: Data classes)
 
-Data classes in Kotlin are primarily used to hold data. Data classes come automatically with additional
-member functions that allow you to print an instance to readable output, compare instances, copy instances, and more.
+Data classes in Kotlin are primarily used to hold data. For each data class, the compiler automatically generates additional member functions that allow you to print an instance to readable output, compare instances, copy instances, and more.
 Data classes are marked with `data`:
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


docs/topics/data-classes.md
```diff
@@ -53,11 +53,9 @@ data class Person(val name: String) {
 }
 ```
 
-In this example, only the `name` property can be used inside the `.toString()`, `.equals()`, `.hashCode()`, and `.copy()` implementations,
-and there is only one component function `.component1()`. The `age` property can't be used inside the `.toString()`, 
-`.equals()`, `.hashCode()`, and `.copy()` implementations because it's declared inside the class body. If two `Person` 
-objects have different ages but the same `name`, then they are treated as equal. This is because the `.equals()` function
-can only check for equality of the `name` property. For example:
+In this example, only the `name` property is used by default inside the `.toString()`, `.equals()`, `.hashCode()`, and `.copy()` implementations,
+and there is only one component function, `.component1()`. The `age` property, declared inside the class body, is not included in these automatically generated functions by default.
```
zamulla, line 57: I feel that this passage could be even more straightforward, like "The `age` property is declared inside the class body and is excluded." Because we only need to come back to the statement at the beginning ("To exclude a property..."). And the following example makes it all clear, so let's get to it faster. What do you think?

docs/topics/data-classes.md
```diff
@@ -1,6 +1,6 @@
 [//]: # (title: Data classes)
 
-Data classes in Kotlin are classes whose main purpose is to hold data. Data classes come automatically with additional
+Data classes in Kotlin are primarily used to hold data. Data classes come automatically with additional
```
zamulla, line 3: Your edit is fine, but I can't help being amused with the phrase "data classes come automatically" ) it could be improved with an actor, like: "For each data class, compiler automatically generates additional member functions..."

zamulla (commented):
As long as we're here, a couple of suggestions.

zamulla (changes requested):
A couple of suggestions to consider.

docs/topics/data-classes.md
daniCsorbaJB: Thank you. :) I agree, it does sound a bit weird, and I like this change. I'm implementing it.

docs/topics/data-classes.md
daniCsorbaJB: I agree. I made it a bit more concise in accordance with your comment. My original intention was to spell out everything in a very obvious way to ensure that we can clear up any confusion that originated the ticket. But I like the change - thank you. 👍  :)

zamulla (approved):

PR #3951, opened 2023-12-18
update kotlin libraries

devcrocod:
* changed the order of Kotlin libraries by importance
* the [krangl](https://github.com/holgerbrandl/krangl) library was removed because it was deprecated 
* fixed typo


koshachy (approved):
Thanks for your contribution!

PR #3965, opened 2024-01-09
fix: clear up the example of usefulness of invariant generics

zamulla:
The comments in the example are not sufficient to explain the train of thought preceding it.

### DIFF
--- a/docs/topics/generics.md
+++ b/docs/topics/generics.md
@@ -26,6 +26,9 @@ val box = Box(1) // 1 has type Int, so the compiler figures out that it is Box<I
 One of the trickiest aspects of Java's type system is the wildcard types (see [Java Generics FAQ](http://www.angelikalanger.com/GenericsFAQ/JavaGenericsFAQ.html)).
 Kotlin doesn't have these. Instead, Kotlin has declaration-site variance and type projections.
 
+
+### Variance and wildcards in Java
+
 Let's think about why Java needs these mysterious wildcards. The problem is explained well in 
 [Effective Java, 3rd Edition](http://www.oracle.com/technetwork/java/effectivejava-136174.html), 
 Item 31: _Use bounded wildcards to increase API flexibility_.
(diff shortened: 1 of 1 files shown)


zarechenskiy (approved):

docs/topics/generics.md
```diff
@@ -36,12 +39,12 @@ compiled but caused an exception at runtime:
 ```java
 // Java
 List<String> strs = new ArrayList<String>();
-List<Object> objs = strs; // !!! A compile-time error here saves us from a runtime exception later.
-objs.add(1); // Put an Integer into a list of Strings
-String s = strs.get(0); // !!! ClassCastException: Cannot cast Integer to String
+List<Object> objs = strs; // Java reports a type mismatch here at compile-time, but what if it didn't?
+objs.add(1); // We would be able to put an Integer into a list of Strings
+String s = strs.get(0); // And then at runtime, Java would throw a ClassCastException: Integer cannot be cast to String
```
sarahhaggarty, line 44: Can you wrap your code comments so that the reader doesn't have to scroll to see the end of the comment?

docs/topics/generics.md
```diff
@@ -93,18 +97,15 @@ you don't get a `String`, but rather an `Object`.
...
 
->"For maximum flexibility, use wildcard types on input parameters that represent producers or consumers",
-> and proposes the following mnemonic:
->
->_PECS stands for Producer-Extends, Consumer-Super._
->
-{type="tip"}
+>"For maximum flexibility, use wildcard types on input parameters that represent producers or consumers,"
+
+and proposes the following mnemonic: _PECS_ stands for _Producer-Extends, Consumer-Super._
```
sarahhaggarty, line 102:
```suggestion
>"For maximum flexibility, use wildcard types on input parameters that represent producers or consumers..."

He proposes the following mnemonic: _PECS_ stands for _Producer-Extends, Consumer-Super._
```
This formatting looks strange when you preview it so I've suggested to break it into two sentences.

sarahhaggarty (commented):
Looks good! Just a couple of minor formatting comments.

sarahhaggarty (approved):
LGTM! 🚀

PR #3976, opened 2024-01-17
ktl-1244-404-page

krutilov:


blocks/404/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const NotFoundContent: FC = () => {
+    const textCn = useTextStyles();
+
+    return (
+        <div className={styles.wrapper}>
+            <img className={styles.image} src={Mascot404.src} alt="404"/>
+            <h1 className={textCn('rs-h1')}>Page not found</h1>
+            <p className={textCn('rs-text-2', { hardness: 'hard' })}>
+                Please use search or try<br className={styles.lineBreak} /> starting from{' '}
+                <a className={textCn('rs-link', { hardness: 'hard', mode: 'rock' })} href="https://kotlinlang.org/">
```
nikpachoo, line 18: I'd like to suggest the following change:
```suggestion
                <a className={textCn('rs-link', { hardness: 'hard', mode: 'rock' })} href="/">
```

pages/404.tsx
```diff
@@ -0,0 +1,52 @@
...
+function NotFoundPage() {
+    const isTS = useTS();
+
+    return (
+        <Layout
+            title={'Kotlin Programming Language'}
+            ogImageName={'general.png'}
+            description={
+                'Kotlin is a programming language that makes coding concise, cross-platform, and fun. It is Google’s preferred language for Android app development.'
+            }
```
nikpachoo, line 32: Do we use the same values for the main page? In that case, we can put them as default values and not define them for each page.

blocks/404/404.module.css
```diff
@@ -0,0 +1,26 @@
+@import "@jetbrains/kotlin-web-site-ui/out/components/breakpoints/media.pcss";
+
+.wrapper {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    padding: 48px 0;
```
zoobestik, line 7: https://github.com/JetBrains/kotlin-web-site-ui/blob/master/src/vars.css#L23-L34

PR #3981, opened 2024-01-19
Remove "Decode inlined stack frames" from native-ios-symbolication.md

SvyatoslavScherbina:
Remove "Decode inlined stack frames" section from "Symbolicating iOS crash reports".

This section seems obsolete. The command no longer works -- it has to be "command script import lldb.macosx" instead of "script import ...".

But even with this change, it is not of much use: it can't find .dSYM bundles for crashlogs copied from iPhone. Some additional manipulations are required to make this work.

Considering that this part is not even Kotlin-specific and there are many tutorials on that topic on the internet, it seems justified to remove the problematic section instead of expanding it.

### DIFF
--- a/docs/topics/native/native-ios-symbolication.md
+++ b/docs/topics/native/native-ios-symbolication.md
@@ -88,20 +88,3 @@ kotlin {
 
 </tab>
 </tabs>
-
-## Decode inlined stack frames
-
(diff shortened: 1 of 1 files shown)


koshachy (approved):
Thanks for pointing out the problem.

PR #3983, opened 2024-01-20
Fix crash on readln()

LukasVykuka:
Add solution for crash on `realn()`.


LukasVykuka:
Fixing in template project.

PR #3988, opened 2024-01-23
Main mascot

zoobestik:


blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+        let cancelable: (body: () => Promise<void>) => ReturnType<typeof body> = (body) => body();
+
+        async function playAnimation() {
+            const [lottie, initialData] = await Promise.all([
+                import('lottie-web/build/player/lottie_svg').then((l) => l.default),
+                import('./option3.json'),
+                sleep(ANIMATION_INITIAL_DELAY),
+            ]);
+
+            await cancelable(async () => {
```
zoobestik, line 76: Every wrapped for `cancelable`. Main idea is in two cases:
 * you can scroll infinity times it'll wait until 1s idle
 * skip second animation loading data and playing if you scroll out.

But... may be second case is discussable. Without this case it can be doable simpler with `setTimeout`.
...Or may be you know another better ways how it's possible to cancel a promise.

blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return (
+        <div ref={inViewRef} className={cn(styles.container, className)}>
+            <span ref={mascotNode} className={styles.animation} />
+        </div>
+    );
+}
+
+export default memo(function Mascot(props: MascotProps) {
+    const [isComplete, setComplete] = useState(false);
+    return isComplete ? null : <MascotContent {...props} onFinish={() => setComplete(true)} />;
```
zoobestik, line 110: Drop `useInView` events when it done.

.idea/misc.xml
```diff
@@ -1,7 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
   <component name="JavaScriptSettings">
```
nikpachoo, line 4: Well, maybe we should exclude .idea directory?

blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { AnimationItem, LottiePlayer } from 'lottie-web/build/player/lottie_svg';
+
+import styles from './mascot.module.css';
+
+type MascotProps = {
+    className?: string;
+};
+
+const ANIMATION_INITIAL_DELAY = 1500 as const;
+const ANIMATION_AFTER_DELAY = 7000 as const;
```
nikpachoo, line 13: "as const" - could you please explain why we need it here? I did't expect to see it with const declarations.

package.json
```diff
@@ -104,6 +106,9 @@
     "webpack-dev-server": "4.3.1",
     "yaml-loader": "0.7.0"
   },
+  "resolutions": {
+    "sharp": "^0.28.3"
+  },
```
nikpachoo, line 111: thanks for fixing this issue

nikpachoo (approved):
I don't have a better solution to the questions raised. Maybe @krutilov or @berezinant can offer something.

blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    useEffect(() => {
+        const node = mascotNode.current;
+
+        let animation: AnimationItem;
+        let play: ReturnType<typeof createAnimation>[1];
+
+        let isOnceStarted = false;
+
+        function done() {
+            if (isOnceStarted) onFinish();
```
berezinant, line 64: Not really about this particular tasks but what do you think about adding `"curly": "all"` to eslint to enforce consistent brace style?

blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        let cancelable: (body: () => Promise<void>) => ReturnType<typeof body> = (body) => body();
+
+        async function playAnimation() {
+            const [lottie, initialData] = await Promise.all([
```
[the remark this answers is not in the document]
berezinant: Given there is a start delay for animation by design this does not seem like an issue

blocks/main/latest-news/mascot/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const mascotNode = useRef<HTMLSpanElement>(null);
+    const { ref: inViewRef, inView } = useInView();
+
+    useEffect(() => {
+        const node = mascotNode.current;
+
+        let animation: AnimationItem;
+        let play: ReturnType<typeof createAnimation>[1];
+
+        let isOnceStarted = false;
```
berezinant, line 61: nit: any chance that
```suggestion
        let wasOnceStarted = false;
```
sounds better?

blocks/main/latest-news/mascot/index.tsx
berezinant: If I understood correctly it is more like a `skippable` rather than `cancelable`?
I've tripped a bit here because I was looking for a promise to be aborted or something but looks like this is not what is going on in reality

berezinant (approved):

blocks/main/latest-news/mascot/index.tsx
zoobestik: Thanks, it's a goooooooood question! :)

Mostly it has some useful cases for *complex type* like `Record` or `Array`.
And I know just single so synthetic case for *primitive type*:
```javascript
const OTHER_CONST = 123;

const TIMEOUT_1 = 1000 as const;
const TIMEOUT_2 = 2000 as const;

type TIMEOUTS = typeof TIMEOUT_1 | typeof TIMEOUT_2; // i've invented `enum`!

function print(n: number) {
  console.log(n);
}

function timeout(n: TIMEOUTS) {
  console.log(n);
}

print(OTHER_CONST);
print(TIMEOUT_1);
print(TIMEOUT_2);

timeout(OTHER_CONST); // oops...
timeout(TIMEOUT_1);
timeout(TIMEOUT_2);
```
But of course, it isn't there.
I just use `as const` for all `ANY_CONST` without mind cause i don't know any side effects.
I'll drop it if it's clear.

nikpachoo: Ah, I see now. Didn't now about "const assertions" in TS. [For the reference](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions).

nikpachoo (approved):

PR #3989, opened 2024-01-24
Kt 56387 sealed classes improvement

daniCsorbaJB:
Introducing sealed classes and interfaces improvements


docs/topics/sealed-classes.md
```diff
@@ -1,22 +1,26 @@
 [//]: # (title: Sealed classes and interfaces)
 
 _Sealed_ classes and interfaces represent restricted class hierarchies that provide more control over inheritance.
+Sealed classes and interfaces, in contrast to `internal` classes, ensure inherent exhaustiveness in `when` expressions within their declared module or package.
```
serras, line 4: I don't think this is in contrast to `internal` classes, but to `open` classes.

```suggestion
Sealed classes and interfaces, in contrast to `open` ones (which is the default), ensure inherent exhaustiveness in `when` expressions.
```

docs/topics/sealed-classes.md
```diff
@@ -1,22 +1,26 @@
...
 All direct subclasses of a sealed class are known at compile time. No other subclasses may appear outside
-the module and package within which the sealed class is defined. For example, third-party clients can't extend your sealed class in their code.
-Thus, each instance of a sealed class has a type from a limited set that is known when this class is compiled.
+the module and package within which the sealed class is defined.
```
serras, line 6: Actually, I'm not sure about having to be in the same package (but definitely they are all in the same module).

docs/topics/sealed-classes.md
```diff
@@ -31,7 +35,42 @@ class DatabaseError(val source: DataSource): IOError()
 object RuntimeError : Error
 ```
 
-A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
+The hierarchy of our sealed interface and sealed class example looks like this:
+![Hierarchy Illustration of Sealed Classes and Interfaces](sealedClasses_interfaces.png){width=600}
+
+A sealed class itself is an [abstract](classes.md#abstract-classes) and cannot be instantiated directly.
```
serras, line 41:
```suggestion
A sealed class itself is always [abstract](classes.md#abstract-classes) and, as a result, cannot be instantiated directly. 
```

docs/topics/sealed-classes.md
```diff
@@ -31,7 +35,42 @@ class DatabaseError(val source: DataSource): IOError()
...
+fun main() {
+    val errors = listOf(Error.NetworkError(), Error.DatabaseError(), Error.UnknownError())
+    errors.forEach { println(it.message) }
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.5"}
+
+In some sense, sealed classes are similar to [`enum`](enum-classes.md) classes: the set of values
+for an enum type is also restricted, but each enum constant exists only as a _single instance_, whereas a subclass
+of a sealed class can have _multiple_ instances, each with its own state.
```
serras, line 61:
```suggestion
Sealed classes share with [`enum`](enum-classes.md) classes a restriction over the set of values.
Note that each enum constant exist only as a _single instance_,
whereas subclasses of a sealed class may have _multiple_ instance, each with its own state.
```

docs/topics/sealed-classes.md
```diff
@@ -80,17 +119,39 @@ you can create subclasses in any source set between the `expect` and `actual` de
...
+expression.
+The `when` expression, used with a sealed class, allows the Kotlin compiler to check exhaustively that all possible cases are covered. In such cases, you don't need to add an `else` clause. 
+Here's an example demonstrating this with a `when` expression:
 
 ```kotlin
+// Sealed class and its subclasses
+sealed class Error
+class FileReadError(val file: String): Error()
+class DatabaseError(val source: String): Error()
+object RuntimeError : Error()
```
serras, line 131: You are switching styles here: before you defined things as nested classes. Not super-important, but may puzzle people who are finding this notion for the first time.

docs/topics/sealed-classes.md
```diff
@@ -80,17 +119,39 @@ you can create subclasses in any source set between the `expect` and `actual` de
...
+//sampleStart
+// Function to log errors
 fun log(e: Error) = when(e) {
-    is FileReadError -> { println("Error while reading file ${e.file}") }
-    is DatabaseError -> { println("Error while reading from database ${e.source}") }
-    is RuntimeError ->  { println("Runtime error") }
-    // the `else` clause is not required because all the cases are covered
+    is FileReadError -> println("Error while reading file ${e.file}")
+    is DatabaseError -> println("Error while reading from database ${e.source}")
+    is RuntimeError -> println("Runtime error")
```
serras, line 138: By convention you use only the name of the object

```suggestion
    RuntimeError -> println("Runtime error")
```

daniCsorbaJB: thank you! - very nice feedback - I'll implement it right away

docs/topics/sealed-classes.md
daniCsorbaJB: True! - I agree - I'll make this one nested as well to stay consistent

serras (approved):

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+    errors.forEach { log(it) }
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.5"}
 
-> `when` expressions on [`expect`](multiplatform-expect-actual.md) sealed classes in the common code of multiplatform projects still 
-> require an `else` branch. This happens because subclasses of `actual` platform implementations aren't known in the 
-> common code.
+> In multiplatform projects, if you have a sealed class with a `when` expression as an 
+> [expected declarations](multiplatform-expect-actual.md) in your common code, you still need an `else` branch.
```
sarahhaggarty, line 203:
```suggestion
> [expected declaration](multiplatform-expect-actual.md) in your common code, you still need an `else` branch. 
```

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
 object RuntimeError : Error
 ```
 
-A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
+This example could represent a library's API that contains error classes to let library users handle errors that it can throw. 
+If the hierarchy of such error classes includes interfaces or abstract classes visible in the public API, then nothing 
+prevents other developers from implementing or extending them in the client code. 
+Since the library doesn't know about errors declared outside of it, it can’t treat them consistently with its own classes. 
+However, with a **sealed** hierarchy of error classes, library authors can be sure that they know all the possible error 
+types and that other ones can’t appear later.
```
sarahhaggarty, line 54:
```suggestion
types and that other error types can't appear later.
```

sarahhaggarty (approved):
This is a huge improvement! Thanks for your hard work. I found just a couple copy-paste errors. Please double-check to replace any apostrophes with single quotes.

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+> In multiplatform projects, if you have a sealed class with a `when` expression as an 
+> [expected declaration](multiplatform-expect-actual.md) in your common code, you still need an `else` branch. 
+> This is because subclasses of `actual` platform implementations may extend sealed classes that 
+> aren't known in the common code.
 >
 {type="note"}
+
+## Use case scenarios
+
+Let's explore some practical scenarios where the application of sealed classes or interfaces can be particularly helpful.
```
koshachy, line 211: Let's simplify the second part of the sentence a bit:

"... where sealed classes and interfaces can be particularty helpful/useful"

Does it work?

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+
+### State management in UI applications
+
+You can use sealed classes to represent different states of a UI in an application.
```
koshachy, line 215: different states of UI -> different/vaious UI states

WDYT?

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+
+## Use case scenarios
+
+Let's explore some practical scenarios where the application of sealed classes or interfaces can be particularly helpful.
+
+### State management in UI applications
+
+You can use sealed classes to represent different states of a UI in an application. 
+This approach allows for structured and safe handling of UI changes. 
+This example demonstrates how to manage various UI states:
```
koshachy, line 217: Do we need this sentence or it could be omited?

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+    	is UIState.Success -> showData(state.data)
+    	is UIState.Error -> showError(state.exception)
+	}
+}
+```
+
+### Payment method handling
+
+In practical business applications, handling various payment methods efficiently is a common requirement. 
+This example illustrates how to implement such business logic using sealed classes.
```
koshachy, line 238: This example illustrates how to implement such business logic using sealed classes. ->

You can use sealed classes with `when` expressions to implement such business logic.

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+fun processPayment(payment: Payment) {
+	when (payment) {
+    	is Payment.CreditCard -> processCreditCardPayment(payment.number, payment.expiryDate)
+    	is Payment.PayPal -> processPayPalPayment(payment.email)
+    	is Payment.Cash -> processCashPayment()
+	}
+}
+```
+
+`Payment` is a sealed class that represents different payment methods in an e-commerce system:
```
koshachy, line 258: Maybe organize these senteces into a list?

* `Payment` ...
* `processPayment()` ...

WDYT?

koshachy (changes requested):
Nice job!
Please, see the comments. Feel free to start a discussion.

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
-A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
+This example could represent a library's API that contains error classes to let library users handle errors that it can throw. 
+If the hierarchy of such error classes includes interfaces or abstract classes visible in the public API, then nothing 
+prevents other developers from implementing or extending them in the client code. 
+Since the library doesn't know about errors declared outside of it, it can’t treat them consistently with its own classes. 
+However, with a **sealed** hierarchy of error classes, library authors can be sure that they know all the possible error 
+types and that other error types can't appear later.
+
+The hierarchy of the example looks like this:
+![Hierarchy Illustration of Sealed Classes and Interfaces](sealedClasses_interfaces.svg){width=700}
```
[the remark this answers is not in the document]
daniCsorbaJB: as discussed - I'm changing the name to lowercase+dashes and creating a new folder for concepts

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
 [Learn more about the hierarchical structure of multiplatform projects](multiplatform-hierarchy.md). 
 
-## Sealed classes and when expression
+## Use sealed classes with when expression
 
 The key benefit of using sealed classes comes into play when you use them in a [`when`](control-flow.md#when-expression)
-expression. 
-If it's possible to verify that the statement covers all cases, you don't need to add an `else` clause to the statement:
+expression.
+The `when` expression, used with a sealed class, allows the Kotlin compiler to check exhaustively that all possible cases are covered.
```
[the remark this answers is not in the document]
daniCsorbaJB: 🤔 - That's a tough one. I like all 3 (the original included).
In my opinion:
1. The original is a bit wordy but conveys all the information. 
2. "The when expression with a sealed class..." It's more concise, but I feel it's also a bit more passive - it might imply that the cases are covered by default, rather than being actively checked by the compiler.
3. "If you use.." I think this one is a bit less direct in its structure, but a bit more active than the original.  

I don't think it's an issue being a bit more wordy in this case to belabor a point, because it's one of the most important functionalities of sealed classes.

docs/topics/sealed-classes.md
daniCsorbaJB: Yes, I think it sounds much better! - I'll update it

docs/topics/sealed-classes.md
daniCsorbaJB: I don't think it's necessary, somewhat eases the transition from explanation to example, but I don't think it's necessary.

docs/topics/sealed-classes.md
daniCsorbaJB: I really like this suggestion - more active and puts more emphasis on the when expression. - I'll update it

docs/topics/sealed-classes.md
daniCsorbaJB: hmm 🤔 I think adding a line break between the processPayment() sentence and the previous ones would aid with readability a bit more. 

(Converting this to a list would look something like this (I think it looks a bit awkward):

* `Payment`: a sealed class that represents different payment methods in an e-commerce system:
  * ˙CreditCard˙
  * ˙Paypal˙
  * ˙Cash˙
* ˙processPayment()˙: a function that demonstrates how to handle different payment methods.

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,350 @@
...
+
+`Payment` is a sealed class that represents different payment methods in an e-commerce system: 
+`CreditCard`, `PayPal`, and `Cash`. Each subclass can have its specific properties, like `number` and `expiryDate` for 
+`CreditCard`, and `email` for `PayPal`. The `processPayment()` function demonstrates how to handle different payment methods.
+This approach ensures that all possible payment types are considered, and the system remains flexible for new payment 
+methods to be added in the future.
+
+### API request-response handling
+
+This example explores a practical implementation of a user authentication system that handles API requests and responses
```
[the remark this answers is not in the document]
daniCsorbaJB: I like it it's a bit more active - I'll update it (will also add info about sealed interfaces)

docs/topics/sealed-classes.md
koshachy: WFM!

docs/topics/sealed-classes.md
koshachy: Up to you.

docs/topics/sealed-classes.md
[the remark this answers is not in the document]
koshachy: Fair enough. Up to you.

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,352 @@
...
+> ```
+> 
 {type="note"}
 
 These restrictions don't apply to indirect subclasses. If a direct subclass of a sealed class is not marked as sealed,
 it can be extended in any way that its modifiers allow:
 
 ```kotlin
-sealed interface Error // has implementations only in same package and module
+// Sealed interface 'Error'has implementations only in the same package and module
```
koshachy, line 143:
```suggestion
// Sealed interface 'Error' has implementations only in the same package and module
```

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,352 @@
...
-A sealed class is [abstract](classes.md#abstract-classes) by itself, it cannot be instantiated directly and can have `abstract` members.
+This example could represent a library's API that contains error classes to let library users handle errors that it can throw. 
+If the hierarchy of such error classes includes interfaces or abstract classes visible in the public API, then nothing 
+prevents other developers from implementing or extending them in the client code. 
+Since the library doesn't know about errors declared outside of it, it can’t treat them consistently with its own classes. 
+However, with a **sealed** hierarchy of error classes, library authors can be sure that they know all the possible error 
+types and that other error types can't appear later.
+
+The hierarchy of the example looks like this:
+![Hierarchy Illustration of Sealed Classes and Interfaces](sealed-classes-interfaces.svg){width=700}
```
koshachy, line 57:
```suggestion
![Hierarchy illustration of sealed classes and interfaces](sealed-classes-interfaces.svg){width=700}
```

docs/topics/sealed-classes.md
```diff
@@ -1,99 +1,352 @@
...
-> `enum` classes can't extend a sealed class (as well as any other class), but they can implement sealed interfaces.
+> `enum` classes can't extend a sealed class, or any other class. However, they can implement sealed interfaces:
+>
+> ```kotlin
+> sealed interface Error
+> 
+> // enum class extending the sealed interface Error
+> enum class ErrorType : Error {
+>    FILE_ERROR, DATABASE_ERROR
+> }
```
koshachy, line 133:
```suggestion
> ```kotlin
> sealed interface Error
> 
> // enum class extending the sealed interface Error
> enum class ErrorType : Error {
>     FILE_ERROR, DATABASE_ERROR
> }
```

koshachy (approved):
Great job!
Some minor suggestions, please take a look.

PR #4006, opened 2024-01-31
Update K2 plugin support in whatsnew-eap

sarahhaggarty:
This PR tidies up the list of K2 compiler supported plugins by separating Kotlin compiler plugins from the rest and providing versions/information for non-Kotlin compiler plugins.


koshachy (approved):
LGTM!

PR #4008, opened 2024-02-01
update: adding test dependencies

danil-pavlov:


docs/topics/js/js-project-setup.md
```diff
@@ -159,21 +159,17 @@ these transitive dependencies as well.
 The dependencies on the [standard library](https://kotlinlang.org/api/latest/jvm/stdlib/index.html)
 are added automatically. The version of the standard library is the same as the version of the Kotlin Multiplatform plugin.
 
-The [`kotlin.test`](https://kotlinlang.org/api/latest/kotlin.test/) API is available for multiplatform tests.
-When you create a multiplatform project, the Project Wizard automatically adds test dependencies to all the source sets.
-
-If you don't use the Project Wizard to create your project, you can add the dependencies manually:
+For multiplatform tests, the [`kotlin.test`](https://kotlinlang.org/api/latest/kotlin.test/) API is available.
+When you create a multiplatform project, you can add test dependencies to all the source sets:
```
zamulla, line 163: The description seems to contradict the code: “add dependencies” although we only add one dependency. The ambiguity is resolved by the comment, but I'd say better not to have it at all and reinforce the message with the comment.

I'd expand with something like “...you can add test dependencies to all the source sets by adding a single dependency in `commonTest`.”

docs/topics/js/js-running-tests.md
```diff
@@ -3,8 +3,7 @@
 The Kotlin Multiplatform Gradle plugin lets you run tests through a variety of test runners that can be specified via the Gradle
 configuration.
 
-When you create a multiplatform project, the Project Wizard automatically adds test dependencies to all the source sets.
-If you created your project without it, you can add the dependencies manually to make test annotations
+When you create a multiplatform project, you can add test dependencies to all the source sets to make test annotations
```
zamulla, line 6: It's confusing here, too: "...you can add dependencies to all the source sets to make functionality available for the JavaScript." You can add a dependency to commonTest, and that's what makes JavaScript work, right?

zamulla (commented):
Overall, LGTM only one phrasing issue to discuss.

zamulla (approved):
Awesome, thanks!

PR #3982, opened 2024-01-19
equality.md: add a clarification to the section on structural equality

scolsen:
Because the structural equality operator `==` desugars to `equals`, there are some cases in which its underlying behavior will actually be *referential*. In fact, this is the default for non-data classes.

The name "structural equality", while it makes sense when considering data classes, may lead to confusion for this reason. I've added a note to clarify that, in spite of the name of the operator, there may be cases in which the underlying behavior of the comparison is not strictly structural in nature.

PR #3969, opened 2024-01-15
equality: Add note for Kotlin/JS

Hill-98:
I found that the documentation for equality doesn't take into account JS platform specificities, which might cause some confusion for beginners who don't know JS, so I added a note.

My English is not good, so if there is any grammar problem, please tell me how to change it.

PR #4009, opened 2024-02-01
update: equality page

AlejandraPedroza:
Update equality page. Incorporated external feedback and other minor changes.

### DIFF
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -67,6 +67,10 @@ class Point(val x: Int, val y: Int) {
     }
 }
 ```
+> When overriding the equals() function, you should also override the [hashCode() function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html#kotlin.Any$hashCode()) 
+> to keep consistency between equality and hashing and ensure a proper behavior of these functions.
+>
+{type="note"}
 
 Functions with the same name and other signatures (like `equals(other: Foo)`) don't affect equality checks with
 the operators `==` and `!=`.
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -1,9 +1,9 @@
 [//]: # (title: Equality)
 
-In Kotlin there are two types of equality:
+In Kotlin, there are two types of equality:
 
-* _Structural_ equality (`==` - a check for `equals()`)
-* _Referential_ equality (`===` - two references point to the same object)
+* _Structural_ equality (`==`) - a check for the `equals()` function
+* _Referential_ equality (`===`) - a check for two references pointing to the same object
 
 ## Structural equality
 
@@ -14,7 +14,7 @@ By convention, an expression like `a == b` is translated to:
 a?.equals(b) ?: (b === null)
 ```
 
-If `a` is not `null`, it calls the `equals(Any?)` function, otherwise (`a` is `null`) it checks that `b`
+If `a` is not `null`, it calls the `equals(Any?)` function. Otherwise (`a` is `null`), it checks that `b`
 is referentially equal to `null`.
 
 Note that there's no point in optimizing your code when comparing to `null` explicitly:
@@ -22,8 +22,15 @@ Note that there's no point in optimizing your code when comparing to `null` expl
 
 To provide a custom equals check implementation, override the
 [`equals(other: Any?): Boolean`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) function.
-Functions with the same name and other signatures, like `equals(other: Foo)`, don't affect equality checks with
+Functions with the same name and other signatures (like `equals(other: Foo)`) don't affect equality checks with
 the operators `==` and `!=`.
+
+While this equality operator (`==`) is called structural equality to distinguish it from the [referential equality](#referential-equality) (`===`),
+its underlying behavior could be referential in nature.
+
+One common scenario where structural equality can exhibit referential behavior is when handling non-data classes, 
+especially those without a custom implementation of the `equals()` function. In fact, this behavior is inherent by default
+in all classes except for [data classes](data-classes.md).
 
 Structural equality has nothing to do with comparison defined by the `Comparable<...>` interface, so only a custom 
 `equals(Any?)` implementation may affect the behavior of the operator. 
@@ -33,6 +40,24 @@ Structural equality has nothing to do with comparison defined by the `Comparable
 Referential equality is checked by the `===` operation and its negated counterpart `!==`. `a === b` evaluates to
 true if and only if `a` and `b` point to the same object. For values represented by primitive types at runtime
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
+
+> In Kotlin/JS, there is no difference between `===` and `==` operators when comparing strings. The reason is that in Kotlin/JS,
+> the `==` operator checks that two values are equal. Meanwhile, the `===` operator also checks that two values are equal 
+> besides verifying that the types of these two values are equal. In JS, Kotlin always uses the `===` operator:
+>
+> ```kotlin
+> fun main() {
+>     val name = "kotlin"
+>     val value1 = name.substring(0, 1)
+>     val value2 = name.substring(0, 1)
+>     
+>     println(if (value1 === value2) "yes" else "no")
+>     // Prints 'yes' in Kotlin/JS
+>     // Prints 'no' in other platforms
+> }
+> ```
+>
+{type="note"}
 
 ## Floating-point numbers equality
 
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -67,7 +67,7 @@ class Point(val x: Int, val y: Int) {
     }
 }
 ```
-> When overriding the equals() function, you should also override the [hashCode() function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html#kotlin.Any$hashCode()) 
+> When overriding the equals() function, you should also override the [hashCode() function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html) 
 > to keep consistency between equality and hashing and ensure a proper behavior of these functions.
 >
 {type="note"}
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -102,6 +102,10 @@ fun main() {
 For values represented by primitive types at runtime
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
 
+> For more information about equality in Kotlin/JS, see the [Kotlin/JS](js-interop.md) documentation.
+>
+{type="tip"}
+
 ## Floating-point numbers equality
 
 Equality checks for `Float` or `Double` operands (nullable or not) follow the 
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -105,7 +105,7 @@ fun main() {
 For values represented by primitive types at runtime
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
 
-> The referential equality is implemented differently in Kotlin/JS. For more information about equality in Kotlin/JS, see the [Kotlin/JS](js-interop.md#equality) documentation.
+> The referential equality is implemented differently in Kotlin/JS. For more information about equality, see the [Kotlin/JS](js-interop.md#equality) documentation.
 >
 {type="tip"}
 
--- a/docs/topics/equality.md
+++ b/docs/topics/equality.md
@@ -46,8 +46,8 @@ function to provide a custom equality logic and, in this way, implement structur
 Value classes and data classes are two specific types in Kotlin that automatically override `equals()`. 
 Value classes override the `equals()` function to implement structural equality by default.
 
-Similarly, data classes override the `equals()` function to implement structural equality by default. However, if a data class 
-is marked as `final` in the parent class, its `equals()` behavior remains unchanged.
+Similarly, data classes override the `equals()` function to implement structural equality by default. However, if the
+`equals()` function is marked as `final` in the parent class, its behavior remains unchanged.
 
 Distinctly, non-data classes (those not declared with the `data` modifier) do not override the 
 `equals()` function by default. Instead, non-data classes implement referential equality behavior inherited from the `Any` class.
@@ -83,10 +83,10 @@ true if and only if `a` and `b` point to the same object:
 
 ```kotlin
 fun main() {
-    val a = "Hello"
-    val b = a
-    val c = "world"
-    val d = "world"
+    var a = "Hello"
+    var b = a
+    var c = "world"
+    var d = "world"
 
     println(a === b)
     // true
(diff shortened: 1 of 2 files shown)


docs/topics/equality.md
```diff
@@ -34,6 +41,24 @@ Referential equality is checked by the `===` operation and its negated counterpa
 true if and only if `a` and `b` point to the same object. For values represented by primitive types at runtime
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
 
+> In Kotlin/JS, there is no difference between `===` and `==` operators when comparing strings. The reason is that in Kotlin/JS,
+> the `==` operator checks that two values are equal. Meanwhile, the `===` operator also checks that two values are equal 
+> besides verifying that the types of these two values are equal. In JS, Kotlin always uses the `===` operator:
```
ilya-g, line 46: This looks like some random fact about K/JS platform implementation aspects and I believe here it would only complicate the matter. We can find a place for that in Js interop (https://kotlinlang.org/docs/js-interop.html) and maybe explain even more weird cases there like https://pl.kotl.in/N-30hX1AM, and leave only a link here.

docs/topics/equality.md
```diff
@@ -14,17 +14,24 @@ By convention, an expression like `a == b` is translated to:
...
-Functions with the same name and other signatures, like `equals(other: Foo)`, don't affect equality checks with
+Functions with the same name and other signatures (like `equals(other: Foo)`) don't affect equality checks with
 the operators `==` and `!=`.
 
+While this equality operator (`==`) is called structural equality to distinguish it from the [referential equality](#referential-equality) (`===`),
+its underlying behavior could be referential in nature.
+
+One common scenario where structural equality can exhibit referential behavior is when handling non-data classes, 
+especially those without a custom implementation of the `equals()` function. In fact, this behavior is inherent by default
+in all classes except for [data classes](data-classes.md).
```
ilya-g, line 33: This is not completely clearly laid out.
By default, the structural `equals()` function is implemented using referential equality in the root `Any` class.
Then classes can override it to define actually structural equality.
Value classes always override equals implementing structural equality.
Data classes also override equals the same way, unless it is final in the parent class.

ilya-g (changes requested):

docs/topics/equality.md
ilya-g: It would also be nice to explain what "structural" actually means, i.e. if one wants to implement structural equals(), how would they do that.

docs/topics/equality.md
AlejandraPedroza: I placed the content in Js interop meanwhile, I'll ask in the TW team for suggestions of placement also. Meanwhile, could you please help me explaining why `println(1.0 as Any === 1 as Any)` is TRUE in Kotlin/JS and not FALSE. I was looking for explanations about this behavior but didn't find. Also, if you think we can add more edge cases, please let me know. Thanks.

docs/topics/equality.md
```diff
@@ -1,50 +1,123 @@
...
 
+In Kotlin, the `equals()` function is inherited by all classes from the `Any` class. By default, the `equals()` function 
+implements [referential equality](#referential-equality). However, classes in Kotlin can override the `equals()` 
+function to provide a custom equality logic and, in this way, implement structural equality.
+
+Value classes and data classes are two specific types in Kotlin that automatically override `equals()`. 
+Value classes override the `equals()` function to implement structural equality by default.
+
+Similarly, data classes override the `equals()` function to implement structural equality by default. However, if a data class 
+is marked as `final` in the parent class, its `equals()` behavior remains unchanged.
```
ilya-g, line 50: ", if **`equals`** is marked as `final` in the parent class"

docs/topics/equality.md
```diff
@@ -1,50 +1,123 @@
...
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
 
+> For more information about equality in Kotlin/JS, see the [Kotlin/JS](js-interop.md) documentation.
+>
+{type="tip"}
+
 ## Floating-point numbers equality
 
-When an equality check operands are statically known to be `Float` or `Double` (nullable or not), the check follows the 
+Equality checks for `Float` or `Double` operands (nullable or not) follow the
```
ilya-g, line 111: It was an important clarification about "statically known". Otherwise it can make an impression that the types of operands are checked to be floating point in runtime.

docs/topics/equality.md
```diff
@@ -1,50 +1,123 @@
...
+{type="tip"}
+
 ## Floating-point numbers equality
 
-When an equality check operands are statically known to be `Float` or `Double` (nullable or not), the check follows the 
+Equality checks for `Float` or `Double` operands (nullable or not) follow the 
 [IEEE 754 Standard for Floating-Point Arithmetic](https://en.wikipedia.org/wiki/IEEE_754). 
 
-Otherwise, structural equality is used, which disagrees with the standard so that `NaN` is equal to itself, `NaN` is considered greater than any other element, including `POSITIVE_INFINITY`, and `-0.0` is not equal to `0.0`.
+For other operand types, structural equality is implemented. This approach disagrees with the IEEE standard, considering
```
ilya-g, line 114: This whole paragraph is about floating point types, so this mode of equality can used for the same operands, just when their type is not known statically, e.g. `a == b`, when the type of either `a` or `b` variable is statically some supertype of Float/Double, but the values are Float/Double.

docs/topics/js/js-interop.md
```diff
@@ -202,12 +202,36 @@ fun usingAsOperator(s: Any) = s as String
 They will be compiled accordingly:
 
 ```javascript
-function usingUnsafeCast(s) {
+function usingUnsafeCast (s) {
```
ilya-g, line 205: I'm not sure if this space required by JS style guide?

docs/topics/equality.md
ilya-g: In, K/JS https://kotlinlang.org/docs/js-to-kotlin-interop.html#kotlin-types-in-javascript five numeric types are all represented with the same JS Number type in runtime, so the values of these types numerically equal are totally indistinguishable. Not sure if we can commit that it is as designed, but this is the current implementation peculiarity.

docs/topics/js/js-interop.md
```diff
@@ -202,12 +202,36 @@ fun usingAsOperator(s: Any) = s as String
...
\ No newline at end of file
+```
+
+## Equality
+
+Kotlin/JS has particular semantics for equality comparison, especially in the case of strings. 
+
+In Kotlin/JS, there is no difference between `===` and `==` operators when comparing strings. The reason is that in Kotlin/JS,
+the `==` operator checks that two values are equal. Meanwhile, the `===` operator checks both that two values are equal 
+and that the types of these two values are also equal. In JS, Kotlin always uses the `===` operator:
```
ilya-g, line 221: I believe here we mixed Kotlin and JS operators.
It's better to say that in K/JS, Kotlin === operator always translates to JS === operator, which checks that types and values are equal.

ilya-g (changes requested):

docs/topics/js/js-interop.md
AlejandraPedroza: Does Kotlin === operator always translates to JS === operator when comparing strings or can I remove "when comparing strings"?
I rephrased it, please take a look.

docs/topics/equality.md
AlejandraPedroza: I rephrased it as it was with a couple of words replaced. As we're linking to the specific pages with further explanation, we can leave it as it is here. What do you think?

docs/topics/js/js-interop.md
ilya-g: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#function_declarations

ilya-g: Although, there are multiple opinions on that in JS ecosystem. Better to consult with someone from JS team on that. If you decide to change the existing style, please do it consistently in all examples and not in this PR.

docs/topics/js/js-interop.md
```diff
@@ -202,12 +202,53 @@ fun usingAsOperator(s: Any) = s as String
...
+    val value2 = name.substring(0, 1)
+
+    println(if (value1 === value2) "yes" else "no")
+    // Prints 'yes' in Kotlin/JS
+    // Prints 'no' in other platforms
+}
+ ```
+
+Because of this usage of the JavaScript `===` operator, there is no Kotlin [structural equality](equality.md#structural-equality) `==` operator 
+when comparing strings in Kotlin/JS.
```
ilya-g, line 238: I don't get where does this follow from. In Kotlin/JS, strings can be compared by content with `==`.

docs/topics/js/js-interop.md
```diff
@@ -202,12 +202,53 @@ fun usingAsOperator(s: Any) = s as String
...
+
+Also, in Kotlin/JS, the [`Byte`, `Short`, `Int`, `Float`, and `Double`](js-to-kotlin-interop.md#kotlin-types-in-javascript) numeric types 
+are all represented with the `Number` JavaScript type in runtime. In consequence, these five types are numerically equal,
```
ilya-g, line 241: I don't think we can say that "types are numerically equal". Some values of these types can be.

docs/topics/equality.md
```diff
@@ -1,50 +1,124 @@
...
 ## Floating-point numbers equality
 
-When an equality check operands are statically known to be `Float` or `Double` (nullable or not), the check follows the 
-[IEEE 754 Standard for Floating-Point Arithmetic](https://en.wikipedia.org/wiki/IEEE_754). 
+When the equality check operands are statically known to be `Float` or `Double` (nullable or not), the check follows the 
+[IEEE 754 Standard for Floating-Point Arithmetic](https://en.wikipedia.org/wiki/IEEE_754).
 
-Otherwise, structural equality is used, which disagrees with the standard so that `NaN` is equal to itself, `NaN` is considered greater than any other element, including `POSITIVE_INFINITY`, and `-0.0` is not equal to `0.0`.
+Otherwise, structural equality is implemented. This approach disagrees with the standard, considering 
+that `NaN` is equal to itself, `NaN` is considered greater than any other element (including `POSITIVE_INFINITY`), and
```
ilya-g, line 115: "NaN is equal to itself", "NaN is greater than any other", and "-0.0 is not equal to 0.0" are all conjuncts. Better to use wording making it more clear.

docs/topics/js/js-interop.md
ilya-g: As far as I can tell, Kotlin === always translates to JS === for all operand types.

docs/topics/js/js-interop.md
AlejandraPedroza: This comes originally from the external feedback we got: https://github.com/JetBrains/kotlin-web-site/pull/3969
I'll remove this string-specific sentence then.

docs/topics/equality.md
AlejandraPedroza: Is it clearer now?

docs/topics/equality.md
```diff
@@ -1,50 +1,128 @@
...
+
+Value classes and data classes are two specific types in Kotlin that automatically override `equals()`. 
+Value classes override the `equals()` function to implement structural equality by default.
+
+Similarly, data classes override the `equals()` function to implement structural equality by default. However, if the
+`equals()` function is marked as `final` in the parent class, its behavior remains unchanged.
+
+Distinctly, non-data classes (those not declared with the `data` modifier) do not override the 
+`equals()` function by default. Instead, non-data classes implement referential equality behavior inherited from the `Any` class.
+To implement structural equality, non-data classes require a custom equality logic to override the `equals()` function.
```
ilya-g, line 54: Just though of an idea to give also a recommendation to override hashCode function if you're overriding equals.

ilya-g (approved):

docs/topics/equality.md
```diff
@@ -1,50 +1,132 @@
...
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is Point) return false
+
+        // Compares properties for structural equality
+        return this.x == other.x && this.y == other.y
+    }
+}
+```
+> When overriding the equals() function, you should also override the [hashCode() function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html#kotlin.Any$hashCode())
```
ilya-g, line 70:
```suggestion
> When overriding the equals() function, you should also override the [hashCode() function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/hash-code.html) 
```
I believe the anchor is not needed - that page fits in one screen. Also it can become obsolete after migrating documentation to the new dokka.

docs/topics/js/js-interop.md
```diff
@@ -210,4 +210,41 @@ function usingAsOperator(s) {
     var tmp$;
     return typeof (tmp$ = s) === 'string' ? tmp$ : throwCCE();
 }
-```
\ No newline at end of file
+```
+
+## Equality
+
+Kotlin/JS has particular semantics for equality comparison.
```
danil-pavlov, line 217:
```suggestion
Kotlin/JS has particular semantics for equality comparison compared to other platforms. 
```

docs/topics/js/js-interop.md
```diff
@@ -210,4 +210,41 @@ function usingAsOperator(s) {
...
+    val value2 = name.substring(0, 1)
+
+    println(if (value1 === value2) "yes" else "no")
+    // Prints 'yes' in Kotlin/JS
+    // Prints 'no' in other platforms
+}
+ ```
+
+Also, in Kotlin/JS, the [`Byte`, `Short`, `Int`, `Float`, and `Double`](js-to-kotlin-interop.md#kotlin-types-in-javascript) numeric types 
+are all represented with the `Number` JavaScript type in runtime. In consequence, the values of these five types are indistinguishable:
```
danil-pavlov, line 238:
```suggestion
are all represented with the `Number` JavaScript type in runtime. Therefore, the values of these five types are indistinguishable:
```

docs/topics/js/js-interop.md
```diff
@@ -210,4 +210,41 @@ function usingAsOperator(s) {
...
+
+ ```kotlin
+fun main() {
+    println(1.0 as Any === 1 as Any)
+    // Prints 'true' in Kotlin/JS
+    // Prints 'false' in other platforms
```
danil-pavlov, line 244:
```suggestion
    // Prints 'false' on other platforms
```

docs/topics/equality.md
```diff
@@ -1,50 +1,132 @@
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 Note that there's no point in optimizing your code when comparing to `null` explicitly:
 `a == null` will be automatically translated to `a === null`.
 
+In Kotlin, the `equals()` function is inherited by all classes from the `Any` class. By default, the `equals()` function 
+implements [referential equality](#referential-equality). However, classes in Kotlin can override the `equals()` 
+function to provide a custom equality logic and, in this way, implement structural equality.
+
+Value classes and data classes are two specific types in Kotlin that automatically override `equals()`.
```
danil-pavlov, line 46:
```suggestion
Value classes and data classes are two specific Kotlin types that automatically override `equals()`. 
```

docs/topics/equality.md
```diff
@@ -1,50 +1,132 @@
...
 `a == null` will be automatically translated to `a === null`.
 
+In Kotlin, the `equals()` function is inherited by all classes from the `Any` class. By default, the `equals()` function 
+implements [referential equality](#referential-equality). However, classes in Kotlin can override the `equals()` 
+function to provide a custom equality logic and, in this way, implement structural equality.
+
+Value classes and data classes are two specific types in Kotlin that automatically override `equals()`. 
+Value classes override the `equals()` function to implement structural equality by default.
+
+Similarly, data classes override the `equals()` function to implement structural equality by default. However, if the
```
danil-pavlov, line 49: I think there's a lot repetition in the last three sentences, maybe we can reduce it to:

"Value classes and data classes are two specific Kotlin types that automatically override equals(). That's why they (can) implement structural equality by default."

danil-pavlov (commented):
The page looks so much better now, thank you! Just a couple of suggestions

docs/topics/js/js-interop.md
AlejandraPedroza: Wrote it as "...for equality checks compared to other platforms" to avoid repeat comparison-compared one after the other. WDYT?

docs/topics/equality.md
```diff
@@ -1,50 +1,131 @@
...
+    // true
+
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+For values represented by primitive types at runtime
 (for example, `Int`), the `===` equality check is equivalent to the `==` check.
 
+> The referential equality is implemented differently in Kotlin/JS. For more information about equality in Kotlin/JS, see the [Kotlin/JS](js-interop.md#equality) documentation.
```
danil-pavlov, line 108:
```suggestion
> The referential equality is implemented differently in Kotlin/JS. For more information about equality, see the [Kotlin/JS](js-interop.md#equality) documentation.
```

danil-pavlov (approved):
Looks great! Just one minor repetition of K/JS left

AlejandraPedroza:
#3982 contribution by @scolsen 
 
#3969 contribution by @Hill-98

PR #4012, opened 2024-02-06
SPM export for remote integration

zamulla:


docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,194 @@
...
+Semantic Version of the package.
+
+Now that both files are accessible, you can try and set up the dependency in Xcode:
+* In an Xcode project, choose **File | Add Package Dependencies...** and provide the Git URL for the repo with
+the `Package.swift` file.
+* In an SPM project, create a dependency following the Apple documentation on [Package.Dependency](https://developer.apple.com/documentation/packagedescription/package/dependency).
+
+## Exporting different modules as a single XCFramework
+
+If your project contains unrelated KMP modules which you would like to export as a single iOS binary, you can create
```
DevAlloy, line 118: I'd suggest pointing out that this is the only working way to export multiple modules so far.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,194 @@
...
+        //...
+    }
+    ```
+
+3. Currently, a framework cannot be assembled if the module being exported does not contain any source code. To work around
+this:
+   1. Create a source file inside the `together` folder, for example, `together/src/commonMain/kotlin/Together.kt`.
+   2. Create an empty `main()` function inside this file:
+       ```kotlin
+      fun main() {
```
abdulowork, line 179: It should be enough to create an empty `stub.kt` file. Let's not ask to write any code in this stub

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,194 @@
...
+* In an Xcode project, choose **File | Add Package Dependencies...** and provide the Git URL for the repo with
+the `Package.swift` file.
+* In an SPM project, create a dependency following the Apple documentation on [Package.Dependency](https://developer.apple.com/documentation/packagedescription/package/dependency).
+
+## Exporting different modules as a single XCFramework
+
+If your project contains unrelated KMP modules which you would like to export as a single iOS binary, you can create
+an umbrella module and combine other modules in it.
+
+For example, you have a `network` and a `database` module, which you combine in an `together` module.
```
abdulowork, line 121: > `together` module

Do I recall correctly that the issue with using `umbrella` name for the module was that `umbrella` is a reserved identifier in the modulemap syntax? I wonder if we should deprecate the `umbrella` terminology in favor of something better

docs/topics/native/native-spm.md
zamulla: It didn't work for me with an empty file, but now it does. Good news!

docs/topics/native/native-spm.md
zamulla: I would rather leave an explicit note underlining that the user should not use `umbrella` as a name for the module.  It's a good name, that's why I used it :) and I have no doubt that a significant amount of people will think the same. Xcode says absolutely nothing when you do that, and there is no error except at the very end when the import just doesn't work, so we can save them a lot of trouble.

But I'll discuss this with the TW team.

abdulowork: I think this issue is actually be quite easy to fix. I filed an issue - https://youtrack.jetbrains.com/issue/KT-66565

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+      // Name of the module that will be imported in the consumer project
+      val xcframeworkName = "Shared"
+   
+      listOf(
+         iosX64(),
+         iosArm64(),
+         iosSimulatorArm64(),
+      ).forEach {
+           it.binaries.framework {
+              baseName = xcframeworkName
```
abdulowork, line 64: Could we please add the recommendation to specify the CFBundleIdentifier; this will avoid the K/N linker warning:
```
// Specify CFBundleIdentifier
binaryOption("bundleId", "org.example.${xcframeworkName}")
```

abdulowork: Same [here](https://github.com/JetBrains/kotlin-web-site/pull/4012/files#diff-a04ec0321e644ade5a5ed2af850e02ff2dd29178e3d1662417a0bb057501abd7R145)

abdulowork, line 64: Could we please add `isStatic = true` here

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+            name: "shared",
+            url: "<link to the uploaded XCFramework ZIP file>",
+            checksum:"<checksum calculated for the ZIP file>")
+      ]
+   )
+   ```
+8. You can check if the package manifest is valid by running the following command next to the `Package.swift` file:
+
+    ```shell
+    swift package describe
```
abdulowork, line 104: We should probably also recommend doing:
```
swift package reset && swift package show-dependencies --format json
```

This command will drop the `.build` directory and try to download the `.xcframework` that is pointed by `.binaryTarget`; it should help catch some simple configuration errors, such as incorrect URL, checksum or auth issues

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+      ).forEach {
+           it.binaries.framework {
+              baseName = xcframeworkName
+              xcf.add(this)
+           }
+        }
+      //...
+   }
+   ```
+2. Sync the updated Gradle configuration.
```
abdulowork, line 71: Let's remove this step; we are talking about a CLI workflow here and it might be unclear what kind of sync we are referring to

docs/topics/native/native-spm.md
daugeldauge: To fully check the correctness of the package, one can: 
1) Add `.testTarget(name: "sharedChech", dependencies: ["shared"])`
2) Add file to `Tests/sharedCheck/Check.swift` with one line `import shared`
3) Run `swift test`. It should pass if the framework and the package are correct
4) Remove what was added in 1) and 2)

But I'm not really sure whether it's a good idea to add it to this guide, because it's quite heavy.

zamulla: I think that would be an excellent "if you want to be thorough..." addition, not in the course of the main step-by-step, which is long already, but as another section at the end. One of those short and definitive recipes one spends hours tracking down on StackOverflow.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
+[//]: # (title: Swift Package export setup)
+
+Kotlin/Native output for an Apple target can be consumed as an SPM dependency.
+For example, if your Kotlin Multiplatform project has an iOS target, you may want to make the binary available
+as a dependency to iOS developers working on native Swift projects in Xcode. Using KMP tooling, you can provide
+an artifact that would seamlessly integrate with Xcode experience.
```
danil-pavlov, line 6: Not sure about "integrate... with experience" phrasing
consider "integrate with Xcode" or "integrate with their (devs mentioned earlier) Xcode projects

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
+[//]: # (title: Swift Package export setup)
+
+Kotlin/Native output for an Apple target can be consumed as an SPM dependency.
```
danil-pavlov, line 3: I'd suggest to break the intro into smaller units:

> You can set up the Kotlin/Native output for an Apple target to be consumed as a Swift package manager (SPM) dependency.
> 
> Consider a Kotlin Multiplatform project that has an iOS target. In that case, you may want to make the binary available as a dependency to iOS developers who work on native Swift projects in Xcode. Using Kotlin Multiplatform tooling, you can provide an artifact that seamlessly integrates with their Xcode projects.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+For example, if your Kotlin Multiplatform project has an iOS target, you may want to make the binary available
+as a dependency to iOS developers working on native Swift projects in Xcode. Using KMP tooling, you can provide
```
danil-pavlov, line 5: Here and below, please avoid the KMP acronym. In the docs, we try to use it as little as possible, as Kotlin has a long history of confusing multiplatform acronyms

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+an artifact that would seamlessly integrate with Xcode experience.
+
+This guide shows how to do this using the [XCFramework building](multiplatform-build-native-binaries.md#build-xcframeworks)
```
danil-pavlov, line 8:
```suggestion
This guide shows how to do this by building [XCFrameworks](multiplatform-build-native-binaries.md#build-xcframeworks)
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+capability of Kotlin Gradle.
```
danil-pavlov, line 9:
```suggestion
with the Kotlin Gradle plugin.
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Prepare locations for the files to be uploaded
```
danil-pavlov, line 11:
```suggestion
## Prepare file locations
```

danil-pavlov: I think we can omit some details in headers to make the right-hand menu more readable

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+To make your framework available to be consumed, you will need to upload two files:
```
danil-pavlov, line 13:
```suggestion
To make your framework consumable, you need to upload two files:
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+* A ZIP archive of the XCFramework itself should be uploaded to a convenient file storage with direct access (for example,
+S3, GitHub Releases, or a Maven repository). Choose the option that is easiest to integrate into your workflow.
```
danil-pavlov, line 15: A couple of concerns about the repository list:

* Is "GitHub Releases" a separate product name? Do we need to underline the releases here? If we want to stick to the official names, I think it should be "Amazon S3" as well
* I think it would be great to both simplify the names and make it consistent with the Maven item:

>  for example, an S3, GitHub, or Maven repository

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+as a dependency to iOS developers working on native Swift projects in Xcode. Using KMP tooling, you can provide
+an artifact that would seamlessly integrate with Xcode experience.
+
+This guide shows how to do this using the [XCFramework building](multiplatform-build-native-binaries.md#build-xcframeworks)
+capability of Kotlin Gradle.
+
+## Prepare locations for the files to be uploaded
+
+To make your framework available to be consumed, you will need to upload two files:
+* A ZIP archive of the XCFramework itself should be uploaded to a convenient file storage with direct access (for example,
```
danil-pavlov, line 14: I think the list has too many modal verbs which makes the instructions sound very strict. You can both highlight the files and use the imperative:

> To make your framework consumable, you need to upload two files:
> 
> * A ZIP archive of the XCFramework. Upload it to a convenient file storage with direct access, for example, an S3, GitHub, or Maven repository. Choose the option that is easiest to integrate into your workflow.
> * The Package.swift file, describing the package. Prepare a Git repository and upload the file. To decide which repository to use, see the pros and cons of several options below.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+S3, GitHub Releases, or a Maven repository). Choose the option that is easiest to integrate into your workflow.
+* The `Package.swift` file describing the package needs to be placed in a Git repository. You need to decide which
```
danil-pavlov, line 16:
```suggestion
* The `Package.swift` file describing the package needs to be placed in a Git repository. To decide which
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+repository to use. Pros and cons for several options are discussed below.
```
danil-pavlov, line 17:
```suggestion
repository to use, see pros and cons of several options below.
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Options for Package.swift distribution
```
danil-pavlov, line 19:
```suggestion
### Swift package distribution
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+The `Package.swift` that describes the Swift Package needs to be placed in a Git repository:
```
danil-pavlov, line 20: This list intro repeats the previous section, but doesn't introduce the options

> Consider the following options for distributing Swift packages in a Git repository:

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+* Store the `Package.swift` file in an independent repository to make its versioning separate from the
+KMP project the file describes. This is the recommended approach, allowing scalability and generally easier
+maintenance.
+* Put the `Package.swift` file next to your KMP code for a more straightforward approach. Keep in mind that versioning
+a Swift Package in the same repository as the code can become challenging as SPM uses Git tags for versioning packages
+and this can conflict with using tags for the project itself. 
+* Store the `Package.swift` within the consumer project's repo to avoid the versioning and maintenance issues.
```
danil-pavlov, line 27: Here and below, in the docs, we usually use the full "repository"

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+  However, this approach can cause problems for multi-repository SPM setups of the consumer project and further automation:
```
danil-pavlov, line 29:
```suggestion
  However, this approach can cause problems with multi-repository SPM setups of the consumer project and further automation:
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+  1. Only one consumer package in a multi-package project can depend on the external module, to avoid dependency conflicts
```
danil-pavlov, line 30: it's not a step-by-step instruction, should be a bulleted list I guess

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+  within the project. So all the logic depending on your KMP module should be encapsulated in a particular consumer package.
```
danil-pavlov, line 31:
```suggestion
  within the project. So, all the logic that depends on your KMP module should be encapsulated in a particular consumer package.
```

docs/topics/native/native-spm.md
danil-pavlov, line 30:
```suggestion
  1. In a multi-package project, only one consumer package can depend on the external module, to avoid dependency conflicts
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+To make your framework available to be consumed, you will need to upload two files:
+* A ZIP archive of the XCFramework itself should be uploaded to a convenient file storage with direct access (for example,
+S3, GitHub Releases, or a Maven repository). Choose the option that is easiest to integrate into your workflow.
+* The `Package.swift` file describing the package needs to be placed in a Git repository. You need to decide which
+repository to use. Pros and cons for several options are discussed below.
+
+### Options for Package.swift distribution
+The `Package.swift` that describes the Swift Package needs to be placed in a Git repository:
+* Store the `Package.swift` file in an independent repository to make its versioning separate from the
```
danil-pavlov, line 21: Again, I'd recommend to highlight the options and split the sentences. This is what I came up with:

> * Store the Package.swift file in an independent repository. It helps to separate its versioning from the Kotlin Multiplatform project. This is the recommended approach. It's scalable and generally easier to maintain.
> * Put the Package.swift file in the same repository as your Kotlin Multiplatform code. It's a more straightforward approach, but keep in mind that, in this case, a Swift package and the code have the same versioning. SPM uses Git tags to version packages, which can conflict with tags used for the project.
> * Store the Package.swift within the consumer project's repository. It helps to avoid versioning and maintenance issues. However, this approach can cause problems with multi-repository SPM setups of the consumer project and further automation:

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+* Store the `Package.swift` within the consumer project's repo to avoid the versioning and maintenance issues.
+
+  However, this approach can cause problems for multi-repository SPM setups of the consumer project and further automation:
+  1. Only one consumer package in a multi-package project can depend on the external module, to avoid dependency conflicts
+  within the project. So all the logic depending on your KMP module should be encapsulated in a particular consumer package.
+  2. If you publish the KMP project using an automated CI process, this process would need to include publishing the
+  updated `Package.swift` file to the consumer repo. Such a phase in CI can be difficult to maintain as it may lead
+  to conflicting updates of the consumer repo.
+
+Now that you understand the options, create or adjust a Git repository as needed.
```
danil-pavlov, line 36: I suggest removing this sentence. It can be confusing as it seems to introduce the next section, but is actually related to the Prepare locations section

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Create the XCFramework and the Package.swift file
```
danil-pavlov, line 38:
```suggestion
## Create the XCFramework and Swift package files
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+
+> The following example assumes that the shared code of your KMP project is stored in the `shared` module.
+> If your project is structured differently, substitute "shared" in code and path examples with the name of your module.
```
danil-pavlov, line 41:
```suggestion
> If your project is structured differently, substitute "shared" in code and path examples with your module's name.  
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+Now that you understand the options, create or adjust a Git repository as needed.
+
+## Create the XCFramework and the Package.swift file
+
+> The following example assumes that the shared code of your KMP project is stored in the `shared` module.
+> If your project is structured differently, substitute "shared" in code and path examples with the name of your module.  
+>
+{type="tip"}
+
+To provide a Swift package, you can follow these steps:
```
danil-pavlov, line 45:
```suggestion
To provide a Swift package:
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+1. Set up publishing of an [XCFramework](multiplatform-build-native-binaries.md#build-xcframeworks). Add an `XCFramework`
+call to your iOS targets description in the `shared/build.gradle.kts` file:
+   ```kotlin
+   import org.jetbrains.kotlin.gradle.plugin.mpp.apple.XCFramework
+   
+   kotlin {
+      // other KMP targets
```
danil-pavlov, line 52:
```suggestion
      // Other Kotlin Multiplatform targets
```

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+        }
+      //...
+   }
+   ```
+2. Sync the updated Gradle configuration.
+3. Run the Gradle task to create the framework:
+   
+   `./gradlew :shared:assembleSharedReleaseXCFramework`
+4. The resulting framework will be created as the `shared/build/XCFrameworks/release/shared.xcframework` folder in your project directory.
+5. Put the `shared.xcframework` folder in a ZIP archive and calculate the checksum for the resulting archive, for example:
```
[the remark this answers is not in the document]
danil-pavlov: the tip at the beginning already introduced the example

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+Semantic Version of the package.
+
+Now that both files are accessible, you can try and set up the dependency in Xcode:
+* In an Xcode project, choose **File | Add Package Dependencies...** and provide the Git URL for the repo with
+the `Package.swift` file.
+* In an SPM project, create a dependency following the Apple documentation on [Package.Dependency](https://developer.apple.com/documentation/packagedescription/package/dependency).
+
+## Exporting multiple modules as a single XCFramework
+
+To make several KMP modules available as iOS binaries, create an umbrella module and combine other modules in it,
```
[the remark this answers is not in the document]
danil-pavlov: Is it "combine them inside the umbrella module" or "combine them with the umbrella module"?

danil-pavlov (changes requested):

abdulowork:
LGTM

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+    ```shell
+    swift package describe
+    ```
+    
+    The output will describe any found errors, or show the parsed tree of the manifest if it is grammatically correct. 
+9. Push the `Package.swift` file to the repository you settled on earlier. Make sure to create and push a git tag with the
+Semantic Version of the package.
+
+Now that both files are accessible, you can try and set up the dependency in Xcode:
+* In an Xcode project, choose **File | Add Package Dependencies...** and provide the Git URL for the repo with
```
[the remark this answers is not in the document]
zamulla: Both! I decided to be a little more specific here because SPM dependencies are a nightmare.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+the `Package.swift` file.
+* In an SPM project, create a dependency following the Apple documentation on [Package.Dependency](https://developer.apple.com/documentation/packagedescription/package/dependency).
+
+## Exporting multiple modules as a single XCFramework
```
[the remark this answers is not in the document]
zamulla: I think it's important to underline the multipleness of modules in the title here because it's a common use case. I'm counting on people to search for it with corresponding keywords.

docs/topics/native/native-spm.md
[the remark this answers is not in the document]
zamulla: Inside, yes. I rephrased to make it clearer.

docs/topics/native/native-spm.md
```diff
@@ -0,0 +1,192 @@
...
+            iosTarget.binaries.framework {
+                export(projects.network)
+                export(projects.database)
+    
+                baseName = frameworkName
+                xcf.add(this)
+            }
+        }
+    
+        // Dependencies set as api to export underlying modules
```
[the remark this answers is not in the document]
zamulla: I meant `api` as opposed to `implementation`, clarified it in the comment.

danil-pavlov (approved):

PR #4014, opened 2024-02-06
Update native-cocoapods.md

IliaSolovei:
It is better to install CocoaPods to `/usr/local/bin`, so they will always be available in PATH, and it will be possible to use them from Xcode


danil-pavlov (commented):
Thank you! Should we fix the "Default Ruby" option the same way?

IliaSolovei:
> Should we fix the "Default Ruby" option the same way?

It is not necessary because default Ruby installs gems in `/usr/local/bin` by default

danil-pavlov (approved):

PR #4020, opened 2024-02-06
KTL-1335: added tests for WebHelp page

berezinant:
webhelp tests now are runnable on teamcity
image


kotlin-website.py
```diff
@@ -236,6 +236,10 @@ def kotlin_reference_pdf():
 def kotlin_docs_pdf():
     return send_file(path.join(root_folder, "assets", "kotlin-reference.pdf"))
 
+@app.route('/docs/<path:path>')
```
berezinant, line 239: should we add some conditions here to exclude such routing on production? @zoobestik

berezinant: added check if it is prodcution

.teamcity/tests/buildTypes/E2ETests.kt
```diff
@@ -1,32 +1,41 @@
...
 
-  steps {
-    script {
-      scriptContent = "./scripts/test/up.sh"
-    }
-
-    script {
-      scriptContent = "./scripts/test/run.sh"
+    dependencies {
+        artifacts(AbsoluteId("Kotlin_KotlinSites_KotlinlangTeamcityDsl_BuildReferenceDocs")) {
```
nikpachoo, line 26: I believe the dependency can be resolved differently. You can import the class and use it (we do it already in the configuration).

.teamcity/tests/buildTypes/E2ETestsViaCompose.kt
```diff
@@ -0,0 +1,46 @@
+package tests.buildTypes
+
+import jetbrains.buildServer.configs.kotlin.BuildType
+import jetbrains.buildServer.configs.kotlin.buildFeatures.PullRequests
+import jetbrains.buildServer.configs.kotlin.buildFeatures.pullRequests
+import jetbrains.buildServer.configs.kotlin.buildSteps.script
+
+
+object E2ETestsViaCompose : BuildType({
```
nikpachoo, line 9: It seems like something can be improved here.

- Do we really need another different configuration? Maybe we can merge it with E2ETests?
- If it is still needed, I would prefer to see in the title what it does, not how :)

.teamcity/tests/buildTypes/E2ETests.kt
```diff
@@ -15,35 +14,30 @@ object E2ETests : BuildType({
     root(vcsRoots.KotlinLangOrg)
   }
 
-  steps {
-    script {
-      scriptContent = "./scripts/test/up.sh"
-    }
+    dependencies {
+        artifacts(BuildReferenceDocs) {
+            cleanDestination = true
```
nikpachoo, line 19: Maybe we can reuse there the last successful build to speed up the process?
We do it in a different build: 
```
buildRule = lastSuccessful(branch = "+:<default>")
```

.teamcity/tests/buildTypes/E2ETests.kt
```diff
@@ -15,35 +14,30 @@ object E2ETests : BuildType({
...
-    script {
-      scriptContent = "./scripts/test/run.sh"
     }
 
-    script {
-      scriptContent = "./scripts/test/stop.sh"
-    }
+  steps {
+      script {
+          scriptContent = "docker compose -f docker-compose-e2e-statics.yml up --exit-code-from playwright"
```
nikpachoo, line 27: I'd recommend putting script content in a file and calling it from there. 
That gives you the ability to highlight the script and set the flags.

kotlin-website.py
```diff
@@ -48,6 +48,7 @@
 build_check_links = True
 build_errors = []
 url_adapter = app.create_url_adapter(None)
+node_env = os.environ.get('NODE_ENV', 'development')
```
nikpachoo, line 51: Do we set "production" for the variable in prod or we need to do that?

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import {  RESOLUTIONS } from '../visual/visual-constants';
+
+test.describe('WebHelp page interactivity', async () => {
+    test.beforeEach(async ({ page }) => {
+        const webHelpPage = new WebHelpPage(page, '/docs/test-page.html');
+        await webHelpPage.init();
+    });
+
+    test(`Should open page  item  properly on desktop`, async ({ page }) => {
+        await page.setViewportSize(RESOLUTIONS[0]);
```
nikpachoo, line 13: Would be great to change RESOLUTIONS[0], to something like "RESOLUTIONS.MOBILE"

nikpachoo (approved):
Great work!
Nothing critical from my side, the changes I proposed can be done as separate tasks.
I would also like to see some note in Readme about the tests, at least a general explanation: how does it work (can I run it locally, and how, the restrictions, the goal of it).

.teamcity/tests/buildTypes/E2ETests.kt
zoobestik: I'm not sure. We want to test branch before release, no? I think it's nice build by chain.

docker-compose-e2e-statics.yml
```diff
@@ -0,0 +1,29 @@
...
+      - network1
+
+  playwright:
+    build:
+      context: .
+      dockerfile: ./dockerfiles/playwright/Dockerfile
+    command: yarn run test:visual:ci
+    environment:
+      BASE_URL: "http://nginx-server"
+      ipc: host
```
zoobestik, line 19: What is the issue resolve `ipc=host`?

docker-compose-e2e-statics.yml
```diff
@@ -0,0 +1,29 @@
...
+    volumes:
+      - ./test/snapshots:/var/www/test/snapshots
+      - ./test-results:/var/www/test-results
+    depends_on:
+      - nginx-server
```
zoobestik, line 24: There is no problem with container soft race without `HEALTHCHECK`?

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    test(`Should copy codeblock content when copy button clicked`, async ({ page }) => {
+        const EXPECTED_CODEBLOCK_CONTENT = `plugins {
+kotlin("kapt") version "1.9.22"
+}`;
+        const codeBlock = page.locator(testSelector('code-block')).first();
+        await codeBlock.hover();
+        const copyButton = codeBlock.locator(testSelector('copy-button')).first();
+        await copyButton.click();
+        // since Playwright doesn't have access to the clipboard, we have to test it manually by creating a textarea and pasting the clipboard content into it
```
zoobestik, line 31: Have you tried `grantPermissions`?
```javascript
await context.grantPermissions(["clipboard-read"]);
const handle = await page.evaluateHandle(() => navigator.clipboard.readText());
const clipboardContent = await handle.jsonValue();
```

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { expect, test } from '@playwright/test';
+import { testSelector } from '../utils';
+import { WebHelpPage } from '../page/webhelp-page';
+import {  RESOLUTIONS } from '../visual/visual-constants';
```
zoobestik, line 4: `  ` suggest me file doesn't linted. 😔

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+test.describe('WebHelp page interactivity', async () => {
+    test.beforeEach(async ({ page }) => {
+        const webHelpPage = new WebHelpPage(page, '/docs/test-page.html');
+        await webHelpPage.init();
```
zoobestik, line 9: There is actually not so clear why we use class for just one method.

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        await page.keyboard.press('Meta+V');
+        const clipboardContent = await page.$eval('#clipboardTextarea', (textAreaElement: HTMLTextAreaElement) => textAreaElement.value);
+        expect(clipboardContent).toEqual(EXPECTED_CODEBLOCK_CONTENT);
+    });
+
+    test(`Should zoom zoomable image when clicked`, async ({ page }) => {
+        const element = page.locator('figure').filter({ has: page.locator('a[href="images/ksp-class-diagram.svg"]') }).first();
+        await element.locator('button').first().click();
+        const lightbox = page.locator('div.light-box').first();
+        expect(await lightbox.count()).toBe(1);
```
zoobestik, line 48: If we test it, it would be nice check inside the lightbox image exists.

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    test(`Should close zoomable image when close button clicked`, async ({ page }) => {
```
zoobestik, line 51: Looks like this test fully covered previous. No? 🤔

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import { WebHelpPage } from '../page/webhelp-page';
+import {  RESOLUTIONS } from '../visual/visual-constants';
+
+test.describe('WebHelp page interactivity', async () => {
+    test.beforeEach(async ({ page }) => {
+        const webHelpPage = new WebHelpPage(page, '/docs/test-page.html');
+        await webHelpPage.init();
+    });
+
+    test(`Should open page  item  properly on desktop`, async ({ page }) => {
```
zoobestik, line 12:
```suggestion
    test(`Should open page item properly on desktop`, async ({ page }) => {
```

test/visual/visual-constants.ts
```diff
@@ -0,0 +1,10 @@
+export const MICRO_ANIMATION_TIMEOUT = 350;
+export const MICRO_ANIMATION_TIMEOUT_LONG = 500;
+
+export const ELEMENT_PADDING_OFFSET = 25;
+
+export const RESOLUTIONS = [
```
zoobestik, line 6: Actually, i don't fully understand `RESOLUTIONS` idea. Is it re-invention of [`playwright emulation`](https://playwright.dev/docs/emulation) or something different?

test/visual/webhelp.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+
+test.describe('WebHelp page appearance', async () => {
+    test.beforeEach(async ({ page }) => {
+        const webHelpPage = new WebHelpPage(page, '/docs/test-page.html');
+        await webHelpPage.init();
+    });
+
+    test(`Should render table of contents properly on desktop`, async ({ page }) => {
+        await page.setViewportSize(RESOLUTIONS[0]);
+        const element = page.locator('nav').locator('ul.toc').first();
```
zoobestik, line 22: Here and in other places, `element` not so clear what u r choose.

zoobestik (changes requested):

zoobestik:
In common cases looks ok. But for me different resolution and spec/visual test way rise a lot of questions.

kotlin-website.py
berezinant: No, it is a new variable only considered in development layer. So at the moment it doesn't matter do we set it's value in production or not. The only possible issue is a little inconsistency — there is value in dev and no value in prod. But nothing will break

test/e2e/webhelp.spec.ts
berezinant: I did so to maintain consistency in style between different tests but I agree that it may look like an overkill here

docker-compose-e2e-statics.yml
berezinant: Experiment shows that it's now required :)

docker-compose-e2e-statics.yml
berezinant: I haven't experienced it but it is possible so I added healtcheck

test/e2e/webhelp.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        await tocItem.click();
+        const linkTocItem = element.locator('a[href="basic-syntax.html"]').first();
+        await linkTocItem.click();
+        expect(page.url()).toContain('basic-syntax.html');
+    });
+
+
+    test(`Should copy codeblock content when copy button clicked`, async ({ page }) => {
+        const EXPECTED_CODEBLOCK_CONTENT = `plugins {
+kotlin("kapt") version "1.9.22"
```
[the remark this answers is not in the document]
berezinant: removed $kotlinVersion from this place it test-page.md

.teamcity/tests/buildTypes/E2ETests.kt
berezinant: ✅

.teamcity/tests/buildTypes/E2ETestsViaCompose.kt
berezinant: ✅

.teamcity/tests/buildTypes/E2ETests.kt
berezinant: ✅

test/e2e/webhelp.spec.ts
berezinant: ✅

test/e2e/webhelp.spec.ts
berezinant: the files in test/ are not linted indeed
I suggest to fix it in another pr

test/e2e/webhelp.spec.ts
berezinant: done

test/e2e/webhelp.spec.ts
berezinant: Here we tests that "close" also works so this is not exactly the same case

zoobestik: Ok, the question why is it two different tests?

berezinant: Just to highlight the functionality it focuses on: one tests openning, the other tests closing
I also do not like that the second tests overlaps in steps with the first, but if we consider analyzing test reports in the future we could possibly see that "opening is working, closing is not working"
If we have only one tests that we can't figure out from the report what is broken: opennign, closing or both

.teamcity/tests/buildTypes/E2ETests.kt
berezinant: So we need to decide something here.
For me it is also sounds more reliable to build needed artifacts explicitly before tests run (this is to leave things as it is now)

test/e2e/webhelp.spec.ts
berezinant: I've tried this and made this work locally but not in Docker unfortunately

zoobestik (approved):

PR #4023, opened 2024-02-10
exercise 3 has the solution within the question part

HafizBeyy:


sarahhaggarty:
Hi @HafizBeyy 
Thanks for your suggestion in this PR. It's actually our intention to provide the body of the function so that you can focus on just adding the default parameter values and named arguments to make the code easier to read. For now, I will close this PR but please let me know if there is anything else that is unclear about the exercise.

PR #4030, opened 2024-02-14
chore: adding version browser clarification

AlejandraPedroza:
Rephrased "In addition to updating the browser settings, make sure you use a Kotlin version before 1.9.20 when using a Chrome version older than version 119" sentence.


zamulla (approved):
Nice.

PR #4034, opened 2024-02-15
update: adding explanation of lazy to booleans and minor changes

AlejandraPedroza:


zarechenskiy (approved):

docs/topics/booleans.md
```diff
@@ -2,35 +2,43 @@
 
 The type `Boolean` represents boolean objects that can have two values: `true` and `false`. 
 
-> On the JVM, booleans stored as primitive type: `boolean`, typically use 8 bits.
+> On the JVM, booleans stored as the primitive `boolean` type typically use 8 bits.
 >
 {type="note"}
 
-`Boolean` has a nullable counterpart `Boolean?` that also has the `null` value.
+`Boolean` has a nullable counterpart (`Boolean?`) with the `null` value.
```
zamulla, line 9: This may sound as if `Boolean?` only has the `null` value. Does this sentence even need the second part, does the audience need to be reminded of what "nullable" is?

Also, I would put this sentence together with the one at line 3, they would make a nice paragraph. Right now, it looks like the tip splits that nice paragraph in two for no reason I can see. What do you think?

docs/topics/booleans.md
```diff
@@ -2,35 +2,43 @@
...
 
-`||` and `&&` work lazily.
+For example:
 
 ```kotlin
 fun main() {
 //sampleStart
     val myTrue: Boolean = true
     val myFalse: Boolean = false
     val boolNull: Boolean? = null
```
zamulla, line 24: I understand that this is supposed to demonstrate how a `Boolean?` can be null, but it looks like an unused variable. Even the playground points this out. Maybe printing it out together would be clearer? Show that 1) it compiles; 2) the value is actually null.

zamulla (approved):
Noice!

PR #4045, opened 2024-02-22
Update jvm-spring-boot-add-db-support.md

Luxios22:
When I follow this tutorial to [this part](https://kotlinlang.org/docs/jvm-spring-boot-add-db-support.html#retrieve-messages-by-id), it cannot work in my local environment and give the following error:
```
Task :compileKotlin FAILED
e: KotlinSpringbootDemoApplication.kt:51:54 Type mismatch: inferred type is Unit but List<Message> was expected
e: KotlinSpringbootDemoApplication.kt:51:57 Type mismatch: inferred type is Unit but List<Message> was expected
e: KotlinSpringbootDemoApplication.kt:51:102 Type mismatch: inferred type is String but RowCallbackHandler was expected
e: KotlinSpringbootDemoApplication.kt:51:106 Passing value as a vararg is only allowed inside a parenthesized argument list
e: KotlinSpringbootDemoApplication.kt:51:108 Cannot infer a type for this parameter. Please specify it explicitly.
e: KotlinSpringbootDemoApplication.kt:51:118 Cannot infer a type for this parameter. Please specify it explicitly.
```
My local `build.gradle`:
```
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile // For `KotlinCompile` task below

plugins {
    id 'org.springframework.boot' version '3.2.2'
    id 'io.spring.dependency-management' version '1.1.4'
    id 'org.jetbrains.kotlin.jvm' version '1.9.22' // The version of Kotlin to use
    id 'org.jetbrains.kotlin.plugin.spring' version '1.9.22' // The Kotlin Spring plugin
}

group = 'org.example'
version = '0.0.1-SNAPSHOT'

java {
    sourceCompatibility = '17'
}

repositories {
    mavenCentral()
}

dependencies {
    implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
    implementation 'org.springframework.boot:spring-boot-starter-web'
    implementation 'com.fasterxml.jackson.module:jackson-module-kotlin' // Jackson extensions for Kotlin for working with JSON
    implementation 'org.jetbrains.kotlin:kotlin-reflect' // Kotlin reflection library, required for working with Spring
    runtimeOnly 'com.h2database:h2'
    testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

tasks.withType(KotlinCompile) { // Settings for `KotlinCompile` tasks
    kotlinOptions { // Kotlin compiler options
        freeCompilerArgs += '-Xjsr305=strict' // `-Xjsr305=strict` enables the strict mode for JSR-305 annotations
        jvmTarget = '17' // This option specifies the target version of the generated JVM bytecode
    }
}

tasks.named('test') {
    useJUnitPlatform()
}
```
After investigating into the `query` function, I saw the following:
```
    @Deprecated
    public void query(String sql, @Nullable Object[] args, RowCallbackHandler rch) throws DataAccessException {
        this.query(sql, this.newArgPreparedStatementSetter(args), rch);
    }
```
I guess the error is because this function is not supported in latest version, but I am not sure for I am just a new learner.  Anyway after I changed the code as shown in this PR, then it worked. So, I create this PR to update the tutorial.

### DIFF
--- a/docs/topics/jvm/jvm-spring-boot-add-db-support.md
+++ b/docs/topics/jvm/jvm-spring-boot-add-db-support.md
@@ -248,7 +248,7 @@ Extend the functionality of the application to retrieve the individual messages 
             Message(response.getString("id"), response.getString("text"))
         }
     
-        fun findMessageById(id: String): List<Message> = db.query("select * from messages where id = ?", id) { response, _ ->
+        fun findMessageById(id: String): List<Message> = db.query("select * from messages where id = ?", arrayOf(id), intArrayOf(Types.VARCHAR)) { response, _ ->
             Message(response.getString("id"), response.getString("text"))
         }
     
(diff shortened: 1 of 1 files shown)


antonarhipov:
@Luxios22 Looks like you've missed it import for the query extension function:

`import org.springframework.jdbc.core.query`

Could it be the case?

Luxios22:
Oh yes, your suggestion works. It seems the IDE did not import it automatically and use another query method. I do not know whether should be noted in the tutorial. Anyway thanks for your clarification.

antonarhipov:
@Luxios22 Yes, it's noted in the tutorial - the green message right under the code snippet. We'll make it more prominent. Meanwhile, we do have the issue registered to improve IJ import behavior for the extension functions, hope it gets resolved soon.

koshachy:
Hey there,

It's mentioned in the tutorial just after the code block (follow the link you provided). We'll discuss within the team how to make it more visible.

Thanks for your feedback!

Luxios22:
Oh sorry I did not notice it, thanks!
I think you may state `import org.springframework.jdbc.core.query` under the explanation in the green part, for when writing the code in IDE may not notice the import statement in the code block. 
Anyway thanks for the quick and patient review!

PR #4049, opened 2024-02-22
Update idioms.md with suggestion to use inline value classes

gw111zz:
Update idioms.md with suggestion to use inline value classes

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -2,21 +2,6 @@
 
 A collection of random and frequently used idioms in Kotlin. If you have a favorite idiom, contribute it by sending a pull request.
 
-## Use inline value classes to wrap single values for type-safety
-
-```kotlin
-@JvmInline
-value class EmployeeId(private val id: String)
-
-@JvmInline
-value class CustomerId(private val id: String)
-```
-
-Now accidentally mixing up `EmployeeId` and `CustomerId` will result in a compilation error. The inline classes will be represented by the 
-underlying type at runtime. `@JvmInline` is only needed for JVM backends.
-
-(See [Inline classes](inline-classes.md) for more information)
-
 ## Create DTOs (POJOs/POCOs)
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @gw111zz 
Thanks for your suggestion! Could you move your idiom lower in the document so that it's underneath the "Create a singleton" idiom?

gw111zz:
> Hi @gw111zz Thanks for your suggestion! Could you move your idiom lower in the document so that it's underneath the "Create a singleton" idiom?

Thanks for your comment @sarahhaggarty . I've made the suggested change.

docs/topics/idioms.md
```diff
@@ -125,6 +125,21 @@ object Resource {
 }
 ```
 
+## Use inline value classes to wrap single values for type-safety
```
sarahhaggarty, line 128:
```suggestion
## Use inline classes for type-safe values
```

docs/topics/idioms.md
```diff
@@ -125,6 +125,21 @@ object Resource {
...
+```kotlin
+@JvmInline
+value class EmployeeId(private val id: String)
+
+@JvmInline
+value class CustomerId(private val id: String)
+```
+
+Now accidentally mixing up `EmployeeId` and `CustomerId` will result in a compilation error. The inline classes will be represented by the 
+underlying type at runtime. `@JvmInline` is only needed for JVM backends.
```
sarahhaggarty, line 139:
```suggestion
If you accidentally mix up `EmployeeId` and `CustomerId`, a compilation error is triggered.

> The `@JvmInline` annotation is only needed for JVM backends.
>
{type="note"}
```

docs/topics/idioms.md
```diff
@@ -125,6 +125,21 @@ object Resource {
...
+
+(See [Inline classes](inline-classes.md) for more information)
```
sarahhaggarty, line 141:
```suggestion
For more information, see [Inline value classes](inline-classes.md).
```

sarahhaggarty (changes requested):
Thanks for your suggestion @gw111zz !
I've suggested some changes to bring the text in line with our style guide. Please can you apply them? Then we can go ahead with the merge! 🚀

sarahhaggarty:
Hi @gw111zz 
Are you able to implement these changes? If I don't hear from you within a week, I'll combine these changes into another PR.

sarahhaggarty:
Changes add in #4088!

PR #4053, opened 2024-02-26
test: updated tests for api reference footer

berezinant:


berezinant:
closing for now, will do it another way later

PR #4059, opened 2024-02-28
Update idioms.md - Adding !! concept  

sachinletsgo:
Adding one of the most important concepts in this doc (!!)

### DIFF
--- a/docs/topics/idioms.md
+++ b/docs/topics/idioms.md
@@ -196,6 +196,14 @@ value?.let {
 }
 ```
 
+## Convert value to a non-nullable type
+
+```kotlin
+val value = ...
+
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @sachinletsgo !
Thanks for your suggestion of a new idiom to our docs. Your suggestion does demonstrate the use of the not-null assertion operator (`!!`) but I'm sorry to say that it's not an idiom itself. In our idioms, we're looking for concise constructs that developers regularly use with operators like this. So for now, I'll close the PR. However, if you do come across a nice construct or slightly bigger example that you regularly use that also incorporates the `!!` operator, feel free to suggest it!

PR #4071, opened 2024-03-06
fix: abstract away the name of the directory

zamulla:
The directory is not necessarily named "shared" even in wizard-generated projects.


koshachy (approved):
Feel free to merge PRs like these without needing approval.

PR #4072, opened 2024-03-06
Added a little clarification - implicit invocation of toString() in string templates

pichuzhkin:
Added info how templates really works - implicit invocation of toString()

### DIFF
--- a/docs/topics/strings.md
+++ b/docs/topics/strings.md
@@ -110,8 +110,9 @@ By default, a pipe symbol `|` is used as margin prefix, but you can choose anoth
 
 ## String templates
 
-String literals may contain _template expressions_ – pieces of code that are evaluated and whose results are concatenated into the string.
-A template expression starts with a dollar sign (`$`) and consists of either a name:
+String literals may contain _template expressions_ – pieces of code that are evaluated and whose results are concatenated into a string.
+When a template expression is processed, Kotlin automatically calls the `.toString()` function on the expression's result
+to convert it into a string. A template expression starts with a dollar sign (`$`) and consists of either a variable name:
 
 ```kotlin
 fun main() {
@@ -119,6 +120,11 @@ fun main() {
     val i = 10
     println("i = $i") 
     // i = 10
+    
+    val letters = listOf("a","b","c","d","e")
+    println("Letters: $letters") 
+    // Letters: [a, b, c, d, e]
+
 //sampleEnd
 }
 ```


nikitabobko:
@pichuzhkin I'm curious what is your motivation for adding this clarification? What benefit does this clarification add?

The documentation should explain things, but it shouldn't turn into a verbose wall of obvious text, otherwise it'd be too hard to read it. There should be a balance.

Have you stumbled across some puzzler? Or the motivation is different?

docs/topics/strings.md
```diff
@@ -111,14 +111,18 @@ By default, a pipe symbol `|` is used as margin prefix, but you can choose anoth
...
 
 ```kotlin
 fun main() {
 //sampleStart
     val i = 10
     println("i = $i") 
     // i = 10
+    val letters = listOf("a","b","c","d","e")
+    println("List of letters: $letters") // evaluates as letters.toString() 
+    // List of letters: [a, b, c, d, e]
```
nikitabobko, line 125: It's better to mention something with an explicit `toString` defined to make the example more obvious. Like this:

```
val obj = object { override fun toString(): String = "Hello world" }
println("obj = $obj")
// obj = Hello world
```

docs/topics/strings.md
```diff
@@ -132,6 +136,9 @@ fun main() {
     val s = "abc"
     println("$s.length is ${s.length}") 
     // abc.length is 3
+    val numbers = listOf(1,2,3,4,5)
+    println("Last two numbers: ${numbers.takeLast(2)}")
+    // Last two numbers: [4, 5]
```
nikitabobko, line 141: I think this addition is redundant. We already made a point about `toString` in the previous example

nikitabobko (changes requested):
Thanks for the elaboration, yes I agree that this part of the doc can be improved. `toString` is not mentioned in the "String templates" section at all, but it'd be good to mention it

> underlying mechanism can't be understood using Ctrl+Click somewhere

I like this benchmark

---

Overall the idea of the PR looks good to me except for the two comments that I left. 

@sarahhaggarty do you have any comments?

docs/topics/strings.md
pichuzhkin: Oh, really, that's very nice - I'm not so familiar with anonymous objects so I couldn't even think of such an example. Maybe we should keep my example and add yours also?

docs/topics/strings.md
pichuzhkin: I agree, that's redundant. Thanks!

docs/topics/strings.md
nikitabobko: > Maybe we should keep my example and add yours also?

What benefit does `listOf` example add?

pichuzhkin: > What benefit does `listOf` example add?

I supposed that demonstration of all three behaviours (let's call it three levels of difficulty) will be most illustrative:
* Easy level - A simple int 
* Medium level -  A collection type 
* Hard level - An overridden toString() - thank you for this contribution

I won't insist on adding listOf() example if I don't convince you with this comment

sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution and insight @pichuzhkin! I added some changes on top of yours to match our style guide. Thanks again!

PR #4073, opened 2024-03-07
Update kotlin-metadata-jvm release version

sandwwraith:
The latest version is 0.9.0. This only affects the header on https://kotlinlang.org/api/kotlinx-metadata-jvm/, as actual documentation contents are built from Kotlin `master`, am I right?


nikpachoo (approved):
@sandwwraith, you are right. The version only affects the header.  
We can eliminate the version from our code if you want by putting it on library's side.

Right now we are using the following code to build the references:
```
./gradlew :kotlinx-metadata-jvm:dokkaHtml -PkotlinxMetadataDeployVersion=${KOTLINX_METADATA_JVM_RELEASE_TAG}
```
So, that might be on the library's side easily, and we call smth like './build-reference.sh' instead of it.

sandwwraith:
A different approach is required now, see https://youtrack.jetbrains.com/issue/KTL-1399

PR #4081, opened 2024-03-14
fix: Native get started missing a plugin

zamulla:
Users need to install the [Native Debugging Support](https://plugins.jetbrains.com/plugin/12775-native-debugging-support) plugin to see the run configuration.

https://youtrack.jetbrains.com/issue/KT-64948/


danil-pavlov (approved):

danil-pavlov (commented):
Great, thank you! I think you can remove the old output screenshots "native-output-N" too

PR #4083, opened 2024-03-14
wasm: remove libraries page and set redirects

AlejandraPedroza:


igoriakovlev (approved):

docs/topics/wasm/wasm-overview.md
```diff
@@ -30,7 +30,11 @@ format is platform-independent because it runs on its own virtual machine. Wasm
 a compilation target to run on the web.
 
 Kotlin/Wasm compiles your Kotlin code into Wasm format. Using Kotlin/Wasm, you can create applications that run on 
-different environments and devices, which support Wasm and meet Kotlin's requirements.
+different environments and devices, which support Wasm and meet Kotlin's requirements. 
+
+Additionally, you can use the most popular Kotlin libraries in Kotlin/Wasm out-of-the-box. Like other Kotlin and Multiplatform
```
daniCsorbaJB, line 35: smallest of notes: 
I think it would be better to use "out of the box" - instead of out-of-the-box here. 
I'd use out-of-the-box when it's an adjective directly modifying a noun, like in "an out-of-the-box solution".

daniCsorbaJB (commented):
Looks good to me! :D - good addition! 🚀 I added one small note

daniCsorbaJB (approved):
Nice work! :D Thanks for the quick change - looks good to me! 🚀

PR #4087, opened 2024-03-20
chore: adding an example to applying plugins for clarification

daniCsorbaJB:


docs/topics/gradle/gradle-configure-project.md
```diff
@@ -18,19 +18,23 @@ from the Gradle plugins DSL:
 <tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
-// replace `<...>` with the plugin name 
+// Replace `<...>` with the plugin name appropriate for your target environment.
```
zamulla, line 21: We don't put periods at the end of line comments (not an explicit rule, but can be seen in [the example](https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit#heading=h.sckrree1jn5r).)

zamulla (approved):
A tiny correction, otherwise good to go.

PR #4089, opened 2024-03-21
fix: `when` must have `else` branch if its subject is a enum/sealed/Boolean AND it's nullable

PetrakovichVictoria:
This is the important moment) Or maybe I don't know English:))

Ok:
Screenshot 2024-03-21 at 11 54 42

Nullable `Bit` and an absent `else` branch, not ok:
Screenshot 2024-03-21 at 11 54 59

### DIFF
--- a/docs/topics/control-flow.md
+++ b/docs/topics/control-flow.md
@@ -92,7 +92,7 @@ val numericValue = when (getRandomBit()) {
 
 In `when` _statements_, the `else` branch is mandatory in the following conditions:
 * `when` has a subject of a `Boolean`, [`enum`](enum-classes.md),
-or [`sealed`](sealed-classes.md) type, or their nullable counterparts.
+or [`sealed`](sealed-classes.md) type, and their nullable counterparts.
 * branches of `when` don't cover all possible cases for this subject.
 
 ```kotlin


docs/topics/control-flow.md
```diff
@@ -92,7 +92,7 @@ val numericValue = when (getRandomBit()) {
 
 In `when` _statements_, the `else` branch is mandatory in the following conditions:
 * `when` has a subject of a `Boolean`, [`enum`](enum-classes.md),
```
PetrakovichVictoria, line 94: Some more details for the example in the "Conversation" tab. For a non-nullable enum, an else branch is not needed (and the current wording with "or" sounds to me like it is needed). For a nullable enum, it's obligatory. This is why I think that `and` works better here.

PetrakovichVictoria:
I’d say it would be better to say sth like 
when has a **nullable** subject of a Boolean, enum, or sealed type.

It’s up to you to decide whether to leave it as is or maybe to use an option as I've just proposed:)

sarahhaggarty:
As we discussed on Slack, I'm going to rewrite this section due to some other related work I'm doing on the tour. So I'll close this PR and create a new one when the updated content is ready. But thanks for pointing out the issue @PetrakovichVictoria ! 🙏

PR #4097, opened 2024-03-25
Update multiplatform-publish-lib.md

martinbonnin:
Use `androidTarget` instead of deprecated `android`


zamulla (approved):
Thanks for the update!

PR #4107, opened 2024-04-02
Update comparison to Java

antonarhipov:
- Java: instead of unnamed variables and record patterns, link to the Java Pattern Matching design document
- Java: package-private visibility
- Kotlin: explicit API mode & Opt-In requirements


koshachy (approved):
lgtm

PR #4114, opened 2024-04-03
fix(ktl-1335): removed kotlin variable from test-page

berezinant:
to ensure proper testing by screenshots

### DIFF
--- a/docs/topics/test-page.md
+++ b/docs/topics/test-page.md
@@ -17,7 +17,7 @@
 
 ```kotlin
 plugins {
-    kotlin("kapt") version "%kotlinVersion%"
+    kotlin("kapt") version "1.9.23"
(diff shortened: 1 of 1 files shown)


docs/topics/test-page.md
```diff
@@ -17,7 +17,7 @@
 
 ```kotlin
 plugins {
-    kotlin("kapt") version "%kotlinVersion%"
```
nikpachoo, line 20: Why do we need to remove the variable? If it is a Docs functionality, I would like to keep it and adapt the tests.

nikpachoo (changes requested):

zoobestik (approved):

nikpachoo (approved):

PR #4115, opened 2024-04-05
feat: RC1 what's new update

danil-pavlov:


docs/topics/whatsnew-eap.md
```diff
@@ -800,12 +856,86 @@ keep in mind that you might encounter deprecation warnings or some new Gradle fe
...
+> Use it only for evaluation purposes. We would appreciate your feedback on it in [YouTrack](https://kotl.in/issue).
+>
+{type="warning"}
+
+Prior to Kotlin %kotlinEapVersion%, configuring compiler options in a multiplatform project with Gradle was only possible
+at a low level, such as per task, compilation, or source set. To make it easier to configure compiler options more generally
+in your projects, Kotlin %kotlinEapVersion% comes with a new Gradle DSL.
+
+With this new DSL, you can configure compiler options at the extension level for all the targets and shared source sets
+like `commonMain` and at a target level for a specific target:
```
koshachy, line 879:
```suggestion
With this new DSL, you can configure compiler options at the extension level for all the targets and shared source sets,
such as `commonMain`, as well as at a target level for a specific target:
```

docs/topics/whatsnew-eap.md
```diff
@@ -800,12 +856,86 @@ keep in mind that you might encounter deprecation warnings or some new Gradle fe
...
+The settings at a higher level are used as a convention (defaults) for a lower level:
+
+* The extension compiler options values are the default for target compiler options, including shared source sets,
+  like `commonMain`, `nativeMain`, and `commonTest`.
+* Target compiler options values are used as the default for compilation unit (task) compiler options, for example,
+  `compileKotlinJvm` and `compileTestKotlinJvm` tasks.
+
+In turn, the configuration made at a lower level overrides related settings at a higher level:
+
+
```
koshachy, line 912:
```suggestion
```

docs/topics/whatsnew-eap.md
```diff
@@ -1085,18 +1215,13 @@ fun writeBooksTo(writer: XMLWriter) {
 
 ## What to expect from upcoming Kotlin EAP releases
 
-The upcoming 2.0.0-RC1 release will further increase the stability of the K2 compiler.
+Starting from **Kotlin 2.0.0-RC1**, you can start using the K2 compiler in production.
```
koshachy, line 1218:
```suggestion
Starting from **Kotlin 2.0.0-RC1**, you can use the K2 compiler in production.
```

docs/topics/whatsnew-eap.md
```diff
@@ -572,7 +572,55 @@ We would appreciate any feedback you may have!
 * [Enable the Send usage statistics option](https://www.jetbrains.com/help/idea/settings-usage-statistics.html) to
   allow JetBrains to collect anonymous data about K2 usage.
 
-## Kotlin/Native: resolving conflicts with Objective-C methods
+## Kotlin/JVM
+
+This version brings the following changes:
+
+* [Generate lambda functions using invokedynamic](#generate-lambda-functions-using-invokedynamic)
```
koshachy, line 579:
```suggestion
* [Generate lambda functions using `invokedynamic`](#generate-lambda-functions-using-invokedynamic)
```

docs/topics/whatsnew-eap.md
```diff
@@ -572,7 +572,55 @@ We would appreciate any feedback you may have!
...
+* [The kotlinx-metadata-jvm library is now Stable](#the-kotlinx-metadata-jvm-library-is-now-stable)
```
koshachy, line 580:
```suggestion
* [The `kotlinx-metadata-jvm` library is now Stable](#the-kotlinx-metadata-jvm-library-is-now-stable)
```

koshachy (approved):
Nice job!
Look through the comments, please.

docs/topics/whatsnew-eap.md
```diff
@@ -800,12 +856,86 @@ keep in mind that you might encounter deprecation warnings or some new Gradle fe
...
+Gradle attribute is published by default with all Kotlin variants.
+
+The attribute helps distinguish JVM and Android variants of Kotlin Multiplatform libraries. It indicates that a certain
+library variant is better suited for a certain JVM environment. The target environment could be "android", "stardard-jvm",
+or "no-jvm".
+
+Publishing this attribute should make consuming Kotlin Multiplatform libraries with JVM and Android targets more robust
+from non-multiplatform clients as well, such as Java-only projects.
+
+If necessary, you can disable attribute publication. To do that, add the following Gradle option to your `gradle.properties` file:
```
zamulla, line 933:
```suggestion
If necessary, you can disable publication of this attribute. To do that, add the following Gradle option to your `gradle.properties` file:
```

So as not to say "disable publication of all attributes."

docs/topics/whatsnew-eap.md
```diff
@@ -800,12 +856,86 @@ keep in mind that you might encounter deprecation warnings or some new Gradle fe
...
+{type="warning"}
+
+Prior to Kotlin %kotlinEapVersion%, configuring compiler options in a multiplatform project with Gradle was only possible
+at a low level, such as per task, compilation, or source set. To make it easier to configure compiler options more generally
+in your projects, Kotlin %kotlinEapVersion% comes with a new Gradle DSL.
+
+With this new DSL, you can configure compiler options at the extension level for all the targets and shared source sets
+like `commonMain` and at a target level for a specific target:
+
+```kotlin
```
zamulla, line 882: I feel that the example comes between two parts of the conceptual explanation (and that leads to "extension level" being explained later than it is mentioned). More natural sequence would be "... was only possible at low level → project configuration now has three layers → here's an example → here's nuance about conventions".

zamulla (changes requested):
Minor suggestions about phrasing and sequence.

docs/topics/whatsnew-eap.md
danil-pavlov: I think this sequence would work for the actual documentation. Here, in What's new, we need to announce the news, the new DSL. The whole section starting with "3 layers" is additional information to explain how the overall picture looks like now, but it can be removed without losing the point of the announcement

docs/topics/whatsnew-eap.md
```diff
@@ -572,7 +572,55 @@ We would appreciate any feedback you may have!
...
+
+Since the first version, Kotlin has generated lambdas as anonymous classes. However, starting from [Kotlin 1.5](whatsnew15.md#lambdas-via-invokedynamic),
+the option for `invokedynamic` generation was available by using the `-Xlambdas=indy` compiler flag. In Kotlin %kotlinEapVersion%,
+`invokedynamic` has become the default method for lambda generation. This method produces lighter binaries and aligns
+Kotlin with JVM optimizations, ensuring applications benefit from ongoing and future improvements in JVM performance.
+
+Currently, it has three limitations compared to ordinary lambda compilation:
+
+* A lambda compiled into `invokedynamic` is not serializable.
+* Calling `toString()` on such a lambda produces a less readable string representation.
```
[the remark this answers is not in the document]
daniCsorbaJB: Sure, we can put it as the last point, but I don't think it can be clarified in a simple way without including a long example which would make it far less readable. 

I think we should include it after it leaves RC in the docs, but in the what's new section, it would feel a bit too much to me.

docs/topics/whatsnew-eap.md
```diff
@@ -572,7 +572,55 @@ We would appreciate any feedback you may have!
...
+
+The `kotlin-metadata-jvm` library provides an API to read and modify metadata of binary files generated by the Kotlin/JVM
+compiler.
+
+## Kotlin/Native
+
+This version brings the following changes:
+
+* [Resolving conflicts with Objective-C methods](#resolving-conflicts-with-objective-c-methods)
+* [Changed log level for compiler arguments in Kotlin/Native](#changed-log-level-for-compiler-arguments)
```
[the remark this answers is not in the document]
danil-pavlov: I checked, and we don't have consistency in these header lists throughout different WNs, as well as this one. I'll revisit it in RC2
For this section in particular, I'd like to keep the sense(usefulness) of the feature in the header, for example, "Conflict resolution with Objective-C methods", but change the second header to look similar

docs/topics/whatsnew-eap.md
zamulla: My main point is about the "extension level"; if the intended audience doesn't need the term introduced here, then it's all good.

zamulla (approved):
👍

PR #4119, opened 2024-04-09
Update kotlin-tour-classes.md

japygo:
Change a `//sampleStart` in `Print as string`.
Because if `val user = User("Alex", 1)` didn't see when test the code then confused to understand the code.


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @japygo!

PR #4120, opened 2024-04-09
Update kotlin-tour-null-safety.md

japygo:
`nullString` is never modified, so it can be declared using 'val'


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @japygo !

PR #4123, opened 2024-04-10
Fix the type of "state" to match the defined sealed class type.

jamhour1g:


daniCsorbaJB (approved):
Hi @jamhour1g! Great catch; thank you for your contribution! 🚀 💯

PR #4140, opened 2024-04-18
fix: change Maven version from 3.5.1 to 3.8.6 as 3.5.1 doesn't exist:)

PetrakovichVictoria:
The version 3.5.1 doesn't exist, see here: https://maven.apache.org/docs/history.html#maven-3-8-x

3.8.6 is chosen by advice from @CherepanovAleksei as we [use](https://jetbrains.team/p/kt/repositories/kotlin/files/ff2fea390bafb765abebc11bb272c8e2e999f46f/libraries/pom.xml?tab=source&line=54&lines-count=1) this version.

### DIFF
--- a/docs/topics/lombok.md
+++ b/docs/topics/lombok.md
@@ -151,7 +151,7 @@ In Maven, use the following settings to launch Lombok with Java's compiler:
 <plugin>
     <groupId>org.apache.maven.plugins</groupId>
     <artifactId>maven-compiler-plugin</artifactId>
-    <version>3.5.1</version>
+    <version>3.8.6</version>
     <configuration>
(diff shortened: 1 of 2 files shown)


CherepanovAleksei (dismissed):
LGTM

koshachy (approved):
Hey @PetrakovichVictoria
Thanks for your contribution!

I'll wait @CherepanovAleksei for an approval.

CherepanovAleksei:
ftr, we need to revise the versions - the proposed version refers to the mvn version, but not to the version maven compiler plugin (https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-compiler-plugin)

PR #4146, opened 2024-04-23
feat: avoid "string" as parameter name to improve readability

mikeloguay:
In general, I think it is a bad idea to use name parameters with the same name of its type. I am coming from C#, and there both `string` and `String` are the reserved words for the type name. It could be confusing and easily avoided using `message`, `word`, or something else instead.


mikeloguay:
Hi @sarahhaggarty, no problem for the delay. I was following the tour and just thought this was a bit confusing.

OK about the rename to `text`. Done!

sarahhaggarty (approved):

mikeloguay:
> Great! Thanks for your contribution @mikeloguay !

To you!

PR #4176, opened 2024-05-07
update: add limitations for open expect classes (KT-22841)

zamulla:
Clarifying the limitations of actualizing an open expect class.

### DIFF
--- a/docs/topics/multiplatform/multiplatform-expect-actual.md
+++ b/docs/topics/multiplatform/multiplatform-expect-actual.md
@@ -406,10 +406,10 @@ expected members.
 
 This means that `actual` overrides cannot change:
 
-* Modality of a class member.
+* Modality of a class member (`open`, `sealed`, `abstract`, and `final`).
   An exception to this rule: `expect final` members can be actualized by `actual open` members.
 * Return type of a function (covariant overrides).
-* Visibility of a member.
+* [Visibility](visibility-modifiers.md) of a member.
   An exception to this rule: visibility of an `actual` member can be more permissive if the corresponding `expect` member is `final`.
 * Visibility of a property setter.
 * Variable keywords (`val` cannot be actualized as `var` and vice versa).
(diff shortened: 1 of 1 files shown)


docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -399,6 +399,24 @@ typically provides an expected `CommonViewModel` class whose actual Android coun
 from the Android framework. See [Use platform-specific APIs](multiplatform-connect-to-apis.md#adapting-to-an-existing-hierarchy-using-expected-actual-classes)
 for a detailed description of this example. -->
 
+#### Limitations of open expected classes
```
nikitabobko, line 402: I don't know if it's common in our documentation, but "expected" feels wrong to me.

`expect` is a keyword and converting it to the adjective sounds incorrect

nikitabobko, line 402: Please drop "open". The rules that you describe are applicable to both open and non-open classes.

docs/topics/multiplatform/multiplatform-expect-actual.md
```diff
@@ -399,6 +399,24 @@ typically provides an expected `CommonViewModel` class whose actual Android coun
...
+expected members.
+
+This means that `actual` overrides cannot change:
+
+* Modality of a class member (`open`, `sealed`, `abstract`, and `final`).
+  An exception to this rule: `expect final` members can be actualized by `actual open` members.
+* Return type of a function (covariant overrides).
+* [Visibility](visibility-modifiers.md) of a member.
+  An exception to this rule: visibility of an `actual` member can be more permissive if the corresponding `expect` member is `final`.
+* Visibility of a property setter.
```
nikitabobko, line 414: I think this item can be skipped (the previous item speaks in terms of "members", so the previous item is a superset. Properties and functions are both considered members)

But it's up to you

nikitabobko (changes requested):

docs/topics/multiplatform/multiplatform-expect-actual.md
zamulla: I don't disagree, and I avoided using the adjective in the text. But writing "expect classes" in the title, when we cannot highlight the keyword, reads like a grammar mistake. All the rest of the modifiers I can remember are adjectives to begin with (internal, abstract, actual...), so they don't have this problem. Here, we need a compromise.

Right now the concept we introduce on the page is "expected declarations," and that's the convention. I think it's possible to have clearer terminology here. We'll keep thinking about this.

nikitabobko (approved):

PR #4180, opened 2024-05-09
update: make enumEntries<T>() Stable

ElviraMustafina:


qwwdfsad (approved):

koshachy (commented):
In this file there are a lot of mentions of the `enumValues<T>()`.

Shouldn't they be replaced by `enumEntries<T>()` as well?

koshachy (approved):

PR #4182, opened 2024-05-10
Docs: Improve Sample Code

Attacktive:
- renamed the local variables:
name `b` was used as the constructor parameter AND a local variable.
- removed `val` from the constructor parameter
image


Attacktive:
Oops. Sorry for the noise. I'm going to open another. :disappointed:

PR #4183, opened 2024-05-10
Docs: Improve Sample Code

Attacktive:
- renamed the local variables:
name `b` was used as the constructor parameter AND a local variable.
- removed `val` from the constructor parameter
Screenshot 2024-05-10 11:32:48


AlejandraPedroza (approved):
Great catches! Thanks for your contribution.

PR #4188, opened 2024-05-13
Remove equals symbol from jvm arg parameters

mezpahlan:
Including the equal symbol for the `-Xms` arg caused my Kotlin Daemon to crash on startup. I tracked it down to the equal symbol and copying from these help docs.


AlejandraPedroza (approved):
Nice catch. Thanks for your contribution!

PR #4193, opened 2024-05-14
update: add info about test source set isolation (KT-66047)

zamulla:


zamulla:
This has been solved in another PR https://github.com/JetBrains/kotlin-web-site/pull/4297

PR #4196, opened 2024-05-15
Fixed expected output string.

nthnn:


koshachy:
Hey there!

This comment string isn't what we expected; the actual output is shown in a comment below. The main point of this code is to demonstrate that `apple` will be overwritten, and the result will be 140.

I'm closing this PR, but feel free to ask any questions.

nthnn:
My bad, dumb mistake. Didn't saw the last entry.

PR #4201, opened 2024-05-15
Update smart cast documentation for Kotlin 2.0.0

sarahhaggarty:
This PR updates smart cast documentation for Kotlin 2.0.0.

### DIFF
--- a/docs/topics/typecasts.md
+++ b/docs/topics/typecasts.md
@@ -1,6 +1,6 @@
 [//]: # (title: Type checks and casts)
 
-In Kotlin, you can perform type checks to check the type of an object at runtime. Type casts convert objects to a 
+In Kotlin, you can perform type checks to check the type of an object at runtime. Type casts enable you to convert objects to a 
 different type.
 
 > To learn specifically about **generics** type checks and casts, for example `List<T>`, `Map<K,V>`, see [Generics type checks and casts](generics.md#generics-type-checks-and-casts).
@@ -9,7 +9,7 @@ different type.
 
 ## is and !is operators
 
-Use the `is` operator or its negated form `!is` to perform a runtime check that identifies whether an object conforms to a given type:
+To perform a runtime check that identifies whether an object conforms to a given type, use the `is` operator or its negated form `!is`:
 
 ```kotlin
 if (obj is String) {
@@ -25,8 +25,9 @@ if (obj !is String) { // Same as !(obj is String)
 
 ## Smart casts
 
-In most cases, you don't need to use explicit cast operators in Kotlin because the compiler tracks the
-`is`-checks and [explicit casts](#unsafe-cast-operator) for immutable values and inserts (safe) casts automatically when necessary:
+In most cases, you don't need to use explicit cast operators because the compiler automatically casts objects for you.
+This is called smart-casting. The compiler tracks the type checks and [explicit casts](#unsafe-cast-operator) for immutable
+values and inserts implicit (safe) casts automatically when necessary:
 
 ```kotlin
 fun demo(x: Any) {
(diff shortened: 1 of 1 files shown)


docs/topics/typecasts.md
```diff
@@ -110,27 +255,29 @@ Smart casts can be used in the following conditions:
...
+To make the example work for possible null values, use a nullable type on the right-hand side of the cast:
 
 ```kotlin
 val x: String? = y as String?
 ```
 
 ## "Safe" (nullable) cast operator
 
-To avoid exceptions, use the *safe* cast operator `as?`, which returns `null` on failure.
+To explicitly cast an object to a nullable type, use the *safe* cast operator `as?`:
```
nikitabobko, line 275: "explicitly cast an object to a nullable type" is technically incorrect

`y as String?` - cast to a nullable type
`y as? String` - it's just a safe-cast

The difference is that if `y` has an unrelated type (e.g. `Int`), safe-cast will return `null`, but cast to nullable will throw an Exception

docs/topics/typecasts.md
```diff
@@ -36,15 +37,64 @@ fun demo(x: Any) {
...
+```kotlin
+when (x) {
+    is Int -> print(x + 1)
+    is String -> print(x.length + 1)
+    is IntArray -> print(x.sum())
+}
+```
+
+From Kotlin 2.0.0, if you declare a variable before using it in your `if`, `when`, or `while` condition, then any
+information collected by the compiler about the variable will be accessible in the condition statement and its block for
```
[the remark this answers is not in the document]
sarahhaggarty: I used your suggestion but with "corresponding" instead.

docs/topics/typecasts.md
```diff
@@ -36,15 +37,64 @@ fun demo(x: Any) {
...
+    }
+}
+
+fun petAnimal(animal: Any) {
+    val isCat = animal is Cat
+    if (isCat) {
+        // In Kotlin 2.0.0, the compiler can access
+        // information about isCat, so it knows that
+        // animal was smart-cast to the type Cat.
+        // Therefore, the purr() function is successfully called.
```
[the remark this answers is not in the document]
sarahhaggarty: I went for "the purr() function can be called". What do you think?

docs/topics/typecasts.md
sarahhaggarty: As discussed, reverted to previous text.

nikitabobko (approved):

docs/topics/typecasts.md
```diff
@@ -56,17 +106,112 @@ if (x is String && x.length > 0) {
...
+### Inline functions
+
+From Kotlin 2.0.0, the compiler can smart-cast variables captured within lambda functions that are passed to [inline functions](inline-functions.md).
+
+Inline functions are treated as having an implicit [`callsInPlace`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.contracts/-contract-builder/calls-in-place.html)
+contract. This means that any lambda functions passed to an inline function are called in place. Since lambda functions
+are called in place, the compiler knows that a lambda function can't leak references to any variables contained within
+its function body.
+
+The compiler uses this knowledge, along with other analyses to decide whether it's safe to smart-cast any of the
```
danil-pavlov, line 143:
```suggestion
The compiler uses this knowledge, along with other analyses, to decide whether it's safe to smart-cast any of the
```

docs/topics/typecasts.md
```diff
@@ -56,17 +106,112 @@ if (x is String && x.length > 0) {
...
+    }
+
+    return processor
 }
 ```
 
+### Exception handling
+
+From Kotlin 2.0.0, smart cast information is passed on to `catch` and `finally` blocks. This change makes your code safer
+as the compiler keeps track of whether your object has a nullable type. For example:
```
danil-pavlov, line 179:
```suggestion
as the compiler tracks whether your object has a nullable type. For example:
```

danil-pavlov (commented):
I think you can remove all the "From Kotlin 2.0.0," mentions since docs are true to the current stable version

PR #4209, opened 2024-05-16
Add compatibility guide for Kotlin 2.0.0

sarahhaggarty:
This PR adds the first phase of the compatibility guide for Kotlin 2.0.0.


docs/topics/compatibility-guides/compatibility-guide-20.md
```diff
@@ -0,0 +1,200 @@
...
+> which prevents the exposure of the Kotlin Gradle plugin DSL functions and properties to levels where they are not intended
+> to be available. The following levels are separated from each other:
+> * Kotlin extension
+> * Kotlin target
+> * Kotlin compilation
+> * Kotlin compilation task
+>
+> **Deprecation cycle**:
+>
+> - 2.0.0: if a build script is configured incorrectly, report a warning
```
Tapchicoma, line 128: actually error initially. For most popular use-cases we introduced deprecated methods in the contexts where they could be misused, but it is too much work to do for every case.  And, unfortunately, it is not possible to control the level of violation.

docs/topics/compatibility-guides/compatibility-guide-20.md
```diff
@@ -0,0 +1,200 @@
...
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: The `kotlin.useK2` Gradle property has been removed.
+>
+> **Deprecation cycle**:
+>
+> - 1.8.20: the `kotlin.useK2` Gradle property is deprecated
```
Tapchicoma, line 178: It was introduced only in `1.9.20` release, so this line should be removed

docs/topics/compatibility-guides/compatibility-guide-20.md
```diff
@@ -0,0 +1,200 @@
...
+### Deprecate old compiler option DSLs
+
+> **Issue**: [KT-67234](https://youtrack.jetbrains.com/issue/KT-67234)
+>
+> **Component**: Gradle
+>
+> **Incompatible change type**: source
+>
+> **Short summary**: The ability to configure the `compilerOptions` property in the `kotlinCompilation` DSL has been deprecated. 
+> The `kotlinOptions` DSL has also been deprecated.
```
Tapchicoma, line 139: I would split `kotlinOptions` DSL deprecation in a separate section to make it more visible. It is quite used DSL and we should highlight this change.

docs/topics/compatibility-guides/compatibility-guide-20.md
```diff
@@ -0,0 +1,200 @@
...
+| [KT-57422](https://youtrack.jetbrains.com/issue/KT-57422) | K2: Prohibit use-site 'get' targeted annotations on property getters                                                       |
+| [KT-34372](https://youtrack.jetbrains.com/issue/KT-34372) | Report missed error for virtual inline method in enum classes                                                       |
+| [KT-47986](https://youtrack.jetbrains.com/issue/KT-47986) | Forbid implicit inferring a type variable into an upper bound in the builder inference context                                                       |
+| [KT-53982](https://youtrack.jetbrains.com/issue/KT-53982) | Keep nullability when approximating local types in public signatures                                                       |
+
+## Tools
+
+### Visibility changes in Gradle
+
+> **Issue**: [KT-67227](https://youtrack.jetbrains.com/issue/KT-67227)
```
Tapchicoma, line 111: Here and below links are leading to "document change" issues, is it intended?

docs/topics/compatibility-guides/compatibility-guide-20.md
antohaby: Compiler option -Xuse-k2 was deprecated in 1.8.20 https://youtrack.jetbrains.com/issue/KT-56239

Maybe we can rephrase it. To emphasise the fact that `kotlin.useK2` is redundant and doesn't do anything.

>  **Incompatible change type**: source

And I believe it is a behavioural change. but it depends.
Let me describe potential cases that:

1. Kotlin 1.9.X; useK2 = false  => does nothing.
2. Kotlin 1.9.X; useK2 = true   => enables K2 compiler.
3. Kotlin 2.0+; useK2 = true => wouldn't have any effect.
4. Kotlin 2.0+; useK2 = false => wouldn't disable K2 compiler back to K1 

So there were potential behavioural change between 1. and 4. (i.e. it 1.9.* useK2=false used to do nothing and in 2.0+ it also keeps doing nothing, but semantically user might expect it to stop using K2).
But since we removed this option in 2.0 there is no Breaking Change :)

And Kotlin Build wouldn't fail in 2.0 release if this Gradle property is present or not. So it doesn't matter much.

danil-pavlov (approved):

Tapchicoma (approved):

PR #4216, opened 2024-05-17
feat: recommendations on klib compatibility

danil-pavlov:


docs/topics/components-stability.md
```diff
@@ -71,9 +71,10 @@ We make sure to document precisely which subcomponents are not stable. We also d
...
 | Kotlin/Native memory manager                          | Stable       | 1.9.20                   |                                                                            |
-| klib binaries                                         | Stable       | 1.9.20                   |                                                                            |
+| klib binaries                                         | Stable       | 1.9.20                   | Not including cinterop klibs, see below                                    |
 | Kotlin Multiplatform                                  | Stable       | 1.9.20                   |                                                                            |
 | Kotlin/Native interop with C and Objective C          | Beta         | 1.3                      |                                                                            |
+| cinterop klib binaries                                | Beta         | 1.3                      |                                                                            |
```
ddolovov, line 77: I thought that it's called `C interop` in other docs. Am I wrong?

docs/topics/kotlin-evolution.md
```diff
@@ -205,4 +205,21 @@ For fully stable versions of the compiler, the default binary compatibility prot
...
+Note that not all target platforms have reached this level of stability, but Kotlin/JVM has.
+
+#### Kotlin klib binaries
+
+Kotlin klib binaries have reached the [Stable](components-stability.md#stability-levels-explained) level in Kotlin 1.9.20.
+However, there are some compatibility details you need to keep in mind:
+
+* klib binaries are backwards compatible starting with Kotlin 1.9.20. For example, the 2.0.* compiler can read binaries
+  produced by the 1.9.* compiler.
+* klib binaries are compatible within the same language or feature version. For example, the 2.0.* compiler can read
```
ddolovov, line 217: I would probably use the same wording as in the previous section (for JVM) to avoid misunderstanding that this rule is different for KLIBs and JARs. Which is: `Older compilers reject klibs that rely on new features (e.g. a 1.9.20 compiler rejects binaries that use 2.0.*-specific language features)`

docs/topics/kotlin-evolution.md
```diff
@@ -205,4 +205,21 @@ For fully stable versions of the compiler, the default binary compatibility prot
...
+  binaries produced by the 2.0.* compiler.
+* Forward compatibility is _not_ guaranteed. For example, the 2.0.* compiler cannot read binaries produced by the 2.1.*
```
ddolovov, line 219: To be strict: the 2.0.* compiler ~~cannot~~ is not guaranteed to read binaries produced by the 2.1.*

docs/topics/components-stability.md
danil-pavlov: I checked our docs, when talking about interop in general, we use 'C interoperability', but with specifics, we use 'cinterop tool', 'cinterop declarations', 'cinterop libraries'

docs/topics/kotlin-evolution.md
danil-pavlov: Thank you! Added a line about older compilers

ddolovov: I meant that the whole line can be replaced by the line that you've actually added.

docs/topics/kotlin-evolution.md
```diff
@@ -205,4 +205,23 @@ For fully stable versions of the compiler, the default binary compatibility prot
...
\ No newline at end of file
+Note that not all target platforms have reached this level of stability, but Kotlin/JVM has.
+
+#### Kotlin klib binaries
+
+Kotlin klib binaries have reached the [Stable](components-stability.md#stability-levels-explained) level in Kotlin 1.9.20.
+However, there are some compatibility details you need to keep in mind:
+
+* klib binaries are backwards compatible starting with Kotlin 1.9.20. For example, the 2.0.* compiler can read binaries
+  produced by the 1.9.* compiler.
```
ddolovov, line 216: ... produced by the 1.9.20 compiler (or 1.9.20+) -- as far as we are talking about starting to maintain these rules starting with 1.9.20.

docs/topics/kotlin-evolution.md
danil-pavlov: Ah, I see. thank you for the clarification!

docs/topics/kotlin-evolution.md
```diff
@@ -205,4 +205,21 @@ For fully stable versions of the compiler, the default binary compatibility prot
...
+Note that not all target platforms have reached this level of stability, but Kotlin/JVM has.
+
+#### Kotlin klib binaries
+
+Kotlin klib binaries have reached the [Stable](components-stability.md#stability-levels-explained) level in Kotlin 1.9.20.
+However, there are some compatibility details you need to keep in mind:
+
+* klib binaries are backwards compatible starting with Kotlin 1.9.20. For example, the 2.0.x compiler can read binaries
+  produced by the 1.9.2x compiler.
+* Older compilers reject klib binaries that rely on new features. For example, the 1.9.20 compiler rejects binaries that
```
erokhins, line 217: Well... I think that we should just remove this phrase because the next one is saying that forward compatibility is not guaranteed at all, and this part in JVM was added for forward compatibility cases

docs/topics/kotlin-evolution.md
```diff
@@ -205,4 +205,21 @@ For fully stable versions of the compiler, the default binary compatibility prot
```
erokhins, line 205: Btw, let's update 
> in the cases when new features are not used, for example, 1.3 can understand most binaries from 1.4, but not 1.5

to 
> in the cases when new features are not used, for example, 1.9 can understand most binaries from 2.0, but not 2.1

Just to highlight that with 2.0 it also works

erokhins (approved):
Added couple comments, the rest is good, thank you!

docs/topics/kotlin-evolution.md
danil-pavlov: OK, let's leave only backward and forward compatibility recommendations then

koshachy (approved):

PR #4221, opened 2024-05-20
feat: add the Kotlin Power-assert plugin docs

koshachy:


docs/topics/power-assert.md
```diff
@@ -0,0 +1,380 @@
...
+        }
+    }
+}
+```
+
+In the output, all the `assert()` function error messages will be printed one after another:
+
+```text
+Charlie has an invalid salary: 40000
+Dave has an invalid age: 150
```
bnorm, line 374: This is not the output I would have expected if power-assert was enabled for the `AssertScope.assert` function. I would expect something more along the lines of the following, if the function was configured correctly.

```text
Charlie has an invalid salary: 40000
assert(employee.salary > 50000) { "${employee.name} has an invalid salary: ${employee.salary}" }
       |        |
       |        40000
       Employee(name="Charlie", age=55, salary=40000)
Dave has an invalid age: 150
assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.age}" }
       |        |
       |        150
       Employee(name="Dave", age=150, salary=70000)
```

I think configuring the function is just missing, and you need something like the following to make this work as expected.

```kotlin
powerAssert {
    functions = listOf("[package AssertScope is from].AssertScope.assert")
}
```

docs/topics/power-assert.md
```diff
@@ -0,0 +1,380 @@
...
+Value should not be empty
+require(value.isNotEmpty()) { "Value should not be empty" }
+        |     |
+        |     false
+        
+```
+
+The message shows the intermediate values that lead to the failure, making it easier to debug.
+
+### Function call tracing
```
bnorm, line 276: I don't think this example is equivalent to the one from kotlin-power-assert. What might have been missing is I excluded the explicit configuring of the `dbg` function from the kotlin-power-assert example for brevity, which may have been a mistake. Let's consider the other example again for a moment.

```kotlin
fun <T> dbg(value: T): T = value

fun <T> dbg(value: T, msg: String): T {
    println(msg)
    return value
}

fun main() {
    println(dbg(1 + 2 + 3))
}
```

with output:

```text
dbg(1 + 2 + 3)
      |   |
      |   6
      3
6
```

This example basically shows that a function doesn't always need to throw an exception to be useful. Instead, we call the `dbg(T)` function, but power-assert rewrites the call to the `dbg(T, String)` function and includes the call diagram. This allows us to "log" the call parameters with `println`.

So if you have the `dbg` function configured with power-assert, it's sort of like `println` that you can wrap stuff with and get it automatically logged to the console with contextual information. This is more of a main code example, rather than a test code example. Hopefully that explains better what this example is attempting to demonstrate.

docs/topics/power-assert.md
```diff
@@ -0,0 +1,380 @@
...
+
+```kotlin
+class ComplexExampleTest {
+
+    data class Person(val name: String, val age: Int)
+
+    @Test
+    fun testComplexAssertion() {
+        val person = Person("Alice", 30)
+        assert(person.name.startsWith("A") && person.name.length > 3 && person.age > 20 && person.age < 29)
```
bnorm, line 208: Consider using `Person("Alice", 10)` or some age less or equal to 20. This will result in the following output (manually typed...), which shows that power-assert handles boolean short-circuiting correctly.

```text
Assertion failed
assert(person.name.startsWith("A") && person.name.length > 3 && person.age > 20 && person.age < 29)
       |      |    |                  |      |    |      |      |      |   |
       |      |    |                  |      |    |      |      |      |   false
       |      |    |                  |      |    |      |      |      10
       |      |    |                  |      |    |      |      Person(name=Alice, age=10)
       |      |    |                  |      |    |      true
       |      |    |                  |      |    5
       |      |    |                  |      Alice
       |      |    |                  Person(name=Alice, age=10)
       |      |    true
       |      Alice
       Person(name=Alice, age=10)
```

You could then maybe add a note about using `and` and `or` expressions if short-circuiting is not desired.

docs/topics/power-assert.md
koshachy: Exactly, I forgot to add both `"org.example.AssertScope.assert", "org.example.assertSoftly"`functions

docs/topics/power-assert.md
koshachy: Maybe I'm doing something wrong, but it doesn't work for me.
I specify the `dbg` function in the `powerAssert {}` block but `println()` shows me the result without any intermediate output.

docs/topics/power-assert.md
bnorm: `org.example.assertSoftly` is not needed and will result in a compiler warning, only `org.example.AssertScope.assert` is needed.

docs/topics/power-assert.md
bnorm: Hmm... interesting. To avoid delaying documentation, I would say exclude this example for now and we can add it later with some additional documentation on what the compiler plugin expects for function signatures. It's kind of an advanced use case that I doubt many people will be interested in right away anyways.

docs/topics/power-assert.md
```diff
@@ -0,0 +1,399 @@
...
+       |      |    true
+       |      Alice
+       Person(name=Alice, age=10)
+```
+
+### Beyond assert function
+
+The Power-assert plugin can transform various functions beyond `assert` which is transformed by default.
+Functions like `require()`, `check()`, `assertTrue()`, `assertEqual()` and others can also be transformed,
+if they have a form that allows taking a `String` or `() -> String` value as the last parameter.
```
bnorm, line 232: Just as a note for a followup, I would like to see this expanded into more details in the future. It's enough for now, but we will probably get questions about the exact requirements, and more details here should help.

docs/topics/power-assert.md
koshachy: Ok, I'll comment this one

docs/topics/power-assert.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+</tab>
+</tabs>
+
+## Configure the plugin
+
+The Power-assert plugin provides several options to customize its behavior:
+
+- **`functions`**: A list of complete function paths that the Power-assert plugin will transform. If empty, only `kotlin.assert()` will be transformed by default.
```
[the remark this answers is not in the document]
koshachy: The wording is important.
I rollbacked the fully-quilified here, it seems OK.

docs/topics/power-assert.md
koshachy: Yeah, let's plan it as a next step! Thanks for you feedback!

bnorm (approved):

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+       Person(name=Alice, age=10)
+```
+
+### Beyond assert function
+
+The Power-assert plugin can transform various functions beyond `assert` which is transformed by default.
+Functions like `require()`, `check()`, `assertTrue()`, `assertEqual()` and others can also be transformed,
+if they have a form that allows taking a `String` or `() -> String` value as the last parameter.
+
+Before adding a new function in a test, specify the function in the `powerAssert {}` block of your build script file.
```
zamulla, line 234:
```suggestion
Before using a new function in a test, specify the function in the `powerAssert {}` block of your build script file.
```

Either this, or "adding **to**..."

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+       |        |   |
+       |        |   false
+       |        150
+       Employee(name=Dave, age=150, salary=70000)
+```
+
+## What's next
+
+* Look through a [simple project with the plugin enabled](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-integration-tests/src/test/resources/testProject/powerAssertSimple)
+   and more [complex project with multiple source sets](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-integration-tests/src/test/resources/testProject/powerAssertSourceSets)
```
zamulla, line 400:
```suggestion
   and a more [complex project with multiple source sets](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-integration-tests/src/test/resources/testProject/powerAssertSourceSets).
```

Should this be a list, with only a single item?

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+
+</tab>
+</tabs>
+
+## Configure the plugin
+
+The Power-assert plugin provides several options to customize its behavior:
+
+* **`functions`**: A list of fully-qualified function paths. The Power-assert plugin will transform the calls to these functions. If empty, only `kotlin.assert()` calls will be transformed by default.
+* **`includedSourceSets`**: A list of Gradle source sets that the Power-assert plugin will transform. If empty, all _test source sets_ will be transformed by default.
```
zamulla, line 58: "If empty" here reads to me like "if an empty list is provided", but what I think is meant is "if the option is not specified" (because that is the default). Wdyt?

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+> It may be changed at any time. Use it only for evaluation purposes.
+> We would appreciate your feedback on it in [YouTrack](https://kotl.in/issue).
+>
+{type="warning"}
+
+The Kotlin Power-assert compiler plugin improves the debugging experience
+by providing detailed failure messages with contextual information.
+It simplifies the process of writing tests by automatically generating 
+intermediate values in failure messages. 
+It helps you understand why a test failed without needing complex assertion libraries.
```
zamulla, line 13: This paragraph mirrors the list that follows, but not quite. I'd argue for making the paragraph shorter ("improves the experience, simplifies the process...") and follow up with the list of features that confirm these claims.

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+
+The Power-assert plugin key features:
+
+* **Enhanced error messages**: The plugin captures and displays the values of variables and 
+   sub-expressions within the assertion to clearly identify the cause of failure.
+* **Simplified testing**: Automatically generates informative failure messages,
+   reducing the need for complex assertion libraries.
+* **Supports multiple functions**: By default, it transforms `assert()` function calls but can also transform other functions,
```
zamulla, line 21:
```suggestion
* **Support for multiple functions**: By default, it transforms `assert()` function calls but can also transform other functions,
```

Just to make it uniform with other features.

zamulla (approved):
Looks good, with a couple of points to consider.

docs/topics/power-assert.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+</tab>
+</tabs>
+
+## Configure the plugin
+
+The Power-assert plugin provides several options to customize its behavior:
+
+* **`functions`**: A list of fully-qualified function paths. The Power-assert plugin will transform the calls to these functions. If empty, only `kotlin.assert()` calls will be transformed by default.
```
koshachy, line 57:
```suggestion
* **`functions`**: A list of fully-qualified function paths. The Power-assert plugin will transform the calls to these functions. If not specified, only `kotlin.assert()` calls will be transformed by default.
```

docs/topics/power-assert.md
koshachy, line 58:
```suggestion
* **`includedSourceSets`**: A list of Gradle source sets that the Power-assert plugin will transform. If not specified, all _test source sets_ will be transformed by default.
```

docs/topics/power-assert.md
koshachy: Seems valid. I'll rewrite it after.
Thanks!

docs/topics/power-assert.md
bnorm: Also, I think this link is wrong. I think `blob` needs to be `tree`. Unless you mean to link to the `build.gradle.kts` file?

PR #4224, opened 2024-05-21
Fix Increment/decrement example in k2 migration doc

jayasuryat:
#### Increment and decrement operators

Prior to Kotlin 2.0.0, the compiler didn't understand that the type of an object can change after using
an increment or decrement operator. As the compiler couldn't accurately track the object type,
your code could lead to unresolved reference errors. In Kotlin 2.0.0, this has been fixed:

```diff
interface Rho {
    operator fun inc(): Sigma = TODO()
}

interface Sigma : Rho {
    fun sigma() = Unit
}

- interface Tau {
+ interface Tau : Rho {
    fun tau() = Unit
}

fun main(input: Rho) {
    var unknownObject: Rho = input

    // Check if unknownObject inherits from the Tau interface
    if (unknownObject is Tau) {

        // Use the overloaded inc() operator from interface Rho,
        // which smart-casts the type of unknownObject to Sigma.
        ++unknownObject

        // In Kotlin 2.0.0, the compiler knows unknownObject has type
        // Sigma, so the sigma() function is called successfully.
        unknownObject.sigma()

        // In Kotlin 1.9.20, the compiler thinks unknownObject has type
        // Tau, so calling the sigma() function throws an error.

        // In Kotlin 2.0.0, the compiler knows unknownObject has type
        // Sigma, so calling the tau() function throws an error.
        unknownObject.tau()
        // Unresolved reference 'tau'

        // In Kotlin 1.9.20, the compiler mistakenly thinks that 
        // unknownObject has type Tau, so the tau() function is
        // called successfully.
    }
}
```
---

If what I understood is correct, in this example, `interface` `Tau` should also extend interface `Rho`


sarahhaggarty:
Hi @jayasuryat !
It's actually possible that there is a class in the wild that inherits from both `Tau` and `Rho` interfaces so the `unknownObject is Tau` check is successful. I've added a note as part of #4247 to explain this. I'll close this PR for now but thanks for pointing this out to us!

PR #4229, opened 2024-05-21
chore: update core api reference docs (2.0.0)

ilya-g:
- use core api reference docs built with legacy Dokka from 2.0.0 project
- add version selector 2.0 for stdlib docs


koshachy (approved):
Thanks!

PR #4232, opened 2024-05-21
Update Kotlin 2.0.0 documentation

sarahhaggarty:
This PR adds the new documentation for Kotlin 2.0.0.


docs/topics/typecasts.md
```diff
@@ -110,16 +254,16 @@ Smart casts can be used in the following conditions:
...
 ```kotlin
 val x: String = y as String
 ```
 
-Note that `null` cannot be cast to `String`, as this type is not [nullable](null-safety.md).
-If `y` is null, the code above throws an exception.
-To make code like this correct for null values, use the nullable type on the right-hand side of the cast:
+If the cast isn't possible, the compiler throws an exception. This is why it's called _unsafe_.
+
+In the previous example, if `y` is `null`, the code above also throws an exception. This is because `null` can't be cast to `String`, as `null` isn't [nullable](null-safety.md).
```
lukellmann, line 265: I think it's supposed to say `String` isn't nullable.
```suggestion
In the previous example, if `y` is `null`, the code above also throws an exception. This is because `null` can't be cast to `String`, as `String` isn't [nullable](null-safety.md).
```

sarahhaggarty: Thanks @lukellmann ! You're right. I fixed this as part of #4247

PR #4251, opened 2024-05-28
Reorder Modifier Keywords for Improved Readability

k-taro56:
I wonder if I might suggest that the order of the modifier keywords in the `coding-conventions.md` document could be rearranged to enhance clarity and maintain consistency with Kotlin's recommended practices.

I believe that this adjustment would ensure that related modifiers are grouped more logically, aiding in both readability and maintainability.


koshachy:
Hey @k-taro56,

There is no strict rule on how to use the `external` modifier in the code, so I'm closing this issue. In our documentation, we use `open external` instead of `external open`.
But thanks for the heads-up. We've created an issue to not only improve the documentation but also IDE-related checks. You can become a watcher for this issue here: https://youtrack.jetbrains.com/issue/KT-69046.

Thank you again!

PR #4253, opened 2024-05-28
chore: add links to Any, Nothing, and Unit types

koshachy:


koshachy:
@ice-phoenix or maybe it is better to provide links to the reference?
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-nothing.html
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/

ice-phoenix:
@koshachy TL;DR: Referencing stdlib is a tiny bit better in my humble opinion.

I would say both options are OK, with different advantages.

* Referencing the stdlib is a bit better, as it has documentation about different utility extension functions available on `Any`.
* Referencing the spec is a bit better, as it has links to how Kotlin treats `Nothing` in its data-flow analysis (e.g., dead code detection).

My intuition tells me that, for our regular users, knowing about available utility functions on `Any` is a bit more useful than knowing about `Nothing` and its handling in DFA.

PR #4261, opened 2024-05-30
feat: more info from interopedia

danil-pavlov:


docs/topics/native/native-objc-interop.md
```diff
@@ -32,27 +32,137 @@ Kotlin modules can be used in Swift/Objective-C code if compiled into a framewor
 * See [Build final native binaries](multiplatform-build-native-binaries.md#declare-binaries) to see how to declare binaries.
 * Check out the [Kotlin Multiplatform sample project](https://github.com/Kotlin/kmm-basic-sample) for an example.
 
-### Hiding Kotlin declarations
+If you don't want to export Kotlin declarations to Objective-C and Swift, you can use special annotations, [`@HiddenFromObjC`](#hiding-kotlin-declarations-from-objective-c-and-swift)
+and [`@ShouldRefineInSwift`](#use-refining-in-swift).
```
pahill, line 36: @ShouldRefineInSwift is for another purpose and the declaration is exported but hidden. Don't you mean @HiddenFromObjC and internal?

pahill (approved):
Just the one comment, the rest LGTM, thank you

docs/topics/native/native-objc-interop.md
danil-pavlov: Thank you! I agree that this paragraph is confusing, it's better to remove it

docs/topics/native/native-objc-interop.md
```diff
@@ -32,27 +32,134 @@ Kotlin modules can be used in Swift/Objective-C code if compiled into a framewor
...
+* For an example on how to use the `@ShouldRefineInSwift` annotation, see the [Kotlin-Swift interopedia](https://github.com/kotlin-hands-on/kotlin-swift-interopedia/blob/main/docs/overview/ShouldRefineInSwift.md).
+
+### Change declaration names
+
+> The `@ObjCName` annotation is [Experimental](components-stability.md#stability-levels-explained) and requires [opt-in](opt-in-requirements.md).
+>
+{type="warning"}
+
+To avoid renaming Kotlin declarations, use the `@ObjCName` annotation. It instructs the Kotlin compiler to use
+a custom Objective-C and Swift name for classes, interfaces, and other Kotlin concepts:
```
zamulla, line 73:
```suggestion
custom Objective-C and Swift names for classes, interfaces, and other Kotlin entities:
```

docs/topics/native/native-objc-interop.md
```diff
@@ -32,27 +32,134 @@ Kotlin modules can be used in Swift/Objective-C code if compiled into a framewor
...
+```
+
+[See another example in the Kotlin-Swift interopedia](https://github.com/kotlin-hands-on/kotlin-swift-interopedia/blob/main/docs/overview/ObjCName.md).
+
+### Provide documentation with KDoc comments
+
+Documentation is essential for understanding the basics and the finer details of any API. Providing documentation for
+the shared Kotlin API allows you to communicate with its users on matters of usage, dos and don'ts, and so on.
+
+By default, [KDocs](kotlin-doc.md) documentation comments are not translated into corresponding comments when generating
```
zamulla, line 94:
```suggestion
By default, [KDocs](kotlin-doc.md) comments are not translated into corresponding comments when generating
```

Seems redundant, but please check my assumption here.

zamulla (changes requested):
Nice! Several polishing suggestions to consider.

docs/topics/native/native-objc-interop.md
danil-pavlov: I think we should keep 'concepts' here, this term is used in the first level of the docs tree

zamulla: Oh, I'm not saying that the term is wrong, but that it's not appropriate here: we're trying to say that `MySwiftArray` can be a custom name for the entity called `MyKotlinArray`. `MyKotlinArray` is an entity of the kind described by the Kotlin concept of a class, but that's not important here, as an entity of any kind can be assigned a custom name.

Maybe it's more clear to rewrite the sentence like this: "It instructs the Kotlin compiler to use the specified custom name for the annotated class, interface, or another Kotlin entity." Right now the plural "classes" can indeed be understood both as referring to the concept of a class and to the set of all implemented classes.

zamulla (approved):
Awesome, thanks 👍

PR #4276, opened 2024-06-06
Remove deleted function 'url()' from CocoaPods DSL reference

daugeldauge:
https://youtrack.jetbrains.com/issue/KT-68657


danil-pavlov:
@daugeldauge, thank you for the PR! Is the mention of `url()` in the `specRepos` block in the table above (line 51) is still valid or should we remove it too?

daugeldauge:
@danil-pavlov it's still valid, it's a different API.

danil-pavlov (approved):

PR #4277, opened 2024-06-06
KTL-1504 No redirect from /api/latest/kotlin.test to /api/latest/kotlin.test/

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1504/No-redirect-from-api-latest-kotlin.test-to-api-latest-kotlin.test

I searched all the documentation for a similar link, but this is the only place to fix it.


zamulla (approved):
👍

PR #4282, opened 2024-06-09
Clarified migration procedure with `open lateinit var` properties

mipo256:
These changes are related to K2 compiler migration procedure.

During migration procedure, we noticed that for `lateinit` properties, the first rule stated in migration procedure is not actually correct. Although it might seem obvious, I think it is worth mention in the doc that all `open var` properties must be initialized during declaration, but except `open lateinit var` properties.

Hope it makes sense.

### DIFF
--- a/docs/topics/k2-compiler-migration-guide.md
+++ b/docs/topics/k2-compiler-migration-guide.md
@@ -580,14 +580,15 @@ This section highlights the following modifications:
 
 **What's changed?**
 
-In Kotlin 2.0, all `open` properties with backing fields must be immediately initialized; otherwise, you'll get a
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @mipo256 !
Thanks for your suggestion. Can you share a bit more detail about what problem you experienced with the migration? Do you use `open lateinit var` a lot in your code?

sarahhaggarty:
Hi @mipo256 !
I haven't heard from you in a couple of weeks so I'm going to close this PR. Please comment again on this PR if you'd like to discuss it further and I can consider reopening it.

PR #4285, opened 2024-06-10
feat: new compatibility issue

danil-pavlov:
- [KT-68926](https://youtrack.jetbrains.com/issue/KT-68926/Docs-Incorrect-framework-version-during-Kotlin-upgrade-downgrade)

### DIFF
--- a/docs/topics/multiplatform/multiplatform-compatibility-guide.md
+++ b/docs/topics/multiplatform/multiplatform-compatibility-guide.md
@@ -866,15 +866,18 @@ framework. So, the changes won't be visible in the iOS app in Xcode.
 
 **What's the workaround?**
 
-1. In Xcode, clean build directories using **Product** | **Clean Build Folder** or by running the following command:
+1. In Xcode, clean build directories using **Product** | **Clean Build Folder**.
+2. In the terminal, run the following command:
 
    ```none
    ./gradlew clean
    ```
 
-2. Build the app again to ensure that the new version of the iOS framework is used. 
+3. Build the app again to ensure that the new version of the iOS framework is used. 
 
 **When will the issue be fixed?**
 
 We're planning to fix this issue in Kotlin 2.0.10 and later versions. You can check if any preview versions of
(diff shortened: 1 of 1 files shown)


docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -852,3 +852,32 @@ and prepare this functionality for the upcoming Kotlin 2.0 release. From now on:
...
+
+   ```none
+   ./gradlew clean
+   ```
+
+3. Build the app again to ensure that the new version of the iOS framework is used. 
+
+**When will the issue be fixed?**
+
+We're planning to fix this issue in Kotlin 2.0.10 and later versions. You can check if any preview versions of
```
zamulla, line 880:
```suggestion
We're planning to fix this issue in Kotlin 2.0.10. You can check if any preview versions of
```

Isn't it implicit that if we fix it in 2.0.10, we count on it to be fixed later on?

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -852,3 +852,32 @@ and prepare this functionality for the upcoming Kotlin 2.0 release. From now on:
 
 Starting with Kotlin 1.9.20, you need to explicitly make a cast to and from the corresponding C and Objective-C forward
 declarations. Also, it's now only possible to import forward declarations by using special packages.
+
+## Incorrect version of iOS framework after Kotlin upgrade
+
+**What's the issue?**
+
+If you use direct integration through the `embedAndSignAppleFrameworkForXcode` task, which connects the iOS framework
```
zamulla, line 860: I think the "What's the issue?" question should be answered as soon as possible to help users understand sooner that it is the issue they encountered or not.

It might help to restructure these paragraphs:
* The changes in Kotlin code are not reflected in the iOS app in Xcode when you use direct integration <...>.
* This can happen when you upgrade Kotlin <...> then make changes <...>: Xcode may incorrectly <...>

zamulla (approved):
Great! One last polishing thing, but the code is green.

PR #4286, opened 2024-06-10
update: java version note

danil-pavlov:
* [KT-68317](https://youtrack.jetbrains.com/issue/KT-68317/)
* [KT-49691](https://youtrack.jetbrains.com/issue/KT-49691/)

### DIFF
--- a/docs/topics/native/native-command-line-compiler.md
+++ b/docs/topics/native/native-command-line-compiler.md
@@ -1,33 +1,37 @@
 [//]: # (title: Get started with Kotlin/Native using the command-line compiler)
 
-## Obtain the compiler
+## Download and install the compiler
 
-The Kotlin/Native compiler is available for macOS, Linux, and Windows. It is available as a command line tool and ships 
-as part of the standard Kotlin distribution and can be downloaded from [GitHub Releases](%kotlinLatestUrl%). It supports
-different targets including Linux, macOS, iOS, and others. [See the full list of supported targets](native-target-support.md). 
+The Kotlin/Native compiler works on macOS, Linux, and Windows. It's available as a command line tool and ships 
+as part of the standard Kotlin distribution. You can download it from our [GitHub Releases](%kotlinLatestUrl%) page.
+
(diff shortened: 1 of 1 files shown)


docs/topics/native/native-command-line-compiler.md
```diff
@@ -1,33 +1,37 @@
...
+Choose a working directory and create there a file named `hello.kt`. Add the following code:
 
 ```kotlin
 fun main() {
-  println("Hello Kotlin/Native!")
+  println("Hello, Kotlin/Native!")
 }
 ```
 
+The application will print "Hello, Kotlin/Native" to the standard output.
```
zamulla, line 30: At first glance I thought that this is what's going to happen after I created the file. The previous version of the section started with this, and it seems a clearer sequence: here's what we want the code to do, and here's the code itself, based on that description.

docs/topics/native/native-command-line-compiler.md
```diff
@@ -36,8 +40,8 @@ kotlinc-native hello.kt -o hello
 
 The value of `-o` option specifies the name of the output file, so this call should generate a `hello.kexe` (Linux and macOS)
 or `hello.exe` (Windows) binary file.
-For the full list of available compiler options, see the [compiler options reference](compiler-reference.md).
+For the full list of available options, see [Kotlin compiler options](compiler-reference.md).
```
zamulla, line 44: Reading this, I would expect the outcome here, as in "and now let's run the program we compiled." But that's your call as to how necessary that is.

zamulla (changes requested):
Looks good, I noted one potentially confusing sequence.

docs/topics/native/native-command-line-compiler.md
danil-pavlov: True, we're not running the program yet at this point

docs/topics/native/native-command-line-compiler.md
danil-pavlov: Added a separation section below

zamulla (approved):
Noice!

PR #4287, opened 2024-06-10
feat: macos development note in kotlin/native

danil-pavlov:
* [KT-57862](https://youtrack.jetbrains.com/issue/KT-57862/Note-about-macOS-and-MPP-setup-in-K-N)


zamulla (approved):
👍

PR #4294, opened 2024-06-17
Update Groovy syntax from using the set() function

sarahhaggarty:
This PR updates any instances in our docs where Groovy syntax uses the `set()` function and replaces it with an assignment operator.

### DIFF
--- a/docs/topics/gradle/gradle-compilation-and-caches.md
+++ b/docs/topics/gradle/gradle-compilation-and-caches.md
@@ -254,7 +254,7 @@ tasks.withType<CompileUsingKotlinDaemon>().configureEach {
 <tab title="Groovy" group-key="groovy">
 
 ```groovy
-tasks.withType(CompileUsingKotlinDaemon::class).configureEach { task ->
+tasks.withType(CompileUsingKotlinDaemon).configureEach { task ->
     task.kotlinDaemonJvmArguments = ["-Xmx1g", "-Xms512m"]
 }
 ```
--- a/docs/topics/gradle/gradle-compilation-and-caches.md
+++ b/docs/topics/gradle/gradle-compilation-and-caches.md
@@ -255,7 +255,7 @@ tasks.withType<CompileUsingKotlinDaemon>().configureEach {
 
 ```groovy
 tasks.withType(CompileUsingKotlinDaemon::class).configureEach { task ->
-    task.kotlinDaemonJvmArguments.set(["-Xmx1g", "-Xms512m"])
+    task.kotlinDaemonJvmArguments = ["-Xmx1g", "-Xms512m"]
 }
 ```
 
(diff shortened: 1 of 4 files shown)


docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -255,7 +255,7 @@ tasks.withType<CompileUsingKotlinDaemon>().configureEach {
 
 ```groovy
 tasks.withType(CompileUsingKotlinDaemon::class).configureEach { task ->
```
ALikhachev, line 257:
```suggestion
tasks.withType(CompileUsingKotlinDaemon).configureEach { task ->
```
Oh, we had this sample incorrect. This is a correction to make it work.

docs/topics/gradle/gradle-configure-project.md
```diff
@@ -317,7 +317,7 @@ kotlin {
 ```groovy
 kotlin {
     jvmToolchain {
-        languageVersion.set(JavaLanguageVersion.of(<MAJOR_JDK_VERSION>))
+        languageVersion = JavaLanguageVersion.of(<MAJOR_JDK_VERSION>))
```
ALikhachev, line 320:
```suggestion
        languageVersion = JavaLanguageVersion.of(<MAJOR_JDK_VERSION>)
```

ALikhachev (changes requested):
Looking at these samples, I see another kind of inconsistency. 

1. In some samples we add import statements to reference classes by short names. (like here https://github.com/JetBrains/kotlin-web-site/blob/56844680abde3eb74de066df6634f2ad19d82074/docs/topics/gradle/gradle-compilation-and-caches.md#L336)
2. In other samples we reference classes by FQN (like here https://github.com/JetBrains/kotlin-web-site/blob/2b6165469b2102b845225b569f1cee1384db6416/docs/topics/gradle/gradle-configure-project.md#L192)
3. The most problematic case is when we reference classes by short names without import statement. (like here https://github.com/JetBrains/kotlin-web-site/blob/56844680abde3eb74de066df6634f2ad19d82074/docs/topics/gradle/gradle-compilation-and-caches.md#L248)

I think it should be addressed separately anyway

sarahhaggarty:
> Looking at these samples, I see another kind of inconsistency.
> 
> 1. In some samples we add import statements to reference classes by short names. (like here https://github.com/JetBrains/kotlin-web-site/blob/56844680abde3eb74de066df6634f2ad19d82074/docs/topics/gradle/gradle-compilation-and-caches.md#L336
>    )
> 2. In other samples we reference classes by FQN (like here https://github.com/JetBrains/kotlin-web-site/blob/2b6165469b2102b845225b569f1cee1384db6416/docs/topics/gradle/gradle-configure-project.md#L192
>    )
> 3. The most problematic case is when we reference classes by short names without import statement. (like here https://github.com/JetBrains/kotlin-web-site/blob/56844680abde3eb74de066df6634f2ad19d82074/docs/topics/gradle/gradle-compilation-and-caches.md#L248
>    ) When one just copy-pastes such samples, they do not work without additional actions
> 
> I think this inconsistency should be addressed separately anyway

Thanks for pointing this out! I'll fix it in another PR :)

ALikhachev (approved):

PR #4297, opened 2024-06-17
update: dependencies graphs in Multiplatform

zamulla:


docs/topics/multiplatform/multiplatform-advanced-project-structure.md
```diff
@@ -168,6 +168,32 @@ There are three important concepts in dependency resolution:
...
+### Aligning versions of common dependencies across source sets
+
+Common dependencies need to be aligned across source sets to make sure that common code is always compiled against
+the version of a library that satisfies all source sets.
+
+In the example above, imagine that you want to add the `androidx.navigation:navigation-compose:2.7.7` dependency to your
+`androidMain` source set. Your project explicitly declares the `kotlinx-coroutines-core:1.7.3` dependency for the `commonMain`
+source set, but the Compose Navigation library with the version 2.7.7 requires Kotlin coroutines 1.8.0 or newer.
+
+Resolving this, Gradle applies `kotlinx-coroutines-core:1.8.0` to the `commonMain` source set. But to make the common code
```
antohaby, line 180: > Gradle applies ... to the `commonMain` source set.

It is unclear why Gradle does that? i mean reader can struggle to infer that information from the sentence above.
So maybe it worth adding some explanation. Something like that:

Since commonMain source set is compiled together with androidMain, it means that commonMain should see the same version of `kotlinx-coroutines-core` as androidMain.  Despite the fact that its direct dependency was just `kotlinx-coroutines-core:1.7.3`.

docs/topics/multiplatform/multiplatform-advanced-project-structure.md
```diff
@@ -168,6 +168,32 @@ There are three important concepts in dependency resolution:
...
+compile consistently across targets, the iOS source sets also need to be constrained to the same dependency version.
```
antohaby, line 181: I'm not an english guru, just asking. 

what's the difference between `across targets` vs `across all targets` in this context?

zamulla: "targets" refers to an undetermined set of targets, and "all targets" — to a determined set (not necessarily defined explicitly).

To me "all targets" seems to implicitly refer to all KMP-supported targets, not all project targets. But looking at this fragment again, I do think it's better to clarify, for example, "across all configured targets."

antohaby (approved):

docs/topics/multiplatform/multiplatform-advanced-project-structure.md
```diff
@@ -168,6 +168,36 @@ There are three important concepts in dependency resolution:
...
+Kotlin Gradle plugin helps to align these dependencies, making sure that the effective dependency version
+is the same for each source set.
+
+In the example above, imagine that you want to add the `androidx.navigation:navigation-compose:2.7.7` dependency to your
+`androidMain` source set. Your project explicitly declares the `kotlinx-coroutines-core:1.7.3` dependency for the `commonMain`
+source set, but the Compose Navigation library with the version 2.7.7 requires Kotlin coroutines 1.8.0 or newer.
+
+Since `commonMain` and `androidMain` are compiled together, Kotlin Gradle plugin chooses between the two version of the
+coroutines library and applies `kotlinx-coroutines-core:1.8.0` to the `commonMain` source set. But to make the common code
+compile consistently across all configured targets, the iOS source sets also need to be constrained to the same dependency version.
```
danil-pavlov, line 185:
```suggestion
compile consistently across all configured targets, iOS source sets also need to be constrained to the same dependency version.
```

docs/topics/multiplatform/multiplatform-advanced-project-structure.md
```diff
@@ -168,6 +168,36 @@ There are three important concepts in dependency resolution:
...
+So Gradle propagates the `kotlinx.coroutines-*:1.8.0` dependency to the `iosMain` source set as well.
+
+![Alignment of dependencies in a group of source sets](multiplatform-source-set-dependency-alignment.svg){width=700}
+
+Dependencies are aligned across source sets of each group: the `*Main` source sets and the [`*Test` source sets](multiplatform-discover-project.md#integration-with-tests).
```
danil-pavlov, line 190: We haven't previously used the term 'group of source sets', and I found it a bit confusing, because 'the group of test source sets' sounds like a part of all test source sets

I'd recommend to write it explicitly each time:

> The Gradle configuration for *Test source sets includes all dependencies of *Main source sets but not vice versa.

> For example, you have the Kotlin coroutines 1.7.3 dependency in your *Main source sets, propagated to every project source set.

docs/topics/multiplatform/multiplatform-advanced-project-structure.md
```diff
@@ -168,6 +168,36 @@ There are three important concepts in dependency resolution:
 
    ![Error on JVM-specific API in common code](dependency-resolution-error.png){width=700}
 
+### Aligning versions of common dependencies across source sets
```
danil-pavlov, line 171: I'm not sure if the new section fits under Dependencies and dependsOn. The existing section describes kinds of dependencies and that's it. so the new one should be a separate section on the second level or be under the Compilations (or closer to it), since it's more concerned with having consistent binaries after the compilation and less with how dependencies are set up

danil-pavlov (changes requested):

danil-pavlov (approved):

PR #4300, opened 2024-06-19
feat: spm export doc update

danil-pavlov:


docs/topics/native/native-spm.md
```diff
@@ -6,47 +6,48 @@ Consider a Kotlin Multiplatform project that has an iOS target. You may want to
...
+#### Project configuration options {initial-collapse-state="collapsed"}
 
-Consider the following options for organizing your Git repositories:
+In this tutorial, you'll store your XCFramework and the `Package.swift` file in separate Git repositories.
+However, there can be other project configurations. You can consider the following options for organizing your Git repositories:
 
-* Store the `Package.swift` file in an independent repository. This allows versioning it separately from the
-Kotlin Multiplatform project the file describes. This is the recommended approach: it allows scaling and generally easier
-to maintain.
+* Store the `Package.swift` file and your ZIP archive with the XCFramework in separate Git repositories. This allows
```
DevAlloy, line 26: The same idea here

docs/topics/native/native-spm.md
```diff
@@ -6,47 +6,48 @@ Consider a Kotlin Multiplatform project that has an iOS target. You may want to
...
+  versioning it separately from the Kotlin Multiplatform project the file describes. This is the recommended approach:
+  it allows scaling and is generally easier to maintain.
```
markst, line 28: I agree with this. targets can be defined as follows:

```swift
targets: [
    .binaryTarget(
        name: "shared",
        path: "./shared.xcframework"
    )
]
```

binaries can be versioned and avoids need for zipping, generating checksums and uploading to remote hosted store.

DevAlloy: According to our latest survey and further interviews, teams use SPM integration mainly for remote integration. This is because of a request from the iOS team: we use SPM in our iOS app, so you need to find a way to integrate KMP in the same way. 

But I believe we can add this as an alternative scenario. Thanks for your suggestion!

PR #4324, opened 2024-07-14
update: Conditions and loops page

cmxiv:
Adds details about how the `when` expression derives its resulting type.


sarahhaggarty:
Hi @cmxiv !

Sorry for the delay in getting back to you about your PR. We’ve been tied up migrating our docs to Writerside, but it's finally done! I talked to our language design team about your PR. While your suggestion is a nice technical detail, we don’t think it’s something that needs to be added to the docs. There isn’t a clear use case where developers would benefit from knowing the inner workings of the compiler when it comes to type inference in when expressions. So, I’m going to close this PR. But thanks for the suggestion, and I hope you’ll keep contributing in the future!

PR #4338, opened 2024-07-23
fix(pdf): update k2-compiler-architecture image with linear gradient

zoobestik:
…and path fix for pdf rendering


sarahhaggarty (approved):
LGTM! 🚀

PR #4339, opened 2024-07-24
update: improving object expression examples

daniCsorbaJB:


docs/topics/object-declarations.md
```diff
@@ -3,9 +3,22 @@
 Sometimes you need to create an object that is a slight modification of some class, without explicitly declaring a new
 subclass for it. Kotlin can handle this with _object expressions_ and _object declarations_.
```
nikitabobko, line 4: Given that the last two bulletpoints in your list are more widely used (see the comment below), it'd make sense to rewrite the intro to highlight the more popular use cases

docs/topics/object-declarations.md
```diff
@@ -3,9 +3,22 @@
...
+* **Limited class inheritance is desired:** You need to create single instances of classes without creating subclasses,
+such as for event listeners or callbacks. This ensures only one instance exists throughout the application.
+* **Type-safe design is required:** You require one-time implementations of interfaces or abstract classes using object expressions.
+This can be useful for scenarios like a button click handler.
+* **Modifying existing class behavior temporarily:** You want to modify the behavior of an existing class without creating a new subclass.
+For example, adding temporary functionality to an object for a specific operation.
+* **Using singletons for shared resources:** You need to ensure that only one instance of a class exists throughout the application.
+For example, managing a database connection pool.
+* **Creating factory methods:** You need a convenient way to create instances efficiently.
+[Companion objects](#companion-objects) allow you to define class-level methods and properties tied to a class, simplifying the creation and management of these instances.
```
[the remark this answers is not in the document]
daniCsorbaJB: thank you 👍 - moved them up

docs/topics/object-declarations.md
```diff
@@ -296,55 +422,114 @@ class MyClass {
...
+        private val defaultName = "Unknown user"
+
+        fun createDefault(): User = User(defaultName)
+    }
+}
+
+val defaultUser = User.createDefault()
+println(defaultUser.name) 
+// Unknown user
+```
```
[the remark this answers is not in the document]
daniCsorbaJB: thank you! - this is great - I updated it 👍

docs/topics/object-declarations.md
```diff
@@ -296,55 +422,114 @@ class MyClass {
...
+    // Calls the show method from the companion object of User2
+    println(reference2.show()) 
+    // User2's Companion Object
+}
 ```
+{kotlin-runnable="true"}
 
-Note that even though the members of companion objects look like static members in other languages, at runtime those
-are still instance members of real objects, and can, for example, implement interfaces:
+Companion objects are instance members of their class, which allows them to implement interfaces:
```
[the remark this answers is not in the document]
daniCsorbaJB: hmm 🤔 nice catch! 

- I found the original description a bit too complicated, so I'd like to clarify this part a little.

What do you think about this one:
Although members of companion objects in Kotlin look like static members from other languages,
they are actually instance members of the companion object. This allows companion objects to implement interfaces:

docs/topics/object-declarations.md
daniCsorbaJB: good point - I rephrased it 👍

docs/topics/object-declarations.md
```diff
@@ -1,23 +1,37 @@
...
+Object expressions and object declarations are best used for scenarios when:
+
+* **Using singletons for shared resources:** You need to ensure that only one instance of a class exists throughout the application.
+For example, managing a database connection pool.
+* **Creating factory methods:** You need a convenient way to create instances efficiently.
+[Companion objects](#companion-objects) allow you to define class-level methods and properties tied to a class, simplifying the creation and management of these instances.
+* **Modifying existing class behavior temporarily:** You want to modify the behavior of an existing class without creating a new subclass.
+For example, adding temporary functionality to an object for a specific operation.
+* **Limited class inheritance is desired:** You need to create single instances of classes without creating subclasses,
+such as for event listeners or callbacks. This ensures only one instance exists throughout the application.
```
nikitabobko, line 17: Sorry, for not leaving this comment earlier. 

I propose to drop this bulletpoint. It's technically incorrect "You need to create single instances of classes without creating subclasses" (anonymous subclass is technically still a new subclass, so we don't limit subclassing)

Other two bulletpoints "Modifying existing class behavior temporarily" and "Type-safe design is required" cover this bulletpoint good enough

Up to you

docs/topics/object-declarations.md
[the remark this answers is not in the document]
nikitabobko: A new suggestion is technically correct, so I'm fine with it :)

nikitabobko (approved):

docs/topics/object-declarations.md
daniCsorbaJB: thank you! 👍 I removed it

docs/topics/object-declarations.md
```diff
@@ -294,65 +427,125 @@ class MyClass {
...
+}
 ```
+{kotlin-runnable="true"}
 
-However, on the JVM you can have members of companion objects generated as real static methods and fields if you use
+However, on the JVM, you can have members of companion objects generated as real static methods and fields if you use
 the `@JvmStatic` annotation. See the [Java interoperability](java-to-kotlin-interop.md#static-fields) section
 for more detail.
 
 ### Semantic difference between object expressions and declarations
```
sarahhaggarty, line 546: I think this should be one level higher 🤔 I'm wondering if it would be better to break up this section and add it to the relevant sections before. You can use these bits of information to motivate why this makes expressions/declarations so useful for certain use cases. WDYT?

docs/topics/object-declarations.md
```diff
@@ -47,38 +60,76 @@ window.addMouseListener(object : MouseAdapter() {
...
+    // Executes the transaction
+    temporaryAccount.execute()
+}
+//sampleEnd
+fun main() {
+    // Creates a BankAccount with an initial balance of 1000
+    val myAccount = BankAccount(1000)
+    // Performs a special transaction on the created account
+    specialTransaction(myAccount)
+    // 1500
```
sarahhaggarty, line 99:
```suggestion
    // Executing special transaction. New balance is 1500.

File: docs/topics/object-declarations.md [hunk 8]
```diff
@@ -47,38 +60,76 @@ window.addMouseListener(object : MouseAdapter() {
...
+    fun printPreferences() {
+        val preferences = getPreferences()
+        println("Theme: ${preferences.theme}, Font Size: ${preferences.fontSize}")
     }
 }
+//sampleEnd
+
+fun main() {
+    val userPreferences = UserPreferences()
+    userPreferences.printPreferences()
```
sarahhaggarty, line 126:
```suggestion
    userPreferences.printPreferences()
    // Theme: Dark, Font Size: 14
```

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
+        override fun notifyUser() {
+            println("Sending detailed notification")
+        }
+        val message: String = "Detailed message content"
     }
 }
+//sampleEnd
+fun main() {
+    val notificationManager = NotificationManager()
+    // No output
```
sarahhaggarty, line 178:
```suggestion
```

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
+
+    // The message property is not accessible here because the return type is Any
+    val notification = notificationManager.getNotification()
+    // No output
```
sarahhaggarty, line 182:
```suggestion
```

docs/topics/object-declarations.md
```diff
@@ -130,40 +212,75 @@ fun countClicks(window: JComponent) {
...
+fun main() {
+    // Creates an instance of ExampleDataProvider
+    val exampleProvider = ExampleDataProvider()
+
+    // To refer to the object, use its name directly
+    DataProviderManager.registerDataProvider(exampleProvider)
+
+    // Retrieves and prints all data providers
+    println(DataProviderManager.allDataProviders.map { it.provideData() })
+}
```
sarahhaggarty, line 263:
```suggestion
    [Example data]
}
```

sarahhaggarty (changes requested):
Awesome job with the examples! 🚀 Most of my review comments are about making some changes to the structure of the page and how some of the information is presented. (Some parts you didn't touch but I couldn't resist some changes there too so I hope you don't mind!) Let me know if anything isn't clear :)

docs/topics/object-declarations.md
```diff
@@ -1,23 +1,35 @@
...
+declaration. These anonymous classes are useful for one-time use. They can be created from scratch, inherit from existing classes,
 or implement interfaces. Instances of anonymous classes are also called _anonymous objects_ because they are defined by
 an expression, not a name.
 
-### Creating anonymous objects from scratch
+### Create anonymous objects from scratch
 
 Object expressions start with the `object` keyword.
 
 If you just need an object that doesn't have any nontrivial supertypes, write its members in curly braces after `object`:
```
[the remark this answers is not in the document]
daniCsorbaJB: Hmm 🤔 - how about this one:

 If the object doesn’t extend any classes or implement interfaces, you can define an object’s members directly inside curly braces after the `object` keyword:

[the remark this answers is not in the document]
daniCsorbaJB: It still uses "doesn't" but feels more natural then turning it into a positive sentence structure to me

docs/topics/object-declarations.md
```diff
@@ -47,38 +60,76 @@ window.addMouseListener(object : MouseAdapter() {
...
 
-interface B { /*...*/ }
+// Defines an interface Transaction with an execute() method
+interface Transaction {
+    fun execute()
+}
+
+// Function to perform a special transaction on a BankAccount
+fun specialTransaction(account: BankAccount) {
+    // Creates an anonymous class inheriting from BankAccount and implementing Transaction
```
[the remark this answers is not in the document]
daniCsorbaJB: I like this suggestion -but what do you think about using  "implement an interface" instead of "inherit from an interface" :
So it would look like this:
// Creates an anonymous object that inherits from the BankAccount class and implements the Transaction interface

docs/topics/object-declarations.md
```diff
@@ -47,38 +60,76 @@ window.addMouseListener(object : MouseAdapter() {
...
 ```
+{kotlin-runnable="true"}
 
-### Using anonymous objects as return and value types
+### Use anonymous objects as return and value types
 
-When an anonymous object is used as a type of a local or [private](visibility-modifiers.md#packages) but not [inline](inline-functions.md)
-declaration (function or property), all its members are accessible via this function or property:
+When you use an anonymous object as the type for a local or [private](visibility-modifiers.md#packages) function or property (but not an [inline function](inline-functions.md)),
+all its members are accessible through that function or property:
```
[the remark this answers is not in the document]
daniCsorbaJB: that's a really good point, rephrased it a little:
all the members of that anonymous object are accessible through that function or property:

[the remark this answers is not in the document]
daniCsorbaJB: good point - I added:
This allows you to return an anonymous object with specific properties,
offering a simple way to encapsulate data or behavior without creating a separate class.

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
+        val message: String = "General notification"
     }
 
-    // The return type is A; x is not accessible
-    fun getObjectA() = object: A {
-        override fun funFromA() {}
-        val x: String = "x"
+    // The return type is Notification because the anonymous object implements only one interface
+    // notifyUser() is accessible because it is part of Notification
+    // The message property is not accessible because it is not declared in the Notification interface
```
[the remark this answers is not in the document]
daniCsorbaJB: I think property is better here.

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
+    val notification = notificationManager.getNotification()
+    // No output
+
+    // The notifyUser() is accessible
+    // The message property is not accessible here because the return type is Notification
+    val emailNotification = notificationManager.getEmailNotification()
+    emailNotification.notifyUser()
+    // Sending email notification
+
+    // The notifyUser() method and message property are not accessible here because the return type is DetailedNotification
```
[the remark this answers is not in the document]
daniCsorbaJB: changing method--> function

docs/topics/object-declarations.md
```diff
@@ -130,40 +212,75 @@ fun countClicks(window: JComponent) {
...
 }
 ```
 
 ## Object declarations
 {id="object-declarations-overview"}
 
-The [Singleton](https://en.wikipedia.org/wiki/Singleton_pattern) pattern can be useful in several cases,
-and Kotlin makes it easy to declare singletons:
+You can create single instances of objects in Kotlin using _object declarations_, which always have a name following the `object` keyword.
+This allows you to define a class and create an instance of it in a single step, which is useful for implementing [Singleton patterns](https://en.wikipedia.org/wiki/Singleton_pattern).
```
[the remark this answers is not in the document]
daniCsorbaJB: fair enough - I described singletons at the beginning (although I still left in the Wikipedia link (I think we can remove it though with the new description)

docs/topics/object-declarations.md
daniCsorbaJB: added it as a comment! 👍

docs/topics/object-declarations.md
```diff
@@ -130,40 +212,75 @@ fun countClicks(window: JComponent) {
...
+    DataProviderManager.registerDataProvider(exampleProvider)
+
+    // Retrieves and prints all data providers
+    println(DataProviderManager.allDataProviders.map { it.provideData() })
+}
+```
+{kotlin-runnable="true"}
 
-The initialization of an object declaration is thread-safe and done on first access.
+> Like variable declarations, object declarations, are not expressions, and they cannot be used on the right-hand side
```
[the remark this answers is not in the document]
daniCsorbaJB: sure I added:
// Syntax error: An object expression cannot bind a name.
val myObject = object MySingleton {
    val name = "Singleton"
}

docs/topics/object-declarations.md
```diff
@@ -294,65 +427,125 @@ class MyClass {
...
+    println(reference2.show()) 
+    // User2's Companion Object
+}
 ```
+{kotlin-runnable="true"}
 
-Note that even though the members of companion objects look like static members in other languages, at runtime those
-are still instance members of real objects, and can, for example, implement interfaces:
+Although members of companion objects in Kotlin look like static members from other languages,
+they are actually instance members of the companion object. This allows companion objects to implement interfaces:
```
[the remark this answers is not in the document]
daniCsorbaJB: added short description:
Although members of companion objects in Kotlin look like static members from other languages,
they are actually instance members of the companion object, meaning they belong to the object itself.

docs/topics/object-declarations.md
daniCsorbaJB: I think it's a very good idea to include it one level higher (makes far more sense). 👍 

Hmm 🤔 I think it's warranted to highlight it in a separate section. It draws more attention that way (it might get lost if we include this info in already information-heavy sections.

docs/topics/object-declarations.md
daniCsorbaJB: I think this is an important part to understand that these produce no output, how about moving the comment?
    // Produces no output
    val notificationManager = NotificationManager()

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
+fun main() {
+    val notificationManager = NotificationManager()
+    // No output
+
+    // The message property is not accessible here because the return type is Any
+    val notification = notificationManager.getNotification()
+    // No output
+
+    // The notifyUser() is accessible
+    // The message property is not accessible here because the return type is Notification
```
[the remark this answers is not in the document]
daniCsorbaJB: same as above

docs/topics/object-declarations.md
```diff
@@ -173,46 +290,51 @@ object DefaultListener : MouseAdapter() {
...
 
 fun main() {
-    println(MyObject) // MyObject@1f32e575
+    println(MyObject) 
+    // MyObject@1f32e575
 }
 ```
+{kotlin-runnable="true"}
 
 Just like [data classes](data-classes.md), you can mark an `object` declaration with the `data` modifier.
```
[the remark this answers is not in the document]
daniCsorbaJB: good point, how about starting with something like this:

However, by marking an object declaration with the `data` modifier,
you can instruct the compiler to return the actual name of the object when calling `.toString()`, the same way it works for [data classes](data-classes.md):

```kotlin
data object MyDataObject {
    val number: Int = 3
}

fun main() {
    println(MyDataObject) 
    // MyDataObject
}
```
{kotlin-runnable="true"}

Additionally, the compiler generates several functions for your `data object`:

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: Works for me!

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: Nice!

docs/topics/object-declarations.md
```diff
@@ -87,32 +138,63 @@ In all these cases, members added in the anonymous object are not accessible. Ov
...
-interface B
 
-class C {
-    // The return type is Any; x is not accessible
-    fun getObject() = object {
-        val x: String = "x"
+interface DetailedNotification
+
+class NotificationManager {
+    // The return type is Any; message is not accessible
```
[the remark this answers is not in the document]
sarahhaggarty: OK!

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: OK!

docs/topics/object-declarations.md
sarahhaggarty: Yes! Much better.

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: Nice!

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: Nice!

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: Thanks!

docs/topics/object-declarations.md
sarahhaggarty: OK, let's see how it looks once you've made the other changes :)

docs/topics/object-declarations.md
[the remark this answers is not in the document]
sarahhaggarty: It's better! But what do you think about this instead?

> In its simplest form, you can define an object’s members directly inside curly braces after the object keyword. This is possible when the object doesn’t extend any classes or implement any interfaces:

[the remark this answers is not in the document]
daniCsorbaJB: As discussed, let's go with you can define an object's members directly inside curly braces after the object keyword:

docs/topics/object-declarations.md
```diff
@@ -294,65 +237,337 @@ class MyClass {
...
+
+If the object doesn't extend any classes or implement interfaces, you can define an object's members directly inside curly braces after the `object` keyword:
+
+```kotlin
+fun main() {
+//sampleStart
+    val helloWorld = object {
+        val hello = "Hello"
+        val world = "World"
+        // Object expressions extend the Any class, which already has a `toString()` function,
```
sarahhaggarty, line 377:
```suggestion
        // Object expressions extend the Any class, which already has a toString() function,
```

docs/topics/object-declarations.md
```diff
@@ -173,48 +100,53 @@ object DefaultListener : MouseAdapter() {
...
+    println(MyDataObject) 
+    // MyDataObject
 }
 ```
+{kotlin-runnable="true"}
+
+Additionally, the compiler generates several functions for your `data object`:
+
+* `.toString()` returns the name of the data object
+* `.equals()`/`.hashCode()` enables equality checks and hash-based collections
```
sarahhaggarty, line 140:
```suggestion
* `toString()` returns the name of the data object
* `equals()`/`hashCode()` enables equality checks and hash-based collections
```

docs/topics/object-declarations.md
```diff
@@ -173,48 +100,53 @@ object DefaultListener : MouseAdapter() {
...
+
+  > You can't provide a custom `equals` or `hashCode` implementation for a `data object`.
+  >
+  {type="note"}
 
-The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
-In most cases, you will only have a single instance of your data object at runtime (after all, a `data object` declares a singleton).
-However, in the edge case where another object of the same type is generated at runtime (for example, by using platform 
-reflection with `java.lang.reflect` or a JVM serialization library that uses this API under the hood), this ensures that 
+The `.equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
```
sarahhaggarty, line 146:
```suggestion
The `equals()` function for a `data object` ensures that all objects that have the type of your `data object` are considered equal.
```

docs/topics/object-declarations.md
```diff
@@ -230,44 +162,52 @@ data object MySingleton
...
 
 #### Differences between data objects and data classes
 
-While `data object` and `data class` declarations are often used together and have some similarities, there are some 
+While `data object` and `data class` declarations are often used together and have some similarities, there are some
 functions that are not generated for a `data object`:
 
-* No `copy()` function. Because a `data object` declaration is intended to be used as singleton objects, no `copy()` 
-  function is generated. The singleton pattern restricts the instantiation of a class to a single instance, which would 
+* No `.copy()` function. Because a `data object` declaration is intended to be used as singletons, no `.copy()`
```
sarahhaggarty, line 197:
```suggestion
* No `copy()` function. Because a `data object` declaration is intended to be used as singletons, no `copy()`
```

docs/topics/object-declarations.md
```diff
@@ -230,44 +162,52 @@ data object MySingleton
...
+  function is generated. Singletons restrict the instantiation of a class to a single instance, which would
   be violated by allowing copies of the instance to be created.
-* No `componentN()` function. Unlike a `data class`, a `data object` does not have any data properties. 
-  Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
+* No `.componentN()` function. Unlike a `data class`, a `data object` does not have any data properties.
+  Since attempting to destructure such an object without data properties would not make sense, no `.componentN()` functions are generated.
```
sarahhaggarty, line 201:
```suggestion
* No `componentN()` function. Unlike a `data class`, a `data object` does not have any data properties.
  Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
```

docs/topics/object-declarations.md
```diff
@@ -230,44 +162,52 @@ data object MySingleton
...
-Data object declarations are particularly useful for sealed hierarchies like 
-[sealed classes or sealed interfaces](sealed-classes.md), since they allow you to maintain symmetry with any data classes 
-you may have defined alongside the object. In this example, declaring `EndOfFile` as a `data object` instead of a plain `object` 
-means that it will get the `toString()` function without the need to override it manually:
+Data object declarations are particularly useful for sealed hierarchies like
+[sealed classes or sealed interfaces](sealed-classes.md).
+They allow you to maintain symmetry with any data classes you may have defined alongside the object.
+
+In this example, declaring `EndOfFile` as a `data object` instead of a plain `object`
+means that it will get the `.toString()` function without the need to override it manually:
```
sarahhaggarty, line 210:
```suggestion
means that it will get the `toString()` function without the need to override it manually:
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +237,337 @@ class MyClass {
...
+
+// A function to perform a special transaction on a BankAccount
+fun specialTransaction(account: BankAccount) {
+    // Creates an anonymous object that inherits from the BankAccount class and implements the Transaction interface
+    // The balance of the provided account is passed to the BankAccount superclass constructor
+    val temporaryAccount = object : BankAccount(account.balance), Transaction {
+
+        override val balance = account.balance + 500  // Temporary bonus
+
+        // Implements the .execute() function from the Transaction interface
```
sarahhaggarty, line 425:
```suggestion
        // Implements the execute() function from the Transaction interface
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +237,337 @@ class MyClass {
...
+//sampleEnd
+fun main() {
+    // This produces no output
+    val notificationManager = NotificationManager()
+
+    // The message property is not accessible here because the return type is Any
+    // This produces no output
+    val notification = notificationManager.getNotification()
+
+    // The notifyUser() is accessible
```
sarahhaggarty, line 528:
```suggestion
    // The notifyUser() function is accessible
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +237,337 @@ class MyClass {
...
+offering a simple way to encapsulate data or behavior without creating a separate class.
+
+If a function or property that returns an anonymous object is `public` or `private`, its actual type is:
+
+* `Any` if the anonymous object doesn't have a declared supertype.
+* The declared supertype of the anonymous object, if there is exactly one such type.
+* The explicitly declared type if there is more than one declared supertype.
+
+In all these cases, members added in the anonymous object are not accessible. Overridden members are accessible if they
+are declared in the actual type of the function or property:
```
sarahhaggarty, line 482:
```suggestion
are declared in the actual type of the function or property. For example:
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +237,337 @@ class MyClass {
...
+
+        override fun mouseEntered(e: MouseEvent) {
+            enterCount++
+        }
+    })
+    // The clickCount and enterCount variables are accessible within the object expression
+}
+```
+
+## Semantic difference between object expressions and declarations
```
sarahhaggarty, line 568:
```suggestion
## Behavior difference between object expressions and declarations
```
I think "behavior" is more accurate than "semantic" here, what do you think?

docs/topics/object-declarations.md
```diff
@@ -173,48 +100,53 @@ object DefaultListener : MouseAdapter() {
...
-This instructs the compiler to generate a number of functions for your object:
-
-* `toString()` returns the name of the data object
-* `equals()`/`hashCode()` pair
-
-  > You can't provide a custom `equals` or `hashCode` implementation for a `data object`.
-  >
-  {type="note"}
+However, by marking an object declaration with the `data` modifier,
+you can instruct the compiler to return the actual name of the object when calling `.toString()`, the same way it works for [data classes](data-classes.md):
```
sarahhaggarty, line 123:
```suggestion
you can instruct the compiler to return the actual name of the object when calling `toString()`, the same way it works for [data classes](data-classes.md):
```

docs/topics/object-declarations.md
```diff
@@ -230,44 +162,52 @@ data object MySingleton
...
     // Kotlin reflection does not permit the instantiation of data objects.
-    // This creates a new MySingleton instance "by force" (i.e. Java platform reflection)
+    // This creates a new MySingleton instance "by force" (using Java platform reflection)
     // Don't do this yourself!
     return (MySingleton.javaClass.declaredConstructors[0].apply { isAccessible = true } as Constructor<MySingleton>).newInstance()
 }
 ```
 
-The generated `hashCode()` function has behavior that is consistent with the `equals()` function, so that all runtime 
+The generated `.hashCode()` function has a behavior that is consistent with the `.equals()` function, so that all runtime
```
sarahhaggarty, line 189:
```suggestion
The generated `hashCode()` function has a behavior that is consistent with the `equals()` function, so that all runtime
```

docs/topics/object-declarations.md
sarahhaggarty, line 201:
```suggestion
* No `componentN()` function. Unlike a `data class`, a `data object` does not have any data properties.
  Since attempting to destructure such an object without data properties would not make sense, no `componentN()` functions are generated.
```

sarahhaggarty (changes requested):
It's really shaping up nicely! Almost there 💪

docs/topics/object-declarations.md
daniCsorbaJB: I like it, I agree - I changed it

docs/topics/object-declarations.md
```diff
@@ -294,65 +234,336 @@ class MyClass {
...
+    // Performs a special transaction on the created account
+    specialTransaction(myAccount)
+    // Executing special transaction. New balance is 1500.
+}
+```
+{kotlin-runnable="true" id="object-expression-anonymous-object"}
+
+### Use anonymous objects as return and value types
+
+When you return an anonymous object from a local or [private](visibility-modifiers.md#packages) function or property (but not an [inline function](inline-functions.md)),
```
sarahhaggarty, line 441:
```suggestion
When you return an anonymous object from a local or [`private`](visibility-modifiers.md#packages) function or property (but not an [inline function](inline-functions.md)),
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +233,336 @@ class MyClass {
...
+
+        override fun mouseEntered(e: MouseEvent) {
+            enterCount++
+        }
+    })
+    // The clickCount and enterCount variables are accessible within the object expression
+}
+```
+
+## Behavior difference between object expressions and declarations
```
sarahhaggarty, line 563:
```suggestion
## Behavior difference between object declarations and expressions
```

docs/topics/object-declarations.md
```diff
@@ -294,65 +233,336 @@ class MyClass {
...
+
+There are differences in the initialization behavior between object expressions and object declarations:
```
sarahhaggarty, line 565:
```suggestion
There are differences in the initialization behavior between object declarations and object expressions:
```
For symmetry. Might be worth swapping the order of the first two bullets as well. I leave it up to you though!

docs/topics/object-declarations.md
```diff
@@ -1,220 +1,148 @@
...
-    override fun mouseClicked(e: MouseEvent) { ... }
+object MyObject
 
-    override fun mouseEntered(e: MouseEvent) { ... }
+fun main() {
+    println(MyObject) 
+    // MyObject@1f32e575
 }
 ```
+{kotlin-runnable="true"}
```
sarahhaggarty, line 116: Can you add an ID here too? :)

sarahhaggarty (approved):
Looks great!
I made a couple of minor suggestions to make it "declarations  and expressions" uniformly everywhere :) Otherwise, LGTM! 🚀 
Thank you for all your hard work on this!

docs/topics/object-declarations.md
```diff
@@ -1,169 +1,85 @@
...
-declaration. Such classes are useful for one-time use. You can define them from scratch, inherit from existing classes,
-or implement interfaces. Instances of anonymous classes are also called _anonymous objects_ because they are defined by
-an expression, not a name.
+Object expressions and object declarations are best used for scenarios when:
 
-### Creating anonymous objects from scratch
+* **Using singletons for shared resources:** You need to ensure that only one instance of a class exists throughout the application.
+For example, managing a database connection pool.
+* **Creating factory methods:** You need a convenient way to create instances efficiently.
+[Companion objects](#companion-objects) allow you to define class-level methods and properties tied to a class, simplifying the creation and management of these instances.
```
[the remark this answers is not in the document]
daniCsorbaJB: ah! Great catch thank you 👍 💯

PR #4362, opened 2024-08-07
feat: static accessors for the default hierarchy

danil-pavlov:
This PR changes the description of the default source set hierarchy, following [KT-56566](https://youtrack.jetbrains.com/issue/KT-56566), plus fixes mentions of the Android target

### DIFF
--- a/docs/topics/multiplatform/multiplatform-hierarchy.md
+++ b/docs/topics/multiplatform/multiplatform-hierarchy.md
@@ -21,7 +21,7 @@ This is a more low-level approach: it's more flexible but requires more effort a
 
 ## Default hierarchy template
 
-Starting with Kotlin 1.9.20, the Kotlin Gradle plugin has a built-in default [hierarchy template](#see-the-full-hierarchy-template).
+The Kotlin Gradle plugin has a built-in default [hierarchy template](#see-the-full-hierarchy-template).
 It contains predefined intermediate source sets for some popular use cases.
 The plugin sets up those source sets automatically based on the targets specified in your project.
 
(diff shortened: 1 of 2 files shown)


antohaby (approved):

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -64,8 +64,10 @@ are no watchOS targets in the project.
 If you add a watchOS target, like `watchosArm64`, the `watchos` source set is created, and the code
 from the `apple`, `native`, and `common` source sets is compiled to `watchosArm64` as well.
 
-The Kotlin Gradle plugin creates type-safe accessors for all of the source sets from the default hierarchy template,
-so you can reference them without `by getting` or `by creating` constructions compared to the [manual configuration](#manual-configuration):
+The Kotlin Gradle plugin provides both type-safe and static accessors for all of the source sets from the default hierarchy
+template, so you can reference them without `by getting` or `by creating` constructions compared to the [manual configuration](#manual-configuration).
```
zamulla, line 68:
```suggestion
template, so you can reference them without `by getting` or `by creating` constructs compared to the [manual configuration](#manual-configuration).
```

I don't believe "construction" has the meaning needed here, but [language constructs](https://en.wikipedia.org/wiki/Language_construct) are definitely a thing.

docs/topics/multiplatform/multiplatform-hierarchy.md
```diff
@@ -80,6 +82,8 @@ kotlin {
         iosMain.dependencies {
             implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:%coroutinesVersion%")
         }
+        linuxX64Main { } // Warning: accessing source set
```
zamulla, line 85: I don't think our guidelines are explicit here, but I feel that a multiline comment especially should be placed before the code, not alongside — it also makes it easier to separate changes to the code itself from changes to comments.

zamulla (commented):
All good! Just one clarification and one formatting nitpick.

docs/topics/multiplatform/multiplatform-hierarchy.md
danil-pavlov: As I see on our Coding conventions page, it's used as a synonym to "constructs", plus a few other mentions in the docs. But we can of course change it here, no problem

zamulla (approved):
👍

PR #4367, opened 2024-08-08
feat: concurrent marking in GC

danil-pavlov:
This PR creates a new section that combines GC pause time optimization options, including the new [KT-58865](https://youtrack.jetbrains.com/issue/KT-58865/Native-implement-concurrent-mark-in-GC)

### DIFF
--- a/docs/topics/native/native-memory-manager.md
+++ b/docs/topics/native/native-memory-manager.md
@@ -69,8 +69,7 @@ To track GC-related pauses in your app:
 
 ### Optimize GC performance
 
-To improve GC performance, you can enable concurrent marking to decrease the GC pause time. It allows to run the marking
-phase of the garbage collection simultaneously with application threads.
+To improve GC performance, you can enable concurrent marking to decrease the GC pause time. This allows the marking phase of garbage collection to run simultaneously with application threads.
 
 The feature is currently [Experimental](components-stability.md#stability-levels-explained). To enable it, set the
 following compiler option in your `gradle.properties` file:
@@ -158,7 +157,7 @@ update might improve memory consumption.
 
 If you continue to experience high memory consumption after updating, several options are available:
 
-* Switch to a different memory allocator by using one of the following compiler option in your Gradle build script:
+* Switch to a different memory allocator by using one of the following compiler options in your Gradle build script:
 
   * `-Xallocator=std` for the system allocator.
   * `-Xallocator=mimalloc` for the [mimalloc](https://github.com/microsoft/mimalloc) allocator.
--- a/docs/topics/native/native-memory-manager.md
+++ b/docs/topics/native/native-memory-manager.md
@@ -4,25 +4,29 @@ Kotlin/Native uses a modern memory manager that is similar to the JVM, Go, and o
 the following features:
 
 * Objects are stored in a shared heap and can be accessed from any thread.
-* Tracing garbage collection (GC) is performed periodically to collect objects that are not reachable from the "roots",
+* Tracing garbage collection is performed periodically to collect objects that are not reachable from the "roots",
   like local and global variables.
 
 ## Garbage collector
 
-Kotlin/Native's GC algorithm is constantly evolving. Currently, it functions as a stop-the-world mark and concurrent sweep
-collector that does not separate the heap into generations.
-
-The GC uses a full parallel mark that combines paused mutators, the GC thread, and optional marker threads to process
-the mark queue. By default, paused mutators and at least one GC thread participate in the marking process.
-You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compilation option.
-However, this may increase the pause time of the garbage collector.
-
-When the marking phase is completed, the GC processes weak references and nullifies reference points to an unmarked object.
-To decrease the GC pause time, you can enable the concurrent processing of weak references by using
-the `-Xbinary=concurrentWeakSweep=true` compilation option.
+Kotlin/Native's garbage collector (GC) algorithm is constantly evolving. Currently, it functions as a stop-the-world mark
+and concurrent sweep collector that does not separate the heap into generations.
 
 The GC is executed on a separate thread and started based on the timer
 and memory pressure heuristics. Alternatively, it can be [called manually](#enable-garbage-collection-manually).
+
+To process the mark queue, the GC uses a full parallel mark that combines application threads, the GC thread,
+and optional marker threads. Application threads and at least one GC thread participate in the marking process.
+By default, application threads must be paused when the GC is marking objects in the heap.
+
+> You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compiler option.
+> However, this may increase the garbage collector's pause time.
+>
+{type="tip"}
+
+When the marking phase is completed, the GC processes weak references and nullifies reference points to an unmarked object.
+
+See how to [monitor](#monitor-gc-performance) and [optimize](#optimize-gc-performance) garbage collection.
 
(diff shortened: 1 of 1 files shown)


docs/topics/native/native-memory-manager.md
```diff
@@ -62,10 +66,27 @@ To track GC-related pauses in your app:
...
+
+  The feature is currently [Experimental](components-stability.md#stability-levels-explained). To enable it, set the
+  following option in your `gradle.properties` file:
+  
+  ```properties
+  kotlin.native.binary.gc=cms
+  ```
+
+* For the next phase of handling weak references, you can enable concurrent processing by using the
+  `-Xbinary=concurrentWeakSweep=true` compiler option.
```
haitaka, line 84: The `concurrentWeakSweep` does not make sense with the CMS GC (i.e. always enabled with CMS), and is the default behavior with the Parallel Mark (default) GC. 
I think, we should remove this item.

haitaka, line 84: BTW, we need to choose a single idiomatic way of setting binary options and stay with it. Here we suggest to use `-Xbinary=...` command line argument, while a few likes above we do it through a `kotlin.native.binary...` property.

docs/topics/native/native-memory-manager.md
```diff
@@ -4,25 +4,29 @@ Kotlin/Native uses a modern memory manager that is similar to the JVM, Go, and o
...
+and concurrent sweep collector that does not separate the heap into generations.
 
-The GC uses a full parallel mark that combines paused mutators, the GC thread, and optional marker threads to process
-the mark queue. By default, paused mutators and at least one GC thread participate in the marking process.
-You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compilation option.
-However, this may increase the pause time of the garbage collector.
+The GC is executed on a separate thread and started based on the timer
+and memory pressure heuristics. Alternatively, it can be [called manually](#enable-garbage-collection-manually).
+
+To process the mark queue, the GC uses a full parallel mark that combines application threads, the GC thread,
```
haitaka, line 18: I would rather say: "GC processes the mark queue on several (execution) threads in parallel, this includes application threads, the GC thread, and optional marker threads."

docs/topics/native/native-memory-manager.md
```diff
@@ -4,25 +4,29 @@ Kotlin/Native uses a modern memory manager that is similar to the JVM, Go, and o
...
+and optional marker threads. Application threads and at least one GC thread participate in the marking process.
+By default, application threads must be paused when the GC is marking objects in the heap.
+
+> You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compiler option.
+> However, this may increase the garbage collector's pause time.
```
haitaka, line 23: And then here: "You can disable the parallelization of the mark phase with the `-Xbinary=gcMarkSingleThreaded=true` compiler option. However, this may increase the garbage collector's pause time on large heaps."

docs/topics/native/native-memory-manager.md
```diff
@@ -141,7 +162,7 @@ update might improve memory consumption.
 
 If you continue to experience high memory consumption after updating, several options are available:
 
-* Switch to a different memory allocator by using one of the following compilation options in your Gradle build script:
+* Switch to a different memory allocator by using one of the following compiler option in your Gradle build script:
 
   * `-Xallocator=mimalloc` for the [mimalloc](https://github.com/microsoft/mimalloc) allocator.
```
haitaka, line 167: Not quite sure how it should be done properly, but please note that we plan to deprecate the mimalloc allocator in 2.1.0. Perhaps, we could stop suggesting as the first workaround option it already now?

docs/topics/native/native-memory-manager.md
```diff
@@ -4,25 +4,29 @@ Kotlin/Native uses a modern memory manager that is similar to the JVM, Go, and o
...
-collector that does not separate the heap into generations.
+Kotlin/Native's garbage collector (GC) algorithm is constantly evolving. Currently, it functions as a stop-the-world mark
+and concurrent sweep collector that does not separate the heap into generations.
 
-The GC uses a full parallel mark that combines paused mutators, the GC thread, and optional marker threads to process
-the mark queue. By default, paused mutators and at least one GC thread participate in the marking process.
-You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compilation option.
-However, this may increase the pause time of the garbage collector.
+The GC is executed on a separate thread and started based on the timer
+and memory pressure heuristics. Alternatively, it can be [called manually](#enable-garbage-collection-manually).
```
haitaka, line 16: The memory pressure is the driving factor here, with the timer being a backup plan. Hence, I would change the order: "The GC is executed on a separate thread and started based on the memory pressure heuristics or by a timer."

haitaka (changes requested):

docs/topics/native/native-memory-manager.md
danil-pavlov: OK, then since it's enabled by default, I mentioned it in the general description of the GC. The only question here is do we want now to add info on how to disable `concurrentWeakSweep`?

danil-pavlov: True, I changed all of them to `kotlin.native` properties, please check if I got it right

docs/topics/native/native-memory-manager.md
danil-pavlov: thank you! I put the standard allocator option on the top. Please check if there's a "kotlin to be documented" tag on the mimalloc deprecation task in YT, so that we don't forget it in 2.1

docs/topics/native/native-memory-manager.md
```diff
@@ -32,10 +37,10 @@ and waits for its completion.
 ### Monitor GC performance
 
 To monitor the GC performance, you can look through its logs and diagnose issues. To enable logging,
-set the following compiler option in the Gradle build script:
+set the following compiler option in your `gradle.properties` file:
 
 ```none
--Xruntime-logs=gc=info
+kotlin.native.runtimeLogs.gc=info
```
haitaka: Unfortunately, it doesn't work this way, this is a different family of options. Unlike the `binary` ones, these have to be passed through the command line (e.g. in the `freeCompilerArgs`) in the form of `-Xruntime=logs=gc=info`.

docs/topics/native/native-memory-manager.md
```diff
@@ -4,25 +4,30 @@ Kotlin/Native uses a modern memory manager that is similar to the JVM, Go, and o
...
+and concurrent sweep collector that does not separate the heap into generations.
 
-The GC uses a full parallel mark that combines paused mutators, the GC thread, and optional marker threads to process
-the mark queue. By default, paused mutators and at least one GC thread participate in the marking process.
-You can disable the full parallel mark with the `-Xbinary=gcMarkSingleThreaded=true` compilation option.
-However, this may increase the pause time of the garbage collector.
+The GC is executed on a separate thread and started based on the memory pressure heuristics or by a timer. Alternatively,
+it can be [called manually](#enable-garbage-collection-manually).
+
+The GC processes the mark queue on several (execution) threads in parallel, including application threads, the GC thread,
```
haitaka, line 18: Sorry for the confusion, let's remove the "(execution)".

docs/topics/native/native-memory-manager.md
haitaka: Thank you. Unfortunately, this trick works only with the `-Xbinary=...` options. All the other such as `-Xruntime-logs=` and `-Xallocator=` have to remain in the command line form.

haitaka (approved):

PR #4374, opened 2024-08-15
Update kotlin-tour-functions.md

aamirfayaz:
in lambda parameter is followed by : and not ->


sarahhaggarty:
Hi @aamirfayaz !
We actually say the "paramater**s**" are followed by ->. Not just the "parameter". So the text is correct as it is. But thanks for the suggestion!

PR #4380, opened 2024-08-16
update: discourage combination of K/N compiler options

danil-pavlov:


anton-bannykh (commented):
[minor] We ended up reporting a warning, not strictly prohibiting the combination, so maybe "should not be used" is more suitable than "cannot be used"? Not exactly an expert here tbh.

anton-bannykh (changes requested):
Apart from should/can looks good.

anton-bannykh (approved):

docs/topics/compiler-reference.md
```diff
@@ -285,7 +285,8 @@ Enable runtime assertions in the generated code.
 
 ### -g
 
-Enable emitting debug information.
+Enable emitting debug information. This option lowers the optimization level and, therefore, should not be combined with
```
sarahhaggarty, line 288:
```suggestion
Enable emitting debug information. This option lowers the optimization level and should not be combined with
```

sarahhaggarty (approved):
Looks good! I made one minor suggestion for conciseness but feel free to ignore it.

PR #4381, opened 2024-08-16
update: adding instructions for github releases

danil-pavlov:


docs/topics/native/native-spm.md
```diff
@@ -132,7 +164,7 @@ To set up the publishing of an XCFramework:
     
     The output will describe any errors found or show the successful download and parsing result if the manifest is correct.
 
-6. Push the `Package.swift` file to your remote repository. Make sure to create and push a Git tag with the
+7. Push the `Package.swift` file to your remote repository. Make sure to create and push a Git tag with the
    semantic version of the package.
 
 ### Add the package dependency
```
DevAlloy, line 170: Let's mention that now we can add the package dependency to a client iOS project.

DevAlloy: Also this project appears suddenly, so possibly we can mention that you can use your own client iOS project or create a new one. It makes sense because we mention ContentView.swift file in "Check your setup section", but this step valid only for new created project.

docs/topics/native/native-spm.md
```diff
@@ -132,29 +164,32 @@ To set up the publishing of an XCFramework:
...
-6. Push the `Package.swift` file to your remote repository. Make sure to create and push a Git tag with the
+7. Push the `Package.swift` file to your remote repository. Make sure to create and push a Git tag with the
    semantic version of the package.
 
-### Add the package dependency
+### Add a package dependency
 
-Now that both files are accessible, you can add the package dependency:
+Now that both files are accessible, you can add a package dependency to an existing client iOS project or create a new
+project. To add a package dependency:
```
zamulla, line 173:
```suggestion
project. To add a dependency on the Swift package you created:
```

"The" here showed that we're adding a dependency specifically on the package we just released, not just any old package dependency. Should we be specific here in some way?

zamulla (approved):
Noice. Left a couple of small suggestions.

docs/topics/native/native-spm.md
```diff
@@ -80,26 +80,58 @@ To set up the publishing of an XCFramework:
...
+                   <img src="github-release-link.png" alt="Copy the link to the uploaded file" width="500"/>
+               </li>
+         </list>
+       </def>
+   </deflist>
+
+3. [Recommended] Check that the link works and that the file can be downloaded. In the terminal, run the following command:
+
+    ```shell
+    curl <downloadable link to the uploaded XCFramework ZIP file>
```
zamulla, line 131: The `shell` language here highlights the "link" within the brackets, not sure if that is better than no highlight at all.

zamulla, line 131: Also "downloadable link" should probably be "download link", although that does change the meaning a bit. Maybe "working download link"?

docs/topics/native/native-spm.md
```diff
@@ -80,26 +80,58 @@ To set up the publishing of an XCFramework:
...
+   
+   <deflist collapsible="true">
+       <def title="Upload to a GitHub release">
+           <list type="decimal">
+               <li>Go to <a href="https://github.com">GitHub</a> and log in to your account.</li>
+               <li>Navigate to the repository where you want to create a release.</li>
+               <li>In the <b>Releases</b> section on the right, click the <b>Create a new release</b> link.</li>
+               <li>Fill in the release information, add or create a new tag, specify the release title and write a description.</li>
+               <li>
+                   <p>Upload the ZIP file with the XCFramework through the <b>Attach binaries by dropping them here or selecting them</b> option at the bottom:</p>
```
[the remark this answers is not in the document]
danil-pavlov: I'm not sure, it's too big for a button, let's call it a field, since you can use it for drag&drop

docs/topics/native/native-spm.md
danil-pavlov: Thank you! rephrased the intro + fixed the article

docs/topics/native/native-spm.md
danil-pavlov: It's essential to keep the meaning that it's not just a link to the file location, but a link that starts the file download when selected or specified, I think 'downloadable' is the shortest way to say that

PR #4392, opened 2024-08-23
Update dead code elimination docs

AlejandraPedroza:
This PR includes updates to the DCD docs such as a warning about deprecation and information about the new usage.

### DIFF
--- a/docs/topics/js/javascript-dce.md
+++ b/docs/topics/js/javascript-dce.md
@@ -1,5 +1,11 @@
 [//]: # (title: Kotlin/JS dead code elimination)
 
+> The DCE tool is deprecated. The DCE tool was designed for the legacy JS backend, which is now obsolete. The current 
+> [JS IR backend](#dce-and-javascript-ir-compiler) supports dead code elimination out of the box, and the @JsExport feature allows specifying which Kotlin 
+> functions and classes to retain during tree shaking.
+>
+{type="warning"}
+
 The Kotlin Multiplatform Gradle plugin includes a _[dead code elimination](https://wikipedia.org/wiki/Dead_code_elimination)_ (_DCE_) tool.
 Dead code elimination is often also called _tree shaking_. It reduces the size or the resulting JavaScript code by
 removing unused properties, functions, and classes.
@@ -14,6 +20,26 @@ Unused declarations can appear in cases like:
 
 The Kotlin Multiplatform Gradle plugin handles DCE automatically when you build a **production bundle**, for example by using the
 `browserProductionWebpack` task. **Development bundling** tasks (like `browserDevelopmentWebpack`) don't include DCE.
+
+## DCE and JavaScript IR compiler
+
+The new way of applying dead code elimination that works with the IR compiler is:
+
+* DCE is disabled when compiling for development, which corresponds to the following Gradle tasks:
+  * `browserDevelopmentRun`
+  * `browserDevelopmentWebpack`
+  * `nodeDevelopmentRun`
+  * `compileDevelopmentExecutableKotlinJs`
+  * `compileDevelopmentLibraryKotlinJs`
+  * Other development tasks you recognize for the word "development" in their name
+* DCE is enabled when compiling for production, which corresponds to the following Gradle tasks:
+  * `browserProductionRun`
+  * `browserProductionWebpack`
+  * `compileProductionExecutableKotlinJs`
+  * `compileProductionLibraryKotlinJs`
+  * Other production tasks you recognize for the word "production" in their name
+
+With the @JsExport annotation, you can specify the declarations you want DCE to treat as roots.
 
 ## Exclude declarations from DCE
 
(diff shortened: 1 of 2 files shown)


broadwaylamb (commented):
I've searched through mentions of the DCE tool across the repo, and found a couple more cases:

https://github.com/JetBrains/kotlin-web-site/blob/617acf0959b4340743ed6fe697b86ef2e864a4a2/docs/topics/multiplatform/multiplatform-dsl-reference.md?plain=1#L211

https://github.com/JetBrains/kotlin-web-site/blob/617acf0959b4340743ed6fe697b86ef2e864a4a2/docs/topics/multiplatform/multiplatform-dsl-reference.md?plain=1#L219-L221

Also, I'm not sure what's the policy about keeping links to obsolete articles in the outline, but if we decide that we don't want them there, this entry should also be removed:
https://github.com/JetBrains/kotlin-web-site/blob/617acf0959b4340743ed6fe697b86ef2e864a4a2/docs/kr.tree#L263

docs/topics/js/javascript-dce.md
```diff
@@ -15,6 +21,26 @@ Unused declarations can appear in cases like:
...
+
+The new way of applying dead code elimination that works with the IR compiler is:
+
+* DCE is disabled when compiling for development, which corresponds to the following Gradle tasks:
+  * `browserDevelopmentRun`
+  * `browserDevelopmentWebpack`
+  * `nodeDevelopmentRun`
+  * `compileDevelopmentExecutableKotlinJs`
+  * `compileDevelopmentLibraryKotlinJs`
+  * Other development tasks you recognize for the word "development" in their name
```
broadwaylamb, line 34:
```suggestion
  * Other Gradle tasks you recognize for the word "development" in their name
```

Otherwise it seems a bit tautological

docs/topics/js/javascript-dce.md
```diff
@@ -15,6 +21,26 @@ Unused declarations can appear in cases like:
...
+* DCE is enabled when compiling for production, which corresponds to the following Gradle tasks:
+  * `browserProductionRun`
+  * `browserProductionWebpack`
+  * `compileProductionExecutableKotlinJs`
+  * `compileProductionLibraryKotlinJs`
+  * Other production tasks you recognize for the word "production" in their name
```
broadwaylamb, line 40:
```suggestion
  * Other Gradle tasks you recognize for the word "production" in their name
```

docs/topics/js/javascript-dce.md
```diff
@@ -1,5 +1,11 @@
 [//]: # (title: Kotlin/JS dead code elimination)
 
+> The DCE tool is deprecated. The DCE tool was designed for the legacy JS backend, which is now obsolete. The current 
+> [JS IR backend](#dce-and-javascript-ir-compiler) supports dead code elimination out of the box, and the @JsExport feature allows specifying which Kotlin 
+> functions and classes to retain during tree shaking.
```
anton-bannykh, line 5: Using "tree shaking" here might be confusing. Although the term is very similar to DCE, it usually refers to JS [tooling](https://webpack.js.org/guides/tree-shaking/) and DCE to refer to Kotlin tooling, which might actually remove the `@JsExport`ed declarations if they don't end up being used by the JS code.

More detail. The difference between DCE and tree shaking is that Kotlin's DCE handles code produced by Kotlin and JS "tree shaking" ususally handles the whole bundle. Those two are not interchangeable and usually work together - DCE help remove Kotlin bloat, and tree shaking handles the whole bundle - both hand-written, and generated. It is less precise with Kotlin code but it can shave off some extra code by taking into account which exported declarations didn't end up being used by the "manual" JS code.

anton-bannykh (approved):

broadwaylamb (approved):

docs/topics/js/javascript-dce.md
AlejandraPedroza: Thanks for the explanation. I'll change tree shaking to DCE.

docs/topics/js/javascript-dce.md
```diff
@@ -15,6 +21,26 @@ Unused declarations can appear in cases like:
...
+  * `compileDevelopmentLibraryKotlinJs`
+  * Other Gradle tasks including "development" in their name
+* DCE is enabled when compiling for production, which corresponds to the following Gradle tasks:
+  * `browserProductionRun`
+  * `browserProductionWebpack`
+  * `compileProductionExecutableKotlinJs`
+  * `compileProductionLibraryKotlinJs`
+  * Other Gradle tasks including "production" in their name
+
+With the @JsExport annotation, you can specify the declarations you want DCE to treat as roots.
```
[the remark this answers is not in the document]
broadwaylamb: No, we call "roots" the declarations that we want to treat as entry points during dead code elimination, so that they and whatever declarations they transitively reference are not eliminated.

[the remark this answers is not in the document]
broadwaylamb: But yeah, we may prefer some term that's easier to understand for ordinary users than some compiler lingo.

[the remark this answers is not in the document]
AlejandraPedroza: I'll leave it as "roots" being the current used concept.

daniCsorbaJB (approved):
nice, thank you! LGTM! 🚀

PR #4397, opened 2024-08-24
Update fun-interfaces.md

sdboyapati:
SAM can have only one abstract method. But it is written wrong on this page. I modified that.


sarahhaggarty:
Thanks for pointing this out @sdboyapati ! To make it more clear I changed the wording to "member functions" instead. But I kept you as the original author. Thanks for your contribution.

PR #4399, opened 2024-08-26
Mention sooner that the tour code runs in-browser

orblivion:
I suggest letting users know _sooner_ that they can do the tour without installing anything yet. This very page suggests that users install Android Studio or IDEA, but doesn't say that the tour lets you play with Kotlin right in the browser.

I for one went ahead and installed Android Studio before clicking the Start button. (Though, I'm making use of it now so it's okay for me in the end).

I wasn't sure how you felt about repeating text between pages or whatever, but I copied some of the language from the note in the next page as a quick idea. Another idea would be to just replicate the entire note.


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your suggestion @orblivion! I think it's a good idea. I modified your suggested text a little to align more with our style. Thanks for your contribution!

PR #4407, opened 2024-08-28
State that both condition must be true

Arthur-Milchior:
The way I understand the sentence, it would mean that any of those condition would suffice. I think it'd be nice to make explicit that both cases must hold

### DIFF
--- a/docs/topics/control-flow.md
+++ b/docs/topics/control-flow.md
@@ -90,7 +90,7 @@ val numericValue = when (getRandomBit()) {
 }
 ```
 
-In `when` _statements_, the `else` branch is mandatory in the following conditions:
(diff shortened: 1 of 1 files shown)


Arthur-Milchior:
Anything I could do to help?

sarahhaggarty:
Hi @Arthur-Milchior !
Sorry it's taken so long to get back to you on this PR. We received some other feedback about this page, specifically that the requirements for statements versus expressions weren't very clear. So, I created a separate PR to address both your suggestions and this other feedback #4480. Please feel free to check out the updated content on the page and let me know if you have any additional feedback. In the meantime, I'll close this PR. Thanks for your suggestions to improve this!

PR #4419, opened 2024-09-04
Grammar inconsistencies around "cases" (uppercase, lowercase, etc.)

samharp:
Adjust grammar usage when describing uppercase, lowercase, camelCase, Pascal case, and other style cases; replace word "case" with synonyms when not describing one of the cases above (e.g. "example", "circumstance", etc.)


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @samharp !
I modified your PR a bit to match our style guide (use active voice,  simple language). After checking with other developers in our team, we decided not to provide the additional explanation about Pascal case, so I removed that as well. Thanks again for your suggestions!

PR #4421, opened 2024-09-06
fix(community): 502 for mobile event render

zoobestik:


zoobestik:
@krutilov, urgent pls

krutilov (approved):

PR #4424, opened 2024-09-09
update: updating warning to be more strict about commiting .kotlin

daniCsorbaJB:


adam-enko (approved):
Looks good, thanks!

docs/topics/gradle/gradle-configure-project.md
```diff
@@ -81,7 +81,8 @@ methods and properties, and supports all the current Gradle features.
 By default, the Kotlin Gradle plugin stores persistent project-specific data at the root of the project,
 in the `.kotlin` directory.
 
-> You may want to add the `.kotlin` directory to your project's `.gitignore` file.
+> Do not commit the `.kotlin` directory to version control.
```
AlejandraPedroza, line 84: Is there a reason why? It would be nice if we could explain with a couple of words. But just a suggestion, up to you.

docs/topics/gradle/gradle-configure-project.md
```diff
@@ -81,7 +81,8 @@ methods and properties, and supports all the current Gradle features.
...
+> For example, if you are using Git, add `.kotlin` to your project's `.gitignore` file.
```
AlejandraPedroza, line 85: Is this only an example of what they can do, or is it the recommended practice? If it's the recommended practice, how about: 
"Instead, if you are using Git, add `.kotlin` to your project's `.gitignore` file."

AlejandraPedroza: Is there something else they can do if they're not using Git?

AlejandraPedroza (approved):
Looks good! Added a couple of suggestions.

docs/topics/gradle/gradle-configure-project.md
daniCsorbaJB: good point - it's an example - there could be other version control systems, but GIt is by far the most popular.

docs/topics/gradle/gradle-configure-project.md
daniCsorbaJB: as discussed it would be better to keep it as unambiguous as possible

adam-enko:
Internal Slack discussion, for future reference https://jetbrains.slack.com/archives/C01K9J6JW2X/p1725539696354249

The key point: 

> Just to recap my understanding, the `.kotlin` dir [used to be](https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects) inside `.gradle/`, which is always git ignored. AFAIK it was only moved because the Gradle devs didn't want plugins to use `.gradle/`. The purpose of the directory hasn't changed, and we never suggested that people should add `.gradle/kotlin/**` files to VCS.

PR #4469, opened 2024-10-02
Update information about Kotlin/Wasm exception catching

JSMonk:


docs/topics/wasm/wasm-js-interop.md
```diff
@@ -359,7 +359,7 @@ external fun <T : JsAny> processData(data: JsArray<T>): T
 
 ## Exception handling
 
-The Kotlin `try-catch` expression can't catch JavaScript exceptions.
+Since version 2.0.0, the Kotlin `try-catch` expression can catch JavaScript exceptions via the `JsException` and `Throwable` types.
```
AlejandraPedroza, line 362:
```suggestion
Starting from version 2.0.0, the Kotlin `try-catch` expression can catch JavaScript exceptions via the `JsException` and `Throwable` types.
```

AlejandraPedroza: Thanks for this update. There's [this issue](https://youtrack.jetbrains.com/issue/KT-71824/Kotlin-Wasm-documentation-update-for-section-Exceptions-handing) about this update and also says we need to describe how it works now and what options do we have now. Is it something we want to mention now?

AlejandraPedroza (commented):
Thanks for this change! I left one comment, please check.

bashor (approved):

PR #4477, opened 2024-10-04
update: forward declarations in c and objc

danil-pavlov:


docs/topics/native/native-c-interop.md
```diff
@@ -646,4 +646,55 @@ fun readData(fd: Int) {
...
+```
+
+```C
+// Second C library
+// Header:
+#include <stdlib.h>
+
+typedef struct {
+    int data;
+} ForwardDeclaredStruct;
```
SvyatoslavScherbina, line 679: This is not exactly the right way to do this. For example, if you include both these libraries to the same C source file, you will get a compilation error, because the type is redefined incompatibly.

Let's get rid of `typedef` instead:
```C
// First C library
#include <stdio.h>

struct ForwardDeclaredStruct;

void consumeStruct(struct ForwardDeclaredStruct* s) {
    printf("Struct consumed\n");
}
```

```C
// Second C library
// Header:
#include <stdlib.h>

struct ForwardDeclaredStruct {
    int data;
};

// Implementation:
struct ForwardDeclaredStruct* produceStruct() {
    struct ForwardDeclaredStruct* s = malloc(sizeof(struct ForwardDeclaredStruct));
    s->data = 42;
    return s;
}
```

docs/topics/native/native-c-interop.md
```diff
@@ -646,4 +646,55 @@ fun readData(fd: Int) {
...
\ No newline at end of file
+exception paths.
+
+### Forward declarations
+
+To import forward declarations, use the `cnames` package. For example, to import a `cstructName` forward declaration
+declared in a C library with a `library.package`, use a special forward declaration package:
+`import cnames.structs.cstructName`.
+
+To transfer objects between libraries, you need to make an explicit `as` cast to and from the corresponding C
```
SvyatoslavScherbina, line 657: > To transfer objects between libraries, you need to make an explicit `as` cast 

This is a bit imprecise and might be misleading. For example, if both libraries have only forward declarations for `cstructName`, then you don't need any cast.

I propose rephrasing this part the following way:
1. First, introduce the second library as well, and mention that it has a real declaration that goes into another package.
2. And after that, you can say that transferring objects between these two libraries requires a cast.

docs/topics/native/native-c-interop.md
```diff
@@ -646,4 +646,55 @@ fun readData(fd: Int) {
...
+    return s;
+}
+```
+
+To transfer objects between them, make the `as` cast in you Kotlin code:
+
+```kotlin
+// Kotlin code:
+fun test() {
+    consumeStruct(produceStruct() as cnames.structs.ForwardDeclaredStruct)
```
SvyatoslavScherbina, line 694: The type should be `CPointer<cnames.structs.ForwardDeclaredStruct>`.

docs/topics/native/native-c-interop.md
```diff
@@ -646,4 +646,55 @@ fun readData(fd: Int) {
...
+}
+```
+
+> Casting is only allowed from the corresponding real class. Otherwise, you'll get an error.
```
SvyatoslavScherbina, line 698: Since the casting actually happens between `CPointer`s, the compiler will consider it as a general unchecked cast, and there will be no error. Let's remove this note at all.

docs/topics/native/native-objc-interop.md
```diff
@@ -228,7 +228,11 @@ A Swift/Objective-C initializer is imported to Kotlin as constructors or as fact
 The latter happens with initializers declared in the Objective-C category or as a Swift extension,
 because Kotlin has no concept of extension constructors.
 
-Kotlin constructors are imported as initializers to Swift/Objective-C. 
+> Remember to expose Swift methods with `@objc` before importing Swift initializers to Kotlin.
```
SvyatoslavScherbina, line 231: (Not sure why this tip is needed at all, but)
Let's replace "expose" with "annotate"?

docs/topics/native/native-objc-interop.md
```diff
@@ -569,6 +573,56 @@ binaries.framework {
 }
 ```
 
+### Forward declarations
+
+To import forward declarations, use the `objcnames` package. For example, to import a `objcstructName` forward declaration
```
SvyatoslavScherbina, line 578: There are no structs among Objective-C forward declarations, only classes and protocols. Within `objcnames.classes` and `objcnames.protocols` packages.

Structs are C concepts and covered by the other document and `cnames.structs` package.

docs/topics/native/native-c-interop.md
SvyatoslavScherbina: The same applies to the Objective-C page.

docs/topics/native/native-objc-interop.md
```diff
@@ -569,6 +573,56 @@ binaries.framework {
...
+```
+
+```ObjC
+// Second objcinterop library
+// Header:
+#import <Foundation/Foundation.h>
+@protocol ForwardDeclaredProtocol
+@end
+// Implementation:
+@implementation ForwardDeclaredProtocolImpl : NSObject
```
SvyatoslavScherbina, line 604: It should be `@interface ForwardDeclaredProtocolImpl : NSObject <ForwardDeclaredProtocol>` instead.

docs/topics/native/native-objc-interop.md
```diff
@@ -569,6 +573,56 @@ binaries.framework {
...
+@end;
```
SvyatoslavScherbina, line 605: `;` is not necessary.

docs/topics/native/native-objc-interop.md
```diff
@@ -569,6 +573,56 @@ binaries.framework {
...
+    return [ForwardDeclaredProtocolImpl new];
+}
+```
+
+To transfer objects between them, make the `as` cast in you Kotlin code:
+
+```kotlin
+// Kotlin code:
+fun test() {
+    consumeStruct(produceStruct() as cnames.structs.ForwardDeclaredStruct)
```
SvyatoslavScherbina, line 617:
```kotlin
consumeProtocol(produceProtocol() as objcnames.protocols.ForwardDeclaredProtocol)
```

docs/topics/native/native-objc-interop.md
danil-pavlov: oh, sorry I forgot about that, there was another request ([KT-45753](https://youtrack.jetbrains.com/issue/KT-45753)), and `@sbogolepov` suggested just to add a note to clarify things

docs/topics/native/native-objc-interop.md
danil-pavlov: Thank you! It's good that we decided to add this because the Obj-C example was copied from the compatibility guide, so we had an error there too. fixed both

docs/topics/native/native-c-interop.md
danil-pavlov: Thank you! Moved this sentence after the second lib

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -926,48 +926,49 @@ and prepare this functionality for the upcoming Kotlin 2.0 release. From now on:
...
+  }
+  ```
+
+  Previously, it was possible to transfer objects between them seamlessly. Now, an explicit `as` cast is required
+  for the forward declaration:
+
+  ```kotlin
+  // Kotlin code:
+  fun test() {
+      consumeProtocol(produceProtocol() as objcnames.protocols.ForwardDeclaredProtocol)
```
SvyatoslavScherbina, line 964: Oh, actually `ForwardDeclaredProtocolProtocol` was the correct name. Sorry I [misled](https://github.com/JetBrains/kotlin-web-site/pull/4477#discussion_r1790357578) you before.

Alternatively, we can declare it as `@protocol ForwardDeclared`, and the name in Kotlin will be `ForwardDeclaredProtocol`.

docs/topics/native/native-objc-interop.md
```diff
@@ -228,7 +228,11 @@ A Swift/Objective-C initializer is imported to Kotlin as constructors or as fact
 The latter happens with initializers declared in the Objective-C category or as a Swift extension,
 because Kotlin has no concept of extension constructors.
 
-Kotlin constructors are imported as initializers to Swift/Objective-C. 
+> Remember to annotate Swift methods with `@objc` before importing Swift initializers to Kotlin.
```
SvyatoslavScherbina, line 231: Technically, initializers are not methods.
So, the more precise way to say this would be:
> Remember to annotate Swift initializers with `@objc` before importing them Kotlin.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
SvyatoslavScherbina: Similar problem in the cast in the other document, `native-objc-interop.md`.

danil-pavlov: Thanks! Let's revert to the previous naming

SvyatoslavScherbina (approved):

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -926,48 +926,49 @@ and prepare this functionality for the upcoming Kotlin 2.0 release. From now on:
...
+  for the forward declaration:
+
+  ```kotlin
+  // Kotlin code:
+  fun test() {
+      consumeProtocol(produceProtocol() as objcnames.protocols.ForwardDeclaredProtocolProtocol)
+  }
+  ```
+
+  > The casting to `objcnames.protocols.ForwardDeclaredProtocolProtocol` is only allowed from the corresponding real class.
```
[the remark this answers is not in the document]
koshachy: If you'd like to keep the original sentence, the first "The" article is not needed.

docs/topics/native/native-c-interop.md
```diff
@@ -646,4 +646,50 @@ fun readData(fd: Int) {
...
+
+// Implementation:
+struct ForwardDeclaredStruct* produceStruct() {
+    struct ForwardDeclaredStruct* s = malloc(sizeof(struct ForwardDeclaredStruct));
+    s->data = 42;
+    return s;
+}
+```
+
+To transfer objects between the two libraries, make the `as` cast in you Kotlin code:
```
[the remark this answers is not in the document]
koshachy: Sounds more natural to me.

docs/topics/native/native-objc-interop.md
```diff
@@ -228,7 +228,11 @@ A Swift/Objective-C initializer is imported to Kotlin as constructors or as fact
 The latter happens with initializers declared in the Objective-C category or as a Swift extension,
 because Kotlin has no concept of extension constructors.
 
-Kotlin constructors are imported as initializers to Swift/Objective-C. 
+> Remember to annotate Swift initializers with `@objc` before importing them to Kotlin.
```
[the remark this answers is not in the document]
danil-pavlov: No, here is just a reminder of something stated above. Some people skip it, so we decided to add a note in this section too

koshachy (approved):

PR #4480, opened 2024-10-09
Conditions and loops: Update when section

sarahhaggarty:
This PR updates the `when` section on the Conditions and loops page to make it clearer when exhaustiveness is required and in which cases.

### DIFF
--- a/docs/topics/control-flow.md
+++ b/docs/topics/control-flow.md
@@ -27,10 +27,12 @@ fun main() {
     // You can also use `else if` in expressions:
     val maxLimit = 1
     val maxOrLimit = if (maxLimit > a) maxLimit else if (a > b) a else b
-
-    //sampleEnd
+  
     println("max is $max")
+    // max is 3
     println("maxOrLimit is $maxOrLimit")
+    // maxOrLimit is 3
+    //sampleEnd
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="if-else-if-kotlin"}
@@ -50,33 +52,124 @@ val max = if (a > b) {
 If you're using `if` as an expression, for example, for returning its value or
 assigning it to a variable, the `else` branch is mandatory.
 
-## When expression
+## When expressions and statements
 
-`when` defines a conditional expression with multiple branches. It is similar to the `switch` statement in C-like languages.
-Its simple form looks like this.
+`when` is a conditional expression that runs code based on multiple possible values or conditions. It is
+similar to the `switch` statement in Java, C, and similar languages. For example:
 
 ```kotlin
-when (x) {
-    1 -> print("x == 1")
-    2 -> print("x == 2")
-    else -> {
-        print("x is neither 1 nor 2")
+fun main() {
+    //sampleStart
+    val x = 2
+    when (x) {
+        1 -> print("x == 1")
+        2 -> print("x == 2")
+        else -> print("x is neither 1 nor 2")
     }
+    // x == 2
+    //sampleEnd
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-conditions-when-statement"}
 
 `when` matches its argument against all branches sequentially until some branch condition is satisfied.
 
-`when` can be used either as an expression or as a statement. If it is used as an expression, the value
-of the first matching branch becomes the value of the overall expression. If it is used as a statement, the values of
-individual branches are ignored. Just like with `if`, each branch can be a block, and its value
-is the value of the last expression in the block.
+You can use `when` in a few different ways. Firstly, you can use `when` either as an **expression** or as a **statement**.
+As an expression, `when` returns a value for later use in your code. As a statement, `when` completes an action
+without returning anything of further use:
 
-The `else` branch is evaluated if none of the other branch conditions are satisfied.
+<table>
+   <tr>
+       <td>Expression</td>
+       <td>Statement</td>
+   </tr>
+   <tr>
+<td>
 
-If `when` is used as an _expression_, the `else` branch is mandatory,
-unless the compiler can prove that all possible cases are covered with branch conditions,
-for example, with [`enum` class](enum-classes.md) entries and [`sealed` class](sealed-classes.md) subtypes).
+```kotlin
+// Returns a string assigned to the 
+// text variable
+val text = when (x) {
+    1 -> "x == 1"
+    2 -> "x == 2"
+    else -> "x is neither 1 nor 2"
+}
+```
+
+</td>
+<td>
+
+```kotlin
+// Returns nothing but triggers a 
+// print statement
+when (x) {
+    1 -> print("x == 1")
+    2 -> print("x == 2")
+    else -> print("x is neither 1 nor 2")
+}
+```
+
+</td>
+</tr>
+</table>
+
+Secondly, you can use `when` with or without a subject. Whether you use a subject with `when` or not, your expression or
+statement behaves the same. We recommend using `when` with a subject when possible, as it makes your code easier to read
+and maintain by clearly showing what you're checking.
+
+<table>
+   <tr>
+       <td>With subject <code>x</code></td>
+       <td>Without subject</td>
+   </tr>
+   <tr>
+<td>
+
+```kotlin
+when(x) { ... }
+```
+
+</td>
+<td>
+
+```kotlin
+when { ... }
+```
+
+</td>
+</tr>
+</table>
+
+Depending on how you use `when`, there are different requirements for whether you need to cover all possible cases in your
+branches.
+
+If you use `when` as a statement, you don't have to cover all possible cases. In this example, some cases aren't covered,
+so nothing happens. However, no error occurs:
+
+```kotlin
+fun main() {
+    //sampleStart
+    val x = 3
+    when (x) {
+        // Not all cases are covered
+        1 -> print("x == 1")
+        2 -> print("x == 2")
+    }
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-when-statement"}
+
+In a `when` statement, the values of individual branches are ignored. Just like with `if`, each branch can be a block, 
+and its value is the value of the last expression in the block.
+
+If you use `when` as an expression, you have to cover all possible cases. In other words, it must be _exhaustive_.
+The value of the first matching branch becomes the value of the overall expression. If you don't cover all cases, 
+the compiler throws an error.
+
+If your `when` expression has a subject, you can use an `else` branch to make sure that all possible cases are covered, but
+it isn't mandatory. For example, if your subject is a `Boolean`, [`enum` class](enum-classes.md), [`sealed` class](sealed-classes.md),
+or one of their nullable counterparts, you can cover all cases without an `else` branch:
 
 ```kotlin
 enum class Bit {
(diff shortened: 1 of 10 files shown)


docs/topics/control-flow.md
```diff
@@ -50,12 +50,58 @@ val max = if (a > b) {
 If you're using `if` as an expression, for example, for returning its value or
 assigning it to a variable, the `else` branch is mandatory.
 
-## When expression
+## When expressions and statements
 
-`when` defines a conditional expression with multiple branches. It is similar to the `switch` statement in C-like languages.
-Its simple form looks like this.
+`when` is a conditional expression that executes code based on multiple possible values or conditions. It is
+similar to the `switch` statement in C-like languages. For example:
```
serras, line 56:
```suggestion
closely related to the `switch` statement in Java, C, and similar languages. For example:
````

I feel like these days "C-like languages" is not as well understood as it was a few years ago.

File: docs/topics/control-flow.md [hunk 2]
```diff
@@ -50,12 +50,58 @@ val max = if (a > b) {
...
 ```kotlin
+fun main() {
+    //sampleStart
+    val x = 2
+    when (x) {
+        1 -> print("x == 1")
+        2 -> print("x == 2")
+        else -> {
+            print("x is neither 1 nor 2")
+        }
```
serras, line 67:
```suggestion
        else -> println("x is neither 1 nor 2")
```

Just to keep the same style across all branches (unless you want to show something different)

docs/topics/control-flow.md
```diff
@@ -50,12 +50,58 @@ val max = if (a > b) {
...
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-conditions-when-statement"}
+
+`when` matches its argument against all branches sequentially until some branch condition is satisfied.
+
+You can use `when` in a few different ways. Firstly, you can use `when` either as an **expression** or as a **statement**.
+As an expression, `when` returns a value that you can use later in your code. As a statement, `when` completes an action
+without returning anything:
```
serras, line 78: _Mega-nitpick_: Maybe it is more about the developer not caring about the return value of that action? As of now, every Kotlin function returns something, but sometimes is the un-interesting `Unit` value.

docs/topics/control-flow.md
```diff
@@ -50,12 +50,58 @@ val max = if (a > b) {
...
+   <tr>
+       <td>Expression</td>
+       <td>Statement</td>
+   </tr>
+   <tr>
+<td>
+
+```kotlin
+// Returns a string
+when (x) {
```
serras, line 90:
```suggestion
val text = when(x) {
```

Maybe this emphasizes the "returned value"?

docs/topics/control-flow.md
```diff
@@ -65,56 +111,96 @@ when (x) {
...
+    }
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-when-statement"}
+
+In a `when` statement, the values of individual branches are ignored. Just like with `if`, each branch can be a block, 
+and its value is the value of the last expression in the block.
+
+If you use `when` as an expression, you have to cover all possible cases. The value of the first matching branch becomes
```
serras, line 170: _General comment_: maybe you want to point people to the new 2.1 feature "guards".

docs/topics/whatsnew1530.md
```diff
@@ -32,7 +32,7 @@ and type inference:
 >
 {style="warning"}
 
-An _exhaustive_ [`when`](control-flow.md#when-expression) statement contains branches for all possible types or values of its subject or for some types plus an `else` branch. In other words, it covers all possible cases.
+An _exhaustive_ [`when`](control-flow.md#when-expressions-and-statements) statement contains branches for all possible types or values of its subject or for some types plus an `else` branch. In other words, it covers all possible cases.
```
serras, line 35:
```suggestion
An _exhaustive_ [`when`](control-flow.md#when-expressions-and-statements) statement contains branches for either (a) all possible types or values of its subject, or (b) for some types plus an `else` branch. In other words, it covers all possible cases.
```

This sentence is a bit hard to read. I'm not sure my suggestion is any better, though.

serras (approved):
I think the explanation of all the possible forms of `when` is much clearer now.

docs/topics/control-flow.md
sarahhaggarty: I'll make sure to do that once Ale's content is ready

docs/topics/control-flow.md
sarahhaggarty: I updated the text to say "return anything of further use". WDYT?

serras: I like it, it's more exact without going into the nitty gritty details of statement vs. expression.

docs/topics/whatsnew1530.md
serras: It reads better. Still by brains parses "for some types with an `else` branch" as "for some (types with an `else` branch)" instead of "(for some types) with (an `else` branch)".

docs/topics/control-flow.md
```diff
@@ -132,7 +213,7 @@ when (x) {
 }
 ```
 
-You can also check a value for being `in` or `!in` a [range](ranges.md) or a collection:
+You can also check whether a value is `in` or `!in` a [range](ranges.md) or a collection:
```
serras, line 216:
```suggestion
You can also check whether a value is `in` (contained in) or `!in` (not contained in) a [range](ranges.md) or a collection:
```

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
 If you're using `if` as an expression, for example, for returning its value or
 assigning it to a variable, the `else` branch is mandatory.
 
-## When expression
+## When expressions and statements
 
-`when` defines a conditional expression with multiple branches. It is similar to the `switch` statement in C-like languages.
-Its simple form looks like this.
+`when` is a conditional expression that executes code based on multiple possible values or conditions. It is
```
AlejandraPedroza, line 55: Do you think it's feasible to change "executes" for "runs"? Or any other word to avoid "execute" as much as possible, as mentioned in JB style guide and other TW style and best practices guidelines.

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
...
+similar to the `switch` statement in Java, C, and similar languages. For example:
 
 ```kotlin
+fun main() {
+    //sampleStart
+    val x = 2
+    when (x) {
+        1 -> print("x == 1")
+        2 -> print("x == 2")
+        else -> print("x is neither 1 nor 2")
```
AlejandraPedroza, line 65: Let's add a code line reflecting the output, also as per our SG

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
...
+    }
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-conditions-when-statement"}
+
+`when` matches its argument against all branches sequentially until some branch condition is satisfied.
+
+You can use `when` in a few different ways. Firstly, you can use `when` either as an **expression** or as a **statement**.
+As an expression, `when` returns a value that you can use later in your code. As a statement, `when` completes an action
```
AlejandraPedroza, line 75:
```suggestion
As an expression, `when` returns a value for later use in your code. As a statement, `when` completes an action
```

AlejandraPedroza: To avoid repeating "you can" many times in a single paragraph.

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
...
-`when` matches its argument against all branches sequentially until some branch condition is satisfied.
+</td>
+</tr>
+</table>
 
-`when` can be used either as an expression or as a statement. If it is used as an expression, the value
-of the first matching branch becomes the value of the overall expression. If it is used as a statement, the values of
-individual branches are ignored. Just like with `if`, each branch can be a block, and its value
-is the value of the last expression in the block.
+Secondly, you can use `when` with or without a subject.
```
AlejandraPedroza, line 113:
```suggestion
Secondly, you can use `when` with or without a subject:
```

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
...
+        // Not all cases are covered
+        1 -> print("x == 1")
+        2 -> print("x == 2")
+    }
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-when-statement"}
+
+In a `when` statement, the values of individual branches are ignored. Just like with `if`, each branch can be a block,
```
AlejandraPedroza, line 162: Can we add information about what happens when not all the cases are covered and the value doesn't match any of the branches? It is not an error, but is there something returned? 

For example:

fun main() {
    //sampleStart
    val x = 3
    when (x) {
        // Not all cases are covered
        1 -> print("x == 1")
        2 -> print("x == 2")
    }
    //sampleEnd
}

docs/topics/control-flow.md
```diff
@@ -50,71 +50,152 @@ val max = if (a > b) {
...
-    Color.RED -> println("red") // no branches for GREEN and BLUE
-    else -> println("not red") // 'else' is required
+when {
+    a > b -> "a is greater than b"
+    a < b -> "a is less than b"
+    else -> "a is equal to b"
 }
 ```
 
+`when` expressions and statements can be used in a number of different ways to simplify your code, handle multiple conditions,
```
AlejandraPedroza, line 195: `when` expressions and statements offer different ways to... 
OR
`when` expressions and statements work in different ways to...
To avoid passive voice

docs/topics/control-flow.md
```diff
@@ -143,9 +224,9 @@ when (x) {
 }
 ```
 
-Another option is checking that a value `is` or `!is` of a particular type. Note that,
-due to [smart casts](typecasts.md#smart-casts), you can access the methods and properties of the type without
-any extra checks.
+Additionally, you can check that a value `is` or `!is` a particular type. Note that,
```
AlejandraPedroza, line 227: How about making it consistent with the previous leading line that was super clear:

Additionally, you can check that a value is or isn't of a particular type via the `is` or `!is` keywords:

docs/topics/whatsnew1530.md
```diff
@@ -32,7 +32,7 @@ and type inference:
 >
 {style="warning"}
 
-An _exhaustive_ [`when`](control-flow.md#when-expression) statement contains branches for all possible types or values of its subject or for some types plus an `else` branch. In other words, it covers all possible cases.
+An _exhaustive_ [`when`](control-flow.md#when-expressions-and-statements) statement contains branches for either all possible types or values of its subject, or for certain types and includes an `else` branch to cover any remaining cases.
```
AlejandraPedroza, line 35: This clarification about the concept of "exhaustive" is important. Should we mention it in control-flow.md#when-expressions-and-statements ?

AlejandraPedroza (changes requested):
This is a great revamp of the` when` docs!

docs/topics/control-flow.md
sarahhaggarty: Added for other samples too.

docs/topics/control-flow.md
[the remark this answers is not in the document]
sarahhaggarty: There's the note below which explains that we recommend using a subject to make it easier to read and maintain your code. Even further below, we explain that a subject with an expression has the advantage that you don't have to use an `else` branch. Basically, using a subject with expressions makes the compiler's life easier because it just has to check the subject. I didn't add this detail because explaining _why_ it's easier leads to describing the inner mechanics of the compiler 😅

docs/topics/control-flow.md
sarahhaggarty: I updated the text to try and explain this better. What do you think? :)

docs/topics/control-flow.md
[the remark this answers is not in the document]
AlejandraPedroza: I see. So the behavior is the same, and the same goal can be achieved with and without subject, but adding a subject is a choice that makes the life of the compiler easier, right? 
I think the gap we can bridge is clarifying that having or not having a subject doesn't change/affect the output of the code and then reinforcing what is the best practice and why. What do you think? Otherwise, even that we have the note about the recommendation, when looking at the 2 options in the table, there's an open question about: which one should I use according to what my code must achieve (beyond the recommended practice) ?

AlejandraPedroza (approved):
Great improvement to the conditions docs! LGTM 👍

PR #4484, opened 2024-10-11
chore: bump KSP and libs versions

koshachy:
Restructure the v.list a bit

### DIFF
--- a/docs/v.list
+++ b/docs/v.list
@@ -24,6 +24,8 @@
 
     <var name="dateTimeVersion" value="0.6.1" type="string"/>
 
+    <var name="lincheckVersion" value="2.34" type="string"/>
+
     <var name="ktorVersion" value="3.0.0" type="string"/>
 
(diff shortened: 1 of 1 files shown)


docs/v.list
```diff
@@ -4,60 +4,55 @@
...
-
-    <var name="dateTimeVersion" value="0.6.0" type="string"/>
-
     <var name="foojayResolver" value="0.5.0" type="string"/>
 
-    <var name="languageVersion" value="2.0" type="string"/>
-
-    <var name="apiVersion" value="2.0" type="string"/>
-
     <var name="lincheckVersion" value="2.34" type="string"/>
```
danil-pavlov, line 56: I think Lincheck should also be in the Libraries and Frameworks section

danil-pavlov (approved):
Love the Ordnung! Just the lincheck comment left

docs/v.list
koshachy: I've moved, thanks!

PR #4486, opened 2024-10-11
Move conditional expressions and ranges practice section

alcarraz:
Conditional expressions and ranges practice, is only about conditional expressions, so it is better placed nearer to Conditional expressions.

Also, ranges is more suitable to be practiced with loops, so it would make more sense to group it with it, or to have its own practice section.


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @alcarraz !

PR #4487, opened 2024-10-11
The first exercise of Functions practice could be confusing

alcarraz:
The exercise has a `// Write your code here`in the function body, but the function header needs to be modified too.

Either:
- the correct parameter and return type should be added to the function header.
- the function should be removed entirely and just put the `//Write your code here` block.
- the instruction should be `// modify the next function to meet the requirements`

This suggestion goes with the second one.


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @alcarraz !

PR #4491, opened 2024-10-15
feat: non-local break and continue

danil-pavlov:


docs/topics/whatsnew-eap.md
```diff
@@ -80,6 +80,49 @@ These features are available in preview, and we encourage you to try them and sh
 
 [See the full list of Kotlin language design features and proposals](kotlin-language-features-and-proposals.md).
 
+### Non-local break and continue
+
+Kotlin %kotlinEapVersion% adds a preview of another long-awaited feature, an opportunity to use non-local `break` and
```
koshachy, line 85: Isn't it too marketish even for a blog post?

Kotlin %kotlinEapVersion% introduces a preview of a new feature: support for non-local` break` and `continue`.

docs/topics/whatsnew-eap.md
```diff
@@ -80,6 +80,49 @@ These features are available in preview, and we encourage you to try them and sh
...
+`continue`. This feature expands the toolset you can use in the scope of inline functions and reduces boilerplate code
```
koshachy, line 86: I found the first part of the sentence is too hard.
Maybe we can simplify it a bit without using "toolset"?
Something like "provides/adds more flexibility within inline functions"?

docs/topics/whatsnew-eap.md
```diff
@@ -80,6 +80,49 @@ These features are available in preview, and we encourage you to try them and sh
...
+in your project.
+
+Previously, you could only use non-local returns. Now, Kotlin also supports `break` and `continue` [jump expressions](returns.md)
```
koshachy, line 89: Maybe the "before/after" will show better the profit of using non-local break and continue? 
What do you think?

docs/topics/whatsnew-eap.md
```diff
@@ -80,6 +80,49 @@ These features are available in preview, and we encourage you to try them and sh
...
+non-locally. That means that you can apply them within lambdas passed as arguments to an inline function that encloses
```
koshachy, line 90:
```suggestion
non-locally. You can apply them within lambdas passed as arguments to an inline function that encloses
```

koshachy (changes requested):
Nice one,
See the comments.
Feel free to start a discussion.

docs/topics/whatsnew-eap.md
danil-pavlov: We can try, but I'm afraid the most important part with the run lambda get's cut. Please check. we can also do two separate blocks one below the other

docs/topics/whatsnew-eap.md
danil-pavlov: I don't think so, and I don't usually use such phrasing, but here it's justified, the feature request is 12 years old.

docs/topics/whatsnew-eap.md
danil-pavlov: I think it's important to keep this connection here; we need to explain what "non-locally" actually means.

PR #4496, opened 2024-10-21
Fixed: Swapped roles of Producer and Consumer in the sentence

BEOKS:
I corrected the sentence where it seemed like the Producer and Consumer were swapped.


lukellmann:
I think it is right the way it was before this PR.

A producer _produces_ (creates or returns) objects. If you have a producer, you only _read_ (get/retrieve) objects from it and don't put objects into it.

```kotlin
object Producer {
    fun produce() = "string"
}

// usage (you only read from it)
val thing = Producer.produce()
```

A consumer _consumes_ (receives) objects. If you have a consumer, you only _write_ (put) objects into it and don't get objects from it.

```kotlin
object Consumer {
    fun consume(thing: String) {
        println(thing)
    }
}

// usage (you only write to it)
Consumer.consume("string")
```

nikitabobko:
Yes, I think the sentence is correct

PR #4500, opened 2024-10-23
Add new section to whatsnew2020.md

sarahhaggarty:
This PR adds a new section about a Gradle topic to whatsnew2020.md.

### DIFF
--- a/docs/topics/whatsnew2020.md
+++ b/docs/topics/whatsnew2020.md
@@ -584,6 +584,27 @@ By default, this property is set to `false` and the `apiElements` variant in Gra
 We would appreciate your feedback on this new approach. Have you noticed any performance improvements while using it?
 Let us know by adding a comment in [YouTrack](https://youtrack.jetbrains.com/issue/KT-61861/Gradle-Kotlin-compilations-depend-on-packed-artifacts).
 
+### Aligned dependency behavior of Kotlin Gradle plugin with java-test-fixtures plugin
+
+Prior to Kotlin 2.0.20, if you used the [`java-test-fixtures` plugin](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures)
+in your project, there was a difference between Gradle and the Kotlin Gradle plugin in how dependencies were propagated. 
+
+The Kotlin Gradle plugin propagated dependencies:
+
+* From the `java-test-fixtures` plugin's `implementation` and `api` dependency types to the `test` source set compilation classpath.
+* From the main source set's `implementation` and `api` dependency types to the `java-test-fixtures` plugin's source set compilation classpath. 
+
+However, Gradle only propagated dependencies in the `api` dependency types.
+
+This difference in behavior led to some projects finding resource files multiple times in the classpath. 
+
+As of Kotlin 2.0.20, the Kotlin Gradle plugin's behavior is aligned with Gradle's `java-test-fixtures` plugin so this 
+problem no longer occurs for this or other Gradle plugins.
+
+As a result of this change, some dependencies in the `test` and `testFixtures` source sets may no longer be accessible.
+If this happens, either change the dependency declaration type from `implementation` to `api` or add a new dependency 
+declaration on the affected source set.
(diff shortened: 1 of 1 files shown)


docs/topics/whatsnew2020.md
```diff
@@ -584,6 +584,27 @@ By default, this property is set to `false` and the `apiElements` variant in Gra
...
 Let us know by adding a comment in [YouTrack](https://youtrack.jetbrains.com/issue/KT-61861/Gradle-Kotlin-compilations-depend-on-packed-artifacts).
 
+### Aligned dependency behavior of Kotlin Gradle plugin with java-test-fixtures plugin
+
+Prior to Kotlin 2.0.20, if you used the [`java-test-fixtures` plugin](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures)
+in your project, there was a difference between Gradle and the Kotlin Gradle plugin in how dependencies were propagated. 
+
+The Kotlin Gradle plugin propagated dependencies:
+
+* From the `java-test-fixtures` plugin's `implementation` and `api` configurations to the `test` sources compilation classpath.
```
danil-pavlov, line 594:
```suggestion
* From the `java-test-fixtures` plugin's `implementation` and `api` configurations to the `test` source set compilation classpath.
```

docs/topics/whatsnew2020.md
```diff
@@ -584,6 +584,27 @@ By default, this property is set to `false` and the `apiElements` variant in Gra
...
+
+However, Gradle only propagated dependencies in the `api` configurations.
+
+This difference in behavior led to some projects finding resource files multiple times in the classpath. 
+
+As of Kotlin 2.0.20, the Kotlin Gradle plugin's behavior is aligned with Gradle's `java-test-fixtures` plugin so this 
+problem no longer occurs for this or other Gradle plugins.
+
+As a result of this change, some dependencies in the `test` and `testFixtures` source sets may no longer be accessible.
+If this happens, either change the dependency declaration type from `implementation` to `api` or add a new dependency
```
danil-pavlov, line 605:
```suggestion
If this happens, either change the dependency type from `implementation` to `api` or add a new dependency 
```

docs/topics/whatsnew2020.md
danil-pavlov: Here and below, in our documentation, we generally refer to such dependency types as `implementation` and `api` as "scopes", not "configurations"

docs/topics/whatsnew2020.md
danil-pavlov: Since we're talking about changing the dependency type here, maybe we shouldn't use "source sets", but refer directly to "scopes" too

and even use general `test*` and `testFixtures*` here, since they could have different types, like `testImplementation`, `testFixtureApi`, and so on

docs/topics/whatsnew2020.md
ALikhachev: Here we refer them just as "dependency types": https://kotlinlang.org/docs/gradle-configure-project.html#dependency-types

Gradle calls them configurations: https://docs.gradle.org/current/userguide/dependency_configurations.html#sec:what-are-dependency-configurations

docs/topics/whatsnew2020.md
ALikhachev: > Since we're talking about changing the dependency type here, maybe we shouldn't use "source sets", but refer directly to "scopes" too

Those scopes are assigned to some source sets anyway. 

> and even use general test* and testFixtures* here, since they could have different types, like testImplementation, testFixtureApi, and so on

But what would we achieve with such patterns?
I either see the current option or some explicit mapping like `implementation -> api`, `testFixturesImplementation -> testApi` (or a suggestion to declare the dependencies directly in `testFixturesImplementation`/`testImplementation`)

sarahhaggarty: To keep it simple, I'll leave it as-is.

danil-pavlov (approved):

PR #4514, opened 2024-10-28
Update null-safety.md

leemckeeman:
This example is demonstrating a compound predicate for a check of a nullable. In the if body, a property of the non-null value is printed, and in the else a fallback message is printed. The comments on this example were switched.

### DIFF
--- a/docs/topics/null-safety.md
+++ b/docs/topics/null-safety.md
@@ -156,10 +156,10 @@ fun main() {
     // Checks for nullability first and then accesses length
     if (b != null && b.length > 0) {
         print("String of length ${b.length}")
-    // Provides alternative if the condition is not met  
+        // String of length 6
     } else {
         print("Empty string")
-        // String of length 6
+        // Provides alternative if the condition is not met      
     }
 //sampleEnd
 }


AlejandraPedroza (changes requested):
Hi!
Thank you so much for your contribution. According to our style guide, the comments describe the purpose of the next code line, and if the code outputs the result using println() or similar, we reflect this in a comment on the next line. In this case, "// Provides alternative if the condition is not met " explains what `} else { ` does, and "// String of length 6" reflects the output of the print, so that is the reason of the position. However, it is not clear due to indentation. I am requesting changes to improve the indentation and make it clearer so that the result snippet is:

```
// Assigns a nullable string to a variable  
val b: String? = "Kotlin"

// Checks for nullability first and then accesses length
if (b != null && b.length > 0) {
    print("String of length ${b.length}")
    
// Provides alternative if the condition is not met  
} else {
    print("Empty string")
}
// String of length 6
```

leemckeeman:
ah, i see! 
// String of length 6
Is at the wrong scope/level of indentation, which seems to tie it to the print (which is why I thought it was in the wrong place). Rather than it belonging immediately below the line that will generate it, it belongs after the entire if/else block, because at the end of the whole block this is the output with the given initialization. MMMM. Ok, got it, thanks for the follow-up!

PR #4521, opened 2024-11-04
Added missed language feature "Expose boxed inline value classes in JVM"

johnlang:
Looks like we could've missed one of the language features on the page back in September, adding lost things now.


koshachy (approved):
LGTM!

PR #4525, opened 2024-11-05
update: adding information about different classpath snapshots

daniCsorbaJB:


docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -13,11 +13,20 @@ On this page, you can learn about the following topics:
...
-build cache](#gradle-build-cache-support) and supports [compilation avoidance](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_compile_avoidance).
+The Kotlin Gradle plugin supports incremental compilation, which is enabled by default for Kotlin/JVM and Kotlin/JS projects.
+Incremental compilation tracks changes to files in the classpath between builds so that only the files affected
+by these changes are compiled.
+This approach works with [Gradle's build cache](#gradle-build-cache-support) and supports [compilation avoidance](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_compile_avoidance).
 
-Incremental compilation is supported for Kotlin/JVM and Kotlin/JS projects, and is enabled by default.
+For Kotlin/JVM, incremental compilation relies on classpath snapshots,
+which capture the API structure of modules to determine when recompilation is necessary.
+You can configure classpath snapshots as **fine-grained** or **coarse-grained**, depending on the level of detail required:
```
mazhukinevgeniy, line 23: It's not a public option. I'd rather say something like this:

```
To optimize the overall pipeline, Kotlin compiler uses two types of classpath snapshots: fine-grained and coarse-grained. 

<some explanations follow>
```

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -13,11 +13,20 @@ On this page, you can learn about the following topics:
...
+
+* **Fine-grained snapshots** track changes within class members, such as fields or properties.
+This setting minimizes the scope of recompilation by targeting only files affected by specific member changes.
```
mazhukinevgeniy, line 26: So, it's not a setting, but rather

> Detecting the member-level changes allows Kotlin compiler to find and recompile the classes that might be using these members directly.

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -13,11 +13,20 @@ On this page, you can learn about the following topics:
...
+To maintain performance, the Kotlin Gradle plugin skips creating fine-grained snapshots for `.jar` files in the Gradle cache.
+* **Coarse-grained snapshots** track changes only at the class level, resulting in smaller snapshots but a broader recompilation scope.
+This setting is useful for classes that change infrequently, such as external libraries.
```
mazhukinevgeniy, line 29: Same as above, it's not a setting. But more importantly, can we somehow tie this to the previous explanation?


> **Fine**
> Detecting the member-level changes allows Kotlin compiler to find and recompile the classes that might be using these members directly.
> <...>
> **Coarse**
> Compared to the fine-grained snapshot, we would not only recompile the classes which depend on the changed members, but we would also recompile all other classes which depend on the changed class.

There might be too much "class" in this sentence, but maybe you'd see a way to rephrase it better.

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -31,14 +40,14 @@ There are several ways to disable incremental compilation:
 
   The parameter should be added to each subsequent build.
 
-Note: Any build with incremental compilation disabled invalidates incremental caches. The first build is never incremental.
+Disabling incremental compilation invalidates incremental caches. The first build is never incremental.
```
mazhukinevgeniy, line 43: The previous version makes sense to me, because the incremental cache becomes invalid only after the build. Do you think it might be too clunky?

mazhukinevgeniy (changes requested):

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: I think it's a bit too clunky - especially starting with Note: feels weird as normally the rest of the docs uses admonition in this case. 

We could try to phrase it more explicitly, how about:
"When you disable incremental compilation, incremental caches become invalid after the build." 

alternatively, perhaps:
"When incremental compilation is disabled, incremental caches are invalidated after the build."

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: I see - I like this version a lot - updated it!

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: So how about we connect it like this:

> **Fine-grained snapshots** track changes within class members, such as fields or properties. By detecting these member-level changes, the Kotlin compiler can recompile only the classes that directly use the modified members. To maintain performance, the Kotlin Gradle plugin skips creating fine-grained snapshots for `.jar` files in the Gradle cache."

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: Hmm 🤔 - yes it repeats class a bit often, but I don't think there is a way to avoid it without risking it becoming ambiguous.

What is your opinion about this version:

> Compared to fine-grained snapshots, coarse-grained snapshots recompile not only the classes that depend on the modified members but also all other classes that depend on the modified class.

It's a bit more explicit but removes any potential confusion from the sentence

docs/topics/gradle/gradle-compilation-and-caches.md
mazhukinevgeniy: How about this?

>Fine-grained snapshots include detailed information about class members, such as properties or methods. When member-level changes are detected, Kotlin compiler can recompile only the classes that depend on the modified members.

To break down the changes:
1) snapshots don't really track changes, they are just that - snapshots of a particular state
2) dependency is a bit wider than direct use, because sometimes we need to compile a class file to see that it won't use a new function

>skips creating fine-grained snapshots for .jar files in the Gradle cache

maybe something like this - "instead creates coarse-grained snapshots for <..>"?

docs/topics/gradle/gradle-compilation-and-caches.md
mazhukinevgeniy: It feels wrong to say that "snapshots recompile something", because they don't.

How about a more basic explanation? A rough version:

> Coarse-grained snapshots store class info without member-level details. When a class-level change is detected, Kotlin compiler knows that some members have changed, but it doesn't know which exactly. So it would recompile all uses of the changed class.

mazhukinevgeniy:
looks good to me, thank you!

mazhukinevgeniy (approved):

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -13,11 +13,21 @@ On this page, you can learn about the following topics:
...
 ## Incremental compilation
 
-The Kotlin Gradle plugin supports incremental compilation. Incremental compilation tracks changes to files in the classpath
-between builds so that only the files affected by these changes are compiled. Incremental compilation works with [Gradle's
-build cache](#gradle-build-cache-support) and supports [compilation avoidance](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_compile_avoidance).
-
-Incremental compilation is supported for Kotlin/JVM and Kotlin/JS projects, and is enabled by default.
+The Kotlin Gradle plugin supports incremental compilation, which is enabled by default for Kotlin/JVM and Kotlin/JS projects.
+Incremental compilation tracks changes to files in the classpath between builds so that only the files affected
+by these changes are compiled.
```
AlejandraPedroza, line 18: Would saying "by the last changes" be accurate? If it is, I think it's clearer.

docs/topics/gradle/gradle-compilation-and-caches.md
```diff
@@ -13,11 +13,21 @@ On this page, you can learn about the following topics:
...
+This approach works with [Gradle's build cache](#gradle-build-cache-support) and supports [compilation avoidance](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_compile_avoidance).
+
+For Kotlin/JVM, incremental compilation relies on classpath snapshots,
+which capture the API structure of modules to determine when recompilation is necessary.
+To optimize the overall pipeline, the Kotlin compiler uses two types of classpath snapshots:
+
+* **Fine-grained snapshots** include detailed information about class members, such as properties or functions.
+When member-level changes are detected, the Kotlin compiler recompiles only the classes that depend on the modified members.
+To maintain performance, the Kotlin Gradle plugin creates coarse-grained snapshots for `.jar` files in the Gradle cache.
+* **Coarse-grained snapshots** only contain the class ABI hash.
```
AlejandraPedroza, line 28:
```suggestion
* **Coarse-grained snapshots:** only contain the class ABI hash.
```

AlejandraPedroza: How known is the acronym ABI? Would it be worth it to write the complete term (Application Binary Interface) ?

AlejandraPedroza (approved):
Just minor suggestions. LGTM! 🚀

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: I think it would be slightly inaccurate, if I understand correctly. Incremental compilation tracks all changes across the classpath, not just the most recent/last one.

docs/topics/gradle/gradle-compilation-and-caches.md
daniCsorbaJB: Hmm 🤔 - I think we could put a link to it (we had a link to it in an another section as well)

PR #4531, opened 2024-11-06
update gradle-compiler-options.md and fix anchors

AlejandraPedroza:


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,8 +15,52 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+  like `commonMain`, `nativeMain`, and `commonTest`.
+* The values of target compiler options are used as the default for compilation unit (task) compiler options, for
+  example, `compileKotlinJvm` and `compileTestKotlinJvm` tasks.
+
+In turn, configurations made at a lower level override related settings at a higher level:
+
+* Task-level compiler options override related configurations at the target or the extension level.
+* Target-level compiler options override related configurations at the extension level.
+
+To confirm which level of compiler arguments is applied to the compilation, use the `DEBUG` level of Gradle [logging](https://docs.gradle.org/current/userguide/logging.html).
```
Tapchicoma, line 40: Maybe add the following:
- for JVM and JS/WASM tasks search log for "Kotlin compiler args:" string
- for Native tasks search log for "Arguments =" string

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,8 +15,52 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+Gradle is making these properties traceable, so understanding what overrides your values will soon be easier.
```
Tapchicoma, line 41: I would drop this for now until Gradle will actually will have traceability support.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,8 +15,52 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+
+> If you are a third-party plugin author, it is recommended that you apply your configuration on the project level to avoid
```
Tapchicoma, line 43: Possibly we could link here to new extensions introduced in Kotlin 2.1.0 - they could be used by plugin authors to do it :thinking:

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,8 +15,52 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+kotlin {
+    compilerOptions {
+        optIn.add("kotlin.RequiresOptIn")
+    }
+}    
+```
+
+If you want to configure a plugin of a target different from JVM/Android and [Kotlin Multiplatform](multiplatform-dsl-reference.md), 
+use the `compilerOptions {}` property of the corresponding Kotlin compilation task. The following examples 
+show how to set this configuration up in both Kotlin and Groovy DSLs:
```
Tapchicoma, line 63: It looks should not belong to "Extension level" section - maybe could be a part of "Compilation Unit" section

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,8 +15,52 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+Gradle is making these properties traceable, so understanding what overrides your values will soon be easier.
+
+> If you are a third-party plugin author, it is recommended that you apply your configuration on the project level to avoid 
+> overriding issues. Also, it is recommended that you document such configuration on your side explicitly.
+>
+{style="tip"}
+
+### Extension level
+
+You can configure compiler options for all the targets and source sets
```
Tapchicoma, line 50:
```suggestion
You can configure compiler options for all the targets and shared source sets
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -43,40 +87,42 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
-
-The names of the tasks in Android Projects contain [build variant](https://developer.android.com/studio/build/build-variants.html)
-names and follow the `compile<BuildVariant>Kotlin` pattern, for example, `compileDebugKotlin` or `compileReleaseUnitTestKotlin`.
-
-For both the JVM and Android projects, it's possible to define options using the project Kotlin extension DSL:
-
-<tabs group="build-script">
-<tab title="Kotlin" group-key="kotlin">
+You can configure compiler options for the JVM/Android and Kotlin Multiplatform targets
+in the `compilerOptions {}` block inside the `target {}` block:
```
Tapchicoma, line 93: `target {}` DSL is valid for JVM/Android, but not valid for Multiplatform - in KMP it should be specific target name, e.g. `jvm { compilerOptions {}}`.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -43,40 +87,42 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
 }
 ```
 
-</tab>
-<tab title="Groovy" group-key="groovy">
+### Compilation unit level
 
-```groovy
-kotlin {
+You can configure compiler options for a specific compilation unit or task in a `compilerOptions {}`
```
Tapchicoma, line 107: Maybe we could also mention here how to access and configure via `KotlinCompilation`:
```
kotlin {
    jvm {
        val main by compilations.getting {
            compileTaskProvider.configure {
                compilerOptions {

                }
            }
        }
    }
}
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -43,40 +87,42 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
 }
 ```
 
-</tab>
-</tabs>
+## Target the JVM
+
+As explained before, you can define compiler options for your JVM/Android projects at the extension, target, and compilation unit levels.
+
+JVM compilation tasks are called `compileKotlin` for production code and `compileTestKotlin`
```
Tapchicoma, line 122:
```suggestion
Default JVM compilation tasks are called `compileKotlin` for production code and `compileTestKotlin`
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -204,14 +249,25 @@ argument to this attribute or a list of arguments:
...
-val compileKotlin: KotlinCompilationTask<*> by tasks
-
-// Single experimental argument
-compileKotlin.compilerOptions.freeCompilerArgs.add("-Xexport-kdoc")
-// Single additional argument, can be a key-value pair
-compileKotlin.compilerOptions.freeCompilerArgs.add("-Xno-param-assertions")
-// List of arguments
-compileKotlin.compilerOptions.freeCompilerArgs.addAll(listOf("-Xno-receiver-assertions", "-Xno-call-assertions"))
+kotlin {
+    target {
```
Tapchicoma, line 253: Maybe better here to use top-level extension level instead of target one?

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -236,23 +292,25 @@ tasks.named('compileKotlin', KotlinCompilationTask) {
...
-tasks
-    .withType<org.jetbrains.kotlin.gradle.tasks.KotlinJvmCompile>()
-    .configureEach {
-        compilerOptions
-            .languageVersion
-            .set(
-                org.jetbrains.kotlin.gradle.dsl.KotlinVersion.%gradleLanguageVersion%
-            )
+kotlin {
+    target {
```
Tapchicoma, line 308: Again imho better here to not use target but extension level example

docs/topics/gradle/gradle-compiler-options.md
AlejandraPedroza: Got it! I see the [task is in progress](https://youtrack.jetbrains.com/issue/KT-72280/DocsBuild-tools-Document-new-extensions-hierarchy-for-KGP), and I don't find a PR with the doc update yet. I'll leave it pending and ask Sarah about it to add the link as soon as we have an anchor to point.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -204,14 +269,23 @@ argument to this attribute or a list of arguments:
...
 // ...
 
-val compileKotlin: KotlinCompilationTask<*> by tasks
+kotlin {
+    compilerOptions {
+        // Single experimental argument
+        freeCompilerArgs.add("-Xexport-kdoc")
+
+        // Single additional argument
+        freeCompilerArgs.add("-Xno-param-assertions")
```
Tapchicoma, line 278: I would prefer to use here more specialized DSL than `freeCompilerArgs` - for example `apiVersion` and `jvmTarget`.

Kotlin:
```
compilerOptions {
    apiVersion.set(KotlinVersion.KOTLIN_2_0)
    jvmTarget.set(JvmTarget.JVM_1_8)
}
```

Groovy:
```
compilerOptions {
    apiVersion = KotlinVersion.KOTLIN_2_0
    jvmTarget = JvmTarget.JVM_1_8
}
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -204,14 +269,23 @@ argument to this attribute or a list of arguments:
...
-compileKotlin.compilerOptions.freeCompilerArgs.add("-Xexport-kdoc")
-// Single additional argument, can be a key-value pair
-compileKotlin.compilerOptions.freeCompilerArgs.add("-Xno-param-assertions")
-// List of arguments
-compileKotlin.compilerOptions.freeCompilerArgs.addAll(listOf("-Xno-receiver-assertions", "-Xno-call-assertions"))
+        // List of arguments
+        freeCompilerArgs.addAll(
+            listOf(
+                "Xno-receiver-assertions",
+                "Xno-call-assertions"
```
Tapchicoma, line 284:
```suggestion
                "-Xno-receiver-assertions",
                "-Xno-call-assertions"
```

docs/topics/gradle/gradle-compiler-options.md
akiya-nagatsuka: @Tapchicoma I guess in kts files you can also use assign operator on lazy api at least in latest Gradle versions, isn't it?
```kotlin
compilerOptions { 
    apiVersion = KotlinVersion.KOTLIN_2_0
    jvmTarget = JvmTarget.JVM_1_8
}
```

Tapchicoma: Yes, it is possible to use, though only starting Gradle 8.2+ versions. As we are still supporting Gradle 7.6.3 we need to provide an example without assign operator.

docs/topics/gradle/gradle-compiler-options.md
AlejandraPedroza: Aren't we removing these code lines if we change the whole snippet to:

`compilerOptions { apiVersion.set(KotlinVersion.KOTLIN_2_0) jvmTarget.set(JvmTarget.JVM_1_8) }`

docs/topics/gradle/gradle-compiler-options.md
akiya-nagatsuka: I see, thank you for explaining your objection

Tapchicoma (approved):

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,51 +15,108 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
 
-Using a build script, you can specify additional compilation options. Use the `compilerOptions` property of a Kotlin compilation task for it. 
-For example:
+When you use the Kotlin Gradle plugin in your project, the Gradle DSL allows comprehensive configuration of compiler options. 
+The Gradle DSL is available for [Kotlin Multiplatform](multiplatform-dsl-reference.md) and [JVM/Android](#target-the-jvm) projects.
 
-<tabs group="build-script">
-<tab title="Kotlin" group-key="kotlin">
+With the Gradle DSL, you can configure compiler options within the build script at three levels: 
+* **Extension level,** as a top-level block. These compiler options are common for all targets and shared source sets.
```
zamulla, line 22: Strictly speaking, only `kotlin {}` is a top-level block here, so I would avoid this term to not have to say things like "top-level within something".

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,51 +15,108 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+* **Target level**, in the target block. These compiler options are for a specific target.
```
zamulla, line 23:
```suggestion
* **Target level**, in the block for a specific target.
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -15,51 +15,108 @@ in the [Working with command-line compiler](command-line.md) tutorial.
...
+* **Compilation unit level,** usually as a specific compilation task.
+
+![Kotlin compiler options levels](compiler-options-levels.svg){width=700}
+
+The settings at a higher level are used as a convention (default) for a lower level:
+
+* The values of extension compiler options are the default for target compiler options, including shared source sets
```
zamulla, line 30: I don't think we introduce the terms "extension compiler options" and "target compiler options" — to avoid confusion I would go the long way of "options set at extension level" and "options set at target level". This also would fix the phrasing of "...options, including source sets" which can be hard to decipher.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -68,24 +125,33 @@ kotlin {
...
 </tabs>
 
+## Target the JVM
+
+As explained before, you can define compiler options for your JVM/Android projects at the extension, target, and compilation unit levels.
+
+Default JVM compilation tasks are called `compileKotlin` for production code and `compileTestKotlin`
+for test code. The tasks for custom source sets are named according to their `compile<Name>Kotlin` patterns.
+
+Regarding Android compilation tasks, you can consult them by typing the `gradlew tasks --all` command in the terminal.
```
zamulla, line 145: The "consult" here might be making this sentence more complicated that it needs to be: we can probably say "You can see the list of Android compilation tasks by running ..."

Although I don't think I understand how Android-specific this command is. This is going to be all tasks, right, not just Android? Should we provide instructions for recognizing Android-specific tasks? Even if they are simply the ones with "Android" in the name :)

zamulla (changes requested):
Nice 👍 

Left some comments to help with clarity and consistency, hopefully.

zamulla (approved):
Thanks for the updates, LGTM!

PR #4537, opened 2024-11-08
feat: shortcut deprecation section

danil-pavlov:


docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -420,7 +420,7 @@ The access to `KotlinCompilation.source` has been deprecated. A code like this p
 kotlin {
     jvm()
     js()
-    ios()
+    iosArm64()
```
AYastrebov, line 423: Let's also add `iosSimulatorArm64()`

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -446,7 +446,7 @@ You can change the code above in one of the following ways:
...
```
AYastrebov, line 449: and here

AYastrebov (changes requested):

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -910,6 +912,41 @@ Here's the planned deprecation cycle:
...
+For example, the `ios()` shortcut created both the `iosArm64` and `iosX64` targets but didn't include the `iosSimulatorArm64`
+target, which is necessary when working on hosts with Apple M chips. However, changing this shortcut was hard to implement
+and could cause issues in existing user projects.
+
+**What's the best practice now?**
+
+The Kotlin Gradle plugin now provides a built-in hierarchy template. Since Kotlin 1.9.20, it's enabled by default
+and contains predefined intermediate source sets for popular use cases.
+
+The difference from shortcuts is that you need to explicitly specify the set of targets first, and the plugin then
```
zamulla, line 932:
```suggestion
Instead of shortcuts you should specify the list of targets, and then the plugin
```

"Explicitly specify" reads redundant to me here, but maybe it isn't?
Also, changed "set" to "list" so that it corresponds to "this list" later.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -910,6 +912,41 @@ Here's the planned deprecation cycle:
...
+automatically sets up intermediate source sets based on this list.
+
+For example, suppose you have `iosArm64` and `iosSimulatorArm64` targets in your project. In that case, the plugin
+automatically creates the `iosMain` and `iosTest` intermediate source sets, and for the `iosArm64` and `macosArm64`
+targets, it creates the `appleMain` and `appleTest` source sets.
```
zamulla, line 937: I don't think I understand where the `macosArm64` target comes from, if we didn't suppose that the project has it from the beginning. Is "`iosArm64` and `iosSimulatorArm64`" one example and "`iosArm64` and `macosArm64`" another?

zamulla (changes requested):
All good, but there is one question that I think is worth clearing up.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
danil-pavlov: Yes, another example to show that different intermediate source sets are created, I should probably split it into two senetences

docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -910,6 +912,41 @@ Here's the planned deprecation cycle:
...
+**What's the best practice now?**
+
+The Kotlin Gradle plugin now provides a built-in hierarchy template. Since Kotlin 1.9.20, it's enabled by default
+and contains predefined intermediate source sets for popular use cases.
+
+Instead of shortcuts, you should specify the list of targets, and then the plugin automatically sets up intermediate
+source sets based on this list.
+
+For example, if you have `iosArm64` and `iosSimulatorArm64` targets in your project, the plugin automatically creates
+the `iosMain` and `iosTest` intermediate source sets. For the `iosArm64` and `macosArm64` targets, the `appleMain` and
```
zamulla, line 936:
```suggestion
the `iosMain` and `iosTest` intermediate source sets. If you have `iosArm64` and `macosArm64` targets, the `appleMain` and
```

I would suggest mirroring the sentence structure here so that it's clearer that this is not a continuation but a separate hypothetical scenario.

zamulla (approved):
Thanks for clarifying. One more note to consider, otherwise LGTM.

PR #4566, opened 2024-11-20
feat: non-local jumps

danil-pavlov:


docs/topics/returns.md
```diff
@@ -39,32 +41,42 @@ loop@ for (i in 1..100) {
...
+### Non-local break and continue
 
-In Kotlin, functions can be nested using function literals, local functions, and object expressions.
-Qualified `return`s allow us to return from an outer function.
-The most important use case is returning from a lambda expression. Recall that when we write the following,
-the `return`-expression returns from the nearest enclosing function - `foo`:
+> This feature is currently [Experimental](components-stability.md#stability-levels-explained).
+> We're planning to stabilize it in future releases.
+> To opt in, use the `-Xnon-local-break-continue` compiler option.
+> We would appreciate your feedback on it in [YouTrack](http://kotl.in/issue).
```
nikitabobko, line 49: It's not clear to me how much sense it makes to point to the YouTrack new issue page in such kind of sentence. At least, it's slightly unexpected.

Maybe a better option would be to point to the issue itself https://youtrack.jetbrains.com/issue/KT-1436

docs/topics/inline-functions.md
```diff
@@ -122,9 +124,30 @@ inline fun f(crossinline body: () -> Unit) {
 }
 ```
 
-> `break` and `continue` are not yet available in inlined lambdas, but we are planning to support them, too.
+### Break and continue
+
+> This feature is currently [Experimental](components-stability.md#stability-levels-explained).
+> We're planning to stabilize it in future releases.
+> To opt in, use the `-Xnon-local-break-continue` compiler option.
+> We would appreciate your feedback on it in [YouTrack](http://kotl.in/issue).
```
nikitabobko, line 132: It's not clear to me how much sense it makes to point to the YouTrack new issue page in such kind of sentence. At least, it's slightly unexpected.

Maybe a better option would be to point to the issue itself https://youtrack.jetbrains.com/issue/KT-1436

docs/topics/returns.md
```diff
@@ -39,32 +41,42 @@ loop@ for (i in 1..100) {
...
+In some cases, you can apply `break` and `continue` *non-locally* without explicitly defining labels.
+Such non-local usages are valid in lambda expressions used in enclosing [inline functions](inline-functions.md):
 
 ```kotlin
-//sampleStart
-fun foo() {
-    listOf(1, 2, 3, 4, 5).forEach {
-        if (it == 3) return // non-local return directly to the caller of foo()
-        print(it)
+fun processList(elements: List&lt;Int&gt;): Boolean {
```
nikitabobko, line 57: The same question about `&lt;`

docs/topics/inline-functions.md
```diff
@@ -54,7 +54,9 @@ however, can be manipulated in any way you like, including being stored in field
 >
 {style="note"}
 
-## Non-local returns
+## Non-local jump expressions
```
nikitabobko, line 57: The documentation on non-local break/continue/return is duplicated in `inline-functions.md` ("Inline functions") and `returs.md` ("Returns and jumps")

I think, it'd be better to place the docs only to one of the pages. and another page could just have a link to the first one.

But if you think that duplication is not a problem, you can keep it "as is". It's up to you.

nikitabobko: To my mind, "non-local jumps" is more of a feature of `inline` functions. And the `returns.md` could be renamed to smth like `jump-labels.md` and be minimized to only talk about labels

docs/topics/inline-functions.md
```diff
@@ -122,9 +124,30 @@ inline fun f(crossinline body: () -> Unit) {
...
+> We would appreciate your feedback on it in [YouTrack](http://kotl.in/issue).
 >
-{style="note"}
+{style="warning"}
+
+Similar to non-local `return`, you can apply `break` and `continue` [jump expressions](returns.md) within lambdas passed
+as arguments to an inline function that encloses the loop:
+
+```kotlin
+fun processList(elements: List&lt;Int&gt;): Boolean {
```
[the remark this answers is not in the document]
danil-pavlov: No, it was necessary for the XML version of this block in What's new, here it's redundant, thank you for noticing!

docs/topics/inline-functions.md
danil-pavlov: In my mind, non-local feature was as important as the labels, so I created an equal structure for both of them in returns.md

But you're right, if it's just a specific feature of inline functions, we can minimize it in returns.md down to just a link to the main Inline functions page with detailed info and avoid duplication

nikitabobko (approved):
lgtm

docs/topics/inline-functions.md
```diff
@@ -122,9 +124,30 @@ inline fun f(crossinline body: () -> Unit) {
...
+> This feature is currently [Experimental](components-stability.md#stability-levels-explained).
+> We're planning to stabilize it in future releases.
+> To opt in, use the `-Xnon-local-break-continue` compiler option.
+> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-1436).
 >
-{style="note"}
+{style="warning"}
+
+Similar to non-local `return`, you can apply `break` and `continue` [jump expressions](returns.md) within lambdas passed
+as arguments to an inline function that encloses the loop:
```
sarahhaggarty, line 137:
```suggestion
Similar to non-local `return`, you can apply `break` and `continue` [jump expressions](returns.md) in lambdas passed
as arguments to an inline function that encloses a loop:
```

docs/topics/returns.md
```diff
@@ -39,32 +39,18 @@ loop@ for (i in 1..100) {
...
+>
+{style="tip"}
+
 ## Return to labels
 
 In Kotlin, functions can be nested using function literals, local functions, and object expressions.
-Qualified `return`s allow us to return from an outer function.
-The most important use case is returning from a lambda expression. Recall that when we write the following,
-the `return`-expression returns from the nearest enclosing function - `foo`:
+Qualified `return`s allow you to return from an outer function.
```
sarahhaggarty, line 50: The `return`s looks a bit weird in terms of formatting. WDYT about making it singular?
"A qualified `return` allows you to return from an outer function."

sarahhaggarty (approved):
Looks good! I left a couple of minor suggestions for you to consider.

PR #4569, opened 2024-11-25
feat: add multi-dollar string interpolation feature

koshachy:


docs/topics/strings.md
```diff
@@ -154,6 +155,76 @@ ${'$'}_9.99
 """
 ```
 
+> To avoid `${'$'}` sequences in multiline strings, you can use Experimental [multi-dollar string interpolation feature](#multi-dollar-string-interpolation).
```
serras, line 158: This works in both single-line and multi-line strings, not only in the latter.

docs/topics/strings.md
```diff
@@ -154,6 +155,76 @@ ${'$'}_9.99
...
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "$id": "https://example.com/product.schema.json",
+      "$dynamicAnchor": "meta"
+      "title": "$${simpleName ?: qualifiedName ?: "unknown"}",
+      "type": "object"
+    }
+    """
+```
+
+Here, the use of `$$` before the multiline string means that only two consecutive dollar signs trigger string interpolation,
```
serras, line 190: As mentioned above, multi-dollar is for any string, not only multiline ones.

docs/topics/strings.md
```diff
@@ -154,6 +155,76 @@ ${'$'}_9.99
...
+
+```kotlin
+val requestData =
+    $$$"""{
+      "currency": "$",
+      "enteredAmount": "42.45 $$",
+      "$$serviceField": "none",
+      "product": "$$$productName"
+    }
+    """
```
serras, line 204: I would show the result of this block of code, because maybe it's not clear that `$$serviceField` will stay with _two_ dollars in the string.

serras (approved):

docs/topics/strings.md
```diff
@@ -154,6 +155,90 @@ ${'$'}_9.99
...
+kotlin {
+    compilerOptions {
+        freeCompilerArgs.add("-Xmulti-dollar-interpolation")
+    }
+}
+```
+
+This feature doesn't affect existing code that uses single-dollar string interpolation. 
+You can continue using single `$`
+as before and apply multi-dollar signs when you need to handle literal dollar signs in strings.
```
sarahhaggarty, line 240:
```suggestion
You can continue using a single `$`
as before and apply multi-dollar signs when you need to handle literal dollar signs in strings.
```

docs/topics/strings.md
```diff
@@ -211,3 +296,13 @@ for example, in localization cases that depend on the user locale.
...
+
+
+---
+
+```kotlin
+// build.gradle.kts
+dependencies {
+    myCompilerScope("org.jetbrains.kotlin:kotlin-compiler-embeddable:%kotlinVersion")
+}
+```
```
sarahhaggarty, line 308: Is this meant to be here?

sarahhaggarty (approved):
Nice! I have a few suggestions. Please take a look!

docs/topics/strings.md
```diff
@@ -154,6 +155,90 @@ ${'$'}_9.99
...
+{style="note"}
+
+### Multi-dollar string interpolation
+
+> The multi-dollar string interpolation is [Experimental](https://kotlinlang.org/docs/components-stability.html#stability-levels-explained).
+> It may be changed at any time. We would appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-2425).
+>
+{style="warning"}
+
+Multi-dollar string interpolation allows you to specify how many consecutive dollar signs are required to trigger interpolation.
```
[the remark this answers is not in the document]
koshachy: I can, just afraid to make the para too long.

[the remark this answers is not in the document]
sarahhaggarty: True. You could break the paragraph after that point? 

```
Multi-dollar string interpolation allows you to specify how many consecutive dollar signs are required to trigger interpolation.
Interpolation is the process of embedding variables or expressions directly into a string.

While you can [escape literals](#escaped-strings) for single-line strings,
multiline strings in Kotlin don't support backslash escaping.
To include dollar signs (`$`) as literal characters, you must use the `${'$'}` construct to prevent string interpolation.
This approach can make code harder to read, especially when strings contain multiple dollar signs.
```

PR #4575, opened 2024-11-27
Update "Tips for improving Kotlin/Native compilation times"

SvyatoslavScherbina:
This PR adds a few more tips to the document, clarifies existing tips and removes an outdated part.

### DIFF
--- a/docs/topics/native/native-improving-compilation-time.md
+++ b/docs/topics/native/native-improving-compilation-time.md
@@ -67,7 +67,7 @@ Here are some recommendations for configuring Gradle for better compilation perf
   Release is highly optimized, and this takes a lot of time: compilation of release binaries takes an order of magnitude
   more time than debug binaries. In a typical development cycle, all the optimizations might be unnecessary. So, if
   your development cycle uses a task with "Release" in its name, consider replacing it with "Debug". Similarly, instead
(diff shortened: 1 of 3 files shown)


abdulowork (approved):

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -8,80 +10,150 @@ Read on for our tips on how to speed up the Kotlin/Native compilation process.
...
+For example, you don't need `iosX64` if you don't run this project on iOS simulators on Intel-based Macs.
+
+> Binaries for different targets are built with `linkDebug*$Target` and `linkRelease*$Target` Gradle tasks.
+> You can look for the executed tasks in the build log or in the
+> [Gradle build scan](https://docs.gradle.org/current/userguide/build_scans.html)
+> by running a Gradle build with the `--scan` option.
+>
+{style="tip"}
+
+### Don't build release binaries
```
SvyatoslavScherbina, line 84: I think the "unless necessary" part of that heading was useful. Without it, this might sound like a recommendation to avoid release binaries completely, which is confusing. Instead, the message should be to avoid them during development, when performance doesn't matter that much. Apps should be released with release binaries in any case.

danil-pavlov (approved):

PR #4586, opened 2024-11-29
Update multiplatform-expect-actual.md

ashraf-M-hassan:


danil-pavlov:
Hi, thank you for the report! I think there's confusion; let me please clarify. So, in the list of rules above, we state that expected declarations go into the common source set, while actual ones go into platform-specific source sets.

In this paragraph, we're stating that you can use intermediate source sets (like `iosMain`) for actual declarations instead of providing actual implementations in each of the target source sets (`iosX64Main`, `iosArm64Main`, and `iosSimulatorArm64Main`). And the Kotlin compiler then uses actual declarations from that intermediate source set for all the corresponding final targets.

PR #4597, opened 2024-12-08
Mention you still need Apple host to test iOS

hfhbd:


danil-pavlov:
Hello Philip, thank you for your PR! The current limitations of this feature need even more clarification, so I've added your changes to another PR, let's continue in #4600

PR #4599, opened 2024-12-09
Fix a bad reference in notes for jvm-spring-boot-using-crudrepository.md

mahozad:
This sentence did not make sense:

> The return type of the `findByIdOrNull()` function...

The *return type* of the function is not an extension function. It is <del>`String`</del> `Message`. The function itself is an extension function.


koshachy (approved):
Thanks!

PR #4600, opened 2024-12-09
update: cinterop limitation and test requirements

danil-pavlov:
This PR clarifies the current limitations of the klib cross-compilation:
* [KT-73574](https://youtrack.jetbrains.com/issue/KT-73574) cinterop dependencies
* #4597

### DIFF
--- a/docs/topics/multiplatform/multiplatform-publish-lib.md
+++ b/docs/topics/multiplatform/multiplatform-publish-lib.md
@@ -50,24 +50,30 @@ in the publication's scope.
 ## Host requirements
 
 Kotlin/Native supports cross-compilation, allowing any host to produce necessary `.klib` artifacts.
+However, there are still some specifics you should keep in mind.
+
+### Compilation for Apple targets
+<primary-label ref="experimental-opt-in"/>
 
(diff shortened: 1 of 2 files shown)


docs/topics/multiplatform/multiplatform-publish-lib.md
```diff
@@ -50,24 +50,30 @@ in the publication's scope.
...
 ```none
 kotlin.native.enableKlibsCrossCompilation=true
 ```
 
-> To build [final binaries](multiplatform-build-native-binaries.md) for Apple targets, you still need to use a Mac machine.
->
-{style="note"}
+Cross-compilation is currently Experimental and has some limitations. You still need to use a Mac machine if:
+
+* Your library has a [cinterop dependency](native-c-interop.md).
```
atyrin, line 67: I'm not sure if it will be obvious that CocoaPods integration is also affected since it is based on the cinterop.

atyrin (approved):

docs/topics/multiplatform/multiplatform-publish-lib.md
danil-pavlov: Thank you! added as a separate item

PR #4614, opened 2024-12-16
update: new ios deployment version

danil-pavlov:
This PR updates iOS deployment versions on CocoaPods integration pages ([KT-73860](https://youtrack.jetbrains.com/issue/KT-73860))


AYastrebov (approved):

zamulla (approved):
LGTM. But let's consider an example other than `FirebaseAuth`, which seems have its own problems, not representative of iOS dependencies as a whole.

PR #4627, opened 2024-12-30
Update opt-in-requirements documentation with new optIn ListProperty

NinoDLC:
Following [the new API recommendations](https://kotlinlang.org/docs/gradle-compiler-options.html#all-compiler-options), we should use `compilerOptions.optIn.add()` instead of `compilerOptions.freeCompilerArgs.add()`.


sarahhaggarty:
Thanks for your contribution @NinoDLC !

PR #4631, opened 2025-01-03
Update 2.1.0 internal compiler APIs doc

martinbonnin:
Follow up from this slack thread: https://kotlinlang.slack.com/archives/C19FD9681/p1735830883666549

### DIFF
--- a/docs/topics/whatsnew21.md
+++ b/docs/topics/whatsnew21.md
@@ -1254,13 +1254,13 @@ making internal compiler symbols available in the build script classpath.
 
 These symbols were intended for internal use only.
 
-Starting with Kotlin 2.1.0, KGP bundles a subset of `org.jetbrains.kotlin:kotlin-compiler-embeddable` class files in its jar file and will progressively remove them. This will prevent compatibility issues and simplify KGP maintenance.
+Starting with Kotlin 2.1.0, KGP bundles a subset of `org.jetbrains.kotlin:kotlin-compiler-embeddable` class files in its jar file and will progressively remove them. This is to prevent compatibility issues and simplify KGP maintenance.
 
-If other parts of your build logic (e.g. plugins such as kotlinter or others) depend on a different version of `org.jetbrains.kotlin:kotlin-compiler-embeddable` than theo one bundle with KGP, there is a high chance of conflict and runtime exception. 
+If other parts of your build logic (e.g. plugins such as kotlinter or others) depend on a different version of `org.jetbrains.kotlin:kotlin-compiler-embeddable` than the one bundle with KGP, there is a high chance of conflict and runtime exception. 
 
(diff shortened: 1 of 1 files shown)


docs/topics/whatsnew21.md
```diff
@@ -1249,16 +1249,21 @@ You can use the `KotlinAndroidExtension` in exactly the same way.
 
 ### Compiler symbols hidden from the Kotlin Gradle plugin API
 
-Starting with Kotlin 2.1.0,
-you will receive a warning if you access compiler module symbols bundled within the Kotlin Gradle plugin (KGP).
```
martinbonnin, line 1253: I think this was technically not exact, right? The warning is based on the presence of a class file that is not bundled in KGP but build logic could still use the symbols bundled with KGP without being detected?

docs/topics/whatsnew21.md
```diff
@@ -1249,16 +1249,21 @@ You can use the `KotlinAndroidExtension` in exactly the same way.
...
-with classloader or process isolation to ensure safe interaction with the compiler.
+These symbols were intended for internal use only.
+
+Starting with Kotlin 2.1.0, KGP bundles a subset of `org.jetbrains.kotlin:kotlin-compiler-embeddable` class files in its jar file and will progressively remove them. This is to prevent compatibility issues and simplify KGP maintenance.
+
+If other parts of your build logic (e.g. plugins such as kotlinter or others) depend on a different version of `org.jetbrains.kotlin:kotlin-compiler-embeddable` than the one bundled with KGP, there is a high chance of conflict and runtime exception. 
+
+To guard against those issues, KGP will now display a warning if `org.jetbrains.kotlin:kotlin-compiler-embeddable` is loaded in the build classpath alongside KGP.
+
+As a long term fix, if you are a plugin author using `org.jetbrains.kotlin:kotlin-compiler-embeddable` classes, we recommend doing so in an isolated classloader.
```
martinbonnin, line 1263: Changed from a requirement to a recommendation. If KGP removes any dependency on compiler APIs and/or the compiler APIs become stable, it might become acceptable to do the work in the main build classpath?

Also Gradle Workers API is a possible solution but it comes with tradeoffs so it's leaving the door open to other solutions.

ALikhachev (approved):

docs/topics/whatsnew21.md
ALikhachev: It might become acceptable but is still subject to conflicts between different plugins that depend on `kotlin-compiler-embeddable`.

koshachy (approved):
Thanks for your contribution!
I've slightly changed the text but kept the same meaning and structure.
Feel free to ask any questions.

martinbonnin:
Works for me! Thanks for the follow up!

PR #4632, opened 2025-01-03
Replace INFO with IGNORE as valid JVM target validation options

khensunny:


koshachy (approved):
Nice catch, thanks!

FTR: https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.dsl.jvm/-jvm-target-validation-mode/

PR #4637, opened 2025-01-07
Update multiplatform-discover-project.md

alajemba-vik:
This PR corrects the terminology used when referring to source set naming conventions. The current text mentions “prefixes” when it should say “suffixes,” as terms like commonMain and iosArm64Main have “Main” at the end, making “suffix” the correct term.


danil-pavlov (approved):
True, thank you for creating the PR!

PR #4658, opened 2025-01-16
Improved Variables section: clarified val initialization

MaddyRizvi:
## This pull request introduces the following changes:

- **Enhanced explanation**:  
  The section has been rewritten to clarify how you can use delayed initialization for `val` variables, especially when their value depends on runtime logic. It now emphasizes the need to specify the data type when the initialization is deferred.

- **Runtime logic clarification**:  
  The explanation now highlights that you can declare a `val` without initializing it immediately. The value can be assigned later based on runtime logic, with the type explicitly defined in such cases.

- **Simplified examples**:  
  The examples have been updated to show how delayed initialization works in practice, offering clearer guidance for developers. They now demonstrate the need to specify the type when a `val` is initialized after declaration.

These changes aim to enhance the documentation by making it more accessible and understandable for developers, helping them easily grasp how to implement delayed initialization for `val` variables.

### DIFF
--- a/docs/topics/basic-syntax.md
+++ b/docs/topics/basic-syntax.md
@@ -202,24 +202,29 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-basic-syntax-inference"}
 
-You can use variables only after initializing them. You can either initialize a variable at the moment of declaration or declare a variable first and initialize it later. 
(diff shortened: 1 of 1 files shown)


koshachy:
#4659 #4656

sarahhaggarty:
I'll close this one as it duplicates #4659.

PR #4659, opened 2025-01-16
Fix: Clarify Range Expression by Adding Parentheses for Clarity

MaddyRizvi:
## Pull Request Explanation

- **Clarified Range Expression**:  
  Parentheses have been added around `(y + 1)` in the range expression `1..(y + 1)` to improve clarity. This ensures that the precedence of the `+` operator in the expression is clear and avoids any confusion regarding the evaluation order.

- **Improved Readability for Beginners**:  
  The added parentheses help beginners better understand the order of operations within the range expression. This change makes it easier to follow the logic and ensures that the range is correctly interpreted as intended.

- **Updated Example for Better Understanding**:  
  The code example has been modified to make the behavior of the range operation more understandable, particularly in how the upper bound is calculated. This update will help developers easily grasp how ranges work in Kotlin, with a focus on operator precedence.

### Purpose of Changes:
The goal of this change is to enhance the clarity of the example and ensure that developers—especially beginners—understand how the range expression works. By adding parentheses, the expression is now more explicit, and it eliminates any ambiguity about the order in which the `+` operation is applied.

These updates aim to improve the overall documentation by providing clearer and more accessible examples for developers, making it easier to understand key concepts in Kotlin.

### DIFF
--- a/docs/topics/basic-syntax.md
+++ b/docs/topics/basic-syntax.md
@@ -453,7 +453,7 @@ fun main() {
 //sampleStart
     val x = 10
     val y = 9
-    if (x in 1..y+1) {
+    if (x in 1..(y + 1)) {
         println("fits in range")
     }
 //sampleEnd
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @MaddyRizvi 
Thanks for your suggestion. Unfortunately what you've suggested doesn't follow our coding conventions so I'm not able to accept this change. So for now, I'll close this PR but let us know if you have any other suggestions in future. Thanks!

PR #4666, opened 2025-01-18
Update delegated-properties.md

NoPalVol:
**Minor improvement: Using 'ref' for delegate receiver**

This pull request suggests a small change to use `ref` instead of `thisRef` in the `Delegate` class documentation on the Kotlinlang.org website.

Using `ref` aligns better with common Kotlin practices for referencing delegate receivers and improves code readability.

### DIFF
--- a/docs/topics/delegated-properties.md
+++ b/docs/topics/delegated-properties.md
@@ -25,12 +25,12 @@ For example:
 import kotlin.reflect.KProperty
 
 class Delegate {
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @NoPalVol 
Thanks for your suggestion! I checked with our developers and they're not aware of such a practice. In our documentation and specification we use `thisRef` so we decided to stick with this for now. But thanks again and I hope you'll consider contributing again in future!

PR #4671, opened 2025-01-21
update: add the overview section for GSoC with Kotlin

koshachy:
[KT-74464](https://youtrack.jetbrains.com/issue/KT-74464)


sarahhaggarty (approved):
LGTM! 🚀

PR #4675, opened 2025-01-23
Create documentation for the Kotlin Metadata JVM library

daniCsorbaJB:
This PR is created to add new documentation for the Kotlin Metadata JVM library. 
Related issue: [KT-66421](https://youtrack.jetbrains.com/issue/KT-66421/DocsJVM-Write-a-page-for-kotlinlang.org-with-a-description-of-kotlin-metadata-jvm-library)


docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+    } catch (e: ClassNotFoundException) {
+        println("Class not found: $className")
+    } catch (e: Exception) {
+        println("Error processing metadata: ${e.message}")
+        e.printStackTrace()
+    }
+}
+```
+
+### Validate binary compatibility with the Kotlin Metadata JVM library
```
sandwwraith, line 160: This whole section feels a little bit weird:

1. Code consists mainly of pieces that are related to reading binary class files manually. While it is a good example, it showcases the usages of the ASM library, not the kotlin-metadata one.
2. The only related kotlin-metadata example is printing visibility, which was already shown directly above in a much simpler way.
3. General phrasing 'validating with tools like BCV' feels wrong. Firstly, BCV does not need this operations to be performed — it encapsulates them and BCV user doesn't have to even know about kotlin-metadata. The BCV input is classfiles, not a list of visibilities. Secondly, (despite not very appropriate name), BCV does not validate anything, it simply compares lists of visibilities. 

I would rather say that this technique can be used to write a tool _similar_ to BCV, but definitely not required to use the BCV itself. What I suggest here:

* Move the example part with ASM manipulation to a separate .md file, mention it as one of the possible variants to obtain a `kotlin.Metadata` instance.
* Merge examples above and below so the resulting example will print visibilities for all functions in both KmClass and KmPackage
* Give a link to BCV to illustrate that using this technique, it is possible to build tools like it.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+    } else {
+        println("No Kotlin metadata found or metadata is not of type Class.")
+    }
+}
+```
+
+## Modify metadata
+
+You can use the `kotlinx-metadata-jvm` library to modify Kotlin class metadata, allowing you to adjust visibility attributes,
+remove specific declarations, or prepare metadata for binary compatibility checks.
```
sandwwraith, line 268: ` prepare metadata for binary compatibility checks` — also seems not related, see my explanation about BCV

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+
```
sandwwraith, line 269: I would also emphasize that modifying metadata is _essential_ if you modify the bytecode in certain ways.
Given that you already have the example of removing private functions, you can write something like this:

"The main goal of modifying metadata is that it should be consistent with existing bytecode. For example, if you are writing a tool like ProGuard, which shrinks the code and may delete some declarations, you have to delete the same declarations from the metadata as well. Let us imagine that we have a JVM tool that deletes private methods from .class files. For that tool to work with Kotlin, it would also have to delete private methods from metadata, which can be done the following way:"

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+            println("No Kotlin Metadata found for class: $className")
+        }
+    } catch (e: ClassNotFoundException) {
+        println("Class not found: $className")
+    } catch (e: Exception) {
+        println("Error processing metadata: ${e.message}")
+        e.printStackTrace()
+    }
+}
+```
```
sandwwraith, line 322: You can also mention that for `readStrict/write` pair there is a convenient `transform` shortcut: https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/transform.html

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+        println("This is not a Kotlin class.")
+    }
+}
+```
+
+After retrieving the `@Metadata` annotation, you can use the [`readLenient()`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/read-lenient.html) and [`readStrict()`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/read-strict.html) functions from the [`KotlinClassMetadata`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/) API to parse it.
+These functions allow you to extract detailed information about classes or files, while addressing different compatibility requirements:
+
+* `readLenient()`: Use this function to read metadata, including metadata generated by newer Kotlin compiler versions. However, it doesn't support modifying or writing metadata.
+* `readStrict()`: Use this function when you need to modify and write metadata. The `readStrict()` function only works with metadata generated by Kotlin compiler versions fully supported by your project.
```
sandwwraith, line 94: I'd also add a link somewhere to the 'Detailed explanation' section of the README in case someone would like to know more about future versions handling

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+}
+```
+
+> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt).
+> 
+{style="tip"}
+
+## What's next
+
+* [See the API reference for the Kotlin Metadata JVM library](https://kotlinlang.org/api/kotlinx-metadata-jvm/).
```
sandwwraith, line 398: Maybe mention 'Module metadata' section of the readme here? It seems lost for now.

sandwwraith (changes requested):

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
+[//]: # (title: Kotlin Metadata JVM library)
+
+The [`kotlinx-metadata-jvm`](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm) library provides tools to read, modify, and generate metadata from Kotlin classes compiled for the JVM.
+This metadata, stored in the [`@Metadata`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-metadata/) annotation within `.class` files,
+is used by libraries and tools to inspect Kotlin-specific constructs such as properties, functions,
```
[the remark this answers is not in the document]
daniCsorbaJB: sounds like a good idea I like it:

How about modifying it like:
This metadata, stored in the [`@Metadata`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-metadata/) annotation within `.class` files,
is used by libraries and tools, such as [`kotlin-reflect`](reflection.md), to inspect Kotlin-specific constructs such as properties, functions,
and classes at runtime.

> The `kotlin-reflect` library relies on metadata to retrieve Kotlin-specific class details at runtime.
> Any inconsistencies between the metadata and the actual `.class` file may lead to incorrect behavior when using reflection.
> 
{style="note"}

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+and classes at runtime.
+You can also use this library to adjust attributes for binary compatibility validation or to generate and embed metadata into `.class` files.
```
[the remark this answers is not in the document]
daniCsorbaJB: nice catch thank you, I originally had something about ASM here but without it doesn't work. 😟 

How about:
You can also use the Kotlin Metadata JVM library to read and analyze metadata for binary compatibility checks or to generate and embed metadata into `.class` files.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Add the library to your project
+
+To include the Kotlin Metadata JVM library in your project, add the appropriate dependency configuration based on your build tool.
```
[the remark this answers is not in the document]
daniCsorbaJB: Good idea, added this line:
> The Kotlin Metadata JVM library follows the same versioning as the Kotlin compiler and standard library. Ensure that the version you use matches your project's Kotlin version.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,399 @@
...
+These functions allow you to extract detailed information about classes or files, while addressing different compatibility requirements:
+
+* `readLenient()`: Use this function to read metadata, including metadata generated by newer Kotlin compiler versions. However, it doesn't support modifying or writing metadata.
+* `readStrict()`: Use this function when you need to modify and write metadata. The `readStrict()` function only works with metadata generated by Kotlin compiler versions fully supported by your project.
+
+    > The `readStrict()` function supports metadata formats up to one version beyond `JvmMetadataVersion.LATEST_STABLE_SUPPORTED`. It throws an error when encountering metadata from newer Kotlin compiler versions to prevent mishandling unknown formats.
```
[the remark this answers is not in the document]
daniCsorbaJB: I see - I'm not sure if I understand correctly, so please correct my if I'm wrong - does this accurately reflect the behavior then:

    > The `readStrict()` function supports metadata formats up to one version beyond `JvmMetadataVersion.LATEST_STABLE_SUPPORTED`, which corresponds to the latest Kotlin version used in the project.
    > For example, if your project depends on `kotlin-metadata-jvm:2.1.0`, `readStrict()` can process metadata up to Kotlin `2.2.x`; otherwise, it throws an error to prevent mishandling unknown formats.
    > 
    > For more information, see [`LATEST_STABLE_SUPPORTED`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-jvm-metadata-version/-companion/-l-a-t-e-s-t_-s-t-a-b-l-e_-s-u-p-p-o-r-t-e-d.html). 
    >
    {style="note"}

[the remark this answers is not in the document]
daniCsorbaJB: I see - I'm not sure if I understand correctly, so please correct my if I'm wrong - does this accurately reflect the behavior then:

    > The `readStrict()` function supports metadata formats up to one version beyond `JvmMetadataVersion.LATEST_STABLE_SUPPORTED`, which corresponds to the latest Kotlin version used in the project.
    > For example, if your project depends on `kotlin-metadata-jvm:2.1.0`, `readStrict()` can process metadata up to Kotlin `2.2.x`; otherwise, it throws an error to prevent mishandling unknown formats.
    > 
    > For more information, see [`LATEST_STABLE_SUPPORTED`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-jvm-metadata-version/-companion/-l-a-t-e-s-t_-s-t-a-b-l-e_-s-u-p-p-o-r-t-e-d.html). 
    >
    {style="note"}

docs/topics/metadata-jvm.md
daniCsorbaJB: Nice one - thank you! - I added it as a tip

daniCsorbaJB: Nice one - thank you! - I added it as a tip

docs/topics/metadata-jvm.md
daniCsorbaJB: I added it to the note below 👍

docs/topics/metadata-jvm.md
daniCsorbaJB: Great suggestion, thank you! 💯 

I restructured the section a little

docs/topics/metadata-jvm.md
daniCsorbaJB: added it like so - what do you think: 

## What's next

* [See the API reference for the Kotlin Metadata JVM library](https://kotlinlang.org/api/kotlinx-metadata-jvm/).
* [Check out the Kotlin Metadata JVM GitHub repostiory](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm).
* [Learn about module metadata and working with `.kotlin_module` files.](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#module-metadata)

docs/topics/metadata-jvm.md
daniCsorbaJB: > You can use the [`transform()`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/transform.html) function instead of separately calling `readStrict()` and `write()` separately.
> This function parses metadata, applies transformations via a lambda, and writes the modified metadata automatically.
> 
{style="tip"}

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+    }.toByteArray()
+
+    // Writes the generated class file to disk
+    java.io.File("Hello.class").writeBytes(classBytes)
+
+    println("Metadata and .class file created successfully.")
+}
+```
+
+> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt).
```
sandwwraith, line 409:
```suggestion
> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/50331fb1496378c82c862db04af597e4198ec645/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt#L43).
```

I think it's better to point link to a specific fixed commit and a specific line with the very same example.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+        }
+    }
+
+    // Serializes metadata into KotlinClassMetadata
+    val annotationData = KotlinClassMetadata.Class(
+        klass, JvmMetadataVersion.LATEST_STABLE_SUPPORTED, 0
+    ).write()
+
+    // Generates a .class file with ASM
+    val classBytes = ClassWriter(0).apply {
```
sandwwraith, line 384: This example produces inconsistent metadata: you've added `<init>` and `hello` functions to KmClass, yet generated bytecode does not contain them. I suggest you remove `hello` KmFunction from example (to keep it small) and add `visitMethod(Opcodes.ACC_PUBLIC, "<init>", "()V", null, null).run { ... }` to generate proper constructor.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+   </dependencies>
+   ...
+</project>
+```
+
+
+## Read and parse metadata
+
+The `kotlin-metadata-jvm` library enables you to inspect Kotlin class metadata from compiled `.class` files.
+This allows you to extract structured information about class or file-level declarations and use it with validation tools like the [Binary Compatibility Validator (BCV)](https://github.com/Kotlin/binary-compatibility-validator).
```
sandwwraith, line 79: I'm afraid this sentence may create an impression that you have to use kotlin-metadata-jvm and extract metadata manually if you want to make use of BCV. This is not true. As I said, BCV already uses metadata under the hood, so regular BCV users don't even have to know about the existence of this library.

This documentation page is rather aimed at skilled people who want to create their own tools with functionality _similar to_ BCV. So you can rewrite this roughly as:

```
This allows you to extract structured information about class or file-level declarations, such as names, visibility, etc. 
With this information, you can write various tools to inspect compiled Kotlin declarations. 
For example, [Binary Compatibility Validator (BCV)](https://github.com/Kotlin/binary-compatibility-validator) tool uses kotlin-metadata-jvm to print out public declarations.
```

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+        }
+    } else {
+        println("No Kotlin metadata found or metadata is not of type Class.")
+    }
+}
+```
+
+## Modify metadata
+
+When using tools like [ProGuard](https://github.com/Guardsquare/proguard) to shrink and optimize bytecode, some declarations may be removed from `.class` files.
```
sandwwraith, line 278: Again, this may create impression that it is impossible to use proguard without using and mastering kotlin-metadata-jvm. It is not true — ProGuard _already_ uses this library and edits metadata correctly. 

However, if you have a _custom_ tool similar to proguard, then you have to manipulate metadata accordingly.

sandwwraith (changes requested):

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+
+While you can retrieve metadata using reflection, you can use another approach to extract it from bytecode using a bytecode manipulation framework such as [ASM](https://asm.ow2.io/).
+
+You can do this by following these steps:
+
+1. Read the bytecode of a `.class` file using the ASM library's `ClassReader` class.
+   This class processes the compiled file and populates a `ClassNode` object, which represents the class structure.
+2. Extract the `@Metadata`  from the `ClassNode` object using the `findAnnotation()` function.
+3. Parse the extracted metadata using the `readLenient()` function.
+4. Inspect the parsed metadata with the `kotlinMetadata` property.
```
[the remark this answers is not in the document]
daniCsorbaJB: indeed - I corrected it

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Extract metadata from bytecode
+
+While you can retrieve metadata using reflection, you can use another approach to extract it from bytecode using a bytecode manipulation framework such as [ASM](https://asm.ow2.io/).
+
+You can do this by following these steps:
+
+1. Read the bytecode of a `.class` file using the ASM library's `ClassReader` class.
+   This class processes the compiled file and populates a `ClassNode` object, which represents the class structure.
+2. Extract the `@Metadata`  from the `ClassNode` object using the `findAnnotation()` function.
```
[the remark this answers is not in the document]
daniCsorbaJB: I see what you mean. I agree it feels confusing like this. 

Instead of adding the sentence in parenthesis - I thought of adding this code comment alongside this:
Extract the `@Metadata`  from the `ClassNode` object. In the code example below, a custom extension function `findAnnotation()` is used for this.

and in the code:

```
// Defines a custom extension function to locate an annotation by its name in a ClassNode
fun ClassNode.findAnnotation(annotationName: String, includeInvisible: Boolean = false): AnnotationNode? {
    val visible = visibleAnnotations?.firstOrNull { it.refersToName(annotationName) }
    if (!includeInvisible) return visible
    return visible ?: invisibleAnnotations?.firstOrNull { it.refersToName(annotationName) }
}

```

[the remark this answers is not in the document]
sandwwraith: Yeah, great idea!

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+
+        // Prints class details
+        println("Class name: ${kmClass.name}")
+        println("Functions:")
+        kmClass.functions.forEach { function ->
+            println("- ${function.name}, Visibility: ${function.visibility}")
+        }
+
+        // Extracts and prints additional metadata details
+        val annotation = classNode.findAnnotation("kotlin/Metadata")
```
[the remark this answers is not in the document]
daniCsorbaJB: yea it's not crucial for this example I don't think - I removed it

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+## Create metadata from scratch
+
+To create metadata for a Kotlin class file from scratch using the Kotlin Metadata JVM library:
+
+1. Create an instance of `KmClass`, `KmPackage`, or `KmLambda`, depending on the type of metadata you want to generate.
+2. Populate the instance with attributes such as the class name, visibility, constructors, and function signatures. Kotlin’s `apply()` function can reduce boilerplate code while setting properties.
+3. Use the populated instance to create a `KotlinClassMetadata` object, which can generate a `@Metadata` annotation.
+4. Use the `ClassWriter` class from [ASM](https://asm.ow2.io/) to embed the metadata into a `.class` file.
+5. Specify the metadata version (for example, `JvmMetadataVersion.LATEST_STABLE_SUPPORTED`) and flags (0 for no flags, or copy flags from existing files if necessary).
+6. Access the annotation fields, such as `kind`, `data1` and `data2` to embed the metadata into a `.class` file or process it further.
```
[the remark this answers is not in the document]
daniCsorbaJB: I merged the two steps together like:
5. Use the `ClassWriter` class from [ASM](https://asm.ow2.io/) to embed metadata fields, such as `kind`, `data1` and `data2` into a `.class` file.

docs/topics/metadata-jvm.md
daniCsorbaJB: ok removed this part:
```
functions += KmFunction("hello").apply {
    visibility = Visibility.PUBLIC
    returnType = KmType().apply {
        classifier = KmClassifier.Class("kotlin/String")
    }
    signature = JvmMethodSignature("hello", "()Ljava/lang/String;")
}
```

added a constructor for `visitMethod()`:

```
visitMethod(Opcodes.ACC_PUBLIC, "<init>", "()V", null, null).apply {
    visitCode()
    visitVarInsn(Opcodes.ALOAD, 0)
    visitMethodInsn(Opcodes.INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false)
    visitInsn(Opcodes.RETURN)
    visitMaxs(1, 1)
    visitEnd()
}
```

So combined, it looks like this right now - please let me know what you think:

```
// Imports the necessary libraries
import kotlin.metadata.*
import kotlin.metadata.jvm.*
import org.objectweb.asm.*

fun main() {
    // Creates a KmClass instance
    val klass = KmClass().apply {
        name = "Hello"
        visibility = Visibility.PUBLIC
        constructors += KmConstructor().apply {
            visibility = Visibility.PUBLIC
            signature = JvmMethodSignature("<init>", "()V")
        }
    }

    // Serializes metadata into KotlinClassMetadata
    val annotationData = KotlinClassMetadata.Class(
        klass, JvmMetadataVersion.LATEST_STABLE_SUPPORTED, 0
    ).write()

    // Generates a .class file with ASM
    val classBytes = ClassWriter(0).apply {
        visit(Opcodes.V1_6, Opcodes.ACC_PUBLIC, "Hello", null, "java/lang/Object", null)
        // Generates a default constructor
        visitMethod(Opcodes.ACC_PUBLIC, "<init>", "()V", null, null).apply {
            visitCode()
            visitVarInsn(Opcodes.ALOAD, 0)
            visitMethodInsn(Opcodes.INVOKESPECIAL, "java/lang/Object", "<init>", "()V", false)
            visitInsn(Opcodes.RETURN)
            visitMaxs(1, 1)
            visitEnd()
        }

        visitAnnotation("Lkotlin/Metadata;", true).apply {
            visit("mv", annotationData.metadataVersion)
            visit("k", annotationData.kind)
            visitArray("d1").apply {
                annotationData.data1.forEach { visit(null, it) }
                visitEnd()
            }
            visitArray("d2").apply {
                annotationData.data2.forEach { visit(null, it) }
                visitEnd()
            }
            visitEnd()
        }
        visitEnd()
    }.toByteArray()

    // Writes the generated class file to disk
    java.io.File("Hello.class").writeBytes(classBytes)

    println("Metadata and .class file created successfully.")
}
```

docs/topics/metadata-jvm.md
daniCsorbaJB: hmm 🤔 - I see what you mean- I like your suggestion - I reworded it a bit:

> You can use the `kotlin-metadata-jvm` library to extract structured information from compiled Kotlin `.class` files, such as class names, visibility, and signatures.
This makes it useful in projects that need to analyze compiled Kotlin code.
For example, the [Binary Compatibility Validator (BCV)](https://github.com/Kotlin/binary-compatibility-validator) relies on `kotlin-metadata-jvm` to print public API declarations.

docs/topics/metadata-jvm.md
daniCsorbaJB: 👍  Modified the paragraph to reflect this:

## Modify metadata

When using tools like [ProGuard](https://github.com/Guardsquare/proguard) to shrink and optimize bytecode, some declarations may be removed from `.class` files.
ProGuard automatically updates metadata to keep it consistent with the modified bytecode.

However, if you're developing a custom tool that modifies Kotlin bytecode in a similar way, you need to ensure that metadata is adjusted accordingly.
You can do this using the `kotlin-metadata-jvm` library, which lets you update declarations, adjust attributes, and remove specific elements.

docs/topics/metadata-jvm.md
sandwwraith: Great, I like it

docs/topics/metadata-jvm.md
sandwwraith: Yes, this looks more correct. Also don't forget `// Use the created KotlinMetadata instance to write @kotlin.Metadata annotation on the class file` comment for `visitAnnotation("Lkotlin/Metadata;", true).apply {` line

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,416 @@
...
+This metadata, stored in the [`@Metadata`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-metadata/) annotation within `.class` files,
+is used by libraries and tools such as [`kotlin-reflect`](reflection.md) to inspect Kotlin-specific constructs such as properties, functions,
+and classes at runtime.
+
+> The `kotlin-reflect` library relies on metadata to retrieve Kotlin-specific class details at runtime.
+> Any inconsistencies between the metadata and the actual `.class` file may lead to incorrect behavior when using reflection.
+> 
+{style="warning"}
+
+You can also use the Kotlin Metadata JVM library to inspect attributes for binary compatibility validation or to generate and embed metadata into `.class` files.
```
sandwwraith, line 13:
```suggestion
You can also use the Kotlin Metadata JVM library to inspect various declaration attributes (such as visibility or modality) or to generate and embed metadata into `.class` files.
```

daniCsorbaJB: I like this version updated the docs - I only have a small comment would it work without the parenthesis? 
You can also use the Kotlin Metadata JVM library to inspect various declaration attributes such as visibility or modality, or to generate and embed metadata into `.class` files.

docs/topics/metadata-jvm.md
daniCsorbaJB: This works great!

What do you think about this  small modification about specifying the JvmMetadataVersion and the flags:
    // Serializes a KotlinClassMetadata.Class instance, including the version and flags, into a @kotlin.Metadata annotation

sandwwraith: yeah, great addition 👍🏻

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+
+* `readLenient()`: Use this function to read metadata, including metadata generated by newer Kotlin compiler versions. This function doesn't support modifying or writing metadata.
+* `readStrict()`: Use this function when you need to modify and write metadata. The `readStrict()` function only works with metadata generated by Kotlin compiler versions fully supported by your project.
+
+    > The `readStrict()` function supports metadata formats up to one version beyond [`JvmMetadataVersion.LATEST_STABLE_SUPPORTED`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-jvm-metadata-version/-companion/-l-a-t-e-s-t_-s-t-a-b-l-e_-s-u-p-p-o-r-t-e-d.html), which corresponds to the latest Kotlin version used in the project.
+    > For example, if your project depends on `kotlin-metadata-jvm:2.1.0`, `readStrict()` can process metadata up to Kotlin `2.2.x`; otherwise, it throws an error to prevent mishandling unknown formats.
+    > 
+    > For more information, see the [Kotlin MetaData GitHub repository](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#detailed-explanation).
```
sandwwraith, line 110:
```suggestion
    > For more information, see the [Kotlin Metadata GitHub repository](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#detailed-explanation). 
```

It's Metadata everywhere else, and I think this is the more correct spelling.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,417 @@
...
+            if (metadata is KotlinClassMetadata.Class) {
+                val kmClass = metadata.kmClass
+
+                // Removes private functions from the class metadata
+                kmClass.functions.removeIf { it.visibility == Visibility.PRIVATE }
+                println("Removed private functions. Remaining functions: ${kmClass.functions.map { it.name }}")
+
+                // Writes the modified metadata back
+                val newMetadata = metadata.write()
+                println("Modified metadata: ${newMetadata}")
```
sandwwraith, line 315: This can make an impression that calling `.write()` is enough, but in reality, you still have to write a new metadata to the classfile. Maybe add an additional commentary?

```kotlin
// Serializes the modified metadata back
val newMetadata = metadata.write()
// Now you have to write serialized metadata into the classfile with, for example, ASM. 
// See the example below to learn how to use kotlin-metadata-jvm with ASM.
println("Modified metadata: ${newMetadata}")
```

sandwwraith (changes requested):

docs/topics/metadata-jvm.md
daniCsorbaJB: hmm 🤔 
How about 
```
                // After modifying the metadata, you need to write it into the class file
                // This example uses ASM with kotlin-metadata-jvm to do this
                val newMetadata = metadata.write()
```

docs/topics/metadata-jvm.md
daniCsorbaJB: I agree- updated it

docs/topics/metadata-jvm.md
sandwwraith: The problem is that "this example" does not use ASM

sandwwraith: Only the next one does

daniCsorbaJB: ah my bad - you are right... 
Hmm 🤔 - my issue with pointing to another example within a code example that it's really easy to get confused without a link, so how about we go with something like:

// After modifying the metadata, you need to write it into the class file
// You can use a bytecode manipulation framework such as ASM to do this

sandwwraith: OK!

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+    }.toByteArray()
+
+    // Writes the generated class file to disk
+    java.io.File("Hello.class").writeBytes(classBytes)
+
+    println("Metadata and .class file created successfully.")
+}
+```
+
+> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/50331fb1496378c82c862db04af597e4198ec645/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt#L43)).
```
sandwwraith, line 412:
```suggestion
> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/50331fb1496378c82c862db04af597e4198ec645/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt#L43).
```

Excessive brace

sandwwraith (approved):
I don't have objections anymore 😄 Great work!

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+}
+```
+
+## Modify metadata
+
+When using tools like [ProGuard](https://github.com/Guardsquare/proguard) to shrink and optimize bytecode, some declarations may be removed from `.class` files.
+ProGuard automatically updates metadata to keep it consistent with the modified bytecode.
+
+However, if you're developing a custom tool that modifies Kotlin bytecode in a similar way, you need to ensure that metadata is adjusted accordingly.
+You can do this using the `kotlin-metadata-jvm` library, which lets you update declarations, adjust attributes, and remove specific elements.
```
koshachy, line 275: Again, **usually** we don't use "you can do this", "to do this"
If there is a way to use a library as a subject — it will be easier to read.

Not sure that my versions are better, so up to you:

The kotlin-metadata-jvm library lets you update declarations, change attributes, and remove elements.

With the kotlin-metadata-jvm library, you can update declarations, adjust attributes, and remove specific elements to adjust/change/update/ metadata.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+
+For example, if you use a JVM tool that deletes private methods from Java class files, you must also delete private functions from Kotlin metadata to maintain consistency.
+To do this:
```
koshachy, line 278:
```suggestion
For example, if you use a JVM tool that deletes private methods from Java class files, you must also delete private functions from Kotlin metadata to maintain consistency:
```

koshachy: Will it work?

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+> This function parses metadata, applies transformations through a lambda, and writes the modified metadata automatically.
+> 
+{style="tip"}
+
+## Create metadata from scratch
+
+To create metadata for a Kotlin class file from scratch using the Kotlin Metadata JVM library:
+
+1. Create an instance of `KmClass`, `KmPackage`, or `KmLambda`, depending on the type of metadata you want to generate.
+2. Populate the instance with attributes such as the class name, visibility, constructors, and function signatures.
```
koshachy, line 344: We can simplify it to "Add atrubuttes to the instance ..."

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+
+    > You can use Kotlin's `apply()` function to reduce boilerplate code while setting properties.
```
koshachy, line 346: Do we really need to specify that this is **Kotlin's** function?

koshachy: maybe it is better to say that this one is a scope function and provide a link to the Scope functions doc.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+    >
+    {style="tip"}
+
+3. Use the populated instance to create a `KotlinClassMetadata` object, which can generate a `@Metadata` annotation.
```
koshachy, line 350:
```suggestion
3. Use the instance to create a `KotlinClassMetadata` object, which can generate a `@Metadata` annotation.
```

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+
+> For a more detailed example, see the [Kotlin Metadata JVM GitHub repository](https://github.com/JetBrains/kotlin/blob/50331fb1496378c82c862db04af597e4198ec645/libraries/kotlinx-metadata/jvm/test/kotlin/metadata/test/MetadataSmokeTest.kt#L43).
+> 
+{style="tip"}
+
+## What's next
+
+* [See the API reference for the Kotlin Metadata JVM library](https://kotlinlang.org/api/kotlinx-metadata-jvm/).
+* [Check out the Kotlin Metadata JVM GitHub repostiory](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm).
+* [Learn about module metadata and working with `.kotlin_module` files.](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#module-metadata)
```
koshachy, line 420:
```suggestion
* [Learn about module metadata and working with `.kotlin_module` files](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#module-metadata).
```

koshachy (approved):
Nice work!

I left a few comments — most of them are nitpicks or small suggestions, so it's up to you whether to apply them.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+> The `kotlin-reflect` library relies on metadata to retrieve Kotlin-specific class details at runtime.
+> Any inconsistencies between the metadata and the actual `.class` file may lead to incorrect behavior when using reflection.
+> 
+{style="warning"}
+
+You can also use the Kotlin Metadata JVM library to inspect various declaration attributes such as visibility or modality, or to generate and embed metadata into `.class` files.
+
+## Add the library to your project
+
+To include the Kotlin Metadata JVM library in your project, add the appropriate dependency configuration based on your build tool.
```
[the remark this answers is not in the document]
daniCsorbaJB: I slightly prefer corresponding in that case - let's go with that

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+   </dependencies>
+   ...
+</project>
+```
+
+
+## Read and parse metadata
+
+You can use the `kotlin-metadata-jvm` library to extract structured information from compiled Kotlin `.class` files, such as class names, visibility, and signatures.
+This makes it useful in projects that need to analyze compiled Kotlin declarations.
```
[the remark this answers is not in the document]
daniCsorbaJB: I like it - I updated it

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+For example, the [Binary Compatibility Validator (BCV)](https://github.com/Kotlin/binary-compatibility-validator) relies on `kotlin-metadata-jvm` to print public API declarations.
+
+You can start exploring Kotlin class metadata by retrieving the `@Metadata` annotation from a compiled class using reflection:
```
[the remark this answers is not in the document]
daniCsorbaJB: While I like both suggested versions, to me both seem to imply that it's the only way to retrieve metadata, so as a first example despite being a bit wordier I'd prefer the original here.

docs/topics/metadata-jvm.md
```diff
@@ -0,0 +1,420 @@
...
+        println("This is not a Kotlin class.")
+    }
+}
+```
+
+After retrieving the `@Metadata` annotation, you can use the [`readLenient()`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/read-lenient.html) and [`readStrict()`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/-companion/read-strict.html) functions from the [`KotlinClassMetadata`](https://kotlinlang.org/api/kotlinx-metadata-jvm/kotlin-metadata-jvm/kotlin.metadata.jvm/-kotlin-class-metadata/) API to parse it.
+These functions allow you to extract detailed information about classes or files, while addressing different compatibility requirements:
+
+* `readLenient()`: Use this function to read metadata, including metadata generated by newer Kotlin compiler versions. This function doesn't support modifying or writing metadata.
+* `readStrict()`: Use this function when you need to modify and write metadata. The `readStrict()` function only works with metadata generated by Kotlin compiler versions fully supported by your project.
```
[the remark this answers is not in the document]
daniCsorbaJB: Hmm 🤔 - I feel like this alters a meaning slightly. 
The function itself doesn't modify the metadata it allows you to modify it.

Basically the difference here is that while both allow the user to read metadata - readLenient is more versatile in reading while readStrict allows modification of the metadata after reading

docs/topics/metadata-jvm.md
daniCsorbaJB: I really like the second one - I'll update it with that one thank you

PR #4679, opened 2025-01-28
Introduce information about DslMarker with function types

daniCsorbaJB:
This PR is created to introduce information on how to apply the `@DslMarker` annotation to function types, most commonly lambdas with receivers, to restrict implicit access to outer receivers in Kotlin DSLs and ensure scoped and safe interactions.

Related ticket: [KT-73249](https://youtrack.jetbrains.com/issue/KT-73249/)

### DIFF
--- a/docs/topics/type-safe-builders.md
+++ b/docs/topics/type-safe-builders.md
@@ -243,7 +243,7 @@ html {
 ```
 
 You can also apply the `@DslMarker` annotation directly to [function types](lambdas.md#function-types).
-To do so, annotate a `@DslMarker` annotation with `@Target(AnnotationTarget.TYPE)`:
+Simply annotate the `@DslMarker` annotation with `@Target(AnnotationTarget.TYPE)`:
 
 ```kotlin
 @Target(AnnotationTarget.TYPE)
@@ -251,7 +251,7 @@ To do so, annotate a `@DslMarker` annotation with `@Target(AnnotationTarget.TYPE
 annotation class HtmlTagMarker
 ```
 
-This allows the `@DslMarker` annotation to be applied to function types, most commonly to lambdas with receivers. For example:
+As a result, the `@DslMarker` annotation can be applied to function types, most commonly to lambdas with receivers. For example:
 
 ```kotlin
 fun html(init: @HtmlTagMarker HTML.() -> Unit): HTML { ... }
@@ -273,7 +273,7 @@ html {
 }
 ```
 
-This ensures that only the nearest receiver’s members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
+Only the nearest receiver's members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
 
 ### Full definition of the com.example.html package
 
--- a/docs/topics/type-safe-builders.md
+++ b/docs/topics/type-safe-builders.md
@@ -254,32 +254,26 @@ annotation class HtmlTagMarker
 This allows the `@DslMarker` annotation to be applied to function types, most commonly to lambdas with receivers. For example:
 
 ```kotlin
-fun HTML.head(init: @HtmlTagMarker Head.() -> Unit): Head {
-    val head = Head()
-    head.init()
-    return head
-}
+fun html(init: @HtmlTagMarker HTML.() -> Unit): HTML { ... }
 
-fun Head.title(init: @HtmlTagMarker Title.() -> Unit): Title {
-    val title = Title()
-    title.init()
-    return title
-}
+fun HTML.head(init: @HtmlTagMarker Head.() -> Unit): Head { ... }
+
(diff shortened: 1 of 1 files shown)


docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,46 @@ html {
...
+```kotlin
+@Target(AnnotationTarget.TYPE)
+@DslMarker
+annotation class HtmlTagMarker
+```
+
+This allows the `@DslMarker` annotation to be applied to function types, most commonly to lambdas with receivers. For example:
+
+```kotlin
+fun HTML.head(init: @HtmlTagMarker Head.() -> Unit): Head {
```
ice-phoenix, line 257: NB: I do not know what is the preferred way of doing code examples in our technical documentation, you are the expert here and your opinion is the most important =)

I would prefer to have a more complete example w.r.t. `@DslMarker`, meaning we should also define the `html` function

```
fun html(init: @HtmlTagMarker HTML.() -> Unit): HTML {
    val html = HTML()
    html.init()
    return html
}
```

ice-phoenix: Also, maybe we could keep the bodies of the functions empty or as `TODO()`? Their definitions are not that important w.r.t. DSL scoping rules, and the ones we have now are kinda not entirely correct.

docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,46 @@ html {
...
+}
+
+fun Head.title(init: @HtmlTagMarker Title.() -> Unit): Title {
+    val title = Title()
+    title.init()
+    return title
+}
+```
+
+When you call these functions, the `@DslMarker` annotation restricts access to outer receivers unless you specify them explicitly:
```
ice-phoenix, line 270: Maybe explain this a bit, as in "...annotation restricts access to outer receivers [within the body of a `@DslMarker`-marked lambdas] unless you...?

docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,46 @@ html {
...
+
+```kotlin
+html {
+    head {
+        title {
+            // Access to head or outer html members is restricted here.
```
ice-phoenix, line 276:
```suggestion
            // Access to `title`, `head` or other functions on outer receivers is restricted here.
```

docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,46 @@ html {
...
+        }
+    }
+}
+```
+
+This ensures that only the nearest receiver’s members are accessible within a lambda, preventing unintended interactions between nested scopes.
```
ice-phoenix, line 282: It's not only about members, it's about all available callables, meaning extensions are also included in the restrictions.

ice-phoenix (changes requested):

docs/topics/type-safe-builders.md
daniCsorbaJB: Great point - thank you - how about adding it as:
> This ensures that only the nearest receiver’s members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.

ice-phoenix (approved):

docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,40 @@ html {
...
+You can also apply the `@DslMarker` annotation directly to [function types](lambdas.md#function-types).
+To do so, annotate a `@DslMarker` annotation with `@Target(AnnotationTarget.TYPE)`:
+
+```kotlin
+@Target(AnnotationTarget.TYPE)
+@DslMarker
+annotation class HtmlTagMarker
+```
+
+This allows the `@DslMarker` annotation to be applied to function types, most commonly to lambdas with receivers. For example:
```
sarahhaggarty, line 254:
```suggestion
As a result, the `@DslMarker` annotation can be applied to function types, most commonly to lambdas with receivers. For example:
```

docs/topics/type-safe-builders.md
```diff
@@ -242,7 +242,40 @@ html {
...
+html {
+    head {
+        title {
+            // Access to title, head or other functions of outer receivers is restricted here.
+        }
+    }
+}
+```
+
+This ensures that only the nearest receiver’s members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
```
sarahhaggarty, line 276:
```suggestion
Only the nearest receiver's members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
```
Removing "this ensures" because it's not 100% clear what "this" refers to.

sarahhaggarty (approved):
Looks good!
I made a few suggestions to avoid using "this" too much at the beginning of sentences :) Please let me know if anything is unclear.

docs/topics/type-safe-builders.md
daniCsorbaJB: good point - updated it

PR #4683, opened 2025-01-30
KTL-2046 Complete the AB experiment with Hero Banner

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-2046/Complete-the-AB-experiment-with-Hero-Banner

### DIFF
--- a/data/hero-banner.yml
+++ b/data/hero-banner.yml
@@ -1,4 +1,4 @@
-isActive: true
+isActive: false
 title: "Ktor 3.0 is out"
 caption: "A better way to build servers & clients"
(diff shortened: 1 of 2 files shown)


data/hero-banner.yml
```diff
@@ -1,4 +1,4 @@
-isActive: true
+isActive: false
```
berezinant, line 1: Just wanted to check, am I correct that we want to stop showing banner?
If so, do we need to have the disabled banner in the code for future purposes or is it better to get rid of it completely for the sake of simplicity?

berezinant (approved):

PR #4688, opened 2025-02-03
Clarify range and step definition

daniCsorbaJB:
This PR is created to add some additional description about ranges in the documentation.
Related issue: [KT-74207](https://youtrack.jetbrains.com/issue/KT-74207/)

### DIFF
--- a/docs/topics/ranges.md
+++ b/docs/topics/ranges.md
@@ -4,20 +4,24 @@ Kotlin lets you easily create ranges of values using the [`.rangeTo()`](https://
 and [`.rangeUntil()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/range-until.html) functions from the 
 `kotlin.ranges` package. 
 
+A range represents an ordered set of values with a defined start and end. By default, it increments by 1 at each step.
+For example, `1..4` represents the numbers 1, 2, 3, and 4.
+
 To create:
-* a closed-ended range, call the `.rangeTo()` function with the `..` operator.
-* an open-ended range, call the `.rangeUntil()` function with the `..<` operator.
+
+* a closed-ended range, call the `.rangeTo()` function with the `..` operator. This includes both the start and end values.
+* an open-ended range, call the `.rangeUntil()` function with the `..<` operator. This includes the start value but excludes the end value.
 
 For example:
 
 ```kotlin
 fun main() {
 //sampleStart
-    // Closed-ended range
+    // Closed-ended range: includes both 1 and 4
     println(4 in 1..4)
     // true
     
-    // Open-ended range
+    // Open-ended range: includes 1, excludes 4
     println(4 in 1..<4)
     // false
 //sampleEnd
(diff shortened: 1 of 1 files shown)


docs/topics/ranges.md
```diff
@@ -4,20 +4,24 @@ Kotlin lets you easily create ranges of values using the [`.rangeTo()`](https://
 and [`.rangeUntil()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/range-until.html) functions from the 
 `kotlin.ranges` package. 
 
+A range represents an ordered set of values with a defined start and end. By default, it increments by 1 at each step.
+For example, `1..4` represents the numbers 1, 2, 3, and 4.
```
zamulla, line 8: This may be out of scope of clarifying definitions, so feel free to move this out of this particular PR, but since we're already here:
* Maybe add the "Range" title so that "Ranges and progressions" are broken up neatly into "Ranges" and "Progressions"?
* Since most of the ranges  is talking about iteration, I think it would be nice to move the "Ranges are particularly useful for iterating..." passage here, after the definition, to present the use case straight away. The details on creating and customizing ranges are for people who won't be satisfied with a simple example.

docs/topics/ranges.md
```diff
@@ -50,11 +54,10 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-ranges-downto"}
 
-It is also possible to iterate over numbers with an arbitrary step (not necessarily 1). This is done via the
-[`step`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/step.html) function.
+You can also iterate over numbers with a custom step using the
+[`step()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/step.html) function, instead of the default increment of `1`:
```
zamulla, line 58: The Kotlin guidelines recommend spelling out numbers from zero to nine, although in this context of arithmetic it might not work as well. But I don't think the "1" here needs the monospaced treatment, since it's just a number. Wdyt?

zamulla (approved):
Nice! I left a couple of comments for possible improvement, but nothing critical.

docs/topics/ranges.md
daniCsorbaJB: Good point, I agree - I was on the fence on this one - I put it in monospace as we are talking about the `step()` function which takes a value as seen in the example - I agree with it not needing the monospace though.

docs/topics/ranges.md
daniCsorbaJB: > Maybe add the "Range" title 
I didn't even think of that, but I completely agree - I added a short intro sentence before it as discussed
regarding the part about moving the passage here - While I think it might be a good idea, but I also think that it would make sense to also move downTo() here, but then it takes too long to get to the open-ended vs close-ended ranges and breaks the structure a bit... 🤔

PR #4690, opened 2025-02-04
async docs: CB & promises - sp fix and clarifications 

Liampronan:
- in the Callbacks section, there's a typo here: `**titled** christmas tree`. I was going to fix to `tilted` and then decided to update that to `pyramid of doom` since that seemed more well-known. 

- in the Promises/Futures section, the existing wording is unclear re: when the promise is returned: `we're promised that at some point it will return with an object called a Promise`. So I clarified that with a few more words, aiming to keep existing simplicity without overcomplicating.


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your suggestions @Liampronan !
Sorry it took a while to respond to your PR. I polished the wording a bit and aligned it with our style guide.
Thanks again!

PR #4709, opened 2025-02-17
Minor documentation edits

kencheng:
* Rename misleading function name in example code.

* Change sentences with multiple meanings to have a single meaning.

* Reword sentences to avoid awkward repetition.

* Be more specific when referring to example code.


sarahhaggarty:
Hi @kencheng !
Thanks for sharing your suggestions with us. Since the adjustments are small, they don’t bring a big shift to the overall content. After careful consideration, I've decided not to accept these changes for now. For future contributions, I’d recommend focusing on more substantial updates or ideas that can really enhance the documentation.

But if you have any other suggestions in the future, please let us know!

Thanks again!

PR #4712, opened 2025-02-21
Update generics.md

nbalasundara:
The use of term 'Upper bound' precedes its usage in text. Clarify.

### DIFF
--- a/docs/topics/generics.md
+++ b/docs/topics/generics.md
@@ -250,7 +250,7 @@ type will be a subtype of that projection.
 
 Kotlin provides so-called _star-projection_ syntax for this:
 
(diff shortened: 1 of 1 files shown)


nbalasundara:
Direct reader to the definition of 'Upper bound' whose usage precedes it.

daniCsorbaJB:
Hi [nbalasundara](https://github.com/nbalasundara), thanks so much for your suggestion—it's a valid point! We've got an [open issue](https://youtrack.jetbrains.com/issue/KT-56900) to restructure this page, and I've added your input (and your name) there. We'll update the page to fix this as part of that larger work, so I'm going to close this PR for now. Thanks again for bringing this to our attention! Let us know if you find any other inconsistencies. Best regards, Daniel

PR #4713, opened 2025-02-22
Update generics.md

nbalasundara:
Star projections are only ever applicable at a use site. However, nowhere in the doc does it explicitly state that.


sarahhaggarty:
Hi @nbalasundara !
Thanks for your suggestion! After discussing it with our developer team, I've decided not to include this change for now. You are right that we could do a better job of explaining the restrictions of star projections, but addressing this would require more than what you've suggested. Since we plan to revamp this page, I've added your feedback to the relevant ticket. If you notice anything else that could be improved, please let us know.

Thanks again!

PR #4727, opened 2025-02-27
Moving the KT-71275 from Removed section to the Done section

johnlang:
According to the Compiler folks it makes sense to mark it as done.


koshachy (approved):
LGTM!

PR #4733, opened 2025-03-03
update: wasm custom formatters enabled by default

AlejandraPedroza:


docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
+#### Use custom formatters
 
-In addition to the default debugging configuration, you can set up custom formatters to display and locate variable values in a more user-friendly and comprehensible manner.
+Custom formatters help display and locate variable values in a more user-friendly and comprehensible manner when debugging Kotlin/Wasm code.
 
-![Kotlin/Wasm improved debugger](wasm-debugger-improved.png){width=600}
+Custom formatters are enabled by default in development builds, so you don't need additional Gradle configurations. 
 
-This implementation is supported across major browsers like Firefox and Chromium-based as
+This feature is supported across major browsers like Firefox and Chromium-based as
```
zamulla, line 142: This could be clearer, I think: "...is supported in Firefox and Chromium-based browsers." Because further we don't talk about anything else, and from a cursory search I'm not sure other major browsers, like Safari, support this at all.

docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
 it uses the [custom formatters API](https://firefox-source-docs.mozilla.org/devtools-user/custom_formatters/index.html).
 
-To set up custom formatters for an improved debugging experience:
+To use this feature, you only need to ensure that custom formatters are enabled in your browser's developer tools:
```
zamulla, line 145:
```suggestion
To use this feature, ensure that custom formatters are enabled in your browser's developer tools:
```

docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
 
-1. Add the following compiler argument to the `wasmJs` compiler options:
+* In the Chrome DevTools, it's placed in **Settings | Preferences | Console**:
```
zamulla, line 147:
```suggestion
* In Chrome DevTools, it's placed in **Settings | Preferences | Console**:
```

In their docs, "Chrome DevTools" is a product name which doesn't need an article.

docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
-           compilerOptions {
-               freeCompilerArgs.add("-Xwasm-debugger-custom-formatters")
-           }
-       }
-   }   
-   ```
+  ![Enable custom formatters in Chrome](wasm-custom-formatters-chrome.png){width=400}
 
-2. Enable the **Custom formatters** feature in your browser:
+* In the Firefox Developer Tools, it's placed in **Settings | Advanced settings**:
```
zamulla, line 151: Here and above, I'm not sure what "it's placed" means — we should probably specify what "it" is, for example "...you can find the checkbox in ..."

docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
 
-   * In the Chrome DevTools, it's placed in **Settings | Preferences | Console**:
+  ![Enable custom formatters in Firefox](wasm-custom-formatters-firefox.png){width=400}
 
-   ![Enable custom formatters in Chrome](wasm-custom-formatters-chrome.png){width=400}
+Custom formatters work for development builds. If you have specific requirements for production builds,
```
zamulla, line 155: I'd make sure it's clear that we're talking about Wasm project development builds, and not browser development builds. Not a big deal, but to be thorough.

docs/topics/wasm/wasm-debugging.md
```diff
@@ -133,41 +133,42 @@ kotlin {
...
-   ![Enable custom formatters in Firefox](wasm-custom-formatters-firefox.png){width=400}
+        compilerOptions {
+            freeCompilerArgs.add("-Xwasm-debugger-custom-formatters")
+        }
+    }
+}
+```
 
-After setting up custom formatters, you can complete the debugging tutorial.
+After enabling custom formatters, you can complete the debugging tutorial.
```
zamulla, line 171: In the beginning we're talking about custom formatters as an optional thing, and here it seems that only with them enabled I can complete the tutorial. We should probably use a softer suggestion here? Like, "you can continue with the debugging tutorial" or something.

docs/topics/wasm/wasm-debugging.md
```diff
@@ -200,7 +201,7 @@ After setting up custom formatters, you can complete the debugging tutorial.
 
    ![Check call stack](wasm-debug-scope.png){width=550}
 
-   For an improved visualization of the variable values, see [Enhance your debugging with custom formatters](#enhance-your-debugging-with-custom-formatters).
+   For an improved visualization of the variable values, see [Use custom formatters](#use-custom-formatters).
```
zamulla, line 204: This link doesn't work when the "Use custom formatters" section is collapsed: the browser is trying to navigate to the anchor, but the content is not displayed. The link should probably lead to the "Configure your browser for debugging" heading while explaining that custom formatters are a part of this and the user needs to dig into it.

zamulla (changes requested):
Nice, thanks! I think we can make the flow clearer in a couple of points, and also there is a link which is not quite broken, but doesn't work.

zamulla (approved):
Thanks!

PR #4734, opened 2025-03-04
Intermediate tour: Null safety chapter

sarahhaggarty:
This PR is to add the fourth chapter of the intermediate tour to the staging branch.

### DIFF
--- a/docs/topics/tour/kotlin-tour-intermediate-null-safety.md
+++ b/docs/topics/tour/kotlin-tour-intermediate-null-safety.md
@@ -1,0 +1,665 @@
+[//]: # (title: Intermediate: Null safety)
+
+In the beginner's tour, you learned how to handle `null` values in your code, but there are also use cases when using 
+null safety features can be helpful.
+
+## Smart casts and safe casts
+
+Kotlin can sometimes infer the type without explicit declaration. When you tell Kotlin to treat a variable or object as if it belongs to a
+specific type, this process is called **casting**. When a type is automatically cast, like when it's inferred, it's called
+**smart casting**.
+
+### is and !is operators
+
+Before we explore how casting works, let's see how you can check if an object has a certain type. For this, you can use the
+`is` and `!is` operators with `when` or `if` conditional expressions:
+
+* `is` checks if the object has the type and returns a boolean value.
+* `!is` checks if the object **doesn't** have the type and returns a boolean value.
+
+For example:
+
+```kotlin
+fun printObjectType(obj: Any) {
+    when (obj) {
+        is Int -> println("It's an Integer with value $obj")
+        !is Double -> println("It's NOT a Double")
+        else -> println("Unknown type")
+    }
+}
+
+fun main() {
+    val myInt = 42
+    val myDouble = 3.14
+    val myList = listOf(1, 2, 3)
+  
+    // The type is Int
+    printObjectType(myInt)
+    // It's an Integer with value 42
+
+    // The type is List, so it's NOT a Double.
+    printObjectType(myList)
+    // It's NOT a Double
+
+    // The type is Double, so the else branch is triggered.
+    printObjectType(myDouble)
+    // Unknown type
+}
+```
+{kotlin-runnable="true" id="kotlin-tour-null-safety-casts"}
+
+> You've already seen this in practice when you learned about sealed classes in the [Open and other special classes](kotlin-tour-intermediate-classes-part-3.md) chapter.
+> 
+{style="tip"}
+
+### as and as? operators
+
+To explicitly _cast_ an object to any other type, use the `as` operator. This includes casting from a nullable 
+type to its non-nullable counterpart. If the cast isn't possible, the program crashes **at runtime**. This is why it's 
+called the **unsafe** cast operator.
+
+```kotlin
+fun main() {
+//sampleStart
+    val a: String? = null
+    val b = a as String
+
+    // Triggers an error at runtime
+    print(b)
+//sampleEnd
+}
+```
+{kotlin-runnable="true" validate="false" id="kotlin-tour-null-safety-as-operator"}
+
+To explicitly cast an object to a non-nullable type, but return `null` instead of throwing an error on failure, use the `as?`
+operator. Since the `as?` operator doesn't trigger an error on failure, it is called the **safe** operator.
+
+```kotlin
+fun main() {
+//sampleStart
+    val a: String? = null
+    val b = a as? String
+
+    // Returns null value
+    print(b)
+    // null
+//sampleEnd
+}
+```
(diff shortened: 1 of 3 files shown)


docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+null safety features can be helpful.
+
+## Smart casts and safe casts
+
+You already know how to explicitly declare the type of a variable or object and that Kotlin sometimes has the ability
+to infer the type without explicit declaration. When you tell Kotlin to treat a variable or object as if it belongs to a
+specific type, this process is called **casting**. When a type is automatically cast, like when it's inferred, it's called
+**smart casting**.
+
+Before we explore how casting works, how can you check if an object has a certain type? For this, you can use the
```
danil-pavlov, line 13: Just a nitpick, I feel like the connection in the first sentence is not very clear, what about:

> Before we explore how casting works, let's learn how you can check if an object has a certain type.

or

> Before exploring how casting works, let's see how to check if an object has a certain type.

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+`is` and `!is` operators with `when` or `if`:
+
+* `is` checks if the object has the type and returns a boolean value.
+* `!is` checks if the object **doesn't** have the type and returns a boolean value.
+
+You've already seen this in practice when you learned about sealed classes in the [Open and other special classes](kotlin-tour-intermediate-classes-part-3.md) chapter.
```
danil-pavlov, line 19: This note breaks the flow a little bit; maybe move it under the code as a tip?

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+
+To:
+
+```kotlin
+fun calculateTotalStringLength(items: List<Any>): Int {
+    return items.sumOf { (it as? String)?.length ?: 0 }
+}
+```
+
+In the first version, the example:
```
danil-pavlov, line 120: It might be even better to have these lists with explanations right next to the code
Plus, visually is would help to avoid to consequent lists

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+For more information about how you can use `return` in your code, see [Returns and jumps](returns.md).
+
+## Null safety practice
+
+### Exercise 1 {initial-collapse-state="collapsed" collapsible="true" id="null-safety-exercise-1"}
+
+You are developing a notification system for an app where users can enable or disable different types of notifications.
+Complete the `getNotificationPreferences()` function so that:
+
+1. The `validUser` variable uses the `as?` operator to check if `user` is a `User` object. If it isn't, return an empty list.
```
danil-pavlov, line 350:
```suggestion
1. The `validUser` variable uses the `as?` operator to check if `user` is an object of the `User` class. If it isn't, return an empty list.
```

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+In both of these examples, if all items are `null` values, an empty list is returned.
+
+Kotlin also provides functions that you can use to find values in collections. If a value isn't found, they return `null`
+values instead of triggering an error:
+
+* [`singleOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/single-or-null.html), looks for only one item by its exact value. If one doesn't exist or there are multiple items with the same value, returns a `null` value.
+* [`maxOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/max-or-null.html), finds the highest value. If one doesn't exist, returns a `null` value.
+* [`minOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/min-or-null.html), finds the lowest value. If one doesn't exist, returns a `null` value.
+
+> The `singleOrNull()`, `maxOrNull()`, and `minOrNull()` functions are designed to be used with collections that **don't**
```
danil-pavlov, line 182: Again, I feel like the note here breaks the flow (and makes it look like the example is directly connected to the note, and not to the list), consider moving it lower

danil-pavlov (changes requested):
Great chapter! Just minor suggestions
Also,  I'm a bit concerned with the length of "Smart casts and safe casts" and "Early returns and the Elvis operator﻿". It's mostly because of code samples,  but it would be great to have subsections there (if it;s possible of course)

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
```diff
@@ -0,0 +1,658 @@
...
+
+## Smart casts and safe casts
+
+You already know how to explicitly declare the type of a variable or object and that Kotlin sometimes has the ability
+to infer the type without explicit declaration. When you tell Kotlin to treat a variable or object as if it belongs to a
+specific type, this process is called **casting**. When a type is automatically cast, like when it's inferred, it's called
+**smart casting**.
+
+Before we explore how casting works, how can you check if an object has a certain type? For this, you can use the
+`is` and `!is` operators with `when` or `if`:
```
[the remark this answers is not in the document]
sarahhaggarty: In the beginner tour, we refer to `when` or `if` without calling them conditions. Can we do the same here? https://kotlinlang.org/docs/kotlin-tour-control-flow.html#conditional-expressions

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
sarahhaggarty: I restructured it as you suggest :)

sarahhaggarty:
Thanks @danil-pavlov !
I added two new headers in "Smart casts and safe casts" but I couldn't find a way to add headers in "Early returns and the Elvis operator﻿". I also added more sample starts/ends to reduce the code sample sizes. Hopefully that makes the chapter seem a bit shorter. WDYT?

docs/topics/tour/kotlin-tour-intermediate-null-safety.md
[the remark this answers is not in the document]
danil-pavlov: Maybe just expressions? It's just that in this sentence, `is` and `!is?` have a term, but `when` and `if` don't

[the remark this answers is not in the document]
sarahhaggarty: I see now. I went with "conditional expressions". Since `when` can be an expression or a statement 😉

PR #4754, opened 2025-03-16
update: add a warning for casting in java-to-kotlin-nullability-guide.md

XYZboom:
Based on some of my recent findings, many Java developers new to Kotlin often confuse `obj as T` with `obj as T?`, especially when interacting with Java code. Accidentally writing `obj as T` instead of `obj as T?` may lead to unexpected NPEs. For this reason, I've added a warning block in this documentation.

You can see some example commits here:
[androidx-8e967e](https://github.com/androidx/androidx/commit/d0368b8bd00b91987340eaea3856a87c248e967e#diff-82fee78f32a586f9babd3fc0b7a548a1cb2977b9fecdf015b878a9c10cca2ee6L50)
[androidAPS-36b99e](https://github.com/nightscout/AndroidAPS/commit/22cad1a761b1955a628355d9006ceca34036b99e)
[androidAPS-49f0f1](https://github.com/nightscout/AndroidAPS/commit/407a7700a5714bcd2f521480179e14694049f0f1)
[androidAPS-f95b34](https://github.com/nightscout/AndroidAPS/commit/4c9052b29a166cb946fcc138afaa92458ff95b34)
[facebook-android-sdk-fd5a42](https://github.com/facebook/facebook-android-sdk/commit/8d4a2e19cbe9f75bb37eed76d7abd7c785fd5a42)

These fix commits demonstrate how easily people might mistakenly use `obj as T` instead of `obj as T?` during migration, which is precisely why I opened this PR.

### DIFF
--- a/docs/topics/jvm/java-to-kotlin-nullability-guide.md
+++ b/docs/topics/jvm/java-to-kotlin-nullability-guide.md
@@ -337,6 +337,13 @@ you would do the check anyway, but no additional boxing is performed this way.
 >
 {style="note"}
 
+> In Java, the unsafe type cast `(String) y` corresponds to `y as String?` in Kotlin, not `y as String`.
+This is because `y as String?` only throws `ClassCastException` when the `y` is not of type `String`,
+whereas `y as String` additionally throws `NullPointerException` (NPE) when `y` is `null`.
+This distinction might lead to accidental NPE swallowing if using try-catch blocks designed to handle `ClassCastException`.
+>
+{style="warning"}
+
 ## What's next?
 
 * Browse other [Kotlin idioms](idioms.md).


XYZboom:
Thanks for your reply! I noticed that we have a migration guide recommending the use of as?. Perhaps my earlier description was not clear enough — this issue actually occurs when Kotlin calls Java code, as shown in the five commits I mentioned in my previous comment. When calling Java from Kotlin, my guess is that developers rely more heavily on the interaction guide rather than the migration guide, which is why I added this warning here.

sarahhaggarty:
Thanks for the additional context @XYZboom! And apologies again for the slow response. It's true that the closest match to Java's `(String) x` is actually `x as String?`. However, we still recommend using the safe cast operator `as?`, because it returns `null` rather than throwing a `ClassCastException`.

Although using `x as String?` might make it easier to migrate your code, so you can continue handling `ClassCastException` the same way as before, it is safer to migrate your code to handle `null` values directly.

XYZboom:
@sarahhaggarty Thanks for your apply. You're right, we should use `x as? String` instead of `x as String?`. But I don't agree with what you said about `x as String?` makes migration easier. It only makes migration more safe and more in line with Kotlin's style. In my opinion, the migration may takes two steps. First we do a same semantic transfer. In this step, the semantics should be consistent to avoid damaging certain designs. The second step is to make the code conform to Kotlin's design style.
For example, the first step migration of the following code is:
```java
public class A {
    Object o;

    public static String asString(A a) {
        if (a == null) {
            return null;
        }
        return (String) a.o;
    }
}
```
```kt
class A {
    var o: Any? = null

    companion object {
        fun asString(a: A?): String? {
            if (a == null) {
                return null
            }
            return a.o as String?
        }
    }
}
```
The second step is:
```kt
class A {
    var o: Any? = null

    companion object {
        fun asString(a: A?): String? {
            return a?.o as? String
        }
    }
}
```
However the second step migration **hides** the `ClassCastException`, the developers must check carefully if any code is affected. For some cases, there may not be a second step migration after the first step migration is completed. Because changing the handling of `ClassCastException` to handling `null` values may require a longer migration time.
As you can see my examples above, none of these 6 examples use `as?` because they only finish the first step migration.
Although automatic migration tools can handle many migration situations, I still believe it is necessary to inform developers to use `as T?` in the first step of migration but not `as T`.

XYZboom:
For example, in [facebook-android-sdk-fd5a42](https://github.com/facebook/facebook-android-sdk/commit/8d4a2e19cbe9f75bb37eed76d7abd7c785fd5a42), if `oldAuthenticationToken` is `null`, it may mean that no value in intent but if `oldAuthenticationToken` is not an `AuthenticationToken`, it may mean that this is an illegal intent.
However, the code for handling illegal intent may not be here, and if written as `as?`, these logics would have to be altered.

sarahhaggarty:
Hi @XYZboom!
Taking your feedback into account and after discussing with our developers, I added some migration advice in the [Casting types safely](https://kotlinlang.org/docs/java-to-kotlin-nullability-guide.html#casting-types-safely) section as part of https://github.com/JetBrains/kotlin-web-site/pull/5078. I hope this covers your original intent!

XYZboom:
@sarahhaggarty Thanks a lot! The newly added section completely covers my ideas.

PR #4761, opened 2025-03-21
Add OpenSea to server overview doc

jalagar:
👋 We recently rewrote our entire tech stack into Kotlin (see https://opensea.io/blog/articles/introducing-os2). Would love to have our brand on the Kotlin website!


jalagar:
@meilalina sure would love to! Can you email me at james.uejio@opensea.io (or even better DM me on X https://x.com/jalagar_eth). Sorry for the late response!

PR #4768, opened 2025-03-25
Update code samples on landing to kotlin.test (from JUnit)

bartekpacia:
:)

I can run tests annotated with `kotlin.test.Test` on Kotlin/Native.

### DIFF
--- a/templates/inc/pages/index/code-examples.html
+++ b/templates/inc/pages/index/code-examples.html
@@ -128,8 +128,6 @@ val messages = listOf(                                        // Create a list
             <div class="kotlin-overview-code-example is_hidden" id="kotlin-code-example-tests">
                 <pre theme="darcula" class="sample" data-mobile-shorter-height="400" auto-indent="false" data-min-compiler-version="1.5"  data-target-platform="junit">
 // Tests
-// The following example works for JVM only
-import org.junit.Test
 import kotlin.test.*
 
 class SampleTest {


nikpachoo:
Hi @bartekpacia,
Thanks for the contribution.
Agreed, we can drop the comment - but removing `import org.junit.Test` breaks the sample.

Screenshot 2025-03-26 at 17 40 28

bartekpacia:
Then I suspect something's wrong with dependencies (but not sure at all, I don't know this codebase).

> Agreed, we can drop the comment - but removing import org.junit.Test breaks the sample.

This doesn't make sense – if we `import org.junit`, then the comment makes sense, because JUnit is a JVM-only dependency.

Here's my project where I use `org.jetbrains.kotlin:kotlin-test:2.1.20` and the tests run fine on Kotlin/Native. This is a pure Kotlin Multiplatform project, no JVM-dependencies.

Screenshot 2025-03-26 at 23 06 12

bartekpacia:
:(

dkrasnoff:
We don't have dependencies like: "org.jetbrains.kotlin:kotlin-test-junit5:2.1.10" on the playground.

Such dependencies contain [required aliases](https://github.com/JetBrains/kotlin/blob/5d6755750ea964efa356987478ce98370c657b96/libraries/kotlin.test/junit5/src/main/kotlin/Annotations.kt#L11).

So, we still need to write the import manually.

nikpachoo:
@bartekpacia Thanks for reporting the issue. 
We’re considering adding `org.jetbrains.kotlin:kotlin-test-junit5` support to the Compiler Server. You can follow the progress here: [KT-77561](https://youtrack.jetbrains.com/issue/KT-77561)

Closing the PR as the proposed changes cannot be accepted.

PR #4771, opened 2025-03-26
KGP compiler arguments for Js update

AlejandraPedroza:
This PR updates the KGP compiler arguments for Js

For more information, see:
[KT-75694](https://youtrack.jetbrains.com/issue/KT-75694) Fix JS compiler options table


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -364,18 +364,18 @@ Also, see [Types for compiler options](#types-for-compiler-options).
...
+|---|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---|-----------------------------|
+| `friendModulesDisabled` | Disable internal declaration export                                                                                                                                                                                                      | | false                       |
+| `main` | Define whether the `main` function should be called upon execution                                                                                                                                                                       | "call", "noCall". Also, see [Types for compiler options](#types-for-compiler-options) | JsMainFunctionExecutionMode.CALL |
```
Tapchicoma, line 370: As it is about Gradle DSL - I would propose not to use strings ("call", "noCall") but rather actual possible types: `JsMainFunctionExecutionMode.CALL` and `JsMainFunctionExecutionMode.NO_CALL`.

Same approach should applied in changes below.

Tapchicoma, line 370: Nit: KDoc here uses "Specify" instead of "Define". Not sure if it is a problem.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -364,18 +364,18 @@ Also, see [Types for compiler options](#types-for-compiler-options).
...
+| `metaInfo` | This attribute is deprecated and will be removed in the future. It is left only for compatibility with the legacy backend                                                                                                                | | false                            |
```
Tapchicoma, line 371: `metaInfo` was removed, so I would not list it here.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -364,18 +364,18 @@ Also, see [Types for compiler options](#types-for-compiler-options).
...
+| `sourceMapPrefix` | Add the specified prefix to paths in the source map                                                                                                                                                                                      |  | null                            |
+| `target` | Generate JS files for specific ECMA version                                                                                                                                                                                              | "es5", "es2015"  | "es5"                        |
+| `typedArrays` | This attribute is deprecated and will be removed in the future. It is left only for compatibility with the legacy backend                                                                                                                                                                                              | | false                            |
```
Tapchicoma, line 378: I would not list it here

AlejandraPedroza: Do you mean removing the `typedArrays` argument from the list?

Tapchicoma (approved):

PR #4785, opened 2025-04-02
update: update Compile Kotlin and Java sources in Maven

koshachy:


docs/topics/maven.md
```diff
@@ -135,25 +137,43 @@ If you need to configure an execution, you need to specify its ID. You can find
...
 <build>
     <plugins>
+        <!-- Kotlin compiler plugin -->
         <plugin>
             <groupId>org.jetbrains.kotlin</groupId>
             <artifactId>kotlin-maven-plugin</artifactId>
             <version>${kotlin.version}</version>
-            <extensions>true</extensions> <!-- You can set this option 
-            to automatically take information about lifecycles -->
+            <extensions>true</extensions>
```
ALikhachev, line 231: It seems that actually this `<extensions>true</extensions>` does not make any sense. According to the previous comments, it could help avoid some configuration, but that does not appear to be true in Kotlin + Java in this particular configuration. 

However, if we modify the example to
```
    <plugins>
      <!-- Kotlin compiler plugin -->
      <plugin>
        <groupId>org.jetbrains.kotlin</groupId>
        <artifactId>kotlin-maven-plugin</artifactId>
        <version>${kotlin.version}</version>
        <extensions>true</extensions>
        <executions>
          <execution>
            <id>default-compile</id>
            <phase>compile</phase>
            <configuration>
              <sourceDirs>
                <sourceDir>${project.basedir}/src/main/kotlin</sourceDir>
                <sourceDir>${project.basedir}/src/main/java</sourceDir>
              </sourceDirs>
            </configuration>
          </execution>
          <execution>
            <id>default-test-compile</id>
            <phase>test-compile</phase>
            <configuration>
              <sourceDirs>
                <sourceDir>${project.basedir}/src/test/kotlin</sourceDir>
                <sourceDir>${project.basedir}/src/test/java</sourceDir>
              </sourceDirs>
            </configuration>
          </execution>
        </executions>
      </plugin>
    </plugins>
```
we can completely skip the Java plugin configuration and it would just work, thanks to the `<extensions>` setup. This configuration indeed seems much easier and shows benefit of `extensions`

Note that I changed execution IDs in this example.

I would suggest introducing something similar to how we show Kotlin/Groovy scripts, aka "tabs" for showing configuration with `<extensions>true</extensions>` and without it.

ALikhachev, line 231: Oh, interesting, I tried just copy-pasting the setup (both before any changes and from this PR) on top of pure Kotlin setup – it leads to incorrect configuration when Java code can see Kotlin code, but Kotlin code cannot see Java.

Either removing the extensions setup or removing of the following configuration helps.
```
<sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
<testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
```

With the abovementioned change of execution id, it works fine. That's because setup from extensions creates executions `default-compile` and `default-test-compile`

ALikhachev:
I would suggest additionally highlighting that in the setup Kotlin + Java, we need to let Kotlin compiler know about the Java sources, so that's why we configure `${project.basedir}/src/main/java` as a source directory for Kotlin. That is required for Kotlin -> Java references.

ALikhachev (approved):

PR #4786, opened 2025-04-02
update: improve Classes page

AlejandraPedroza:
This PR revamps the content and code examples in the Classes page following external feedback. 
Some of the key changes are:

- Improve the whole page and revamp the code snippets.

- Explain the use case for multiple init blocks.

- Explain constructor chaining.

- Add examples for primary constructors. 

- Clarify the open keyword.

- Explain how to access a class property from a class member function.

- Use the term 'object' consistently.

- Clarify that abstract classes are.

- Provide example companion objects.

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -1,245 +1,747 @@
 [//]: # (title: Classes)
 
-Classes in Kotlin are declared using the keyword `class`:
+Like other object-oriented languages, Kotlin uses _classes_ to encapsulate data and behavior 
+for reusable, structured code.
+
+Kotlin offers a concise syntax for declaring classes. To declare a class, use the `class` keyword 
+followed by the class name:
 
 ```kotlin
 class Person { /*...*/ }
 ```
 
-The class declaration consists of the class name, the class header (specifying its type parameters, the primary constructor, 
-and some other things), and the class body surrounded by curly braces. Both the header and the body are optional; if the
-class has no body, the curly braces can be omitted.
+The class declaration consists of: 
+* **Class header**, including:
+  * `class` keyword
+  * Class name
+  * Type parameters (if any)
+  * [Primary constructor](#primary-constructor) (optional)
+* **Class body** (optional), surrounded by curly braces `{}`, and including class members:
+  * [Secondary constructors](#secondary-constructors)
+  * [Functions](functions.md)
+  * [Properties](properties.md)
+  * [Nested and inner classes](nested-classes.md)
+  * [Object declarations](object-declarations.md)
+
+Here's an example of a class declaration with a property:
 
 ```kotlin
-class Empty
+// Class header with a primary constructor that initializes the 'name' property
+class Person(val name: String) {
+    // Class body with the `age` property 
+    var age: Int = 0
+}
 ```
 
-## Constructors
+In this example, the `Person` class is defined with a primary constructor that declares a read-only property 
+`name` of type `String`. In the body, the class includes a mutable `age` property initialized to `0`. 
+This class represents a person with a fixed name and an age you can update after 
+[object creation](#creating-objects-of-classes).
 
-A class in Kotlin has a _primary constructor_ and possibly one or more _secondary constructors_. The primary constructor
-is declared in the class header, and it goes after the class name and optional type parameters.
+Both the class header and class body can be minimal. If the class does not have a body, you can omit the curly braces `{}`:
 
 ```kotlin
-class Person constructor(firstName: String) { /*...*/ }
+// Class with only a name and no body
+class Person
 ```
 
-If the primary constructor does not have any annotations or visibility modifiers, the `constructor` keyword can be omitted:
+## Constructors and initializer blocks
+
+A class in Kotlin has a [_primary constructor_](#primary-constructor) and, optionally, one or more [_secondary constructors_](#secondary-constructors). 
+
+The primary constructor is the main way to initialize a class, and you have to declare it in the class header. 
+The secondary constructor provides additional initialization logic, and you have to declare it in the class body.
+
+### Primary constructor
+
+The primary constructor is the main way to initialize a class, 
+setting up the initial state of an object when [it is created](#creating-objects-of-classes). 
+
+Declare the primary constructor in the class header after the class name:
 
 ```kotlin
-class Person(firstName: String) { /*...*/ }
+class Person constructor(name: String) { /*...*/ }
 ```
 
-The primary constructor initializes a class instance and its properties in the class header. The class header can't contain
-any runnable code. If you want to run some code during object creation, use _initializer blocks_ inside the class body.
-Initializer blocks are declared with the `init` keyword followed by curly braces. Write any code that you want to run
-within the curly braces.
+In this example, the primary constructor initializes the class `Person` by defining a parameter `name` of type `String`.
 
-During the initialization of an instance, the initializer blocks are executed in the same order as they appear in the
-class body, interleaved with the property initializers:
+This primary constructor ensures that when [creating an object](#creating-objects-of-classes) of `Person`, a 
+value for `name` must be provided, defining an initial required input.
+
+If the primary constructor does not have any annotations or [visibility modifiers](visibility-modifiers.md#constructors), 
+you can omit the `constructor` keyword:
 
 ```kotlin
-//sampleStart
-class InitOrderDemo(name: String) {
-    val firstProperty = "First property: $name".also(::println)
-    
-    init {
-        println("First initializer block that prints $name")
-    }
-    
-    val secondProperty = "Second property: ${name.length}".also(::println)
-    
-    init {
-        println("Second initializer block that prints ${name.length}")
+class Person(name: String) { /*...*/ }
+```
+
+You can define simple properties directly in the primary constructor. This approach makes it easy to create
+classes that need properties initialized directly during object creation.
+
+To declare a read-only property, use the `val` keyword in the parentheses before the argument name. 
+For a mutable property, use the `var` keyword:
+
+```kotlin
+class Person(val name: String, var age: Int) { /*...*/ }
+```
+
+In this example, the `Person` class has a primary constructor with two properties: `name` and `age`. 
+The `name` property is declared as read-only using `val`, while the `age` property is declared as mutable using `var`, meaning that
+the value of `age` can change later on.
+Both properties are initialized directly when an object of `Person` [is created](#creating-objects-of-classes).
+
+A class can define member [functions](functions.md) that, in turn, access class properties.
+To make a primary constructor parameter accessible in member functions, declare it as a property using `val` or `var`:
+
+```kotlin
+// Class with a primary constructor defining properties
+class Person(val name: String, var age: Int) {
+    // Member function accessing class properties
+    fun introduce(): String {
+        return "Hi, I'm $name and I'm $age years old."
     }
 }
-//sampleEnd
+```
+
+See classes in action. Considering the same code example, if you [create an object](#creating-objects-of-classes) of the `Person` class by passing values for
+`name` and `age`, the output is:
+
+```kotlin
+// Class with a primary constructor defining properties
+class Person(val name: String, var age: Int) {
+    // Member function accessing class properties
+    fun introduce(): String {
+        return "Hi, I'm $name, and I'm $age years old."
+    }
+}
 
 fun main() {
-    InitOrderDemo("hello")
+    // Creates an object of the 'Person' class
+    val person = Person("Alice", 25)
+
+    // Calls the function
+    println(person.introduce()) 
+    // Hi, I'm Alice, and I'm 25 years old.
 }
 ```
 {kotlin-runnable="true"}
 
-Primary constructor parameters can be used in the initializer blocks. They can also be used in property initializers
-declared in the class body:
-
-```kotlin
-class Customer(name: String) {
-    val customerKey = name.uppercase()
-}
-```
-
-Kotlin has a concise syntax for declaring properties and initializing them from the primary constructor:
-
-```kotlin
-class Person(val firstName: String, val lastName: String, var age: Int)
-```
-
-Such declarations can also include default values of the class properties:
-
-```kotlin
-class Person(val firstName: String, val lastName: String, var isEmployed: Boolean = true)
-```
-
-You can use a [trailing comma](coding-conventions.md#trailing-commas) when you declare class properties:
+Additionally, you can use a [trailing comma](coding-conventions.md#trailing-commas) when declaring many class properties:
 
 ```kotlin
 class Person(
-    val firstName: String,
+    val name: String,
     val lastName: String,
-    var age: Int, // trailing comma
+    var age: Int,
 ) { /*...*/ }
 ```
 
-Much like regular properties, properties declared in the primary constructor can be mutable (`var`) or read-only (`val`).
-
-If the constructor has annotations or visibility modifiers, the `constructor` keyword is required and the modifiers go before it:
+Using the trailing comma example, if you [create an object](#creating-objects-of-classes) of the `Person`
+class by passing values for `name`, `lastName`, and `age`, you get:
 
 ```kotlin
-class Customer public @Inject constructor(name: String) { /*...*/ }
+// Class with a primary constructor that initializes 'name', 'lastName', and 'age'
+class Person(
+    val name: String,
+    val lastName: String,
+    var age: Int,
+)
+
+fun main() {
+    // Creates an object of the 'Person' class with given values
+    val person = Person("John", "Doe", 30)
+
+    // Accesses properties of the object
+    println("${person.name} ${person.lastName} is ${person.age} years old.")
+    // John Doe is 30 years old.
+}
+```
+{kotlin-runnable="true"}
+
+You can also assign values to properties in the primary constructor:
+
+```kotlin
+class Person(val name: String = "John", var age: Int = 30) { /*...*/ }
 ```
 
-Learn more about [visibility modifiers](visibility-modifiers.md#constructors).
+Assigning values in primary constructors ensures that if no values are passed during [object creation](#creating-objects-of-classes), the property uses the default values:
+
+```kotlin
+// Class with a primary constructor including default values for 'name' and 'age'
+class Person(val name: String = "John", var age: Int = 30)
+
+fun main() {
+    // Creates an object using default values
+    val person = Person()
+    println("Name: ${person.name}, Age: ${person.age}")
+    // Name: John, Age: 30
+}
+```
+{kotlin-runnable="true"}
+
+Also, you can use the primary constructor parameters to initialize additional class properties directly in the class body:
+
+```kotlin
+// Class with a primary constructor including values for 'name' and 'age'
+class Person(val name: String = "John", var age: Int = 30) {
+    // Property 'description' initialized using the primary constructor
+    val description: String = "Name: $name, Age: $age"
+}
+```
+
+In this example, the primary constructor `(val name: String = "John", var age: Int = 30)` initializes the class's properties 
+and sets their initial values. Then, `description` is an additional string property, 
+initialized in the class body using the constructor parameters.
+
+Here's the same example with an [object created](#creating-objects-of-classes):
+
+```kotlin
+// Class with a primary constructor including values for 'name' and 'age'
+class Person(
+    val name: String = "John",
+    var age: Int = 30
+) {
+    // Property 'description' initialized using the primary constructor
+    val description: String = "Name: $name, Age: $age"
+}
+
+fun main() {
+    // Create an instance of the 'Person' class
+    val person = Person()
+    // Accesses the 'description' property from the 'person' object
+    println(person.description)
+    // Name: John, Age: 30
+}
+```
+{kotlin-runnable="true"}
+
+### Initializer blocks
+
+The primary constructor can't contain runnable code. Its purpose is to initialize classes and set class 
+properties. For any logic or behavior that requires actual code execution, you need to place it somewhere else.
+
+If you want to run some code during [object creation](#creating-objects-of-classes), use _initializer blocks_ inside the class body.
+
+Declare initializer blocks with the `init` keyword followed by curly braces `{}`. Write any code that you want to run
+within the curly braces:
+
+```kotlin
+// Class with a primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    init {
+        // Initializer block executed when an object is created
+        println("Person created: $name, age $age")
+    }
+}
+```
+
+In the example above, after the class declaration, there's an initializer block (`init{}`) containing runnable code 
+to print a message that uses the class properties.
+
+Considering the same code example, if you [create an object](#creating-objects-of-classes) of the `Person` class
+by passing values for `name` and `age`, the output is:
+
+```kotlin
+//sampleStart
+// Class with a primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    init {
+        // Initializer block executed when an object is created
+        println("Person created: $name, age $age.")
+        // Person created: John, age 30.
+    }
+}
+
+fun main() {
+    // Creates an object of the 'Person' class
+    Person("John", 30)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+You can add as many initializer blocks as you need, and they will be executed in the order in which they appear in the class body, 
+interleaved with property initializers (`init{}`).
+
+Here's an example of a class with two initializer blocks containing logic that runs when the [object is created](#creating-objects-of-classes). In this case,
+the object of the `Person` class is created by passing values for `name` and `age`:
+
+```kotlin
+//sampleStart
+// Class with a primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    // First initializer block
+    init {
+        // Runs first when an object is created
+        println("Person created: $name, age $age.")
+        // Person created: John, age 30.
+    }
+
+    // Second initializer block
+    init {
+        // Runs after the first initializer block
+        if (age < 18) {
+            println("$name is a minor.")
+        } else {
+            println("$name is an adult.")
+            // John is an adult.
+        }
+    }
+}
+
+fun main() {
+    // Creates an object of the 'Person' class
+    Person("John", 30)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+You can use the primary constructor parameters in the initializer blocks. For example, in the code above, the first and second initializers use
+the `name` and `age` parameters from the primary constructor. 
 
 ### Secondary constructors
 
-A class can also declare _secondary constructors_, which are prefixed with `constructor`:
+In Kotlin, a secondary constructor is an additional constructor that a class can have beyond its primary constructor. 
+Secondary constructors allow you to provide different ways to initialize an object when the primary constructor alone 
+isn't enough.
+
+Secondary constructors are useful when interacting with Java code that requires traditional constructors or if
+you need multiple ways to initialize a class.
+
+To declare a secondary constructor, use the `constructor`
+keyword inside the class body. After the keyword, add the constructor parameters within parenthesis. 
+Then, use curly braces with the constructor logic inside:
 
 ```kotlin
-class Person(val pets: MutableList<Pet> = mutableListOf())
+// Class header with primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    // Class body with secondary constructor
+    constructor(name: String) : this(name, 0) { /*...*/ }
+}
+```
 
-class Pet {
-    constructor(owner: Person) {
-        owner.pets.add(this) // adds this pet to the list of its owner's pets
+See secondary constructors in action. The following code snippet [creates an object](#creating-objects-of-classes)
+`Person("Bob")` based on the secondary constructor. The secondary constructor assigns a default age (`8`) and takes the 
+`name` parameter. The constructor logic prints a message when the secondary constructor is called:
+
+```kotlin
+// Class header with primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    // Secondary constructor
+    constructor(name: String) : this(name, 8) {
+        println("$name created with default age: $age")
+        // Bob created with default age: 8
+    }
+}
+
+fun main() {
+    // Creates an object using the secondary constructor, which sets 'age' to 8
+    Person("Bob")
+}
+```
+{kotlin-runnable="true"}
+
+In the code above, the secondary constructor calls or delegates to the primary constructor via the [`this` keyword](this-expressions.md), 
+passing `name` and the default value of `age`.
+
+In Kotlin, secondary constructors must delegate to the primary constructor. This delegation ensures that all primary 
+constructor initialization logic is executed before any secondary constructor logic runs. 
+
+Constructor delegation can be: 
+* **Direct**, where the secondary constructor calls the primary constructor immediately.
+* **Indirect**, where one secondary constructor calls another, which in turn delegates to the primary constructor.
+
+> For further information about delegation via superclass constructor, see [Calling the superclass implementation](inheritance.md#calling-the-superclass-implementation).
+>
+{style="note"}
+
+Here's an example demonstrating how direct and indirect delegation works:
+
+```kotlin
+// Class header with primary constructor that initializes 'name' and 'age'
+class Person(val name: String, var age: Int) {
+    // Secondary constructor with direct delegation to the primary constructor
+    constructor(name: String) : this(name, 0) {
+        println("Person created with default name: $name")
+    }
+
+    // Secondary constructor with indirect delegation: 'this("Bob")' -> 'constructor(name: String)' -> primary constructor
+    constructor() : this("Bob") {
+        println("New person created with default name: $name")
     }
 }
 ```
 
-If the class has a primary constructor, each secondary constructor needs to delegate to the primary constructor, either
-directly or indirectly through another secondary constructor(s). Delegation to another constructor of the same class is
-done using the `this` keyword:
+The same code example with two [objects created](#creating-objects-of-classes), one for the direct delegation and the other for
+the indirect delegation, looks like this:
 
 ```kotlin
-class Person(val name: String) {
-    val children: MutableList<Person> = mutableListOf()
-    constructor(name: String, parent: Person) : this(name) {
-        parent.children.add(this)
+// Class header with primary constructor that initializes 'name' and 'age'
+class Person(
+    val name: String,
+    var age: Int
+) {
+    // Secondary constructor with direct delegation to the primary constructor
+    constructor(name: String) : this(name, 0) {
+        println("Person created with default age: $age and name: $name.")
+        // Person created with default age: 0 and name: Alice.
+        // Person created with default age: 0 and name: Bob.
+    }
+
+    // Secondary constructor with indirect delegation: 'this("Bob")' -> 'constructor(name: String)' -> primary constructor
+    constructor() : this("Bob") {
+        println("New person created with default age: $age and name: $name.")
+        // New person created with default age: 0 and name: Bob.
     }
 }
+
+fun main() {
+    // Creates an object based on the direct delegation
+    Person("Alice")
+
+    // Creates an object based on the indirect delegation
+    Person()
+}
 ```
+{kotlin-runnable="true"}
 
-Code in initializer blocks effectively becomes part of the primary constructor. Delegation to the primary constructor
-happens at the moment of access to the first statement of a secondary constructor, so the code in all initializer blocks
-and property initializers is executed before the body of the secondary constructor.
+In classes with initializer blocks (`init {}`), the code within these blocks becomes part of the primary constructor.
+When a secondary constructor is invoked, it delegates to the primary constructor first. As a result, all initializer blocks
+and property initializers run before the body of the secondary constructor. 
 
+Whenever an [object of a class is created](#creating-objects-of-classes), all initializer blocks run before any secondary constructor logic.
 Even if the class has no primary constructor, the delegation still happens implicitly, and the initializer blocks are
 still executed:
 
 ```kotlin
 //sampleStart
-class Constructors {
+// Class header with no primary constructor
+class Person {
+    // Initializer block executed when an object is created
     init {
-        println("Init block")
+        // Runs before the secondary constructor
+        println("Initializer block executed")
+        // Initializer block executed
     }
 
+    // Secondary constructor that takes an integer parameter
     constructor(i: Int) {
-        println("Constructor $i")
+        // Runs after the initializer block
+        println("Person $i is created")
+        // Person 1 created
     }
 }
-//sampleEnd
 
 fun main() {
-    Constructors(1)
+    // Creates an object of the 'Person' class using the secondary constructor
+    Person(1)
 }
+//sampleEnd
 ```
 {kotlin-runnable="true"}
 
-If a non-abstract class does not declare any constructors (primary or secondary), it will have a generated primary constructor
-with no arguments. The visibility of the constructor will be public.
+When the `Person(1)` object is created, the initializer block (`init {}`) executes first, followed by the secondary constructor's print statement.
 
+If there's a class that is a [non-abstract class](#abstract-classes) and does not declare any constructors (primary or secondary), it will have a generated primary constructor
+with no arguments:
+
+```kotlin
+// Class with no explicit constructors
+class Person {
+    // No primary or secondary constructors declared
+}
+
+fun main() {
+    // Creates an object of the 'Person' class using the automatically generated primary constructor
+    val person = Person()
+}
+```
+
+The visibility of the constructor will be public, meaning it can be accessed from anywhere.
 If you don't want your class to have a public constructor, declare an empty primary constructor with non-default visibility:
 
 ```kotlin
-class DontCreateMe private constructor() { /*...*/ }
+class Person private constructor() { /*...*/ }
 ```
 
-> On the JVM, if all of the primary constructor parameters have default values, the compiler will generate an additional parameterless constructor which will use the default values. This makes it easier to use Kotlin with libraries such as Jackson or JPA that create class instances through parameterless constructors.
+> On the JVM, if all primary constructor parameters have default values, the compiler generates 
+> an additional parameterless constructor that uses those default values. 
+> 
+> This makes it easier to use Kotlin with libraries such as [Jackson](https://github.com/FasterXML/jackson) 
+> or [Spring Data JPA](https://spring.io/projects/spring-data-jpa), which create class objects through 
+> parameterless constructors.
 >
+> In the following example, Kotlin automatically generates a parameterless constructor `Person()` that uses the default value `""`:
+> 
 > ```kotlin
-> class Customer(val customerName: String = "")
+> class Person(val personName: String = "")
 > ```
 >
 {style="note"}
 
-## Creating instances of classes
+## Creating objects of classes
 
-To create an instance of a class, call the constructor as if it were a regular function. You can assign the created instance to a [variable](basic-syntax.md#variables):
+Like in other object-oriented programming languages, a class in Kotlin serves as a blueprint or template 
+that defines the properties (attributes) and behaviors (functions) of objects. When you create an instance 
+of a class, you are creating a concrete object based on that blueprint.
+
+To create an object of a class, use the class name followed by parentheses (`()`), similar to calling a function:
 
 ```kotlin
-val invoice = Invoice()
-
-val customer = Customer("Joe Smith")
+// Creates an object of the 'Person' class
+val person1 = Person()
 ```
 
-> Kotlin does not have a `new` keyword.
+In Kotlin, you can create objects:
+
+* **Without arguments** (`()`): creates an object using a constructor with default values.
+* **With arguments** (`("value")`): creates an object by passing specific values to the constructor.
+
+As you can see, you can assign the created object to a mutable (`var`) or immutable (`val`) [variable](basic-syntax.md#variables):
+
+```kotlin
+// Creates an object by passing a specific value and assigns it to a mutable variable
+var person2 = Person("Joe")
+
+// Creates an object using the default constructor's value and assigns it to an immutable variable
+val person1 = Person()
+```
+
+Typically, you create objects within the `main()` function because it serves as the entry point of your Kotlin program. 
+
+When the program runs, execution starts from the `main()` function, so this is where you instantiate objects to interact with them.
+When creating objects of a class, you're invoking constructors (for example, `Person()` or `Person("Joe")`) directly inside
+the `main()` function.
+
+Let's look at an example of how object creation works.
+
+The following code defines a `Person` class with a property for storing a name. 
+When an object of the class is created, an initialization block runs to print a message. The example demonstrates 
+creating an object with both the default constructor's value and a specific value:
+
+```kotlin
+// Class header with primary constructor that initializes 'name' with a default value
+class Person(val name: String = "Sebastian")
+
+fun main() {
+    // Creates an object using the default constructor's value
+    val person1 = Person()
+    println("Person object created: ${person1.name}")
+    // Person object created: Sebastian
+
+    // Creates an object by passing a specific value
+    val person2 = Person("Joe")
+    println("Person object created: ${person2.name}")
+    // Person object created: Joe
+}
+```
+{kotlin-runnable="true"}
+
+> In Kotlin, unlike other object-oriented programming languages like Java, 
+> there is no need for the `new` keyword when creating objects of a class.
 >
 {style="note"}
 
-The process of creating instances of nested, inner, and anonymous inner classes is described in [Nested classes](nested-classes.md).
+Additionally, you can create objects inside another function and call that function from `main()`.
 
-## Class members
-
-Classes can contain:
-
-* [Constructors and initializer blocks](classes.md#constructors)
-* [Functions](functions.md)
-* [Properties](properties.md)
-* [Nested and inner classes](nested-classes.md)
-* [Object declarations](object-declarations.md)
+For information about creating objects of nested, inner, and anonymous inner classes,
+see the [Nested classes](nested-classes.md) section.
 
 ## Inheritance
 
-Classes can be derived from each other and form inheritance hierarchies.
-[Learn more about inheritance in Kotlin](inheritance.md).
+Class inheritance in Kotlin allows you to create a new class (derived class) from an existing class (base class), 
+inheriting its properties and functions while adding or modifying behavior.
+
+> For detailed information about inheritance hierarchies and the use of the `open` keyword, see the [Inheritance](inheritance.md) section.
+> 
+{style="note"}
 
 ## Abstract classes
 
-A class may be declared `abstract`, along with some or all of its members.
-An abstract member does not have an implementation in its class.
-You don't need to annotate abstract classes or functions with `open`.
+An abstract class in Kotlin is a class that cannot be instantiated directly and is designed to be inherited 
+by other classes. 
+
+An abstract class can define abstract properties and functions, which must be implemented 
+by subclasses.
+
+You can declare an abstract class using the `abstract` modifier:
 
 ```kotlin
-abstract class Polygon {
-    abstract fun draw()
+abstract class Person
+```
+
+Like any other class, abstract classes can also have constructors. 
+These constructors initialize class properties and enforce required parameters for subclasses:
+
+```kotlin
+abstract class Person(val name: String, val age: Int)
+```
+
+An abstract class can have both abstract and non-abstract members (properties and functions). 
+To declare a member as abstract, you must use the `abstract` keyword explicitly.
+
+Abstract members do not have an implementation in the abstract class. The `override`
+keyword ensures that a subclass provides an actual implementation for an abstract function:
+
+```kotlin
+// Abstract class with a primary constructor that defines 'name' and 'age'
+abstract class Person(
+    val name: String,
+    val age: Int
+) {
+    // Abstract member (doesn't provide default implementation, must be implemented by subclasses)
+    abstract fun introduce()
+
+    // Non-abstract member (has a default implementation)
+    fun greet() {
+        println("Hello, my name is $name.")
+        // Hello, my name is Alice.
+    }
 }
 
-class Rectangle : Polygon() {
-    override fun draw() {
-        // draw the rectangle
+// Subclass that provides an implementation for the abstract member
+class Student(
+    name: String,
+    age: Int,
+    val school: String
+) : Person(name, age) {
+    override fun introduce() {
+        println("I am $name, $age years old, and I study at $school.")
+        // I am Alice, 20 years old, and I study at Engineering University.
+    }
+}
+
+fun main() {
+    // Creates an object of the Student class
+    val student = Student("Alice", 20, "Engineering University")
+    // Calls the non-abstract member
+    student.greet()
+    // Calls the overridden abstract member
+    student.introduce()
+}
+```
+{kotlin-runnable="true"}
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited. 
+
+Similarly, you can override a non-abstract `open` member with an abstract one:
+
+## Open keyword
+
+In Kotlin, the `open` keyword indicates that a class or a member (function or property) can be overridden in subclasses. 
+By default, Kotlin classes and their members are _final_, meaning they cannot be inherited or overridden unless explicitly marked as `open`:
+
+```kotlin
+// Base class with the 'open' keyword to allow inheritance
+open class Person(
+    val name: String
+) {
+    // Open function that can be overridden in a subclass
+    open fun introduce() {
+        println("Hello, my name is $name.")
+    }
+}
+
+// Subclass inheriting from 'Person' and overriding the 'introduce()' function
+class Student(
+    name: String,
+    val school: String
+) : Person(name) {
+    override fun introduce() {
+        println("Hi, I'm $name, and I study at $school.")
     }
 }
 ```
 
-You can override a non-abstract `open` member with an abstract one.
+In addition, you need to add the `open` keyword to every property
+or function that can be overridden.
+
+If you override a member of a base class, the overriding member
+will also be open by default. If you want to change this and forbid the subclasses of your
+class from overriding your implementation, you can explicitly mark the overriding
+member as `final`:
 
 ```kotlin
-open class Polygon {
-    open fun draw() {
-        // some default polygon drawing method
+// Base class with the 'open' keyword to allow inheritance
+open class Person(val name: String) {
+    // Open function that can be overridden in a subclass
+    open fun introduce() {
+        println("Hello, my name is $name.")
     }
 }
 
-abstract class WildShape : Polygon() {
-    // Classes that inherit WildShape need to provide their own
-    // draw method instead of using the default on Polygon
-    abstract override fun draw()
+// Subclass inheriting from 'Person' and overriding the 'introduce()' function
+class Student(name: String, val school: String) : Person(name) {
+    // The 'final' keyword prevents further overriding in subclasses
+    final override fun introduce() {
+        println("Hi, I'm $name, and I study at $school.")
+    }
 }
+```
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited. 
+
+Similarly, you can override a non-abstract `open` member with an abstract one:
+
+```kotlin
+// Abstract class with a primary constructor that defines 'name'
+abstract class Person(val name: String) {
+  // Non-abstract open member that allows overriding
+  open fun greet() {
+    println("Hello, my name is $name.")
+  }
+
+  // Abstract member
+  abstract fun introduce()
+}
+
+// Subclass that overrides the 'greet()' function with an abstract one
+abstract class Student(name: String, val school: String) : Person(name) {
+  // Overrides the non-abstract 'greet()' function with an abstract one
+  abstract override fun greet()
+
+  // Overrides the abstract 'introduce()' function with an abstract one
+  override fun introduce() {
+    println("I am $name, and I study at $school.")
+  }
+}
+
 ```
 
 ## Companion objects
 
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
+In Kotlin, a [companion object](object-declarations.md#companion-objects) is a type of object declaration 
+that allows you to access its members using the class name without needing an object from the class.
 
-Even more specifically, if you declare a [companion object](object-declarations.md#companion-objects) inside your class,
+If you need to write a function that can be called without having a class object but still needs access 
+to the internals of the class, you can define it inside `companion object` within that class:
+
+```kotlin
+// Class with a primary constructor that defines the 'name' property
+class Person(
+    val name: String
+) {
+    // Class body with a companion object
+    companion object {
+        fun createAnonymous() = Person("Anonymous")
+    }
+}
+
+fun main() {
+    // Calls the function without creating an object of the class
+    val anonymous = Person.createAnonymous()
+    println(anonymous.name)
+    // Anonymous
+}
+```
+{kotlin-runnable="true"}
+
+If you declare `companion object` inside your class,
 you can access its members using only the class name as a qualifier.
+
+> [See further details about companion objects here](object-declarations.md#companion-objects).
+>
+{style="note"}


docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
 
-## Creating instances of classes
+## Creating objects of classes
 
-To create an instance of a class, call the constructor as if it were a regular function. You can assign the created instance to a [variable](basic-syntax.md#variables):
+Like in other object-oriented programming languages, a class in Kotlin serves as a blueprint or template 
+that defines the properties (attributes) and behaviors (functions) of objects. When you create an instance 
+of a class, you are creating a concrete object based on that blueprint.
+
+To create an object of a class, use the class name followed by parentheses (`()`), similar to calling a function:
```
ice-phoenix, line 489: In the previous paragraphs we were already creating objects of classes using this syntax. Maybe we should briefly introduce how to call a constructor before, and here we are explaining that in more detail?

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
+var person2 = Person("Joe")
+
+// Creates an object using the default constructor's value and assigns it to an immutable variable
+val person1 = Person()
 ```
 
-> Kotlin does not have a `new` keyword.
->
-{style="note"}
+Typically, you create objects within the `main()` function because it serves as the entry point of your Kotlin program.
```
ice-phoenix, line 511: This statement is not true, you create objects whenever you need to do that, and not only within the `main` function

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
+    final override fun introduce() {
+        println("Hi, I'm $name, and I study at $school.")
+    }
+}
+```
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited. 
+
+Similarly, you can override a non-abstract `open` member with an abstract one:
```
ice-phoenix, line 687: Tbh, I do not think this example adds anything useful for one reading the documentation. Such pattern of overriding `open` with `abstract` is very strange and should not be used unless there are very specific circumstances.

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
 ```
+{kotlin-runnable="true"}
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited. 
+
+Similarly, you can override a non-abstract `open` member with an abstract one:
 
-You can override a non-abstract `open` member with an abstract one.
+## Open keyword
```
ice-phoenix, line 631: I would say this section makes more sense in the `Inheritance` part of the documentation

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
 ## Companion objects
 
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
+In Kotlin, a [companion object](object-declarations.md#companion-objects) is a type of object declaration 
+that allows you to access its members using the class name without needing an object from the class.
 
-Even more specifically, if you declare a [companion object](object-declarations.md#companion-objects) inside your class,
+If you need to write a function that can be called without having a class object but still needs access 
+to the internals of the class, you can define it inside `companion object` within that class:
```
ice-phoenix, line 720: How do you define "internals of the class" here? A companion object can access its own internals, but class itself does not have any internals until you have created an instance of it.

ice-phoenix: Also, here the terminology begins being a bit confusing, as "class object" and "companion object" look very similar, but mean significantly different things

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
 [//]: # (title: Classes)
 
-Classes in Kotlin are declared using the keyword `class`:
+Like other object-oriented languages, Kotlin uses _classes_ to encapsulate data and behavior 
+for reusable, structured code.
```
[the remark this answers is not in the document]
ice-phoenix: I would also prefer to see a very simple example in the beginning of "class -> ctor -> created object instance" which we then iteratively expand on, by adding more properties, functions, ctors, init blocks or other things.

ice-phoenix (changes requested):
My main question is: what are the things we want the reader to know, after they've finished reading the updated "Classes" page? Is it basic syntax for classes, details about constructors, details about initialization, smth else? Right now the section is a bit all over the place, I feel.

In other words, when we set to improve the page, what did we want to achieve?

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
+the `name` and `age` parameters from the primary constructor. 
 
 ### Secondary constructors
 
-A class can also declare _secondary constructors_, which are prefixed with `constructor`:
+In Kotlin, a secondary constructor is an additional constructor that a class can have beyond its primary constructor. 
+Secondary constructors allow you to provide different ways to initialize an object when the primary constructor alone 
+isn't enough.
+
+Secondary constructors are useful when interacting with Java code that requires traditional constructors or if
```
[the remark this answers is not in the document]
AlejandraPedroza: Is it better?

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
+fun main() {
+    // Creates an object of the 'Person' class
+    Person("John", 30)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+You can add as many initializer blocks as you need, and they will be executed in the order in which they appear in the class body, 
+interleaved with property initializers (`init{}`).
```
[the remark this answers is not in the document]
AlejandraPedroza: Initializer block, I placed it where it corresponds.

docs/topics/classes.md
AlejandraPedroza: Added it in primary constructor, right before the first runnable example of object creation.

docs/topics/classes.md
```diff
@@ -1,245 +1,747 @@
...
+    Person()
+}
+```
+{kotlin-runnable="true"}
+
+In classes with initializer blocks (`init {}`), the code within these blocks becomes part of the primary constructor.
+When a secondary constructor is invoked, it delegates to the primary constructor first. As a result, all initializer blocks
+and property initializers run before the body of the secondary constructor. 
+
+Whenever an [object of a class is created](#creating-objects-of-classes), all initializer blocks run before any secondary constructor logic.
```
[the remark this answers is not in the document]
AlejandraPedroza: Do you mean:
First the full code, clean, without inline output comments.
Then separately show what the program prints when executed.

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
-val invoice = Invoice()
+// Creates an object of the 'Person' class
+val person1 = Person()
+```
 
-val customer = Customer("Joe Smith")
+In Kotlin, you can create objects:
+
+* **Without arguments** (`()`): creates an object using a constructor with default values.
+* **With arguments** (`("value")`): creates an object by passing specific values to the constructor.
```
ice-phoenix, line 527:
```suggestion
* **With arguments** (`(value)`): creates an object by passing specific values to the constructor.
```

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+
+As you can see, you can assign the created object to a mutable (`var`) or immutable (`val`) [variable](basic-syntax.md#variables):
```
ice-phoenix, line 529:
```suggestion
As you can see, you can assign the created object to a mutable (`var`) or read-only (`val`) [variable](basic-syntax.md#variables):
```

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+
+```kotlin
+// Creates an object by passing a specific value and assigns it to a mutable variable
+var person2 = Person("Joe")
+
+// Creates an object using the default constructor's value and assigns it to an immutable variable
```
ice-phoenix, line 535:
```suggestion
// Creates an object using the default constructor's value and assigns it to a read-only variable
```

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+val person1 = Person()
 ```
 
-> Kotlin does not have a `new` keyword.
->
-{style="note"}
+You can create objects wherever you need them, inside the `main()` function, within other functions, or as part of a class.
```
ice-phoenix, line 539: What do you mean when you say `as part of a class`?

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
 
-> Kotlin does not have a `new` keyword.
->
-{style="note"}
+You can create objects wherever you need them, inside the `main()` function, within other functions, or as part of a class. 
 
-The process of creating instances of nested, inner, and anonymous inner classes is described in [Nested classes](nested-classes.md).
+When creating objects inside the `main()` function, when the program runs, execution starts from `main()`. 
+Here is where you instantiate objects to interact with them. When creating objects of a class, you're invoking constructors 
+(for example, `Person()` or `Person("Joe")`) directly inside the `main()` function.
```
ice-phoenix, line 543: I feel that, for some reason, here we begin talking about smth unrelated to classes, i.e., about the purposes of the `main` function, when it is invoked, etc. Maybe we should create a separate page about it then, and simply reference it here?

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+
+    // Creates an object by passing a specific value
+    val person2 = Person("Joe")
+    println("Person object created: ${person2.name}")
+    // Person object created: Joe
+}
+```
+{kotlin-runnable="true"}
+
+> In Kotlin, unlike other object-oriented programming languages like Java,
```
ice-phoenix, line 569: Nit: "unlike ... like" reads weird

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+}
+```
+{kotlin-runnable="true"}
+
+> In Kotlin, unlike other object-oriented programming languages like Java, 
+> there is no need for the `new` keyword when creating objects of a class.
+>
+{style="note"}
+
+Additionally, you can create objects inside another function and call that function from `main()`.
```
ice-phoenix, line 574: I think we've already said that before, and repeating the same fact again feels weird.

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+
+```kotlin
+abstract class Person(val name: String, val age: Int)
 ```
 
-You can override a non-abstract `open` member with an abstract one.
+An abstract class can have both abstract and non-abstract members (properties and functions). 
+To declare a member as abstract, you must use the `abstract` keyword explicitly.
+
+Abstract members do not have an implementation in the abstract class. The `override`
```
ice-phoenix, line 612: I think we should add a sentence about how one provides an implementation by writing an override in an inheritor. That will also bridge us to the `override` keyword which is used for overrides.

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
 ```kotlin
-open class Polygon {
-    open fun draw() {
-        // some default polygon drawing method
+// Abstract class with a primary constructor that defines 'name' and 'age'
+abstract class Person(
+    val name: String,
+    val age: Int
+) {
+    // Abstract member (doesn't provide default implementation, must be implemented by subclasses)
```
ice-phoenix, line 621:
```suggestion
    // Abstract member (doesn't provide implementation, must be implemented by subclasses)
```

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+    abstract fun introduce()
+
+    // Non-abstract member (has a default implementation)
```
ice-phoenix, line 624:
```suggestion
    // Non-abstract member (has an implementation)
```

docs/topics/classes.md
ice-phoenix: It's a bit more precise to talk about base (not default) implementations, as "default" term is already used in Kotlin for other things.

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
 
 ## Companion objects
 
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
+In Kotlin, a [companion object](object-declarations.md#companion-objects) is a type of object declaration 
+that allows you to access its members using the class name without needing an object from the class.
+
+If you need to write a function that can be called without creating an object of a class but still needs access to the 
+class's companion object members, you can define it inside an [object declaration](object-declarations.md) within the class:
```
ice-phoenix, line 665:
```suggestion
class's companion object members, you can define it inside an companion [object declaration](object-declarations.md) within the class:
```

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+[For more details about the `open` keyword, see Inheritance](inheritance.md#open-keyword).
 
 ## Companion objects
 
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
+In Kotlin, a [companion object](object-declarations.md#companion-objects) is a type of object declaration 
+that allows you to access its members using the class name without needing an object from the class.
+
+If you need to write a function that can be called without creating an object of a class but still needs access to the
```
ice-phoenix, line 664: I would rephrase this to smth like "If you need to write a function that can be called without creating an object of a class but is still logically connected to the class (such as a factory function), you can define ..."

ice-phoenix (changes requested):
Thanks for the updates! In general, I like the new version.

I narrowed down my main concern to the struggle between "being succinct and to the point targeting existing developers" in the old version vs "being very detailed and explaining very basic things targeting novice developers" in the new one. I don't know how to find a good balance here besides taking a look at the actual feedback of users and what they are missing.

Additionally, I think it's important to remember that, while all and every feedback is important, a lot of people who were completely satisfied with the current succinct and to-the-point version didn't leave their positive feedback. We might "over-optimize" for the smaller part of our target audience which is more vocal and more novice, and they want very basic things explained in detail.

> Note: feel free to disregard my general feedback and only incorporate the technical notes, as you are the person who definitely knows more about how to write good technical documentation. If you believe it should be done in a specific way, then that's what we should do.

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+```kotlin
+class Person(val name: String, var age: Int) { /*...*/ }
+```
+
+In this example, the `Person` class has a primary constructor with two properties: `name` and `age`. 
+The `name` property is declared as read-only using `val`, while the `age` property is declared as mutable using `var`, meaning that
+the value of `age` can change later on.
+Both properties are initialized directly when an object of `Person` [is created](#creating-objects-of-classes).
+
+By declaring class properties in the primary constructor using `val` or `var`, you make them accessible to
```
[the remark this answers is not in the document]
ice-phoenix: Also, after taking a look at the feedback, if we still want to make this page more introductory, I think we should add an explicit example demonstrating the difference between

```
class Person(name: String)
// vs
class Person(val name: String)
```

aka the difference between primary constructor parameters and primary constructor properties.

docs/topics/classes.md
AlejandraPedroza: Changed it for "inside another class"

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
 [//]: # (title: Classes)
 
-Classes in Kotlin are declared using the keyword `class`:
+Like other object-oriented languages, Kotlin uses _classes_ to encapsulate data (properties) and behavior (functions)
+for reusable, structured code.
+
+Classes are blueprints or templates for objects, which you 
+create via [constructors](#constructors-and-initializer-blocks). When you [create an instance of a class](#creating-objects-of-classes), you are creating
```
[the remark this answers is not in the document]
AlejandraPedroza: The decision was made based on two reasons: 

- "Object" appears in the book Kotlin in Action indistinctly from "instance

- We received repetitive feedback about confusing because of the word "instance"

However, after reading your technical suggestion, the final use is "object"

docs/topics/classes.md
```diff
@@ -1,245 +1,692 @@
...
+    println(person.age)
+    // 0
+}
 ```
+{kotlin-runnable="true"}
+
+In this example, the `Person` class is defined with a primary constructor that declares a read-only property 
+`name` of type `String`. In the body, the class includes a mutable `age` property initialized to `0`. 
+
+This class represents a person with a fixed name and an age you can update after
```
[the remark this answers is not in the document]
AlejandraPedroza: The broader goal is to raise the overall quality of our content: better examples help beginners progress to the intermediate level and help experienced developers solve typical problems. The aim is to make the content more accessible and useful for both groups.

But you are right, each snippet was followed by explanations that repeated what the code and code comments already showed. This made it overdocumented. So the explanations were removed when applicable.

AlejandraPedroza:
@ice-phoenix 

Thanks for your detailed review!
The aim is to make the content more accessible and useful for both groups. Better examples and explanations will help beginners progress to the intermediate level and help experienced developers solve typical problems. 
Also, when improving pages, we prioritize practical, widely useful information. We take user feedback into account, but assess very specific cases with low impact. Based on this approach, we work with SMEs to balance what and how we document: tech writers as content experts, SMEs as technical experts.
I appreciate your technical suggestions. I applied them overall. 
From the content side, I reviewed why the page felt over-documented, and produced a clearer, more concise version.

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+  * [Functions](functions.md)
+  * [Properties](properties.md)
+  * [Nested and inner classes](nested-classes.md)
+  * [Object declarations](object-declarations.md)
+
+Both the class header and class body can be minimal. 
+If the class does not have a body, you can omit the curly braces `{}`:
+
+```kotlin
+// Class with name and age properties declared in the primary constructor, but without a body
```
ice-phoenix: Nit: on my kinda wide monitor I have to scroll to read the full line, maybe make such long comments shorter and split between multiple lines?

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+If the class does not have a body, you can omit the curly braces `{}`:
+
+```kotlin
+// Class with name and age properties declared in the primary constructor, but without a body 
+class Person(val name: String, var age: Int)
+```
+
+Here's an example of a class with a header, body, and an [instance created](#creating-instances-of-classes) from it:
 
 ```kotlin
```
ice-phoenix, line 43: For some reason, this code block is formatted a bit differently from other ones, maybe it's worth taking a look.

image

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
 ```
+{kotlin-runnable="true"}
 
-## Constructors
+## Creating instances of classes
 
-A class in Kotlin has a _primary constructor_ and possibly one or more _secondary constructors_. The primary constructor
-is declared in the class header, and it goes after the class name and optional type parameters.
+An instance is created
+when you use the class as a blueprint to build a real thing to work with in your program.
```
ice-phoenix: real thing -> real object?

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+
+To create an instance of a class, use the class name followed by parentheses (`()`), similar to calling a function:
```
ice-phoenix: Maybe add a reference to the Functions section of our docs?

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+
+The primary constructor can declare parameters as properties. Use the `val` keyword before the argument name to declare a read-only property 
+and the `var` keyword for a mutable property.
+
+```kotlin
+class Person(val name: String, var age: Int) { /*...*/ }
+```
+
+These constructor parameter properties are stored as part of the instance and are accessible from outside the class. 
+It's also possible to declare primary constructor parameters only, not stored in the instance, and available only within the class:
```
ice-phoenix: "...available only within the class body"

This is correct enough and ambiguous enough, so that we do not need to actually properly explain where exactly primary constructor parameters are and aren't available.

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+class Person2(val name: String) {
+    fun greet() {
+        println("Hello, $name")
+    }
+}
+
+// Primary constructor parameter only (not stored as a property)
+class Person1(name: String) {
+    fun greet() {
+        println("Hello, $name")
```
ice-phoenix: This code does not compile :crying_cat_face:

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
 
-Much like regular properties, properties declared in the primary constructor can be mutable (`var`) or read-only (`val`).
+### Initializer blocks
 
-Plain constructor parameters (that are not properties) are accessible in:
-* The class header.
-* Initialized properties within the class body.
-* Initializer blocks.
+The primary constructor can't contain runnable code. Its purpose is to initialize the class and set its
+properties. If you want to run some code during [instance creation](#creating-instances-of-classes), use _initializer blocks_ inside the class body.
```
ice-phoenix:
```suggestion
The purpose of the primary constructor is to initialize the class and set its
properties, and usually you can achieve this without needing to run complex code blocks.
If you do need to do this in a primary constructor during [instance creation](#creating-instances-of-classes), use _initializer blocks_ inside the class body.
```

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+}
 ```
+{kotlin-runnable="true"}
+
+> The expression `age.toIntOrNull() ?: 0` converts the `age` string to an integer. For more information about this expression, see [Null safety](null-safety.md).
+>
+{style="tip"}
+
+In the code above, the secondary constructor delegates to the primary constructor via the `this` keyword, 
+passing `name` and the default `age` value.
```
ice-phoenix: "...and the converted to integer `age` value"

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+    // Creates an instance of the Person class
+    val person = Person()
+    // Accesses the description property
+    println(person.description)
+    // Name: John, Age: 30
+}
+```
+{kotlin-runnable="true"}
+
+Additionally, you can use a [trailing comma](coding-conventions.md#trailing-commas) in constructor declarations:
```
ice-phoenix: Maybe "As with functions, you can ..."?

docs/topics/classes.md
```diff
@@ -1,263 +1,560 @@
...
+    println(person.name)
+    // Alice
+    println(person.age)
+    // 0
+}
 ```
+{kotlin-runnable="true"}
 
-## Constructors
+## Creating instances of classes
```
ice-phoenix: Maybe introduce constructors first and then talk about calling them?

docs/topics/classes.md
ice-phoenix: Also, maybe add that quite often `init` blocks are used for data validation, e.g., by calling `require` function?

```
class Person(val age: Int) {
    init { require(age > 0, "age must be positive") }
}
```

ice-phoenix (changes requested):
Thank you for the update, it looks much more focused and to the point, compared to the previous version! I've left some comments, but overall I don't have a lot to add.

In all the places where we are technically not entirely correct, we are ambiguous enough and pragmatic enough; to fix these places, we would need to introduce too many additional things and details, which is not really appropriate for this page, I feel.

docs/topics/classes.md
AlejandraPedroza: This is the issue why the previous version seemed too packed with information and code samples. In that version, all the runnable codes demonstrating instance creation required showed a concept that hadn’t been introduced yet, so it required explanation any time it appeared. In the current structure, creating an instance is explained before all the code samples that demonstrate it.

Additionally, as you can see from the last paragraph, the content is arranged to avoid unnecessary breaks. As soon as constructors are mentioned, there’s a smooth transition to the corresponding explanation. I tried different structures, and this one provides the most natural flow while avoiding repetition.

docs/topics/classes.md
AlejandraPedroza: This mismatch occurs between runnable and non-runnable code in all of our docs. I confirmed with the web team that this is the expected behavior while they fix it.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-Classes in Kotlin are declared using the keyword `class`:
+Like other object-oriented languages, Kotlin uses _classes_ to encapsulate data (properties) and behavior (functions)
+for reusable, structured code.
+
+Classes are blueprints or templates for objects, which you 
+create via [constructors](#constructors-and-initializer-blocks). 
+When you [create an instance of a class](#creating-instances-of-classes), you are creating 
+a concrete object based on that blueprint.
+
+Kotlin offers a concise syntax for declaring classes. To declare a class, use the `class` keyword
```
sarahhaggarty:
```suggestion
Kotlin offers concise syntax for declaring classes. To declare a class, use the `class` keyword 
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+  * [Primary constructor](#primary-constructor) (optional)
+* **Class body** (optional), surrounded by curly braces `{}`, and including **class members** such as:
+  * [Secondary constructors](#secondary-constructors)
+  * [Initializer blocks](#initializer-blocks)
+  * [Functions](functions.md)
+  * [Properties](properties.md)
+  * [Nested and inner classes](nested-classes.md)
+  * [Object declarations](object-declarations.md)
+
+Both the class header and class body can be minimal.
```
sarahhaggarty: What do you mean by "minimal" here?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+* **Class body** (optional), surrounded by curly braces `{}`, and including **class members** such as:
+  * [Secondary constructors](#secondary-constructors)
+  * [Initializer blocks](#initializer-blocks)
+  * [Functions](functions.md)
+  * [Properties](properties.md)
+  * [Nested and inner classes](nested-classes.md)
+  * [Object declarations](object-declarations.md)
+
+Both the class header and class body can be minimal. 
+If the class does not have a body, you can omit the curly braces `{}`:
```
sarahhaggarty:
```suggestion
If the class doesn't have a body, you can omit the curly braces `{}`:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 ```
 
-## Constructors
+Here's an example of a class with a header, body, and an [instance created](#creating-instances-of-classes) from it:
 
-A class in Kotlin has a _primary constructor_ and possibly one or more _secondary constructors_. The primary constructor
-is declared in the class header, and it goes after the class name and optional type parameters.
+```kotlin
+// Person class with a primary constructor 
+// that nitializes the name property
```
sarahhaggarty:
```suggestion
// that initializes the name property
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+If the class does not have a body, you can omit the curly braces `{}`:
 
 ```kotlin
-class Empty
+// Class with primary constructor, but without a body 
+class Person(val name: String, var age: Int)
 ```
 
-## Constructors
+Here's an example of a class with a header, body, and an [instance created](#creating-instances-of-classes) from it:
```
sarahhaggarty:
```suggestion
Here's an example that declares a class with a header and body, then [creates an instance](#creating-instances-of-classes) from it:
```
This doesn't sound natural to me. I've suggested an alternative. I think it's because the "class with a header and body" is the important part. It separates it from the instance creation which is a different level of abstraction.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+```kotlin
+// Person class with a primary constructor 
+// that nitializes the name property
+class Person(val name: String) {
+    // Class body with age property
+    var age: Int = 0
+}
+
+fun main() {
+    // Creates an instance of Person class by calling the constructor
```
sarahhaggarty:
```suggestion
    // Creates an instance of the Person class by calling the constructor
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+}
+```
+{kotlin-runnable="true"}
+
+## Creating instances of classes
+
+An instance is created
+when you use the class as a blueprint to build a real object to work with in your program.
+
+To create an instance of a class, use the class name followed by parentheses (`()`), similar to calling a [function](functions.md):
```
sarahhaggarty:
```suggestion
To create an instance of a class, use the class name followed by parentheses `()`, similar to calling a [function](functions.md):
```
I don't think double parentheses are necessary here :)

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-class Person constructor(firstName: String) { /*...*/ }
+// Creates an instance of the Person class
+val person1 = Person()
 ```
 
-If the primary constructor does not have any annotations or visibility modifiers, the `constructor` keyword can be omitted:
+In Kotlin, you can create instances:
+
+* **Without arguments** (`()`): creates an instance using the default values, if they are declared in the class.
+* **With arguments** (`(value)`): creates an instance by passing specific values.
```
sarahhaggarty:
```suggestion
* **Without arguments** (`Person()`): creates an instance using the default values, if they are declared in the class.
* **With arguments** (`Person(value)`): creates an instance by passing specific values.
```
I think it's clearer if we add a class name to demonstrate where the arguments are when creating an instance.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+You can assign the created instance to a mutable (`var`) or read-only (`val`) [variable](basic-syntax.md#variables):
```
sarahhaggarty, line 81: Out of curosity, is there any benefit to assigning it to a read-only variable? Does it make it safer if you don't want the instance to be modified after the instance is created? 🤔

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+> In Kotlin, unlike other object-oriented programming languages such as Java,
+> there is no need for the `new` keyword when creating class instances.
+>
+{style="note"}
+
+For information about creating instances of nested, inner, and anonymous inner classes,
+see the [Nested classes](nested-classes.md) section.
+
+When creating class instances, you call one of its constructors and provide arguments for the constructor parameters.
```
sarahhaggarty: I think it makes sense to move this into the next section as an intro and transition sentence. Maybe something like this:
"When you create a class instance, you use one of its constructors. A class in Kotlin can have..."
WDYT?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+For more details, see the next section about [Constructors and initializer blocks](#constructors-and-initializer-blocks).
```
sarahhaggarty:
```suggestion
```
I don't think this is necessary as the link is right next to the section itself :)

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+## Constructors and initializer blocks
+
+A class in Kotlin can have a [_primary constructor_](#primary-constructor) and one or more [_secondary constructors_](#secondary-constructors).
+
+The primary constructor is the main way to initialize a class; you declare it in the class header.
+A secondary constructor provides additional initialization logic; you declare it in the class body.
+
+Both primary and secondary constructors are optional, and at least one constructor should be declared in a class.
```
sarahhaggarty:
```suggestion
Both primary and secondary constructors are optional, but a class must have at least one constructor.
```
Can we make this stronger with "must"?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+### Primary constructor
+
+The primary constructor sets up the initial state of an instance when [it is created](#creating-instances-of-classes).
```
sarahhaggarty:
```suggestion
The primary constructor sets up the initial state of an instance when [it's created](#creating-instances-of-classes).
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+To declare a primary constructor, place it in the class header after the class name:
+
+```kotlin
+class Person constructor(name: String) { /*...*/ }
+```
+
+If the primary constructor does not have any annotations or [visibility modifiers](visibility-modifiers.md#constructors),
```
sarahhaggarty:
```suggestion
If the primary constructor doesn't have any annotations or [visibility modifiers](visibility-modifiers.md#constructors), 
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+When creating class instances, you call one of its constructors and provide arguments for the constructor parameters.
+For more details, see the next section about [Constructors and initializer blocks](#constructors-and-initializer-blocks).
+
+## Constructors and initializer blocks
+
+A class in Kotlin can have a [_primary constructor_](#primary-constructor) and one or more [_secondary constructors_](#secondary-constructors).
+
+The primary constructor is the main way to initialize a class; you declare it in the class header.
+A secondary constructor provides additional initialization logic; you declare it in the class body.
```
sarahhaggarty:
```suggestion
The primary constructor is the main way to initialize a class. You declare it in the class header.
A secondary constructor provides additional initialization logic. You declare it in the class body.
```
I think we should try to avoid using semicolons unless they're really necessary :)

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-    
-    init {
-        println("First initializer block that prints $name")
+// Class header with a primary constructor 
+// that initializes name with a default value
+class Person(val name: String = "Sebastian")
+
+fun main() {
+    // Creates an instance using the default constructor's value
+    val person1 = Person()
```
sarahhaggarty:
```suggestion
    val anonymousUser = Person()
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+    // Creates an instance by passing a specific value
+    val person2 = Person("Joe")
+
+    // Accesses the instances' name property
+    println(person1.name)
```
sarahhaggarty:
```suggestion
    println(anonymousUser.name)
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+// Class header with a primary constructor 
+// that initializes name with a default value
+class Person(val name: String = "Sebastian")
+
+fun main() {
+    // Creates an instance using the default constructor's value
+    val person1 = Person()
+
+    // Creates an instance by passing a specific value
+    val person2 = Person("Joe")
```
sarahhaggarty:
```suggestion
    val namedUser = Person("Joe")
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+    // Accesses the instances' name property
+    println(person1.name)
+    // Sebastian
+    println(person2.name)
```
sarahhaggarty:
```suggestion
    println(namedUser.name)
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    val person = Person("Alice")
+
+    // Accesses the instance's properties
+    println(person.name)
+    // Alice
+    println(person.age)
+    // 0
+}
+```
+{kotlin-runnable="true"}
```
sarahhaggarty: Can you add IDs to the runnable samples? Then we'll be able to see in GA which ones have been run.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+* **Without arguments** (`()`): creates an instance using the default values, if they are declared in the class.
+* **With arguments** (`(value)`): creates an instance by passing specific values.
+
+You can assign the created instance to a mutable (`var`) or read-only (`val`) [variable](basic-syntax.md#variables):
 
 ```kotlin
-class Person(firstName: String) { /*...*/ }
+// Creates an instance using the default value 
+// and assigns it to a mutable variable
+var person1 = Person()
```
sarahhaggarty:
```suggestion
var anonymousUser = Person()
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+// Creates an instance by passing a specific value 
+// and assigns it to a read-only variable
+val person2 = Person("Joe")
```
sarahhaggarty:
```suggestion
val namedUser = Person("Joe")
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+An instance is created
+when you use the class as a blueprint to build a real object to work with in your program.
+
+To create an instance of a class, use the class name followed by parentheses (`()`), similar to calling a [function](functions.md):
 
 ```kotlin
-class Person constructor(firstName: String) { /*...*/ }
+// Creates an instance of the Person class
+val person1 = Person()
```
sarahhaggarty:
```suggestion
val anonymousUser = Person()
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+```kotlin
+class Person(val name: String, var age: Int) { /*...*/ }
+```
+
+These constructor parameter properties are stored as part of the instance and are accessible from outside the class. 
+It's also possible to declare primary constructor parameters only, not stored in the instance, and available only within the class body:
+
+```kotlin
+// Primary constructor parameter that is also a property
+class Person2(val name: String) {
```
sarahhaggarty:
```suggestion
class PersonWithProperty(val name: String) {
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    fun greet() {
+        println("Hello, $name")
     }
+}
+
+// Primary constructor parameter only (not stored as a property)
+class Person1(name: String) {
```
sarahhaggarty:
```suggestion
class PersonWithAssignment(name: String) {
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    // Must be assigned to a property to be usable later
+    val n: String = name
```
sarahhaggarty:
```suggestion
    val displayName: String = name
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
     
-    val secondProperty = "Second property: ${name.length}".also(::println)
-    
-    init {
-        println("Second initializer block that prints ${name.length}")
+    fun greet() {
+        println("Hello, $n")
```
sarahhaggarty:
```suggestion
        println("Hello, $displayName")
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
     }
 }
-//sampleEnd
+```
 
-fun main() {
-    InitOrderDemo("hello")
+Properties declared in the primary constructor are accessible to
+[member functions](functions.md) of a class:
```
sarahhaggarty:
```suggestion
Properties declared in the primary constructor are accessible by
[member functions](functions.md) of the class:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+The primary constructor can declare parameters as properties. Use the `val` keyword before the argument name to declare a read-only property 
+and the `var` keyword for a mutable property:
+
+```kotlin
+class Person(val name: String, var age: Int) { /*...*/ }
+```
+
+These constructor parameter properties are stored as part of the instance and are accessible from outside the class. 
+It's also possible to declare primary constructor parameters only, not stored in the instance, and available only within the class body:
```
sarahhaggarty: Can you make this sentence a separate paragraph? I think it would be great to explain explicitly that constructor parameters that aren't properties don't have a `val` or `var`. So this one sentence would be broken up into at least two before the example.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
 ```kotlin
-class Customer(name: String) {
-    val customerKey = name.uppercase()
-}
+class Person(val name: String = "John", var age: Int = 30) { /*...*/ }
 ```
 
-Kotlin has a concise syntax for declaring properties and initializing them from the primary constructor:
+If no value is passed during [instance creation](#creating-instances-of-classes), the property uses the default value:
```
sarahhaggarty:
```suggestion
If no value is passed to the constructor during [instance creation](#creating-instances-of-classes), properties use their default value:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    // Creates an instance using default values
+    val person = Person()
+    println("Name: ${person.name}, Age: ${person.age}")
+    // Name: John, Age: 30
+}
 ```
+{kotlin-runnable="true"}
 
-Such declarations can also include default values of the class properties:
+Use the primary constructor parameters to initialize additional class properties directly in the class body:
```
sarahhaggarty:
```suggestion
You can use primary constructor parameters to initialize additional class properties directly in the class body:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 ```kotlin
-class Person(val firstName: String, val lastName: String, var isEmployed: Boolean = true)
+// Class with a primary constructor 
+// including default values for name and age
+class Person(
+    val name: String = "John",
+    var age: Int = 30
+) {
+    // description property initialized in the class body 
+    // using the primary constructor parameters
```
sarahhaggarty:
```suggestion
    // Initializes the description property
    // from the primary constructor parameters
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    val person = Person()
+    // Accesses the description property
+    println(person.description)
+    // Name: John, Age: 30
+}
 ```
+{kotlin-runnable="true"}
 
-You can use a [trailing comma](coding-conventions.md#trailing-commas) when you declare class properties:
+As with functions, you can use [trailing comma](coding-conventions.md#trailing-commas) in constructor declarations:
```
sarahhaggarty:
```suggestion
As with functions, you can use a [trailing comma](coding-conventions.md#trailing-commas) in constructor declarations:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
-Much like regular properties, properties declared in the primary constructor can be mutable (`var`) or read-only (`val`).
+### Initializer blocks
 
-Plain constructor parameters (that are not properties) are accessible in:
-* The class header.
-* Initialized properties within the class body.
-* Initializer blocks.
+The purpose of the primary constructor is to initialize the class and set its
+properties. In most cases, this initialization doesn’t require complex code.
```
sarahhaggarty:
```suggestion
properties. In most cases, this initialization doesn't require complex code.
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
-For example:
+If you need to run complex code blocks when the primary constructor is executed during [instance creation](#creating-instances-of-classes), 
+use _initializer blocks_ inside the class body.
```
sarahhaggarty: Maybe for improved flow:

> The primary constructor initializes the class and sets its properties. In most cases, you can handle this with simple code.
>
> If you need to perform more complex operations during instance creation, place that logic in _initializer blocks_ inside the class body. These blocks run when the primary constructor executes.

WDYT?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-    val perimeter = 2 * width + 2 * height
+Declare initializer blocks with the `init` keyword followed by curly braces `{}`. 
+Write within the curly braces any code that you want to run during initialization:
 
+```kotlin
+// Class with a primary constructor that initializes name and age
+class Person(val name: String, var age: Int) {
     init {
-        println("Rectangle created with width = $width and height = $height")
+        // Initializer block executed when an instance is created
```
sarahhaggarty:
```suggestion
        // Initializer block runs when an instance is created
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+fun main() {
+    // Creates an instance of the Person class
+    Person("John", 30)
+}
 ```
+{kotlin-runnable="true"}
 
-If the constructor has annotations or visibility modifiers, the `constructor` keyword is required and the modifiers go before it:
+Add as many initializer blocks (`init {}`) as you need. They run in the order in which they appear in the class body, 
+interleaved with property initializers:
```
sarahhaggarty: What does "interleaved with property initializers" mean here? 🤔

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+fun main() {
+    // Creates an instance of the Person class
+    Person("John", 30)
+}
+//sampleEnd
 ```
+{kotlin-runnable="true"}
+
+You can use the primary constructor parameters in the initializer blocks. For example, in the code above, the first and second initializers use
```
sarahhaggarty:
```suggestion
You can use primary constructor parameters in initializer blocks. For example, in the code above, the first and second initializers use
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+the `name` and `age` parameters from the primary constructor. 
+
+A common use of `init` blocks is data validation. For example, by calling the [`require` function](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/require.html):
```
sarahhaggarty:
```suggestion
A common use case for `init` blocks is data validation. For example, by calling the [`require` function](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/require.html):
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
-A class can also declare _secondary constructors_, which are prefixed with `constructor`:
+In Kotlin, a secondary constructor is an additional constructor that a class can have beyond its primary constructor. 
+Secondary constructors allow you to provide different ways to initialize an instance when the primary constructor alone 
+isn't enough.
 
-```kotlin
-class Person(val pets: MutableList<Pet> = mutableListOf())
+Secondary constructors are useful for [Java interoperability](java-to-kotlin-interop.md) and when
+you need multiple ways to initialize a class.
```
sarahhaggarty: There's a bit of duplication about the fact that you can use secondary constructors to initialize a class in multiple ways. How about:

> In Kotlin, secondary constructors are additional constructors that a class can have beyond its primary constructor. Secondary constructors are useful when you need multiple ways to initialize a class or for [Java interoperability](java-to-kotlin-interop.md).

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+To declare a secondary constructor, use the `constructor`
+keyword inside the class body with the constructor parameters within parentheses. Use curly braces with the constructor logic inside:
```
sarahhaggarty:
```suggestion
keyword inside the class body with the constructor parameters within parentheses `()`. Add the constructor logic within curly braces `{}`:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
-class Pet {
-    constructor(owner: Person) {
-        owner.pets.add(this) // adds this pet to the list of its owner's pets
+```kotlin
+// Class header with a primary constructor that initializes name and age
+class Person(val name: String, var age: Int) {
+
+    // Secondary constructor that takes a String age
+    // and converts it to an integer
```
sarahhaggarty:
```suggestion
    // Secondary constructor that takes age as a
    // String and converts it to an Int
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    constructor(name: String, age: String) : this(name, age.toIntOrNull() ?: 0) {
+        println("$name created with converted age: $age")
+        // Bob created with converted age: 8
     }
 }
+
+fun main() {
+    // Uses the secondary constructor with a String age
```
sarahhaggarty:
```suggestion
    // Uses the secondary constructor with age as a String
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    Person("Bob", "8")
+}
 ```
+{kotlin-runnable="true"}
+
+> The expression `age.toIntOrNull() ?: 0` converts the `age` string to an integer. For more information about this expression, see [Null safety](null-safety.md).
```
sarahhaggarty:
```suggestion
> The expression `age.toIntOrNull() ?: 0` uses the Elvis operator. For more information, see [Null safety](null-safety.md#elvis-operator).
```
To avoid mentioning about the expression twice, shall we narrow down the tip to be about the Elvis operator?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+Given that secondary constructors delegate to the primary constructor first, all initializer blocks
+and property initializers run before the body of the secondary constructor. Even if the class has no primary constructor, 
+the delegation still happens implicitly:
 
 ```kotlin
-//sampleStart
-class Constructors {
+// Class header with no primary constructor
+class Person {
+    // Initializer block executed when an instance is created
```
sarahhaggarty:
```suggestion
    // Initializer block runs when an instance is created
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    // Creates an instance of the Person class
+    Person(1)
 }
 ```
 {kotlin-runnable="true"}
 
-If a non-abstract class does not declare any constructors (primary or secondary), it will have a generated primary constructor
-with no arguments. The visibility of the constructor will be public.
+If there's a class that does not declare any constructors (primary or secondary), it will have an implicit primary constructor
+with no parameters:
```
sarahhaggarty:
```suggestion
Classes that don't declare any constructors (primary or secondary), have an implicit primary constructor
with no parameters:
```
To use present tense and make it a little more concise.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+## Inheritance
 
-val customer = Customer("Joe Smith")
-```
+Class inheritance in Kotlin allows you to create a new class (derived class) from an existing class (base class), 
+inheriting its properties and functions while adding or modifying behavior.
 
-> Kotlin does not have a `new` keyword.
->
+> For detailed information about inheritance hierarchies and the use of the `open` keyword, see the [Inheritance](inheritance.md) section.
```
sarahhaggarty:
```suggestion
> For detailed information about inheritance hierarchies and how to use the `open` keyword, see the [Inheritance](inheritance.md) section.
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+> 
 {style="note"}
 
-The process of creating instances of nested, inner, and anonymous inner classes is described in [Nested classes](nested-classes.md).
+## Abstract classes
 
-## Class members
+An abstract class in Kotlin is a class that cannot be instantiated directly and is designed to be inherited 
+by other classes.
```
sarahhaggarty:
```suggestion
In Kotlin, abstract classes are classes that can't be instantiated directly. They are designed to be inherited by other classes which define their actual behavior. This behavior is called an _implementation_.
```
I think the text misses an explanation of what we mean when we talk about implementation. I suggested an alternative to strengthen the narrative. Let me know what you think :)

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+by subclasses.
 
-* [Constructors and initializer blocks](#constructors)
-* [Functions](functions.md)
-* [Properties](properties.md)
-* [Nested and inner classes](nested-classes.md)
-* [Object declarations](object-declarations.md)
+Abstract classes can also have constructors. 
+These constructors initialize class properties and enforce required parameters for subclasses.
+Declare an abstract class using the `abstract` modifier:
```
sarahhaggarty:
```suggestion
Declare an abstract class using the `abstract` keyword:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-## Abstract classes
+An abstract class can have both abstract and non-abstract members (properties and functions). 
+To declare a member as abstract, you must use the `abstract` keyword explicitly. 
 
-A class may be declared `abstract`, along with some or all of its members.
-An abstract member does not have an implementation in its class.
-You don't need to annotate abstract classes or functions with `open`.
+Abstract members do not have an implementation 
+in the abstract class. To provide one, you declare an `override` function 
+in the subclass or inheriting class:
```
sarahhaggarty:
```suggestion
Abstract members don't have an implementation
in the abstract class. You define the implementation in a subclass or inheriting class with an `override` function or property:
```
"To provide one" sounds a bit formal so I suggested an alternative.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 ```kotlin
-abstract class Polygon {
-    abstract fun draw()
+// Abstract class with a primary constructor that declares name and age
+abstract class Person(
+    val name: String,
+    val age: Int
+) {
+    // Abstract member 
+    // Doesn't provide implementation, must be implemented by subclasses
```
sarahhaggarty:
```suggestion
    // Doesn't have an implementation. 
    // It must be implemented by subclasses
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    // Calls the non-abstract member
+    student.greet()
+    // Calls the overridden abstract member
+    student.introduce()
+}
 ```
+{kotlin-runnable="true"}
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited.
```
sarahhaggarty:
```suggestion
You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly inheritable by default. 
```
I think it's more understandable to talk about whether a class/function is inheritable or not rather than whether it is open or not.

Also, maybe this part would be better after the explanation to use the `abstract` keyword for abstract members?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+
+You don't need to annotate abstract classes or functions with the `open` keyword because they are implicitly 
+open and designed to be inherited. 
+
+[For more details about the `open` keyword, see Inheritance](inheritance.md#open-keyword).
+
+## Companion objects
+
+In Kotlin, a [companion object](object-declarations.md#companion-objects) is a type of object declaration 
+that allows you to access its members using the class name without needing an object from the class.
```
sarahhaggarty:
```suggestion
In Kotlin, each class can have a [companion object](object-declarations.md#companion-objects). Companion objects are a type of object declaration 
that allows you to access its members using the class name without creating a class instance.
```
It's quite a long sentence so I suggest breaking it up.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
-
-Even more specifically, if you declare a [companion object](object-declarations.md#companion-objects) inside your class,
+If you declare `companion object` inside your class,
 you can access its members using only the class name as a qualifier.
+
+> [See further details about companion objects here](object-declarations.md#companion-objects).
+>
+{style="note"}
```
sarahhaggarty:
```suggestion
> For more information, see [Companion objects](object-declarations.md#companion-objects).
>
{style="tip"}
```

To avoid saying "here".

docs/topics/inheritance.md
```diff
@@ -38,6 +40,57 @@ class MyView : View {
...
+```kotlin
+// Base class with the open keyword to allow inheritance
+open class Person(val name: String) {
+    // Open function that can be overridden in a subclass
+    open fun introduce() {
+        println("Hello, my name is $name.")
+    }
+}
+
+// Subclass inheriting from Person and overriding the introduce() function
```
sarahhaggarty:
```suggestion
//  A subclass that inherits from Person and overrides the introduce() function
```

docs/topics/inheritance.md
```diff
@@ -38,6 +40,57 @@ class MyView : View {
...
+class Student(name: String, val school: String) : Person(name) {
+    // The final keyword prevents further overriding in subclasses
```
sarahhaggarty:
```suggestion
    // The final keyword prevents further overrides in subclasses
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
-The class declaration consists of the class name, the class header (specifying its type parameters, the primary constructor, 
-and some other things), and the class body surrounded by curly braces. Both the header and the body are optional; if the
-class has no body, the curly braces can be omitted.
+The class declaration consists of: 
+* **Class header**, including but not limited to:
+  * `class` keyword
+  * Class name
+  * Type parameters (if any)
+  * [Primary constructor](#primary-constructor) (optional)
```
sarahhaggarty, line 23:
```suggestion
  * The `class` keyword
  * A class name
  * Type parameters (if any)
  * A [primary constructor](#primary-constructor) (optional)
```
I think it reads nicer with articles :)

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 
 ```kotlin
 class Person { /*...*/ }
 ```
 
-The class declaration consists of the class name, the class header (specifying its type parameters, the primary constructor, 
-and some other things), and the class body surrounded by curly braces. Both the header and the body are optional; if the
-class has no body, the curly braces can be omitted.
+The class declaration consists of: 
+* **Class header**, including but not limited to:
```
sarahhaggarty, line 19:
```suggestion
* A **class header**, including but not limited to:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+  * `class` keyword
+  * Class name
+  * Type parameters (if any)
+  * [Primary constructor](#primary-constructor) (optional)
+* **Class body** (optional), surrounded by curly braces `{}`, and including **class members** such as:
```
sarahhaggarty, line 24:
```suggestion
* A **class body** (optional), surrounded by curly braces `{}`, and including **class members** such as:
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+    // Accesses the instances' name property
+    println(person1.name)
+    // Sebastian
+    println(person2.name)
+    // Joe
+}
+```
+{kotlin-runnable="true"}
+
+> In Kotlin, unlike other object-oriented programming languages such as Java,
```
sarahhaggarty:
```suggestion
> In Kotlin, unlike other object-oriented programming languages,
```
To make it more language-agnostic :)

docs/topics/classes.md
sarahhaggarty: I think it sounds better without the imperative in this case, as we're explaining an option.

docs/topics/classes.md
sarahhaggarty: To make it present tense.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
 ```
+{kotlin-runnable="true"}
 
-## Companion objects
-
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
-
-Even more specifically, if you declare a [companion object](object-declarations.md#companion-objects) inside your class,
+If you declare `companion object` inside your class,
```
sarahhaggarty:
```suggestion
If you declare a `companion object` inside your class,
```
Does "companion object" need to be in inlined code format here?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-    Constructors(1)
+    // Creates an instance of the Person class
+    Person(1)
 }
 ```
 {kotlin-runnable="true"}
 
-If a non-abstract class does not declare any constructors (primary or secondary), it will have a generated primary constructor
-with no arguments. The visibility of the constructor will be public.
+If there's a class that does not declare any constructors (primary or secondary), it will have an implicit primary constructor
```
sarahhaggarty: I think it would be good to put the rest of this content under a separate header. It doesn't really fit under "Secondary constructors". Maybe "Without constructors"?

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-        println("Rectangle created with width = $width and height = $height")
+        // Initializer block executed when an instance is created
+        println("Person created: $name, age $age.")
+        // Person created: John, age 30.
     }
 }
+
+fun main() {
+    // Creates an instance of the Person class
+    Person("John", 30)
```
sarahhaggarty, line 282:
```suggestion
    Person("John", 30)
    // Person created: John, age 30.
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
-        owner.pets.add(this) // adds this pet to the list of its owner's pets
+```kotlin
+// Class header with a primary constructor that initializes name and age
+class Person(val name: String, var age: Int) {
+
+    // Secondary constructor that takes a String age
+    // and converts it to an integer
+    constructor(name: String, age: String) : this(name, age.toIntOrNull() ?: 0) {
+        println("$name created with converted age: $age")
+        // Bob created with converted age: 8
```
sarahhaggarty:
```suggestion
```

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
     }
 }
+
+fun main() {
+    // Uses the secondary constructor with a String age
+    Person("Bob", "8")
```
sarahhaggarty, line 358:
```suggestion
    Person("Bob", "8")
    // Bob created with converted age: 8
```
I think it makes more sense to add the printed statement here for the constructor.

docs/topics/classes.md
```diff
@@ -1,263 +1,587 @@
...
+// Class header with a primary constructor that initializes name and age
+class Person(
+    val name: String,
+    var age: Int
+) {
+    // Secondary constructor with direct delegation 
+    // to the primary constructor
+    constructor(name: String) : this(name, 0) {
+        println("Person created with default age: $age and name: $name.")
+        // Person created with default age: 0 and name: Alice.
```
sarahhaggarty: Can you move the code comments to where they will appear when triggered by the main function?

sarahhaggarty (changes requested):
Huge effort Ale! 🚀  It's definitely looking much better than the original. Thanks for the hard work!
Please take a look at my suggestions. 
There's some for improving names in code samples as well as adding code comments for printed statements within the main() function of samples.
I made a couple minor structural suggestions and asked a couple of questions for clarification.

docs/topics/classes.md
AlejandraPedroza: It means that the class header and the class body can be reduced to their simplest possible form, containing only what is strictly necessary. Class header: just the class name. Class body: just a few required members, or even completely absent. 

Do you think it's clearer if we explain it a bit more?

docs/topics/classes.md
AlejandraPedroza: By declaring an instance with `val`, its reference is read-only and can not be reassigned. However, the object itself is not immutable, and its properties can still be modified, for example, if these properties are declared as `var` within the class.

docs/topics/classes.md
AlejandraPedroza: It means that the code is executed in the same top-to-bottom order as it is written in the class, mixing property initializers and init blocks.

docs/topics/classes.md
AlejandraPedroza: I tested the version with articles, but I think they makes the list a bit harder to grasp the info. Since the articles aren’t consistent across all items, and the list is presented as bullet points, adding them actually makes the content harder to scan. I’d prefer to keep the current style without articles for clarity and consistency.

docs/topics/classes.md
sarahhaggarty: Got it! I think it sounds more natural to say "can be kept minimal". An even more natural version would be "You can keep both the class header and body to a bare minimum." But it isn't as concise and I'm not sure how it is for non-native speakers. What do you think?

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
+
+The primary constructor sets up the initial state of an instance when [it's created](#creating-instances-of-classes).
+
+To declare a primary constructor, place it in the class header after the class name:
+
+```kotlin
+class Person constructor(name: String) { /*...*/ }
+```
+
+If the primary constructor doesn't have any annotations or [visibility modifiers](visibility-modifiers.md#constructors),
```
sarahhaggarty:
```suggestion
If the primary constructor doesn't have any [annotations](annotations.md) or [visibility modifiers](visibility-modifiers.md#constructors), 
```

docs/topics/classes.md
sarahhaggarty: Ah, I see! Would "along with" be a possible alternative instead? I think that would make it clearer 🤔

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
-## Abstract classes
+You don't need to annotate abstract classes or functions with the `open`
+keyword because they are implicitly inheritable by default.
+For more details about the `open` keyword, see [Inheritance](inheritance.md#open-keyword).
 
-A class may be declared `abstract`, along with some or all of its members.
-An abstract member does not have an implementation in its class.
-You don't need to annotate abstract classes or functions with `open`.
+Abstract members don't have an implementation
+in the abstract class.
```
sarahhaggarty:
```suggestion
in the abstract class.
```

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
+    // Accesses the instance's properties
+    println(person.name)
+    // Alice
+    println(person.age)
+    // 0
+}
+```
+{kotlin-runnable="true" id="class-with-header-and-body"}
+
+## Creating instances of classes
```
sarahhaggarty:
```suggestion
## Creating instances
```
Nitpick! What do you think about making this header shorter?

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
 
-val customer = Customer("Joe Smith")
-```
+Class inheritance in Kotlin allows you to create a new class (derived class) from an existing class (base class), 
+inheriting its properties and functions while adding or modifying behavior.
 
-> Kotlin does not have a `new` keyword.
->
+> For detailed information about inheritance hierarchies and how to use of the `open` keyword, see the [Inheritance](inheritance.md) section.
+>
```
sarahhaggarty:
```suggestion
For detailed information about inheritance hierarchies and how to use of the `open` keyword, see the [Inheritance](inheritance.md) section.
```
To make it consistent with the other sections where the references aren't in callouts.

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
-val customer = Customer("Joe Smith")
-```
+Class inheritance in Kotlin allows you to create a new class (derived class) from an existing class (base class), 
+inheriting its properties and functions while adding or modifying behavior.
 
-> Kotlin does not have a `new` keyword.
->
+> For detailed information about inheritance hierarchies and how to use of the `open` keyword, see the [Inheritance](inheritance.md) section.
+> 
 {style="note"}
```
sarahhaggarty:
```suggestion
```

docs/topics/classes.md
```diff
@@ -1,263 +1,595 @@
...
-If you need to write a function that can be called without having a class instance but that needs access to the internals
-of a class (such as a factory method), you can write it as a member of an [object declaration](object-declarations.md) inside that class.
-
-Even more specifically, if you declare a [companion object](object-declarations.md#companion-objects) inside your class,
+If you declare a companion object inside your class,
 you can access its members using only the class name as a qualifier.
+
+> For more information, see [Companion objects](object-declarations.md#companion-objects).
+>
+{style="tip"}
```
sarahhaggarty:
```suggestion
For more information, see [Companion objects](object-declarations.md#companion-objects).
```
To make it consistent with the other sections where the references aren't in callouts.

docs/topics/classes.md
sarahhaggarty: Thanks! I checked with dev advocates and it is safer to assign to a `val` by default but I think this can be added to the Basic syntax page and doesn't need to be added here.

sarahhaggarty (approved):
Awesome work! This is looking really good now. I only have a few very minor suggestions left :) Please take a look and then LGTM! 🚀

PR #4787, opened 2025-04-02
Reference correct argument name in Grouping doc

Meridiam:
Change doc to correctly say that in the `groupBy` example, the function passed via the `valueTransform` argument is used to transform the values to uppercase, not the function passed via the `keySelector` argument.


daniCsorbaJB (approved):
Nice catch! Thanks @Meridiam

PR #4792, opened 2025-04-07
Correct minAndroidGradleVersion for Kotlin v2.1.20

3flex:
The minimum version is still 7.3.1 in v2.1.20:

https://github.com/JetBrains/kotlin/blob/v2.1.20/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/diagnostics/AgpCompatibilityCheck.kt#L16


Tapchicoma (approved):

daniCsorbaJB (approved):
Nice catch! Thanks @3flex

PR #4802, opened 2025-04-11
update: updating coroutines landing page for coroutines restructuring

daniCsorbaJB:
This update involves restructuring the Coroutines overview into a landing page for [KT-76566](https://youtrack.jetbrains.com/issue/KT-76566/DocsCoroutines-Rewrite-Kotlin-coroutines-overview-page)

### DIFF
--- a/docs/topics/coroutines-overview.md
+++ b/docs/topics/coroutines-overview.md
@@ -35,7 +35,7 @@ To launch new coroutines, use coroutine builders like [`.launch()`](https://kotl
 These builders are extension functions on [`CoroutineScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/),
 which defines the coroutine’s lifecycle and provides the coroutine context.
 
-You can learn more about these builders in [Coroutine basics](coroutine-basics.md) and [Composing suspend functions](coroutines-and-channels.md).
+You can learn more about these builders in [Coroutine basics](coroutines-basics.md) and [Composing suspend functions](coroutines-and-channels.md).
 
 ### Coroutine context and behavior
 


docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,126 @@
...
+which includes tools for launching coroutines, handling concurrency, working with asynchronous streams, and more.
 
-New to Kotlin? Take a look at the [Getting started](getting-started.md) page.
+If you're new to coroutines in Kotlin, start with the [Coroutine basics](coroutines-basics.md) guide before diving into more complex topics.
+It introduces the key concepts of suspending functions, coroutine builders, and structured concurrency through simple examples.
 
-### Documentation
+> Check out these sample projects to see how coroutines are used in practice: 
+>
+> * [kotlinx.coroutines examples and sources](https://github.com/Kotlin/kotlin-coroutines/tree/master/examples)
```
danil-pavlov, line 21: This repo is archived. Is there a similar examples collection in `https://github.com/Kotlin/kotlinx.coroutines`?

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,126 @@
...
+
+### Suspending functions and coroutine builders
+
+Coroutines in Kotlin are built on suspending functions, which allow code to pause and resume without blocking a thread.
+The `suspend` keyword marks functions that can perform long-running operations asynchronously.
+
+To launch new coroutines, you can use coroutine builders like [`launch()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) and [`async()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/async.html).
+These builders let you run tasks concurrently, return results, or enforce structured lifecycles.
+
+You can learn more about these builders in [Coroutine basics](coroutine-basics.md) and [Composing suspend functions](coroutines-and-channels.md).
```
danil-pavlov, line 93: Maybe composing-suspending-functions.md would be better for this link instead of coroutines-and-channels.md? And this tutorial could be featured inside Asynchronous flow next to channels?

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,126 @@
...
+To safely manage shared mutable state, the `kotlinx.coroutines` library provides constructs like [`Mutex`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) and atomic variables.
+Learn more in [Shared mutable state and concurrency](shared-mutable-state-and-concurrency.md).
+
+## What's next
+
+* Learn the fundamentals of coroutines, suspending functions, and builders in the [Coroutine basics guide](coroutines-basics.md).
+* Explore how to combine suspending functions and build coroutine pipelines in [Composing suspending functions](coroutine-context-and-dispatchers.md).
+* Learn how to [debug coroutines](debug-coroutines-with-idea.md) using the built-in tools in IntelliJ IDEA.
+* For flow-specific debugging, see [Debug Kotlin Flow using IntelliJ IDEA - tutorial](debug-flow-with-idea.md).
+* See how to write unit tests for coroutine-based code on Android in [Testing Kotlin coroutines on Android](https://developer.android.com/kotlin/coroutines/test).
```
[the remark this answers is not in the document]
danil-pavlov: Btw, regarding Android materials, do you think we should focus on testing for Android here? the original page had a couple of other related links
* [Best practices for coroutines in Android](https://developer.android.com/kotlin/coroutines/coroutines-best-practices)
* [Additional Android resources for Kotlin coroutines and flow](https://developer.android.com/kotlin/coroutines/additional-resources)

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,126 @@
...
+
+For direct communication between coroutines, use [`Channel`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/), which allows one coroutine to send values to another.
+See [Channels](channels.md) for more information.
+
+When multiple coroutines need to access or update the same data, they _share mutable state_.
+Without coordination, this can lead to race conditions, where operations interfere with each other in unpredictable ways.
+To safely manage shared mutable state, the `kotlinx.coroutines` library provides constructs like [`Mutex`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) and atomic variables.
+Learn more in [Shared mutable state and concurrency](shared-mutable-state-and-concurrency.md).
+
+## What's next
```
danil-pavlov, line 120: The original page has a couple more links that seem important, but can, of course, be mentioned on the Basics page instead, WDYT?

* [Guide to UI programming with coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/ui/coroutines-guide-ui.md)
* [Full kotlinx.coroutines API reference](https://kotlinlang.org/api/kotlinx.coroutines/)

danil-pavlov (changes requested):
Love the new intro! nice and clear

docs/topics/coroutines-overview.md
daniCsorbaJB: ah yes much better 👍 good idea

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,126 @@
...
+In Kotlin, asynchronous programming is built around _coroutines_, which let you write asynchronous code in a natural, sequential style using suspending functions.
+Coroutines are lightweight alternatives to threads.
+They can suspend without blocking system resources and are cheap to create, making them better suited for fine-grained concurrency.
 
-## How to start
+Most coroutine features are provided by the [`kotlinx.coroutines`](https://github.com/Kotlin/kotlinx.coroutines) library,
+which includes tools for launching coroutines, handling concurrency, working with asynchronous streams, and more.
 
-New to Kotlin? Take a look at the [Getting started](getting-started.md) page.
+If you're new to coroutines in Kotlin, start with the [Coroutine basics](coroutines-basics.md) guide before diving into more complex topics.
```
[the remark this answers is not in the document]
daniCsorbaJB: good idea - added it 👍

docs/topics/coroutines-overview.md
daniCsorbaJB: After asking the team I removed it and only left in the kotlinconf app

danil-pavlov (approved):

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,72 @@
...
 
-## Sample projects
+### Suspending functions and coroutine builders
 
-- [kotlinx.coroutines examples and sources](https://github.com/Kotlin/kotlin-coroutines/tree/master/examples)
-- [KotlinConf app](https://github.com/JetBrains/kotlinconf-app)
+Coroutines in Kotlin are built on suspending functions, which allow code to pause and resume without blocking a thread.
+The `suspend` keyword marks functions that can perform long-running operations asynchronously.
 
+To launch new coroutines, you can use coroutine builders like [`launch()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) and [`async()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/async.html).
```
dkhalanskyjb, line 34: Both `launch` and `async` have a `CoroutineScope` as the receiver, which is very important and has not been introduced by this point. Simply trying to call `launch` will not compile.

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,72 @@
...
 
-Asynchronous or non-blocking programming is an important part of the development landscape. When creating server-side,
-desktop, or mobile applications, it's important to provide an experience that is not only fluid from the user's perspective,
-but also scalable when needed.
+Applications often need to perform multiple tasks at the same time, such as responding to user input, loading data, or updating the screen.
+To support this, they rely on asynchronous programming, which allows operations to run independently without blocking each other.
 
-Kotlin solves this problem in a flexible way by providing [coroutine](https://en.wikipedia.org/wiki/Coroutine) support
-at the language level and delegating most of the functionality to libraries.
+The most common way to run tasks asynchronously is by using threads, which are independent paths of execution managed by the operating system.
```
dkhalanskyjb, line 6: This is *absolutely correct*, but unfortunately, also very misleading, due to how the meaning of the word "asynchronous" shifted over time. Multithreading is an example of asynchronous computations, no doubt, but today, in most cases, "asynchronous programming" means a very specific thing (which `kotlinx.coroutines` is an example of): along with the request for some thread to do something, also include code that should run after that operation has completed.

Typically, this looks something like this:
```kotlin
// this function returns immediately, and when `readFile` has finished,
// it also schedules `println` to be called
executeOnIoThread(
  whatToExecute = { readFile() },
  andThen = { contents -> println(contents) })
```
Kotlin's `suspend` functions allow writing something to the same effect as if it was linear code:
```kotlin
val contents = withContext(Dispatchers.IO) {
    readFile() // executes on the IO thread
}
// actually, `withContext` returns immediately ("suspends"),
// but after `readFile()` finishes, it schedules the subsequent code to be called:
println(contents)
```

So, this is what's called "asynchronous programming".

Instead of "rely on asynchronous programming", I'd say "concurrency", of which threads are indeed an unambiguous example.

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,72 @@
...
+Together, these elements make up the [_coroutine context_](coroutine-context-and-dispatchers.md), which is inherited by default from the coroutine's parent.
+This context forms a hierarchy that enables structured concurrency, where related coroutines can be [canceled](cancellation-and-timeouts.md) together or [handle exceptions](exception-handling.md) as a group.
+
+### Asynchronous flow and shared mutable state
+
+Kotlin Coroutines provides the [`Flow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/) interface for working with asynchronous streams of values.
+While suspending functions return only a single result, flows let you emit and collect multiple values over time without blocking a thread.
+For more information, see [Asynchronous flow](flow.md).
+
+For direct communication between coroutines, use [`Channel`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/), which allows one coroutine to send values to another.
```
dkhalanskyjb, line 56: > one coroutine ... to another

Channels are more flexible than that: they allow *many* coroutines to send values, and for *many* other coroutines to receive them.

Also, for the use case of one coroutine sending values and the other one receiving them, `Flow` is often a better fit.

* One coroutine sending values, one coroutine receiving them: a cold flow can do that in the case when the values are only sent when someone actively wants to receive them (which is the majority of all cases).
* Some coroutines sending values, multiple coroutines receiving *every emitted value* (a fanout of value emissions): this is what `SharedFlow` is for.
* Some coroutines sending values, some coroutines receiving values, *each value only going to one of the coroutines*: such sharing of work calls for a `Channel`.

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,72 @@
...
+See [Channels](channels.md) and the [Coroutines and channels tutorial](coroutines-and-channels.md) for more information.
+
+When multiple coroutines need to access or update the same data, they _share mutable state_.
+Without coordination, this can lead to race conditions, where operations interfere with each other in unpredictable ways.
+To safely manage shared mutable state, the `kotlinx.coroutines` library provides constructs like [`Mutex`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) and atomic variables.
```
dkhalanskyjb, line 61: `kotlinx.coroutines` does not provide atomic variables, the standard library does (though only recently; before that, `kotlinx.atomicfu` was the library providing them). Also, both mutexes and atomic variables are too low-level and error-prone for most scenarios. `StateFlow` is *very* often the better way of sharing mutable state between coroutines, I think it's better to highlight that instead.

dkhalanskyjb (changes requested):

docs/topics/coroutines-overview.md
daniCsorbaJB: 🤔 Great point :

What do you think about slightly restructuring the beginning of this section to something like this, would this be better 🤔 : 

Coroutines always run in a [`CoroutineScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/), which defines their lifecycle and provides the coroutine context.

Builder functions like `launch()` and `async()` are extension functions on `CoroutineScope`. When called, they create a coroutine and automatically add a set of elements that define how it behaves:

dkhalanskyjb: Looks good!

docs/topics/coroutines-overview.md
daniCsorbaJB: Thank you, that's very helpful 💯 

I've restructured the beginning a bit (moving away from asynchronous programming to concurrency/concurrently):

> Applications often need to perform multiple tasks at the same time, such as responding to user input, loading data, or updating the screen.
> To support this, they rely on concurrency, which allows operations to run independently without blocking each other.
> 
> The most common way to run tasks concurrently is by using threads, which are independent paths of execution managed by the operating system.
> However, threads are relatively heavy, and creating many of them can lead to performance issues.
> 
> To support efficient concurrency, Kotlin uses asynchronous programming built around _coroutines_, which let you write asynchronous code in a natural, sequential style using suspending functions.
> Coroutines are lightweight alternatives to threads.

docs/topics/coroutines-overview.md
daniCsorbaJB: Sounds good to me - let's drop those from the overview then 👍

daniCsorbaJB: Would something like this work? 🤔 
(In this case I'll comment out the link to the [Shared mutable state doc](https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#thread-safe-data-structures). I'll put it back after we updated that page as well)

> When multiple coroutines need to access or update the same data, they _share mutable state_.
> Without coordination, this can lead to race conditions, where operations interfere with each other in unpredictable ways.
> To safely manage shared mutable state, use [`StateFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/#) to wrap the shared data.
> Then you can update it from one coroutine and collect its latest value from others.

docs/topics/coroutines-overview.md
daniCsorbaJB: Hmm 🤔 
Based on that I'm thinking of something like this - what is your opinion? 

> ### Asynchronous flow and shared mutable state
> 
> Kotlin Coroutines provides the [`Flow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/) interface for working with asynchronous streams of values that one coroutine produces and another collects.
> This lets the collecting coroutine process values without blocking a thread.
> For more information, see [Asynchronous flow](flow.md).
> 
> Use [`Channel`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/) when you want several coroutines to send and receive values with each value delivered to only one coroutine.
> See [Channels](channels.md) and the [Coroutines and channels tutorial](coroutines-and-channels.md) for more information.
> 
> When multiple coroutines need to access or update the same data, they _share mutable state_.
> Without coordination, this can lead to race conditions, where operations interfere with each other in unpredictable ways.
> To safely manage shared mutable state, use [`StateFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/#) to wrap the shared data.
> Then, you can update it from one coroutine and collect its latest value from others.
> 
>

dkhalanskyjb: This works as a practical approximation, yes. We can keep it like this, and it will already be okay, or we could polish it some more.

This is a bit misleading:
> Use Flow when you want a coroutine to consume a stream of values produced by another coroutine.

It is not necessarily another coroutine. It could be the same one. It could be several ones. That said, it's often one coroutine, so if you do want to consume values produced by another coroutine, sure, `Flow` could fit your use case well.

Here's another potential improvement I see: I'm wary that `SharedFlow` is not mentioned, though: everything else is, so for completeness, adding just one more entity could be beneficial. On the other hand, this would become a bit unwieldy if it were longer.

Here's a crystal clear distinction between the ways of communicating across coroutines:

* A cold `Flow` allows exactly one coroutine to receive all values produced just for its sake (who exactly produces the values depends on the specific `Flow` construct you use).
* `Channel` allows several coroutines to send values, and several coroutines to receive values, with each value going to exactly a single coroutine.
* A hot `Flow` (`SharedFlow`) allows several coroutines to receive values, but as opposed to channels, the same value will be broadcast to (shared between) all the coroutines that subscribed to the updates.
* A `StateFlow` is a variation of `SharedFlow` (and so ensures that the upcoming values will be broadcast to everyone!) that is specifically tailored to sharing mutable state and setting and obtaining the latest value. Multiple coroutines can update the value.

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,75 @@
...
+### Coroutine context and behavior
+
+Launching a coroutine from a `CoroutineScope` creates a context that governs its execution.
+Builder functions like `.launch()` and `.async()` automatically create a set of elements that define how the coroutine behaves:
+
+* The [`Job`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/) interface tracks the coroutine's lifecycle and enables structured concurrency.
+* [`CoroutineDispatcher`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/) controls where the coroutine runs, such as on a background thread or the main thread in UI applications.
+* [`CoroutineExceptionHandler`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-exception-handler/) handles uncaught exceptions.
+
+Together, these elements make up the [_coroutine context_](coroutine-context-and-dispatchers.md), which is inherited by default from the coroutine's parent.
```
dkhalanskyjb: There are more potential context elements, though, these are only the main ones. To avoid misleading, some form of "and others" should be introduced.

docs/topics/coroutines-overview.md
```diff
@@ -1,39 +1,75 @@
...
+This context forms a hierarchy that enables structured concurrency, where related coroutines can be [canceled](cancellation-and-timeouts.md) together or [handle exceptions](exception-handling.md) as a group.
```
dkhalanskyjb, line 50:
```suggestion
This context forms a hierarchy that enables structured concurrency, where related coroutines can be [cancelled](cancellation-and-timeouts.md) together or [handle exceptions](exception-handling.md) as a group.
```
I know it's not the correct sort of English, but we already have two "L"s everywhere, so consistency demands this, I'm afraid.

dkhalanskyjb (approved):
Good to go as the first approximation. I hope we can revisit this once we work through the rest of the sections. After all, the landing page should reflect them properly.

docs/topics/coroutines-overview.md
daniCsorbaJB: hmm 🤔 I see what you mean - I see both used in the coroutine guides (although you are 100% right that cancelled is generally more used). That said, since the rest of kotlinlang.org/docs consistently uses American English, I’d suggest we align with that and update the spelling to canceled throughout as part of the rewrite.

dkhalanskyjb: The issue is, it's not just about the coroutines guide, it's what function names the library exposes as well:
* https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/is-cancelled.html
* https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.coroutines.cancellation/-cancellation-exception/
* https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/
* https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html

daniCsorbaJB: That is a conundrum — I'm not sure.. I'll double check with the team! 👍

PR #4805, opened 2025-04-14
Replace ./gradlew with gradle in native-get-started.md

namchuai:
gradlew is not there. after install proper gradle, should invoke gradle directly.


danil-pavlov:
Thanks for your suggestion! The rest of the documentation uses `.gradlew` commands, and we need to keep it consistent. We might change it as part of a bigger documentation update in the future. I'll close the PR now, but thanks for sharing this idea with us!

PR #4808, opened 2025-04-16
update: clarify null-handling behavior of==operator

daniCsorbaJB:
This update is intended to clarify the null-handling behavior of the `==` operator for [KT-74113](https://youtrack.jetbrains.com/issue/KT-74113)

### DIFF
--- a/docs/topics/operator-overloading.md
+++ b/docs/topics/operator-overloading.md
@@ -182,8 +182,12 @@ which can be overridden to provide custom equality check implementation. Any oth
 >
 {style="note"}
 
-The `==` operation is special: it is translated to a complex expression that screens for `null`'s.
-`null == null` is always true, and `x == null` for a non-null `x` is always false and won't invoke `x.equals()`.
+The `==` operation translates `a == b` to `a?.equals(b) ?: (b === null)`, which safely handles comparisons involving `null` values.
+The behavior of the comparison depends on the nullability of the values being compared:
+
+* If `a` is `null`, the comparison returns `true` only if `b` is also `null`, and no function is called. 
+* If `a` is a nullable type and isn't `null` at runtime, the compiler calls `a.equals(b)`, even if `b` is `null`.
+* If `a` is a non-nullable type and `b` is known to be `null`, the compiler replaces the comparison with `false` at compile time and skips the function call.
 
 ### Comparison operators
 


docs/topics/operator-overloading.md
```diff
@@ -176,14 +176,15 @@ For the assignment operations, for example `a += b`, the compiler performs the f
...
 
-> `===` and `!==` (identity checks) are not overloadable, so no conventions exist for them.
+> `===` and `!==` (identity checks) aren't overloadable, so no conventions exist for them.
 >
 {style="note"}
 
-The `==` operation is special: it is translated to a complex expression that screens for `null`'s.
-`null == null` is always true, and `x == null` for a non-null `x` is always false and won't invoke `x.equals()`.
+Kotlin calls `.equals()` when neither operand is the `null` literal and the comparison isn’t between two floating-point types.
+Otherwise, `===` is used for `null` literal checks, and non-null floating-point values are compared by numeric value.
```
sarahhaggarty: Instead of "the `null` literal" can we say "a `null` value"?
And "`===` is used to check for `null` values" rather than "`===` is used for `null` literal checks"?

And can you make the second sentence active? "Kotlin uses `===`" or something like that?

docs/topics/operator-overloading.md
```diff
@@ -176,14 +176,15 @@ For the assignment operations, for example `a += b`, the compiler performs the f
 | `a != b` | `!(a?.equals(b) ?: (b === null))` |
 
 These operators only work with the function [`equals(other: Any?): Boolean`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html), 
-which can be overridden to provide custom equality check implementation. Any other function with the same name (like `equals(other: Foo)`) will not be called.
+which you can override to provide a custom equality check implementation.
+Any other function with the same name (like `equals(other: Foo)`) is ignored.
 
-> `===` and `!==` (identity checks) are not overloadable, so no conventions exist for them.
+> `===` and `!==` (identity checks) aren't overloadable, so no conventions exist for them.
```
sarahhaggarty, line 185: What do you think about moving this note to the bottom of the section? I think it would make the content easier to read 🤔

sarahhaggarty (changes requested):
I have some suggestions for the text :)

docs/topics/operator-overloading.md
daniCsorbaJB: > Instead of "the null literal" can we say "a null value"?

🤔  It's an important difference here. With two expressions A and B, A == B expands to A?.equals(B) ?: (B === null) even with a null value. If either operand is literal `null`, A == B instead expands to A === B.

> And can you make the second sentence active? "Kotlin uses ===" or something like that?

Sure thing 👍

sarahhaggarty (approved):
LGTM! 🚀

PR #4813, opened 2025-04-23
feat: update the navigation for the What's new section

koshachy:
update the What's new navigation section:
* better file structure
* old Kotlin versions placed in the Other resources section
* placed compat guides near the corresponding Kotlin What's new pages

### DIFF
--- a/docs/kr.tree
+++ b/docs/kr.tree
@@ -382,6 +382,20 @@
 			<toc-element toc-title="FAQ" topic="ksp-faq.md"/>
 		</toc-element>
 	</toc-element>
+	<toc-element toc-title="Kotlin evolution and roadmap">
+		<toc-element topic="roadmap.md"/>
+		<toc-element toc-title="Language features and proposals" topic="kotlin-language-features-and-proposals.md"/>
+		<toc-element toc-title="Language evolution principles" accepts-web-file-names="kotlin-evolution.html" topic="kotlin-evolution-principles.md"/>
+		<toc-element topic="components-stability.md"/>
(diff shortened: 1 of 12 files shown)


docs/kr.tree
```diff
@@ -404,24 +389,53 @@
...
-			<toc-element topic="compatibility-guide-16.md"/>
-			<toc-element topic="compatibility-guide-15.md"/>
-			<toc-element topic="compatibility-guide-14.md"/>
-			<toc-element topic="compatibility-guide-13.md"/>
-			<toc-element topic="compatibility-modes.md"/>
+		<toc-element toc-title="Earlier Kotlin versions">
+			<toc-element toc-title="Kotlin 2.0.x">
+				<toc-element toc-title="Kotlin 2.0.20" topic="whatsnew2020.md"/>
+				<toc-element toc-title="Kotlin 2.0.0" topic="whatsnew20.md"/>
+				<toc-element topic="compatibility-guide-20.md"/>
```
danil-pavlov, line 398: I think for the left-side menu, the short name "Compatibility guide" would fit better, so the structure of the release is:

**Kotlin 2.0.x**
* Kotlin 2.0.20
* Kotlin 2.0.0
* Compatibility guide

But the page itself would have the full name "Compatibility guide for Kotlin 2.0.x"

koshachy: Do you think we should update the name for the freshest version as well?
I mean that one on line 29.

danil-pavlov: Yeah, definitely, for consistency

PR #4821, opened 2025-04-28
Fix for java-test-fixtures plugin in 2.1.21

sarahhaggarty:
This PR explains that developers should upgrade to 2.1.21 if they are KMP users that use the Java test fixtures Gradle plugin in their projects.

### DIFF
--- a/docs/topics/multiplatform/multiplatform-compatibility-guide.md
+++ b/docs/topics/multiplatform/multiplatform-compatibility-guide.md
@@ -78,6 +78,10 @@ Starting with Gradle 8.7, the Application plugin will no longer work with the Ko
 If you want to use both the Kotlin Multiplatform Gradle plugin and other Gradle plugins for
 Java in your multiplatform project, see [Deprecated compatibility with Kotlin Multiplatform Gradle plugin and Java plugins](multiplatform-compatibility-guide.md#deprecated-compatibility-with-kotlin-multiplatform-gradle-plugin-and-gradle-java-plugins).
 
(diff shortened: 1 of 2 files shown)


docs/topics/multiplatform/multiplatform-compatibility-guide.md
```diff
@@ -78,6 +78,10 @@ Starting with Gradle 8.7, the Application plugin will no longer work with the Ko
 If you want to use both the Kotlin Multiplatform Gradle plugin and other Gradle plugins for
 Java in your multiplatform project, see [Deprecated compatibility with Kotlin Multiplatform Gradle plugin and Java plugins](multiplatform-compatibility-guide.md#deprecated-compatibility-with-kotlin-multiplatform-gradle-plugin-and-gradle-java-plugins).
 
+If you use the [Java test fixtures](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) Gradle plugin
+and a Gradle version higher than 8.7, the plugin doesn't work. Instead, upgrade to [Kotlin 2.1.21](releases.md#release-details)
```
zamulla, line 82: Maybe I'm taking this out of context, but it reads a bit strange: "instead of using the plugin with Gradle > 8.7, upgrade to Kotlin 2.1.21."

Should we add the older version explicitly? "If you use the plugin, Gradle version higher than 8.7, and Kotlin version below 2.1.21, the plugin doesn't work. Upgrade..."

zamulla (changes requested):
Just one bit that I feel needs ironing out.

docs/topics/multiplatform/multiplatform-compatibility-guide.md
sarahhaggarty: Fair point! Updated the text. WDYT?

PR #4829, opened 2025-05-05
update: adding wasm dependencies in a kmp project

danil-pavlov:


docs/topics/multiplatform/multiplatform-add-dependencies.md
```diff
@@ -243,6 +243,9 @@ kotlin {
             // SQLDelight will be available only in the iOS source set, but not in Android or common
             implementation("com.squareup.sqldelight:native-driver:%sqlDelightVersion%")
         }
+        wasmJsMain.dependencies {
```
AlejandraPedroza, line 246: How about adding a line of `implementation` with a placeholder, to be consistent with the rest?

docs/topics/multiplatform/multiplatform-add-dependencies.md
```diff
@@ -268,6 +271,9 @@ kotlin {
                 implementation 'com.squareup.sqldelight:native-driver:%sqlDelightVersion%'
             }
         }
+        wasmJsMain {
```
AlejandraPedroza, line 274: Same here

AlejandraPedroza (approved):
LGTM! I left one suggestion for your consideration.

docs/topics/multiplatform/multiplatform-add-dependencies.md
danil-pavlov: I think it's enough for a source set placeholder, similar to androidMain above

PR #4831, opened 2025-05-07
Add enum class as valid K/JS exportable type

lppedd:


JSMonk (approved):

AlejandraPedroza (approved):
Thanks for your contribution! 🚀

lppedd:
@AlejandraPedroza could you wait before merging? I need to clarify something with Artem and it makes sense to wait imo.

@JSMonk I've just realized that there is an issue in the types table: `BooleanArray` is mapped to `Int8Array`, but in practice it looks like it's compiled to a JS `Array<Boolean>`. Is this a docs issue, or is it some kind of missed optimization?

Edit: seems like `Int8Array` is in the docs since a long time but the compiler always emitted the boolean array (which also makes sense for interop).

JSMonk:
> @AlejandraPedroza could you wait before merging? I need to clarify something with Artem and it makes sense to wait imo.
> 
> @JSMonk I've just realized that there is an issue in the types table: `BooleanArray` is mapped to `Int8Array`, but in practice it looks like it's compiled to a JS `Array<Boolean>`. Is this a docs issue, or is it some kind of missed optimization?
> 
> Edit: seems like `Int8Array` is in the docs since a long time but the compiler always emitted the boolean array (which also makes sense for interop).

Oh, it's really a documentation issue. May I ask you to add the change regarding `BooleanArray`, too, but it should be a separate MR?
Thank you so much for noticing it 🙏

lppedd:
Sure, I'll open a new PR. Let's merge this one as is.

AlejandraPedroza:
Hi! I am merging this PR, as agreed. Feel free to create a new one to fix the types table.
Thanks for your help!

PR #4833, opened 2025-05-08
'Default arguments' -> 'Default parameters'

nikitabobko:
Two fixes:
1. Parameters refer to the variables on the declaration site. Arguments are the expressions that "fill in" the parameters. That's why it's called "default parameters" not "default arguments". You can even see that further in the paragraph it's already referred as a "default parameter"
2. Lambda is not a parameter. It's an expression, and it's an argument that corresponds to a parameter with functional type.

---

I didn't touch anything in `docs/topics/whatsnew/*` because I don't consider them a part of the documentation but rather a piece of history that is written the way it's written.

Similarly, I didn't touch the following files:
- ./docs/topics/gsoc-2023.md
- ./docs/topics/k2-compiler-migration-guide.md

because they don't feel like a documentation to me. But let me know if you actually want me to update them

Further, I acknowledge that I'm changing the title of the section in `./docs/topics/functions.md` file which will break the existing URL links. I don't know what are the policy for doing that (if there are any). So please let me know if I just can't change it, or if there is a more "migration friendly" way to do that.

Feel free to amend the commits as you please or let me know if there are changes that you want me to make, thanks

### DIFF
--- a/docs/topics/js/js-interop.md
+++ b/docs/topics/js/js-interop.md
@@ -133,7 +133,7 @@ class Bar : Foo() {
 There are some limitations:
 
 - When a function of an external base class is overloaded by signature, you can't override it in a derived class.
-- You can't override a function with optional parameters.
+- You can't override a function including parameters with default values.
 - Non-external classes can't be extended by external classes.
 
 ### external interfaces
--- a/docs/topics/js/js-interop.md
+++ b/docs/topics/js/js-interop.md
@@ -92,9 +92,9 @@ external class MyClass {
 }
 ```
 
-### Declare optional parameters
+### Declare parameters with default values
 
-If you are writing an external declaration for a JavaScript function which has an optional parameter, use `definedExternally`.
+If you are writing an external declaration for a JavaScript function which has a parameter with default values, use `definedExternally`.
 This delegates the generation of the default values to the JavaScript function itself:
 
 ```kotlin
--- a/docs/topics/js/js-interop.md
+++ b/docs/topics/js/js-interop.md
@@ -94,7 +94,7 @@ external class MyClass {
 
 ### Declare parameters with default values
 
-If you are writing an external declaration for a JavaScript function which has a parameter with default values, use `definedExternally`.
+If you are writing an external declaration for a JavaScript function which has a parameter with a default value, use `definedExternally`.
 This delegates the generation of the default values to the JavaScript function itself:
 
 ```kotlin
--- a/docs/topics/js/js-interop.md
+++ b/docs/topics/js/js-interop.md
@@ -133,7 +133,7 @@ class Bar : Foo() {
 There are some limitations:
 
 - When a function of an external base class is overloaded by signature, you can't override it in a derived class.
-- You can't override a function with default arguments.
+- You can't override a function with optional parameters.
 - Non-external classes can't be extended by external classes.
 
 ### external interfaces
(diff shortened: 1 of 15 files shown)


nikitabobko:
BTW, the spec also calls it "default parameters" https://kotlinlang.org/spec/declarations.html#named-positional-and-default-parameters

nikitabobko:
Hmm, sorry, I maybe rushed a bit https://jetbrains.team/im/review/1AlVqg2AG16Y?message=C28bz0VH91z&channel=ei0qL1MCT31

I will make sure that we internally agree on the terminology and write back

nikitabobko:
That's the middle ground that we kinda agreed on:

- It's definitely ok to call it "optional parameters", but unfortunately it's not a wide-spread term
- It's ok to say it in full "parameters with default values"
- It's ok to say "default value" (if it's clear from the context)
- "default parameters" and "default arguments" are slightly problematic and it's better to avoid them

https://jetbrains.slack.com/archives/C06E082M6/p1746713347805569

AlejandraPedroza:
Thanks for the detailed clarification! We'll work on it in the next few days.

AlejandraPedroza:
Hi again, @nikitabobko 
I reviewed your PR, and so far it looks good. Thanks for updating the Kotlin docs! (:

- Regarding modifications to `docs/topics/whatsnew/*` , `./docs/topics/gsoc-2023.md`, ` ./docs/topics/k2-compiler-migration-guide.md`, and any other document containing the mention of 'default arguments': could I ask you to update them as well? From the user's perspective, this is content living in the same portal and will be less confusing, more intuitive, and easier to understand if we stick to a consistent term across the content in kotlinlang.org. 
- Regarding the safest way to update a section heading (anchor): when redirecting anchors, you can make a "find and replace" in all the files in our repo. For instance, I did a quick search and found some mentions to `#default-arguments` in four files.  Could you please update all the mentions with the new section name?

Please let me know if I can help. 

Thanks.

nikitabobko:
> Regarding the safest way to update a section heading (anchor): when redirecting anchors, you can make a "find and replace" in all the files in our repo

Yes, it's clear that I can update the links across the site, and I believe I already did that, but I will recheck. Anyway, my concern was about other sites that refer to kotlinlang.org

nikitabobko:
I had to force push because of merge conflicts. The first commit is untouched, it's just rebased

> Regarding modifications to `docs/topics/whatsnew/*` , `./docs/topics/gsoc-2023.md`, `./docs/topics/k2-compiler-migration-guide.md`, and any other document containing the mention of 'default arguments': could I ask you to update them as well?

I've updated these docs as well. I've yet skipped `./docs/topics/k2-compiler-migration-guide.md` because it's hard to come up with reasonable title for the issue, and the text in md document is literally a copy of YT issue title https://youtrack.jetbrains.com/issue/KTLC-97

> Could you please update all the mentions with the new section name?

Done. I indeed missed a couple of other anchors

docs/topics/js/js-interop.md
```diff
@@ -92,9 +92,9 @@ external class MyClass {
 }
 ```
 
-### Declare optional parameters
+### Declare parameters with default values
 
-If you are writing an external declaration for a JavaScript function which has an optional parameter, use `definedExternally`.
+If you are writing an external declaration for a JavaScript function which has a parameter with default values, use `definedExternally`.
```
nikitabobko: values -> value

Singular parameter has singular value

PR #4853, opened 2025-05-17
Update the id of the Ranges chapter to avoid the conflict (WRS-7029)

siarb:
**Root cause:**
The main title automatically receives the `id="ranges"` based on the file name `ranges.md`. This creates a conflict when the `## Ranges` sub-header which also assigns the same id for chapter.

To avoid this conflict, explicitly assign a different ID to the sub-header that does not match the main title’s auto-generated ID.


siarb:
@sarahhaggarty Could you take a look please?
It helps to avoid the problem described in the ticket [WRS-7029](https://youtrack.jetbrains.com/issue/WRS-7029).

PR #4855, opened 2025-05-17
Update when expression example

steshko:
Text above code example explains that a when expression must be exhaustive and if it isn't a compilation error will occur. However, the code example is a when statement in which else branch can be removed without a compilation error.


steshko:
Although it is pretty explanatory that the current example is meant to be used as an expression as each branch simply returns a string, its an inferance that the reader will have to make.

sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @steshko!

PR #4866, opened 2025-05-23
Add a page describing compatibility issues with native interoperability

SvyatoslavScherbina:


docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+
+Using such a transitive native dependency is prone to more compatibility issues.
+For example, a change made by the Kotlin library developer might alter the Kotlin representation of the native library
+incompatibly, leading to compatibility issues when you update the Kotlin library.
+
+So, instead of relying on a transitive dependency, please configure interoperability with the same native
+library directly.
+Use another package name for the native library, similar to how described [here](#use-custom-package-name).
+
+## Future evolution of interoperability
```
sbogolepov: Minor: not sure if such paragraph is needed. Beta status assumes that the feature will evolve.

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+This section covers usages of native types in library APIs specifically.
+The recommendations in the section don't apply to the application code.
+
+They also don't apply to library implementations. In other words, for a library
+```kotlin
+// Be extra careful! Native types are used in the library API.
+public fun createUIView(): UIView
+public fun handleThirdPartyNativeType(c: ThirdPartyNativeType)
+
+// Be careful as usual: native types are not used in the API.
```
sbogolepov: "in the _public_/_library_ API?"

sbogolepov (approved):

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: I don't insist on having this. But I thought it might be nice to repeat.
We also kind of have certain breaking changes in mind already, that was one of the ideas behind this paragraph.

Maybe other reviewers would also have some input.

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+
+There is an important distinction:
+- On each platform, Kotlin provides access to "system" native libraries through so-called
+  ["platform libraries"](native-platform-libs.md)
+- All other native libraries require additional configuration to be used from Kotlin, and this document calls them
+  "third-party libraries".
+
+These two kinds of native libraries have different compatibility properties.
+
+## Platform libraries
```
timofeys1, line 18: Maybe cross-linking to (or even moving here) [the section](https://github.com/JetBrains/kotlin-web-site/blob/3ee698b1f124524819f964131a9677f9733b2960/docs/topics/native/native-objc-interop.md#strong-linking) about using APIs with incompatible deployment target makes sense? This seems like something K/N could handle more gracefully in the future and is related to the platform libraries update

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+Native APIs of libraries shipped with Xcode SDKs change with an Xcode version.
+Even when such changes are source- and binary-compatible within the native language,
+they might become breaking for Kotlin due to the way the interoperability is implemented.
+
+As a result, updating the Kotlin version in a project might bring a breaking change in platform libraries.
+This might matter in two cases:
+- There is a source breaking change in platform libraries that affects the compilation of the source code in your
+  project.
+  Usually it should be trivial to fix.
+- There is a binary breaking change in platform libraries that affects some of your dependencies.
```
timofeys1: I believe these will now always manifest as partial linkage warnings/errors, right? Do we want to be more specific here and say that these issues will look like:
```
warning: <library> @ /code.kt:2:9: Function 'bar' can not be called: No function found for symbol 'foo/bar|bar(){}[100]'
warning: <missing declarations>: No function found for symbol 'foo/bar|bar(){}[100]'
```
and will produce a runtime exception?

we could also recommend promoting the warning to an error with the following:
```kotlin
kotlin {
    iosSimulatorArm64 {
        binaries.configureEach {
            freeCompilerArgs += "-Xpartial-linkage-loglevel=error"
        }
    }
}
```
if the user wants their linkage to be strict about these breakages

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+It often doesn't work at all with an unsupported version of Xcode.
+
+### Using the Xcode version older than recommended
+
+Typically, Kotlin works fine with older Xcode versions.
+There might be occasional issues, most frequently one of the two:
+- The Kotlin API of the native library refers to a non-existent type
+  (example: [KT-71694](https://youtrack.jetbrains.com/issue/KT-71694)).
+- The Kotlin API of the native library includes a type that belongs to a system library.
+  I.e., everything compiles well, but some system native type is added to your native library package.
```
timofeys1: Should we recommend always applying some filtering using `modules` or `headerFilter` in the `.def` file to mitigate this? This is not Xcode-only issue, right?

One other tool we could mention here is:
```
/path/to/konan/bin/klib dump-metadata/dump-metadata-signatures /path/to/cinterop.klib
```
which lets you easily see the API surface of your cinterop library (and whether something leaked by accident)

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Using a transitive third-party native dependency
+
+When a Kotlin library used in your project imports a third-party native library as its implementation detail,
+your project also gets access to that native library.
+This happens because Kotlin/Native doesn't support the separation between `api` and `implementation` dependencies,
+so native libraries always end up being `api` dependencies.
+
+Using such a transitive native dependency is prone to more compatibility issues.
+For example, a change made by the Kotlin library developer might alter the Kotlin representation of the native library
```
timofeys1: > a change made by the Kotlin library developer

Did you mean C/Objective-C library developer instead? Otherwise, this statement is also true for non-cinterop libraries.

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+incompatibly, leading to compatibility issues when you update the Kotlin library.
```
timofeys1: > leading to compatibility issues when you update the Kotlin library

I would expect to see the same issues as [here](https://github.com/JetBrains/kotlin-web-site/pull/4866/files#diff-bf4cddfca77f4bd73256d4db8da1649e9dad6043dd881431767a5e6dc388be5eR40-R43), is that correct? So you would either see your sources failing to compile, or your consumers might see the partial linkage behavior if you are a library author. Could this manifest as some other incompatibility?

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+
+If you publish a Kotlin library that uses third-party native libraries, here are a few things you might find useful.
+
+### Use custom package name
+Using custom package names for third-party native libraries might help to prevent compatibility issues.
+
+When importing a native library to Kotlin, it gets a Kotlin package name.
+If this package name is not unique, users of your library might have a clash: for example, if the same native
+library is imported with the same package name elsewhere in the user's project or other dependencies, those two usages
+will clash.
```
timofeys1: Should we put examples of specific errors like:
```
Linking globals named 'kniprot_...': symbol multiply defined!
```
here to make this page better SEOed?

I also remember one consequence of such a collision is that you will not be able to see your cinterop API, but only your dependency's cinterop API

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+
+When importing a native library, it is possible to include the associated static library (`.a` file) using
+the `-staticLibrary` flag or the `staticLibraries` property in a `.def` file.
+
+In that case a user of your library won't need to deal with native dependencies and linker options to use it.
+But there is a limitation: users won't have a way to configure the usage of the included static library in any way:
+neither exclude it nor replace (substitute) it.
+
+So, they won't be able to resolve potential clashes with other Kotlin libraries that include the same static library
+or to adjust the version of the static library.
```
timofeys1: I think they could, but it would not be very convenient. If they had a collision:
```
-library cinterop-with-libFooV1-a.klib -library cinterop-with-libFooV2-a.klib
```
they could prepend
```
-library cinterop-override-with-libFooV3-a.klib
```
and the override will come first in the `filelist`. Would this solve the issue of adjusting the version / replacing the library?

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,180 @@
...
+native APIs to Kotlin.
+Just like above, the JetBrains team makes reasonable efforts to avoid breaking changes in such cases as well.
+
+## Importing third-party native libraries when the Xcode version doesn't match
+
+Apart from system libraries, Kotlin/Native allows importing third-party native libraries.
+For example,
+by using [CocoaPods integration](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-cocoapods-dsl-reference.html)
+or defining a [cinterops configuration](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html#cinterops).
+This has implications for the compatibility of your project with different Xcode versions.
```
timofeys1: One aspect that might not be entirely clear to someone not familiar with C/Objective-C and the K/N integration with Xcode is: where exactly does Xcode come into play? Maybe it's worth pointing out that for Apple targets, when we run `.h` -> `.klib` translation, the `.h` files are resolved against the selected Xcode sysroot for this particular target, rather than a sysroot that we vendor, as is the case for other targets. That's the reason Xcode update affects your cinterops and also the reason why cross-cinterop from non-Apple hosts to Apple targets is not currently supported.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: I don't consider moving it since the document has a slightly different scope. But I find cross-linking a good idea. Will do.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > This is not Xcode-only issue, right?

No, this is Xcode-only.

> Should we recommend always applying some filtering using modules or headerFilter in the .def file to mitigate this?

Not sure. Isn't it too much detail?
In particular, the CocoaPods plugin applies filtering by default. 

> One other tool we could mention here is:

Seems a bit too advanced and low-level for this document.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > Did you mean C/Objective-C library developer instead?

I mean specifically the Kotlin library developer. If there is a change in the C/Obj-C library, the Kotlin library developer still needs to make a change to bring it: bump the native library version.

> Otherwise, this statement is also true for non-cinterop libraries.

Yes and no. This document concerns cinterop libraries specifically and how they might change incompatibly without intent to do so, in an unexpected case. If a Kotlin library changes incompatibly, the Kotlin library developer is responsible for this, and when using a library, you need to be aware of its compatibility policy.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > I would expect to see the same issues as [here](https://github.com/JetBrains/kotlin-web-site/pull/4866/files#diff-bf4cddfca77f4bd73256d4db8da1649e9dad6043dd881431767a5e6dc388be5eR40-R43), is that correct?

Yes.

> Could this manifest as some other incompatibility?

It shouldn't.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > I also remember one consequence of such a collision is that you will not be able to see your cinterop API, but only your dependency's cinterop API

No idea.

> Should we put examples of specific errors like:

Will do.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > Would this solve the issue of adjusting the version / replacing the library?

Not sure. How?

timofeys1: > How?

On symbol collision, the linker should use the first static library in the filelist, and I assume the filelist order of bundled static libraries is defined by `-library` order. So by ordering or placing your cinterop library with a particular version of a static library, you should be able to override the static library bundled in your dependency

docs/topics/native/native-c-interop-stability.md
timofeys1: > Isn't it too much detail?

Maybe. I was reading this sentence
> some system native type is added to your native library package.

and I (hope I) understand how this happens only because I have some experience with updating Xcode in K/N, so mentioning the `klib` tool and specific mitigations makes this information more tangible and actionable.

If we don't want to mention the `klib` tool, we could mention that when you encounter this issue, you will, for example, see these system APIs in the IDE completion for your cinterop library.

abdulowork (approved):

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: Ah, I see.
>  I assume the filelist order of bundled static libraries is defined by `-library` order.

I think the compiler topologically sorts them at some point.

docs/topics/native/native-c-interop-stability.md
SvyatoslavScherbina: > If we don't want to mention the `klib` tool, we could mention that when you encounter this issue, you will, for example, see these system APIs in the IDE completion for your cinterop library.

I thought that was kind of implied by "some system native type is added to your native library package." 😅 
I deliberately wanted to keep this document short and high-level. In particular, don't use the term "cinterop library", since it is an implementation detail.

I'm not sure how to describe the specific symptom while keeping things short and high-level.

What do you think of

> The Kotlin API of the native library includes a type that belongs to a system library.
I.e., everything compiles well, but some system native type is added to your native library package. For example, this system type can be visible in autocompletion in the IDE when and where you don't expect it.

?

timofeys1: > What do you think of

LGTM, I think this is worth mentioning so that the reader has some way to explore whether this happened to their library

docs/kr.tree
```diff
@@ -181,6 +181,7 @@
 		<toc-element toc-title="Native">
 			<toc-element accepts-web-file-names="basic-kotlin-native-app.html,kotlin-native-with-clion.html,targeting-multiple-platforms.html,native-gradle.html,native-command-line-compiler.html" topic="native-get-started.md"/>
 			<toc-element topic="native-definition-file.md"/>
+			<toc-element toc-title="C and Objective-C libraries" topic="native-c-interop-stability.md"/>
```
SvyatoslavScherbina: I think the title doesn't exactly match the contents now.
Based on the title, one might assume this page covers the topic of using C and Objective-C libraries, while in fact it only covers the stability of it.

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,195 @@
...
+Using an Xcode version newer than recommended might break some Kotlin features. Importing third-party
+native libraries is most affected by this . It often doesn't work at all with an unsupported version of Xcode.
+
+#### Xcode version is older than recommended
+
+Typically, Kotlin works well with older Xcode versions. There might be occasional issues, most often when
+the Kotlin API of the native library:
+
+* Refers to a non-existent type as in [KT-71694](https://youtrack.jetbrains.com/issue/KT-71694).
+* Includes a type that belongs to a system library.
```
SvyatoslavScherbina: The wording is now a bit confusing. This can be read as "when you make the Kotlin API of the native library refer to a non-existent type, this will lead to issues". I.e. as if such a reference was the cause of issues, while in fact it is the result of issues.

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Publishing a library that uses third-party library
+
+If you publish a Kotlin library that uses third-party native libraries, there are a few things you can do to avoid
+compatibility issues.
+
+#### Use custom package name
+
+Using custom package names for third-party native libraries might help to prevent compatibility issues.
+
```
SvyatoslavScherbina, line 149: Btw, is "to" between "help" and "prevent" necessary here?

docs/topics/native/native-c-interop-stability.md
```diff
@@ -0,0 +1,195 @@
...
+  might further limit forward compatibility.
+
+* Kotlin Multiplatform libraries provide backward compatibility
+  (when a newer compiler can use libraries produced with an older version).
+
+  Using a native library in the Kotlin library normally shouldn't affect its backward compatibility.
+  But it opens the possibility of more compiler bugs affecting compatibility.
+
+#### Add static libraries
+
```
SvyatoslavScherbina, line 182: The title is confusing. It doesn't match the "there are a few things you can do to avoid compatibility issues" intro.

It also reads like the recommendation to add static libraries, while the suggestion below is actually the opposite.

Maybe something like "Avoid embedding static libraries"?

docs/topics/native/native-c-interop-stability.md
danil-pavlov: Not really, optional

docs/topics/native/native-c-interop-stability.md
danil-pavlov: True, let's replace

docs/topics/native/native-c-interop-stability.md
danil-pavlov: Thank you, updated the whole section

docs/kr.tree
danil-pavlov: I was trying to add a shorter version for the left menu, let's add 'support'

SvyatoslavScherbina: Ok.
Can we make the title in the document itself more clear? I.e. include the word "stability" in it.

danil-pavlov: My concern about 'stability' is that it sounds like it's the stability of the libraries themselves, and not of the import / usage stability
What about "C and Objective-C library import" for the shorter version and "Stability of C and Objective-C library import" for the full one?

SvyatoslavScherbina: > What about "C and Objective-C library import" for the shorter version and "Stability of C and Objective-C library import" for the full one?

Yes, sounds great!

danil-pavlov (approved):

PR #4870, opened 2025-05-28
feat: add section for read write annotations kotlin metadata

daniCsorbaJB:
This update adds a section to the Kotlin Metadata JVM page about the a new experimental feature for reading and writing annotations in Kotlin metadata

Related ticket: [KT-31857](https://youtrack.jetbrains.com/issue/KT-31857) Provide easy way to retrieve annotations for kotlinx-metadata

### DIFF
--- a/docs/topics/metadata-jvm.md
+++ b/docs/topics/metadata-jvm.md
@@ -182,7 +182,7 @@ fun main() {
 You can store annotations in Kotlin metadata and access them using the `kotlin-metadata-jvm` library.
 This removes the need to match annotations by signature, making access more reliable for overloaded declarations.
 
-To make annotations available in the metadata for your compiled files, add the following compiler option:
+To make annotations available in the metadata of your compiled files, add the following compiler option:
 
 ```kotlin
 -Xannotations-in-metadata
@@ -199,7 +199,7 @@ kotlin {
 }
 ```
 
-With this option enabled, the Kotlin compiler writes annotations into metadata alongside the JVM bytecode,
+When you enable this option, the Kotlin compiler writes annotations into metadata alongside the JVM bytecode,
 making them accessible to the `kotlin-metadata-jvm` library.
 
 The library provides the following APIs for accessing annotations:


docs/topics/metadata-jvm.md
```diff
@@ -176,6 +176,78 @@ fun main() {
 }
 ```
 
+### Write and read annotations in metadata
+
+You can store annotations in Kotlin metadata and access them using the `kotlin-metadata-jvm` library.
+This removes the need to match annotations by signature, making access more reliable for overloaded declarations.
+
+To make annotations available in the metadata for your compiled files, add the following compiler option:
```
AlejandraPedroza: It would be nice to clarify where to add the compiler option.

AlejandraPedroza: "To make annotations available in the metadata for your compiled files"
Isn't it "of" instead of "for"?

docs/topics/metadata-jvm.md
```diff
@@ -176,6 +176,78 @@ fun main() {
...
+```kotlin
+// build.gradle.kts
+kotlin {
+    compilerOptions {
+        freeCompilerArgs.add("-Xannotations-in-metadata")
+    }
+}
+```
+
+With this option enabled, the Kotlin compiler writes annotations into metadata alongside the JVM bytecode,
```
AlejandraPedroza: When you enable this option - To avoid passive voice

docs/topics/metadata-jvm.md
```diff
@@ -176,6 +176,78 @@ fun main() {
...
+making them accessible to the `kotlin-metadata-jvm` library.
+
+The library provides the following APIs for accessing annotations:
+
+* `KmClass.annotations`
```
AlejandraPedroza, line 207: It would be useful for readers to have the link of the spec in kotlinlang.org for each API, what do you think?

docs/topics/metadata-jvm.md
```diff
@@ -176,6 +176,78 @@ fun main() {
...
+
+fun main() {
+    val metadata = Message::class.java.getAnnotation(Metadata::class.java)
+    val kmClass = (KotlinClassMetadata.readStrict(metadata) as KotlinClassMetadata.Class).kmClass
+    println(kmClass.annotations)
+    // [@Label(value = StringValue("Message class"))]
+}
+```
+
+> If you use the `kotlin-metadata-jvm` library in your projects, we recommend updating and testing your code to support annotations.
```
AlejandraPedroza, line 243: "it is recommended"?

AlejandraPedroza (changes requested):
Nice update! Left some comments for your consideration.

docs/topics/metadata-jvm.md
daniCsorbaJB: hmm 🤔 I prefer we recommend here. It should be obvious that JetBrains recommends this I think

PR #4872, opened 2025-06-01
update: Add section about constructor parameters and properties in classes

choonkending:
# High Level Overview

[KT-77990](https://youtrack.jetbrains.com/issue/KT-77990)

When I was going through the documentation, I had a bit of confusion around "When would we include `val` or `var`  vs when we shouldn't?" I could not really find it anywhere in the docs, hence I am proposing to include a section explaining the difference between parameters and properties in the constructor. 

While it may seem obvious, I hope this clarification can help!

I debated between making this it's own tiny section - but figured it flowed naturally from the current description of properties. 

First time contributing to these docs, so happy for any feedback/correction/discussion.

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -93,6 +93,20 @@ class Person(
 
 Much like regular properties, properties declared in the primary constructor can be mutable (`var`) or read-only (`val`).
 
+When you declare a parameter without a `val` or `var` in the primary constructor, it is simply a constructor parameter, and **not a property in the class**. The parameter is only accessible within the constructor's scope.  
+
+```kotlin
+// The following will not compile 
+class RectangleWithParameters(width: Int, height: Int) {
+    fun area() = width * height
+}
+
+// The following will compile
+class RectangleWithProperties(val width: Int, val height: Int) {
+    fun area() = width * height
+}
+```
+
 If the constructor has annotations or visibility modifiers, the `constructor` keyword is required and the modifiers go before it:
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


choonkending:
Bumping this PR - it's not high priority but something valuable imo. Let me know if you reckon it should be reworded/clarified further.

docs/topics/classes.md
```diff
@@ -93,6 +93,20 @@ class Person(
...
+class RectangleWithParameters(width: Int, height: Int) {
+    fun area() = width * height
+}
+
+// The following will compile
+class RectangleWithProperties(val width: Int, val height: Int) {
+    fun area() = width * height
+}
+```
+
```
sarahhaggarty:
````suggestion
Plain constructor parameters (that are not properties) are accessible in:
* The class header.
* Initialized properties within the class body.
* Initializer blocks.

For example:

```kotlin
// width and height are plain constructor parameters
class RectangleWithParameters(width: Int, height: Int) {
    val perimeter = 2 * width + 2 * height

    init {
        println("Rectangle created with width = $width and height = $height")
    }
}
```

````

The constructor's scope needs some definition because there are cases where you can access the constructor parameters outside of the class header. I suggested some text here. Can you update your commit with my suggestion?

sarahhaggarty (changes requested):
Thanks for your suggestion @choonkending ! Sorry it took so long to get to your PR. Your text is correct but could do with some expansion so I made a suggestion. Let me know if you have any questions!

choonkending:
@sarahhaggarty Thank you very much for the suggestion. I've updated the commits to include your suggestion and followed the convention the other commits seem to be having. Let me know what else is required 🙏

sarahhaggarty (approved):

sarahhaggarty:
Great! Thanks for your contribution @choonkending !

PR #4873, opened 2025-06-02
update: language updates for 2.2.0 release

AlejandraPedroza:
This PR contains doc updates about context parameters, guard conditions, and nested type aliases.

### DIFF
--- a/docs/topics/context-parameters.md
+++ b/docs/topics/context-parameters.md
@@ -3,6 +3,7 @@
 <primary-label ref="experimental-general"/>
 
 > Context parameters replace an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm).
+> You can find their main differences in the [design document for context parameters](https://github.com/Kotlin/KEEP/blob/master/proposals/context-parameters.md#summary-of-changes-from-the-previous-proposal).
 > To migrate from context receivers to context
 > parameters, you can use assisted support in IntelliJ IDEA, as described in
 > the related [blog post](https://blog.jetbrains.com/kotlin/2025/04/update-on-context-parameters/).
@@ -13,8 +14,6 @@ Context parameters allow functions and properties to declare dependencies that a
 surrounding context.
 
 With context parameters, you don't need to manually pass around values, such as services or dependencies, that are shared and rarely change across sets of function calls.
-
-Context parameters are not introduced as receivers in the body of a function. As a result, you need to use the name of the context parameters to access their members, unlike with context receivers, where the context is implicitly available.
 
 To declare context parameters for properties and functions, use the `context` keyword
 followed by a list of parameters, with each parameter declared as `name: Type`. Here is an example with a dependency on the `UserService` interface:
--- a/docs/topics/type-aliases.md
+++ b/docs/topics/type-aliases.md
@@ -99,7 +99,6 @@ Nested type aliases follow specific rules to ensure clear and consistent behavio
 * Their scope is the same as [nested classes](nested-classes.md). You can define them inside classes,
   and they hide any parent type aliases with the same name as they don't override.
 * Nested type aliases can be marked as `internal` or `private` to limit their visibility.
-* When a type alias refers to an `inner class`, Kotlin generates a constructor in the static scope for nested type aliases.
 * Nested type aliases are not supported in Kotlin Multiplatform's [`expect/actual` declarations](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-expect-actual.html).
 
 ### How to enable nested type aliases


serras (approved):

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+<primary-label ref="experimental-general"/>
+
+Context parameters allow functions and properties to declare dependencies that are implicitly available in the
+surrounding context.
+
+You can declare context parameters for properties and functions using the `context` keyword
+followed by a list of parameters, each of the form `name: Type`. Here is an example with a dependency on the `UserService` interface:
+
+```kotlin
+// `UserService` defines the dependency required in context
```
daniCsorbaJB:
```suggestion
// UserService defines the dependency required in context 
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+```kotlin
+// Uses `_` as context parameter name
+context(_: UserService)
+fun logWelcome() {
+    // Resolution still finds the appropriate `log` function from UserService
+    outputMessage("Welcome!")
+}
+```
+
+> Context parameters replaces an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm).
```
daniCsorbaJB: Hmm I think it should be either: "The context parameters feature replaces" or "Context parameters replace"

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+fun logWelcome() {
+    // Resolution still finds the appropriate `log` function from UserService
+    outputMessage("Welcome!")
+}
+```
+
+> Context parameters replaces an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm). 
+> To migrate from context receivers to context
+> parameters, you can use assisted support in IntelliJ IDEA, as described in
+> the [blog post](https://blog.jetbrains.com/kotlin/2025/04/update-on-context-parameters/).
```
daniCsorbaJB:
```suggestion
> the related [blog post](https://blog.jetbrains.com/kotlin/2025/04/update-on-context-parameters/).
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
+[//]: # (title: Context parameters)
+
+<primary-label ref="experimental-general"/>
+
+Context parameters allow functions and properties to declare dependencies that are implicitly available in the
+surrounding context.
+
+You can declare context parameters for properties and functions using the `context` keyword
```
daniCsorbaJB:
```suggestion
To declare context parameters for properties and functions, use the `context` keyword
```

daniCsorbaJB: (suggestion)
Just stylistic, a bit more active this way

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+followed by a list of parameters, each of the form `name: Type`. Here is an example with a dependency on the `UserService` interface:
```
daniCsorbaJB:
```suggestion
followed by a list of parameters, each specified in the following format `name: Type`. Here is an example with a dependency on the `UserService` interface:
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+// `UserService` defines the dependency required in context 
+interface UserService {
+    fun log(message: String)
+    fun findUserById(id: Int): String
+}
+
+// Declares a function with a context parameter
+context(users: UserService)
+fun outputMessage(message: String) {
+    // Uses `log` from the context
```
daniCsorbaJB:
```suggestion
    // Uses log from the context
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+    users.log("Log: $message")
+}
+
+// Declares a property with a context parameter
+context(users: UserService)
+val firstUser: String
+    // Uses `findUserById` from the context
```
daniCsorbaJB:
```suggestion
    // Uses findUserById from the context    
```

docs/topics/context-parameters.md
daniCsorbaJB: Same applies in the rest of the doc

docs/topics/context-parameters.md
daniCsorbaJB: or perhaps  "each parameter declared as `name: Type`" 🤔

docs/topics/context-parameters.md
daniCsorbaJB: Also, what do you think of placing this at the beginning of the document? 🤔 - if someone who is familiar with context receivers they would immediately know that this is the replacement. This way it might take them a while to realize that the two are essentially the same?

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+>
+{style="tip"}
+
+#### Context parameters resolution
+
+Kotlin resolves context parameters at the call site by searching for matching context values in the current scope. Kotlin matches them by their type.
+If multiple compatible values exist at the same scope level, the compiler reports an ambiguity:
+
+```kotlin
+// `UserService` defines the dependency required in context
```
daniCsorbaJB: same comments here as above regarding `UserService`

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+
+    // Implements `UserService`
+    val serviceB = object : UserService {
+        override fun log(message: String) = println("B: $message")
+    }
+
+    // Both `serviceA` and `serviceB` match the expected `UserService` type at the call site
+    context(serviceA, serviceB) {
+        outputMessage("This will not compile")
+        // Ambiguity error
```
daniCsorbaJB: Could we write the exact error message here? (not the entire, but the line the users might see here) 🤔 
Alternatively, we could place this comment before the line outputMessage("This will not compile") 
// This results in an Amiguity error

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+```
+
+#### Restrictions
+
+Context parameters are in continuous improvement; some of the current restrictions are:
+
+* Constructors cannot declare context parameters
```
daniCsorbaJB:
```suggestion
* Constructors can't declare context parameters.
```

daniCsorbaJB: just to match the rest of the list (we write can't for the following two so I suggest writing can't here as well)

Also since these sentences have complete sentences (subject-verb) I think we need the period at the end.

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+* Properties with context parameters can't have backing fields or initializers
```
daniCsorbaJB:
```suggestion
* Properties with context parameters can't have backing fields or initializers.
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+* Properties with context parameters can't use delegation
```
daniCsorbaJB:
```suggestion
* Properties with context parameters can't use delegation.
```

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+
+However, context parameters in Kotlin represent a significant improvement in managing dependencies through simplified dependency injection,
```
daniCsorbaJB:
```suggestion
Despite these restrictions, context parameters simplify managing dependencies through simplified dependency injection,
```

daniCsorbaJB: (suggestion) 
a bit more concise this way - wdyt? 🤔

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+    context(serviceA, serviceB) {
+        outputMessage("This will not compile")
+        // Ambiguity error
+    }
+}
+```
+
+#### Restrictions
+
+Context parameters are in continuous improvement; some of the current restrictions are:
```
daniCsorbaJB:
```suggestion
Context parameters are in continuous improvement, and some of the current restrictions include:
```

daniCsorbaJB: (suggestions)
Just stylstic suggestion - whichever you prefer

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,121 @@
...
+        freeCompilerArgs.add("-Xcontext-parameters")
+    }
+}
+```
+
+> Specifying both `-Xcontext-receivers` and `-Xcontext-parameters` compiler options simultaneously leads to an error.
+>
+{style="warning"}
+
+This feature is planned to be stabilized and improved in future Kotlin releases.
```
daniCsorbaJB: we could link to the stability levels here: https://kotlinlang.org/docs/components-stability.html#stability-levels-explained

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
+and simplifying internal implementations.
+
+### Rules for nested type aliases
+
+Nested type aliases have a few rules to ensure clear and consistent usage:
+
+* Nested type aliases must follow all existing [type alias](type-aliases.md) rules.
+* Their scope is the same as [nested classes](nested-classes.md). You can define them inside classes,
```
daniCsorbaJB, line 99: The link with type-aliases.md will not work here :( I think we should remove it.

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
 ```
 {kotlin-runnable="true"}
 
+## Nested type aliases
+
+<primary-label ref="beta"/>
+
+In Kotlin you can define type aliases inside other declarations, as long as they
```
daniCsorbaJB:
```suggestion
In Kotlin, you can define type aliases inside other declarations, as long as they
```

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+```kotlin
+class Dijkstra {
+   typealias VisitedNodes = Set<Node>
+
+   private fun step(visited: VisitedNodes, ...) = ...
+}
+```
+
+Capturing means that the type alias refers to a type parameter defined in the outer class:
```
daniCsorbaJB, line 72:
```suggestion
Capture means that the type alias refers to a type parameter defined in the outer class:
```

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+```kotlin
+class Graph<Node> {
+    // Incorrect because captures `Node`
+    typealias Path = List<Node>
+}
+```
+
+To fix this issue, extract the captured parameter and declare it explicitly in the type alias:
```
daniCsorbaJB:
```suggestion
To fix this issue, declare the type parameter directly in the type alias:
```

daniCsorbaJB: I think the part about extracting it doesn't add extra information here so we could drop it. What do you think? 🤔

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+   private fun step(visited: VisitedNodes, ...) = ...
+}
+```
+
+Capturing means that the type alias refers to a type parameter defined in the outer class:
+
+```kotlin
+class Graph<Node> {
+    // Incorrect because captures `Node`
```
daniCsorbaJB: Same comment as in context-parameters.md remove backtics from comments

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+    typealias Path<Node> = List<Node>
+}
+```
+
+Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
+and simplifying internal implementations.
+
+### Rules for nested type aliases
+
+Nested type aliases have a few rules to ensure clear and consistent usage:
```
daniCsorbaJB:
```suggestion
Nested type aliases follow specific rules to ensure clear and consistent behavior:
```

daniCsorbaJB: (suggestion)
Just a stylistic suggestion - whichever you prefer.

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+* Nested type aliases must follow all existing [type alias](type-aliases.md) rules.
+* Their scope is the same as [nested classes](nested-classes.md). You can define them inside classes,
+  and they hide any parent type aliases with the same name as they don't override.
+* In terms of visibility, the alias can't expose more than its referenced types allow.
+* These type aliases can be internal or private, so you can introduce them just for local usage.
+* If you alias an `inner class`, Kotlin generates a constructor in the static scope for nested type aliases.
```
daniCsorbaJB:
```suggestion
* When a type alias refers to an `inner class`, Kotlin generates a constructor in the static scope for nested type aliases.
```

daniCsorbaJB: I don't think we use alias as a verb in the docs so I think we could avoid it by rephrasing this slightly - what do you think? 🤔

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+
+### Rules for nested type aliases
+
+Nested type aliases have a few rules to ensure clear and consistent usage:
+
+* Nested type aliases must follow all existing [type alias](type-aliases.md) rules.
+* Their scope is the same as [nested classes](nested-classes.md). You can define them inside classes,
+  and they hide any parent type aliases with the same name as they don't override.
+* In terms of visibility, the alias can't expose more than its referenced types allow.
+* These type aliases can be internal or private, so you can introduce them just for local usage.
```
daniCsorbaJB:
```suggestion
* Nested type aliases can be marked as `internal` or `private` to limit their visibility.
```

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+and simplifying internal implementations.
+
+### Rules for nested type aliases
+
+Nested type aliases have a few rules to ensure clear and consistent usage:
+
+* Nested type aliases must follow all existing [type alias](type-aliases.md) rules.
+* Their scope is the same as [nested classes](nested-classes.md). You can define them inside classes,
+  and they hide any parent type aliases with the same name as they don't override.
+* In terms of visibility, the alias can't expose more than its referenced types allow.
```
daniCsorbaJB: I might switch these two bullet points to first introduce that they can have different visibilities, then tell the users that they must follow this rule.

daniCsorbaJB (changes requested):
Nice addition, thank you! 🚀 — I left a couple of comments, please let me know what you think

docs/topics/type-aliases.md
AlejandraPedroza: I'll leave it in gerund to have the noun form. Sounds more natural. WDYT?

docs/topics/type-aliases.md
```diff
@@ -54,3 +54,71 @@ fun main() {
...
+    // Incorrect because captures Node
+    typealias Path = List<Node>
+}
+```
+
+To fix this issue, declare the type parameter directly in the type alias:
+
+```kotlin
+class Graph<Node> {
+    // Correct because Node is a type alias' parameter
```
daniCsorbaJB:
```suggestion
    // Correct because Node is a type alias parameter
```

docs/topics/type-aliases.md
daniCsorbaJB: sure thing 👍 works for me

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,125 @@
...
+> the related [blog post](https://blog.jetbrains.com/kotlin/2025/04/update-on-context-parameters/).
+>
+{style="tip"}
+
+Context parameters allow functions and properties to declare dependencies that are implicitly available in the
+surrounding context.
+
+With context parameters, you don't need to manually pass around values, such as services or dependencies, that are shared and rarely change across sets of function calls.
+
+Context parameters are not introduced as receivers in the body of a function. As a result, you need to use the name of the context parameters to access their members, unlike with context receivers, where the context is implicitly available.
```
serras: I'm sorry to raise this again, but I really think that the documentation should **not** be written from the point of view of somebody who know about context receivers. Because if you don't know about them, why would you expect them to be introduced as receivers? (since they are just parameters).

My suggestion is to drop this paragraph entirely, and move most of its content to the note above (I'll make a suggestion).

docs/topics/context-parameters.md
```diff
@@ -0,0 +1,125 @@
+[//]: # (title: Context parameters)
+
+<primary-label ref="experimental-general"/>
+
+> Context parameters replace an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm).
```
serras, line 5:
```suggestion
> Context parameters replace an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm).
> The main differences are summarized in the [design for context parameters](https://github.com/Kotlin/KEEP/blob/master/proposals/context-parameters.md#summary-of-changes-from-the-previous-proposal).
```

PR #4874, opened 2025-06-02
update: annotations defaulting rule and the new `all` meta-target

zamulla:


docs/topics/annotations.md
```diff
@@ -173,23 +173,115 @@ The full list of supported use-site targets is:
...
+    ```kotlin
+    fun @receiver:Fancy String.myExtension() { ... }
+    ```
+    
   * `param` (constructor parameter)
   * `setparam` (property setter parameter)
   * `delegate` (the field storing the delegate instance for a delegated property)
 
-To annotate the receiver parameter of an extension function, use the following syntax:
+### `all` use-site meta-target
```
AlejandraPedroza:
```suggestion
### `all` meta-target
```

docs/topics/annotations.md
```diff
@@ -173,23 +173,115 @@ The full list of supported use-site targets is:
...
+not only to the parameter and the property or field but also to the corresponding getter and setter.
+
+Specifically, the annotation marked with `all` is propagated, if applicable:
+
+* To the constructor parameter (`param`) if the property is defined in the primary constructor.
+* To the property itself (`property`).
+* To the backing field (`field`) if the property has one.
+* To the getter (`get`).
+* To the setter parameter (`setparam`) if the property is defined as `var`.
+* To the Java-only target `RECORD_COMPONENT` if the class has the `@JvmRecord` annotation.
```
AlejandraPedroza:
```suggestion
* To the Java-only target `RECORD_COMPONENT` if the class has the `@JvmRecord` annotation.

In the example below, the `@Email` annotation is applied to all relevant targets of each property:

```kotlin
data class User(
    val username: String,

    // Applies `@Email` to `param`, `property`, `field`, `get`, and `set_param` (if `var`)
    @all:Email val email: String,
) {
    // Applies `@Email` to `property`, `field`, and `getter` (no `param` since it's not in the constructor)
    @all:Email val secondaryEmail: String? = null
}
```

```

File: docs/topics/annotations.md [hunk 3]
```diff
@@ -173,23 +173,115 @@ The full list of supported use-site targets is:
...
+
+* It does not propagate an annotation to types, potential extension receivers, or context receivers or parameters.
+* It cannot be used with multiple annotations:
+    ```kotlin
+    @all:[A B] // forbidden, use `@all:A @all:B`
+    val x: Int = 5
+    ```
+* It cannot be used with [delegated properties](delegated-properties.md).
+
+TODO: add instructions for enabling
```
AlejandraPedroza:
```suggestion
To enable the `@all` meta-target in your project, use the following compiler option in the command line:

```Bash
-Xannotation-target-all
```

Or add it to the `compilerOptions {}` block of your Gradle build file:

```kotlin
// build.gradle.kts
kotlin {
    compilerOptions {
        freeCompilerArgs.add("-Xannotation-target-all")
    }
}
```
```

File: docs/topics/annotations.md [hunk 4]
```diff
@@ -173,23 +173,115 @@ The full list of supported use-site targets is:
...
+
+Specifically, the annotation marked with `all` is propagated, if applicable:
+
+* To the constructor parameter (`param`) if the property is defined in the primary constructor.
+* To the property itself (`property`).
+* To the backing field (`field`) if the property has one.
+* To the getter (`get`).
+* To the setter parameter (`setparam`) if the property is defined as `var`.
+* To the Java-only target `RECORD_COMPONENT` if the class has the `@JvmRecord` annotation.
+
```
AlejandraPedroza, line 204:
```suggestion

You can use the `@all` meta-target with any property, both inside and outside the primary constructor. 

```

AlejandraPedroza (commented):
I'm adding some suggestions about `@all` target for you to consider.

docs/topics/annotations.md
zamulla: Should we add another line with the `@all:Email var email: String` example? I remembered that the "`set_param` (if `var`)" made me do a double take, doesn't make a lot of sense to me to save space here.

docs/topics/annotations.md
```diff
@@ -173,23 +173,143 @@ The full list of supported use-site targets is:
...
+data class User(
+    val username: String,
+    // Applies `@Email` to `param`, `property`, `field` and `get`
+    @all:Email val email: String,
+    // Applies `@Email` to `param`, `property`, `field`, `get`, and `set_param`
+    @all:Email var name: String,
+) {
+    // Applies `@Email` to `property`, `field`, and `getter` (no `param` since it's not in the constructor)
+    @all:Email val secondaryEmail: String? = null
+}
```
AlejandraPedroza: It seems we're missing ``` to close the sample code.

docs/topics/annotations.md
```diff
@@ -173,23 +173,143 @@ The full list of supported use-site targets is:
...
+
```
AlejandraPedroza:
```suggestion

```suggestion

You can use the `@all` meta-target with any property, both inside and outside the primary constructor.

```

serras (approved):

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+* `param`
+* `property`
+* `field`
+
+For example:
 
 ```kotlin
-fun @receiver:Fancy String.myExtension() { ... }
+data class User(val username: String,
+                // @Email here is equivalent to @param:Email
```
sarahhaggarty:
```suggestion
                // @Email is equivalent to @param:Email
```

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+                @Email val email: String) {
+    // @Email here is equivalent to @field:Email
```
sarahhaggarty:
```suggestion
    // @Email is equivalent to @field:Email
```

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+    @Email val secondaryEmail: String? = null
+}
 ```
 
-If you don't specify a use-site target, the target is chosen according to the `@Target` annotation of the annotation
-being used. If there are multiple applicable targets, the first applicable target from the following list is used:
+Kotlin %kotlinEapVersion% introduced an experimental defaulting rule which should
```
sarahhaggarty:
```suggestion
Kotlin 2.2.0 introduced an experimental defaulting rule which should
```
As this content should only be merged with the final release, it's best to mention the actual release version so the EAP version doesn't automatically update later.

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+
+* If the constructor parameter target (`param`) is applicable, it is used.
+* If the property target (`property`) is applicable, it is used.
+* If the field target (`field`) is applicable while `property` isn't, `field` is used.
+
+Using the same example:
+
+```kotlin
+data class User(val username: String,
+                // @Email here is now equivalent to @param:Email @field:Email
```
sarahhaggarty:
```suggestion
                // @Email is now equivalent to @param:Email @field:Email
```

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+                @Email val email: String) {
+    // @Email here is still equivalent to @field:Email
```
sarahhaggarty:
```suggestion
    // @Email is still equivalent to @field:Email
```

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+        }
+    }
+    ```
+
+### `all` meta-target
+
+<primary-label ref="experimental-opt-in"/>
+
+The `all` target makes it more convenient to cover all important cases when the same annotation needs to apply
+not only to the parameter and the property or field but also to the corresponding getter and setter.
```
sarahhaggarty:
```suggestion
The `all` target makes it easier to apply the same annotation not only to the parameter and the property or field, but also to the corresponding getter and setter.
```

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+    @all:Email var name: String,
+) {
+    // Applies `@Email` to `property`, `field`, and `getter` (no `param` since it's not in the constructor)
+    @all:Email val secondaryEmail: String? = null
+}
+```
+
+You can use the `all` meta-target with any property, both inside and outside the primary constructor.
+
+The `all` target comes with some limitations:
```
sarahhaggarty, line 306: WDYT about adding a subheader here "Limitations"?

docs/topics/annotations.md
```diff
@@ -169,27 +169,150 @@ class Example {
...
+
+* It does not propagate an annotation to types, potential extension receivers, or context receivers or parameters.
+* It cannot be used with multiple annotations:
+    ```kotlin
+    @all:[A B] // forbidden, use `@all:A @all:B`
+    val x: Int = 5
+    ```
+* It cannot be used with [delegated properties](delegated-properties.md).
+
+To enable the `@all` meta-target in your project, use the following compiler option in the command line:
```
sarahhaggarty: WDYT about adding a header here? "How to enable" or "How to opt in"?

sarahhaggarty (commented):
I added some suggestions for you to consider :)

docs/topics/annotations.md
```diff
@@ -169,27 +169,149 @@ class Example {
...
+* `field`
+
+For example:
 
 ```kotlin
-fun @receiver:Fancy String.myExtension() { ... }
+data class User(val username: String,
+                // @Email here is equivalent to @param:Email
+                @Email val email: String) {
+    // @Email is equivalent to @field:Email
```
serras, line 212: Unless you have the definition of the `Email` annotation, we have no way to know whether those choices are correct. In fact, in general annotations in Kotlin use `property` instead of `field`.

serras: I would suggest explicitly introducing `Email` somewhere above.

docs/topics/annotations.md
```diff
@@ -169,27 +169,149 @@ class Example {
...
+
+```kotlin
+data class User(
+    val username: String,
+    // Applies `@Email` to `param`, `property`, `field` and `get`
+    @all:Email val email: String,
+    // Applies `@Email` to `param`, `property`, `field`, `get`, and `set_param`
+    @all:Email var name: String,
+) {
+    // Applies `@Email` to `property`, `field`, and `getter` (no `param` since it's not in the constructor)
```
serras: This contradicts the explanation above for `Email`. If the `Email` annotation is applied to `property` here, then it's also applied to `property` (**not** to `field`) above.

docs/topics/annotations.md
zamulla: Yep, omitted the definition for some reason. Added the same link to Jakarta as in the KEEP.

serras: Could the definition be inlined in the docs? Otherwise the reader needs to jump somewhere else to understand the whole idea.

zamulla: Absolutely, I meant using the same example in general. Added the definition code block for both `all` and defaulting rule sections, check it out.

docs/topics/annotations.md
zamulla: I added the exact definition for `@Email` and removed `property` from the comments here.

PR #4876, opened 2025-06-02
feat: new compiler warning management

danil-pavlov:
This PR updates the compiler reference for the `-Xwarning-level` release in 2.2.0

### DIFF
--- a/docs/topics/compiler-reference.md
+++ b/docs/topics/compiler-reference.md
@@ -159,7 +159,7 @@ Suppress all warnings during compilation.
 
 #### -Werror
 
-Turn all warnings to compilation errors.
+Treat all warnings as compilation errors.
 
 #### -Wextra
 
@@ -187,7 +187,7 @@ You can adjust warning reporting in your project by combining module-wide rules 
 | `-Werror -Xwarning-level=DIAGNOSTIC_NAME:warning`  | Raise all warnings to errors except for the specified ones. |
 | `-Wextra -Xwarning-level=DIAGNOSTIC_NAME:disabled` | Enable all additional checks except for the specified ones. |
 
-In case you have many warnings you want to exclude from general rules, you can list them in a separate file through [`@argfile`](#argfile).
+If you have many warnings to exclude from the general rules, you can list them in a separate file using [`@argfile`](#argfile).
 
 ## Kotlin/JVM compiler options
 


docs/topics/compiler-reference.md
```diff
@@ -136,22 +123,61 @@ $ kotlinc @options/compiler.options hello.kt
...
+#### -Werror
+
+Turn all warnings to compilation errors.
+
+#### -Wextra
+
+Enable [additional declaration, expression, and type compiler checks](whatsnew21.md#extra-compiler-checks) that
+emit warnings if true.
+
+#### -Xsuppress-warning
```
demiurg906: Add a note that this flag is deprecated in favor of `-Xwarning-level`.
Also it makes sense to change the order of sections, so `-Xwarning-level` would go first

docs/topics/compiler-reference.md
```diff
@@ -136,22 +123,61 @@ $ kotlinc @options/compiler.options hello.kt
...
-Activates the Kotlin REPL.
+Configure the severity level of specific compiler warnings:
 
 ```bash
-kotlinc -Xrepl
+kotlinc -Xwarning-level=DIAGNOSTIC_NAME:(error|warning|disabled)
 ```
 
+* `error`: raises only the specified warning to an error.
+* `warning`: emits a warning for the specified diagnostic, for example, [additional compiler checks](whatsnew21.md#extra-compiler-checks).
```
demiurg906: > , for example, [additional compiler checks](whatsnew21.md#extra-compiler-checks).

I would remove that part. We are planning to deprecate the concept of extra checkers

docs/topics/compiler-reference.md
```diff
@@ -136,22 +123,61 @@ $ kotlinc @options/compiler.options hello.kt
...
+* `disabled`: suppresses only specified warning module-wide. Works the same as [`-Xsuppress-warning`](#xsuppress-warning).
```
demiurg906: "Works the same as [`-Xsuppress-warning`](#xsuppress-warning) but for a specific warning"

docs/topics/compiler-reference.md
danil-pavlov: Maybe we should remove all mentions of `-Xsuppress-warning` right now? since it was experimental, we don't need a separate announcement I think

demiurg906: Nice, I like it

demiurg906 (approved):

docs/topics/compiler-reference.md
```diff
@@ -136,22 +123,52 @@ $ kotlinc @options/compiler.options hello.kt
...
+### Warning management
 
-Activates the Kotlin REPL.
+#### -nowarn
+
+Suppress all warnings during compilation.
+
+#### -Werror
+
+Turn all warnings to compilation errors.
```
sarahhaggarty:
```suggestion
Treat all warnings as compilation errors.
```

docs/topics/compiler-reference.md
```diff
@@ -136,22 +123,52 @@ $ kotlinc @options/compiler.options hello.kt
...
+
+You can adjust warning reporting in your project by combining module-wide rules with specific ones:
+
+| Command                                            | Description                                                 |
+|----------------------------------------------------|-------------------------------------------------------------|
+| `-nowarn -Xwarning-level=DIAGNOSTIC_NAME:warning`  | Suppress all warnings except for the specified ones.        |
+| `-Werror -Xwarning-level=DIAGNOSTIC_NAME:warning`  | Raise all warnings to errors except for the specified ones. |
+| `-Wextra -Xwarning-level=DIAGNOSTIC_NAME:disabled` | Enable all additional checks except for the specified ones. |
+
+In case you have many warnings you want to exclude from general rules, you can list them in a separate file through [`@argfile`](#argfile).
```
sarahhaggarty:
```suggestion
If you have many warnings to exclude from the general rules, you can list them in a separate file using [`@argfile`](#argfile).
```

PR #4878, opened 2025-06-02
Update JVM records for 2.2.0

sarahhaggarty:
This PR adds content  about annotating JVM records as part of Kotlin 2.2.0.

### DIFF
--- a/docs/topics/jvm/jvm-records.md
+++ b/docs/topics/jvm/jvm-records.md
@@ -59,11 +59,61 @@ To declare a data class with the `@JvmRecord` annotation, it must meet the follo
 * The class cannot be local.
 * The primary constructor of the class must be as visible as the class itself.
 
-### Enabling JVM records
+### Enable JVM records
 
 JVM records require the `16` target version or higher of the generated JVM bytecode.
 
 To specify it explicitly, use the `jvmTarget` compiler option in [Gradle](gradle-compiler-options.md#attributes-specific-to-jvm) or [Maven](maven.md#attributes-specific-to-jvm).
+
+## Annotate record components in Kotlin
+
+<primary-label ref="experimental-general"/>
+
+In Java, [annotations](annotations.md) on a record component are automatically propagated to the backing field, getter, setter, and constructor parameter.
+You can replicate this behavior in Kotlin by using the [`all`](annotations.md#all-meta-target) use-site target.
+
+> To use the `all` use-site target, you must opt in. Either use the `-Xannotation-target-all` compiler option or add
+> the following to your `build.gradle.kts` file:
+>
+> ```kotlin
+> kotlin {
+>     compilerOptions {
+>         freeCompilerArgs.add("-Xannotation-target-all")
+>     }
+> }
+> ```
+>
+{style="warning"}
+
+For example:
+
+```kotlin
+@JvmRecord
+data class Person(val name: String, @all:Positive val age: Int)
+```
+
+When you use `@JvmRecord` with `@all:`, Kotlin:
+
+* Propagates the annotation to the property, backing field, constructor parameter, getter and setter.
+* Applies the annotation to the record component as well, if the annotation supports Java's `RECORD_COMPONENT`.
+
+## Make annotations work with record components
+
+To make an [annotation](annotations.md) available for both Kotlin properties **and** Java record components, add the
+following meta-annotations to your annotation declaration:
+
+* For Kotlin: [`@Target`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.annotation/-target/index.html)
+* For Java record components: [`@java.lang.annotation.Target`](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Target.html)
+
+For example:
+
+```kotlin
+@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
+@java.lang.annotation.Target(ElementType.CLASS, ElementType.RECORD_COMPONENT)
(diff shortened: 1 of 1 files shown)


docs/topics/jvm/jvm-records.md
```diff
@@ -59,12 +59,52 @@ To declare a data class with the `@JvmRecord` annotation, it must meet the follo
...
+components.
+
+### Propagate annotations of a record component
+
+<primary-label ref="experimental-general"/>
+
+<!-- Add warning about `@all` needing opt-in -->
+
+In Java, annotations on a record component are automatically propagated to the backing field, getter, and constructor parameter.
+You can replicate this behavior in Kotlin by using the [`@all:`](annotations.md) use-site target.
```
zamulla: This may be purely a presentation issue, but as I understand it, the use-site target is `all`, not `@all` since `@` is a part of the annotation. At least on the linked page on annotations we mention them without the `@`.

The current version is more evocative I'd say, but it may obscure the syntax.

docs/topics/jvm/jvm-records.md
```diff
@@ -59,12 +59,52 @@ To declare a data class with the `@JvmRecord` annotation, it must meet the follo
...
 * The primary constructor of the class must be as visible as the class itself.
 
-### Enabling JVM records
+### Enable JVM records
 
 JVM records require the `16` target version or higher of the generated JVM bytecode.
 
 To specify it explicitly, use the `jvmTarget` compiler option in [Gradle](gradle-compiler-options.md#attributes-specific-to-jvm) or [Maven](maven.md#attributes-specific-to-jvm).
 
+## Annotate records in Kotlin
```
serras: (Sorry if this is too much change)

I feel that this part of the documentation puts the stress in creating annotations that work on record components, but most of the information that people may be looking for is how to annotate record components themselves. So I would suggest to put things in the reverse order.

docs/topics/jvm/jvm-records.md
```diff
@@ -59,12 +59,52 @@ To declare a data class with the `@JvmRecord` annotation, it must meet the follo
...
+```kotlin
+@Target(AnnotationTarget.CLASS, AnnotationTarget.PROPERTY)
+@java.lang.annotation.Target(ElementType.CLASS, ElementType.RECORD_COMPONENT)
+annotation class exampleClass
+```
+
+The `exampleClass` annotation can now be applied to Kotlin classes and properties, as well as Java classes and record
+components.
+
+### Propagate annotations of a record component
```
serras: This part is missing one piece of information: you may also annotate the record component with a Java annotation in the usual way if the annotation supports both `FIELD` and `RECORD_COMPONENT` use site targets.

docs/topics/jvm/jvm-records.md
```diff
@@ -59,12 +59,52 @@ To declare a data class with the `@JvmRecord` annotation, it must meet the follo
...
+For example:
+
+```kotlin
+@JvmRecord
+data class Person(val name: String, @all:Positive val age: Int)
+```
+
+When you use `@JvmRecord` with `@all:`, Kotlin:
+
+* Propagates the annotation to the property, backing field, constructor parameter, and getter.
```
serras:
```suggestion
* Propagates the annotation to the property, backing field, constructor parameter, getter, and setter parameter.
```

docs/topics/jvm/jvm-records.md
sarahhaggarty: OK! I restructured it so that it swaps the order. How does it look now?

docs/topics/jvm/jvm-records.md
sarahhaggarty: Can you share a code sample with me that demonstrates this so I can add it to the text?

serras: Imagine you have an annotation defined in Java as follows:

```java
@Target({ElementType.FUNCTION, ElementType.FIELD, ElementType.RECORD_COMPONENT})
public @interface Annotation { }
```

And you define a record in Kotlin:

```kotlin
@JvmRecord data class Person(@field:Annotation val name: String, val age: Int)
```

Then, because (1) the annotation is applied to the field in the Kotlin side, and (2) the annotation may be applied to record components, as declared in the Java side, then the annotation there is applied to _both_ backing field and record component.

Note however that in general you don't want to apply this detailed rule, but actually to mimic the behavior of records in Java. For that reason we suggest to always use `@all` when defining records, since it propagates annotations in a similar way to how Java does it.

docs/topics/jvm/jvm-records.md
serras: It does, thanks!

docs/topics/jvm/jvm-records.md
```diff
@@ -65,45 +65,55 @@ JVM records require the `16` target version or higher of the generated JVM bytec
...
 
-### Propagate annotations of a record component
+* Propagates the annotation to the property, backing field, constructor parameter, getter and setter.
+* Applies the annotation to the record component as well, if the annotation supports Java's `RECORD_COMPONENT`.
 
-<primary-label ref="experimental-general"/>
+## Make annotations work with record components
 
-<!-- Add warning about `@all` needing opt-in -->
+To make an [annotation](annotations.md) available on both Kotlin properties **and** Java record components, add the
```
zamulla:
```suggestion
To make an [annotation](annotations.md) available for both Kotlin properties **and** Java record components, add the
```

zamulla (approved):
Just one preposition I'm not sure of, otherwise all good!

PR #4881, opened 2025-06-03
Add information about `@JvmExposeBoxed` annotation for 2.2.0

sarahhaggarty:
This PR adds information about the `@JvmExposeBoxed` annotation for 2.2.0.

### DIFF
--- a/docs/topics/compiler-reference.md
+++ b/docs/topics/compiler-reference.md
@@ -62,6 +62,8 @@ To show advanced options, use `-X`.
 
 ### -X
 
+<primary-label ref="experimental-general"/>
+
 Display information about the advanced options and exit. These options are currently unstable: 
 their names and behavior may be changed without notice.
 
@@ -124,6 +126,8 @@ Enable usages of API that [requires opt-in](opt-in-requirements.md) with a requi
 fully qualified name.
 
 ### -Xrepl
+
+<primary-label ref="experimental-general"/>
 
 Activates the Kotlin REPL.
 
@@ -217,6 +221,8 @@ Use a custom JDK home directory to include into the classpath if it differs from
 
 ### -Xjdk-release=version
 
+<primary-label ref="experimental-general"/>
+
 Specify the target version of the generated JVM bytecode. Limit the API of the JDK in the classpath to the specified Java version. 
 Automatically sets [`-jvm-target version`](#jvm-target-version).
 Possible values are `1.8`, `9`, `10`, ..., `24`.
@@ -254,6 +260,14 @@ into the classpath.
 ### -script-templates _classnames[,]_
 
 Script definition template classes. Use fully qualified class names and separate them with commas (**,**).
+
+### -Xjvm-expose-boxed
+
+<primary-label ref="experimental-general"/>
+
+Generate boxed versions of all inline value classes in the module, along with boxed variants of functions that use them,
+making both accessible from Java. For more information, see [Inline value classes](java-to-kotlin-interop.md#inline-value-classes)
+in the guide to calling Kotlin from Java.
 
 ## Kotlin/JS compiler options
 
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -10,9 +10,9 @@ On this page, we'll describe the ways to tailor the interop of your Kotlin code 
 
 A Kotlin property is compiled to the following Java elements:
 
- * a getter method, with the name calculated by prepending the `get` prefix
- * a setter method, with the name calculated by prepending the `set` prefix (only for `var` properties)
- * a private field, with the same name as the property name (only for properties with backing fields)
+ * a getter method, with the name calculated by prepending the `get` prefix.
+ * a setter method, with the name calculated by prepending the `set` prefix (only for `var` properties).
+ * a private field, with the same name as the property name (only for properties with backing fields).
 
 For example, `var firstName: String` compiles to the following Java declarations:
 
@@ -28,9 +28,9 @@ public void setFirstName(String firstName) {
 }
 ```
 
-If the name of the property starts with `is`, a different name mapping rule is used: the name of the getter will be
-the same as the property name, and the name of the setter will be obtained by replacing `is` with `set`.
-For example, for a property `isOpen`, the getter will be called `isOpen()` and the setter will be called `setOpen()`.
+If the name of the property starts with `is`, a different name mapping rule is used: the name of the getter is
+the same as the property name, and the name of the setter is obtained by replacing `is` with `set`.
+For example, for a property `isOpen`, the getter is called `isOpen()` and the setter is called `setOpen()`.
 This rule applies for properties of any type, not just `Boolean`.
 
 ## Package-level functions
@@ -109,7 +109,7 @@ org.example.Utils.getDate();
 ## Instance fields
 
 If you need to expose a Kotlin property as a field in Java, annotate it with the [`@JvmField`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-field/index.html) annotation.
-The field will have the same visibility as the underlying property. You can annotate a property with `@JvmField` if it:
+The field has the same visibility as the underlying property. You can annotate a property with `@JvmField` if it:
 * has a backing field
 * is not private
 * does not have `open`, `override` or `const` modifiers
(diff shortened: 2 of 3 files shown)


docs/topics/compiler-reference.md
```diff
@@ -218,6 +226,13 @@ into the classpath.
 
 Script definition template classes. Use fully qualified class names and separate them with commas (**,**).
 
+### -Xjvm-expose-boxed
+
+<primary-label ref="experimental-general"/>
+
+Generate boxed versions of all inline value classes in the module, making them accessible from Java. For more information, see [Inline value classes](java-to-kotlin-interop.md#inline-value-classes)
```
serras: This is not entirely correct. This generated boxed versions of all inline value classes, but also versions of the functions using those boxed inline value classes that can be consumed from Java.

sarahhaggarty: I updated the text to mention this! Please check 🙏

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,42 @@ fun emptyList(): List<Nothing> = listOf()
...
+* Class
+* Constructor
+* Function
+
+For example:
+
+```kotlin
+@JvmExposeBoxed
+@JvmInline
+value class MyInt(val value: Int)
```
hfhbd, line 625: Could you reuse the `PositiveInt` sample instead of using a new sample?

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,42 @@ fun emptyList(): List<Nothing> = listOf()
...
+
+### Inline value classes
+
+<primary-label ref="experimental-general"/>
+
+Kotlin compiles [inline value classes](inline-classes.md) to use **unboxed representations**, which are often inaccessible from Java.
+For example, Java isn't able to call a constructor for the `PositiveInt` class:
+
+```kotlin
+@JvmInline value class PositiveInt(val number: Int)
```
hfhbd: Add a new line before value to match the same sample below. BTW it makes sense to add an init block with a check, to show the use-case of `PositiveInt` class at all.

Also, could you add a non compiling java sample to highlight the benefit from `@JvmExposeBoxed`?

docs/topics/jvm/java-to-kotlin-interop.md
sarahhaggarty: Hello!
I replaced `PositiveInt` with `MyInt` throughout.

docs/topics/jvm/java-to-kotlin-interop.md
sarahhaggarty: Updated to use `MyInt` instead and added a brief non-compiling Java sample.

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,55 @@ fun emptyList(): List<Nothing> = listOf()
...
+@JvmExposeBoxed
+@JvmInline
+value class MyInt(val value: Int)
+
+@OptIn(ExperimentalStdlibApi::class)
+@JvmExposeBoxed
+fun MyInt.timesTwoBoxed(): MyInt = MyInt(this.value * 2)
+```
+
+With these annotations, Kotlin generates a Java-accessible constructor for the `MyInt` class **and** an overload for the
```
serras: Technically it's not an overload, we usually called them "variant"

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -132,14 +132,14 @@ class JavaClient {
 ```
 
 [Late-Initialized](properties.md#late-initialized-properties-and-variables) properties are also exposed as fields. 
-The visibility of the field will be the same as the visibility of `lateinit` property setter.
+The visibility of the field is the same as the visibility of `lateinit` property setter.
```
zamulla:
```suggestion
The visibility of the field is the same as the visibility of the `lateinit` property setter.
```

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -265,8 +265,8 @@ interface ChatBot {
 }
 ```
 
-`@JvmStatic` annotation can also be applied on a property of an object or a companion object
-making its getter and setter methods static members in that object or the class containing the companion object.
+You can also apply `@JvmStatic` annotation on the property of an object or a companion object making its getter and setter
```
zamulla: As far as I can see, general Java content mostly uses "apply to" rather than "apply on" for annotations.

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -358,15 +358,15 @@ Only compatibility bridges and `DefaultImpls` classes are generated.
...
 
-* `private` members are compiled to `private` members
+* `private` members are compiled to `private` members.
 * `private` top-level declarations are compiled to `private` top-level declarations. Package-private accessors are also included,
 if accessed from within a class. 
-* `protected` remains `protected` (note that Java allows accessing protected members from other classes in the same package
-and Kotlin doesn't, so Java classes will have broader access to the code)
-* `internal` declarations become `public` in Java. Members of `internal` classes go through name mangling, to make
+* `protected` remains `protected`. (Note that Java allows accessing protected members from other classes in the same package
+and Kotlin doesn't, so Java classes will have broader access to the code).
```
zamulla:
```suggestion
and Kotlin doesn't, so Java classes will have broader access to the code.)
```

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,53 @@ fun emptyList(): List<Nothing> = listOf()
 // is translated to
 // List emptyList() { ... }
 ```
+
+### Inline value classes
+
+
+Kotlin compiles [inline value classes](inline-classes.md) to use **unboxed representations**, which are often inaccessible from Java.
+For example, Java isn't able to call a constructor for the `MyInt` class:
```
zamulla: Should we rephrase it towards the user? "You can't call this constructor from Java"? For me the interaction gets muddled when I add Java as an actor here. Kotlin generates something, that's fine, but I'm the one calling the shots, I mean constructors. Or is that a problem in the context of interop in general?

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,53 @@ fun emptyList(): List<Nothing> = listOf()
...
+```
+
+You can use the [`@JvmExposeBoxed`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-expose-boxed/) annotation so that Kotlin generates a public constructor that Java can call directly.
+You can apply the annotation at the following levels to ensure fine-grained control over what's exposed to Java:
+
+* Class
+* Constructor
+* Function
+
+Before using the `@JvmExposeBoxed` annotation in your code, you must opt in by using: `@OptIn(ExperimentalStdlibApi::class)`.
```
zamulla:
```suggestion
Before using the `@JvmExposeBoxed` annotation in your code, you must opt in by using `@OptIn(ExperimentalStdlibApi::class)`.
```

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -586,3 +585,53 @@ fun emptyList(): List<Nothing> = listOf()
...
+value class MyInt(val value: Int)
+```
+
+So the following Java code fails:
+
+```java
+MyInt input = new MyInt(5);
+```
+
+You can use the [`@JvmExposeBoxed`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-expose-boxed/) annotation so that Kotlin generates a public constructor that Java can call directly.
```
zamulla: I like putting the solution in the beginning (=put this sentence after the very first one in the section), because otherwise it may look like we're talking about a problem with no apparent solution for a couple of paragraphs. What do you think?

sarahhaggarty: I tried rewriting it so that I introduce the solution at the beginning before providing a deeper explanation. WDYT?

docs/topics/jvm/java-to-kotlin-interop.md
sarahhaggarty: Rewritten as you suggested!

docs/topics/jvm/java-to-kotlin-interop.md
zamulla: Noice and toit 👍

zamulla (approved):
🚀

PR #4883, opened 2025-06-03
Add Build tools API

sarahhaggarty:
This PR adds a page about the experimental Build tools API introduced in Kotlin 2.2.0.

### DIFF
--- a/docs/kr.tree
+++ b/docs/kr.tree
@@ -308,6 +308,7 @@
 				<toc-element topic="gradle-plugin-variants.md"/>
 			</toc-element>
 			<toc-element accepts-web-file-names="using-maven.html" topic="maven.md"/>
+			<toc-element topic="build-tools-api.md"/>
 			<toc-element accepts-web-file-names="using-ant.html" topic="ant.md"/>
 		</toc-element>
 		<toc-element toc-title="Dokka">
--- a/docs/topics/build-tools-api.md
+++ b/docs/topics/build-tools-api.md
@@ -1,0 +1,108 @@
+[//]: # (title: Build tools API)
+
+<primary-label ref="experimental-general"/>
+
+<tldr>Currently, the BTA supports Kotlin/JVM only.</tldr>
+
+Kotlin 2.2.0 introduces an experimental Build tools API (BTA) that simplifies how build systems integrate with the 
+Kotlin compiler.
+
+Previously, adding full Kotlin support to a build system (like incremental compilation, Kotlin compiler plugins, 
+daemons, and Kotlin Multiplatform) required significant effort. The BTA aims to reduce this complexity by providing
+a unified API between build systems and the Kotlin compiler ecosystem.
+
+The BTA defines a single entry point that build systems can implement. This removes the need to deeply integrate with internal compiler details.
+
+> The BTA itself is not yet publicly available for direct use in your own build tool integrations.
+> If you're interested in the proposal or want to share feedback, see the [KEEP](https://github.com/Kotlin/KEEP/issues/421).
+> Follow the status of its implementation in [YouTrack](https://youtrack.jetbrains.com/issue/KT-76255).
+> 
+{style="warning"}
+
+## Integration with Gradle
+
+The Kotlin Gradle plugin (KGP) has experimental support for the BTA, and you need to opt in to use it.
+
+> We'd appreciate your feedback on your experience with the KGP in [YouTrack](https://youtrack.jetbrains.com/issue/KT-56574).
+> 
+{style="note"}
+
+### How to enable
+
+Add the following property to your `gradle.properties` file:
+
+```kotlin
+kotlin.compiler.runViaBuildToolsApi=true
+```
+
+### Configure different compiler versions
+
+With the BTA, you can now use a different Kotlin compiler version than the version used by the KGP. This is useful when:
+
+* You want to try new Kotlin features but haven't updated your build scripts yet.
+* You need the latest plugin fixes but want to stay on an older compiler version for now.
+
+Here's an example of how to configure this in your `build.gradle.kts` file:
+
+```kotlin
+import org.jetbrains.kotlin.buildtools.api.ExperimentalBuildToolsApi
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
+
+plugins {
+	kotlin("jvm") version "2.2.0"
+}
+
+group = "org.jetbrains.example"
+version = "1.0-SNAPSHOT"
+
+repositories {
+	mavenCentral()
+}
+
+kotlin {
+	jvmToolchain(8)
+	@OptIn(ExperimentalBuildToolsApi::class, ExperimentalKotlinGradlePluginApi::class)
+	compilerVersion.set("2.1.21") // <-- different version than 2.2.0
+}
+```
+
+#### Compatible Kotlin compiler and KGP versions
+
+The BTA supports:
+
+* The three previous major Kotlin compiler versions.
+* One major version forward.
+
+For example, in KGP 2.2.0, the supported Kotlin compiler versions are:
+
+* 1.9.25
+* 2.0.x
+* 2.1.x
+* 2.2.x
+* 2.3.x
+
+#### Limitations
+
+Using different compiler versions together with compiler plugins may lead to Kotlin compiler exceptions. The Kotlin team
+plans to address this in future Kotlin releases.
+
+### Enable incremental compilation with "in process" strategy
+
+The KGP supports three [compiler execution strategies](gradle-compilation-and-caches.md#defining-kotlin-compiler-execution-strategy).
+Ordinarily, the "in-process" strategy (which runs the compiler in the Gradle daemon) doesn't support incremental compilation.
+
+With the BTA, the "in-process" strategy now supports incremental compilation. To enable it, add the following property to
+your `gradle.properties` file:
+
+```kotlin
+kotlin.compiler.execution.strategy=in-process
+```
+
+## Integration with Maven
+
+From Kotlin 2.2.0, the BTA is enabled by default in the [`kotlin-maven-plugin`](maven.md).
+
+Although the BTA doesn't give direct benefits for Maven users yet, it provides a solid foundation for developing features like:
+
+* [Kotlin daemon support](https://youtrack.jetbrains.com/issue/KT-77587)
+* [Incremental compilation stabilization](https://youtrack.jetbrains.com/issue/KT-77086)


docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,104 @@
...
+Currently, the BTA supports Kotlin/JVM **only**.
+
+> The BTA itself is not yet publicly available for direct use in your own build tool integrations.
+> If you're interested in the proposal or want to share feedback, see the [KEEP](https://github.com/Kotlin/KEEP/issues/421).
+> 
+{style="warning"}
+
+## Integration with Gradle
+
+The Kotlin Gradle plugin (KGP) already supports the BTA.
```
ALikhachev:
```suggestion
The Kotlin Gradle plugin (KGP) already supports the BTA. The support is an [experimental feature](components-stability.md#stability-levels-explained) that may be changed at any time. We would appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-56574)
```

ALikhachev (changes requested):

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,104 @@
...
+Traditionally, adding full Kotlin support to a build system (like incremental compilation, Kotlin compiler plugins, 
+daemons, and Kotlin Multiplatform) required significant effort. The BTA aims to reduce this complexity by providing
+a unified API between build systems and the Kotlin compiler ecosystem.
+
+The BTA defines a single entry point that build systems can implement. This removes the need to deeply integrate with internal compiler details.
+
+Currently, the BTA supports Kotlin/JVM **only**.
+
+> The BTA itself is not yet publicly available for direct use in your own build tool integrations.
+> If you're interested in the proposal or want to share feedback, see the [KEEP](https://github.com/Kotlin/KEEP/issues/421).
```
Tapchicoma, line 17: @ALikhachev should we also link here BTA implementation issue: https://youtrack.jetbrains.com/issue/KT-76255/Implement-Build-Tools-API? 

So interested users could track progress of it?

docs/topics/build-tools-api.md
sarahhaggarty: Added experimental label also to this section and a separate section for sharing feedback on the experience with the KGP.

docs/topics/build-tools-api.md
sarahhaggarty: Added!

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+* Forward-compatible with 2.2.x and 2.3.x
+
+#### Limitations
+
+Using different compiler versions together with compiler plugins may lead to Kotlin compiler exceptions. The Kotlin team
+plans to address this in future Kotlin releases.
+
+#### Share feedback
+
+While the integration of the BTA with the Kotlin Gradle plugin is still experimental, we'd appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-56574).
```
ALikhachev: I'd suggest moving it to a higher level / a different place. It's currently nested into `Configure different compiler versions`, but we would like overall feedback about BTA in KGP, not only about changing the versions

sarahhaggarty: Good point. Moved it up a level so now users can also see it in the TOC!

ALikhachev (approved):

docs/kr.tree
```diff
@@ -308,6 +308,7 @@
 			</toc-element>
 			<toc-element accepts-web-file-names="using-maven.html" topic="maven.md"/>
 			<toc-element accepts-web-file-names="using-ant.html" topic="ant.md"/>
+			<toc-element topic="build-tools-api.md"/>
```
danil-pavlov, line 311: Ant is deprecated now, and the page will be removed soon, so you can put your new page higher

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
+[//]: # (title: Build tools API)
+
+<primary-label ref="experimental-general"/>
+
+<!-- Add diagram -->
+
+Kotlin 2.2.0 introduces an experimental Build tools API (BTA) that simplifies how build systems integrate with the 
+Kotlin compiler.
+
+Traditionally, adding full Kotlin support to a build system (like incremental compilation, Kotlin compiler plugins,
```
danil-pavlov:
```suggestion
Previously, adding full Kotlin support to a build system (like incremental compilation, Kotlin compiler plugins, 
```

danil-pavlov: Or "Traditionally... requires"

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+Kotlin 2.2.0 introduces an experimental Build tools API (BTA) that simplifies how build systems integrate with the 
+Kotlin compiler.
+
+Traditionally, adding full Kotlin support to a build system (like incremental compilation, Kotlin compiler plugins, 
+daemons, and Kotlin Multiplatform) required significant effort. The BTA aims to reduce this complexity by providing
+a unified API between build systems and the Kotlin compiler ecosystem.
+
+The BTA defines a single entry point that build systems can implement. This removes the need to deeply integrate with internal compiler details.
+
+Currently, the BTA supports Kotlin/JVM **only**.
```
danil-pavlov: I think this sentence could be a tldr; block at the top

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
+[//]: # (title: Build tools API)
+
+<primary-label ref="experimental-general"/>
```
danil-pavlov, line 3: 2 experimental labels on the page seem too much. could we keep Experimental for Gradle and nothing at the top? or just Experimental for the whole page?

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+#### Compatible Kotlin compiler and KGP versions
+
+The BTA supports:
+
+* The three previous major Kotlin compiler versions.
+* One major version forward.
+
+For example, in KGP 2.2.0, the supported Kotlin compiler versions are:
```
danil-pavlov, line 76: I think it's the 1st time that I see KGP having it's own version written like that. Plus, the intro doesn't correspond well with the last item on the list. consider this:

> For example, in Kotlin 2.2.0, the BTA is compatible with (the following compiler versions):

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+* 2.0.x 
+* 1.9.25
+* Forward-compatible with 2.2.x and 2.3.x
+
+#### Limitations
+
+Using different compiler versions together with compiler plugins may lead to Kotlin compiler exceptions. The Kotlin team
+plans to address this in future Kotlin releases.
+
+### Enable incremental compilation with "in process" compiler execution strategy
```
danil-pavlov: This title is too long for the right menu. Can we cut it to just "Enable incremental compilation" and leave the details to the paragraph below?

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Share feedback
+
+While the integration of the BTA with the Kotlin Gradle plugin is still experimental, we'd appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-56574).
+
+## Integration with Maven
+
+From Kotlin 2.2.0, the BTA is enabled by default in the `kotlin-maven-plugin`.
+
+Although the BTA doesn't provide direct benefits for Maven users yet, it provides a solid foundation for developing features like:
```
danil-pavlov:
```suggestion
Although the BTA doesn't give direct benefits for Maven users yet, it provides a solid foundation for developing features like:
```

danil-pavlov: double 'provide'

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+kotlin.compiler.execution.strategy=in-process
+```
+
+### Share feedback
+
+While the integration of the BTA with the Kotlin Gradle plugin is still experimental, we'd appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-56574).
+
+## Integration with Maven
+
+From Kotlin 2.2.0, the BTA is enabled by default in the `kotlin-maven-plugin`.
```
danil-pavlov:
```suggestion
From Kotlin 2.2.0, the BTA is enabled by default in the [`kotlin-maven-plugin`](maven﻿.md).
```

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+
+Although the BTA doesn't provide direct benefits for Maven users yet, it provides a solid foundation for developing features like:
+
+* [Kotlin daemon support](https://youtrack.jetbrains.com/issue/KT-77587/Maven-Introduce-Kotlin-daemon-support-and-make-it-enabled-by-default)
```
danil-pavlov:
```suggestion
* [Kotlin daemon support](https://youtrack.jetbrains.com/issue/KT-77587)
```

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+* [Incremental compilation stabilization](https://youtrack.jetbrains.com/issue/KT-77086/Stabilize-incremental-compilation-in-Maven)
```
danil-pavlov:
```suggestion
* [Incremental compilation stabilization](https://youtrack.jetbrains.com/issue/KT-77086)
```

docs/topics/build-tools-api.md
```diff
@@ -0,0 +1,111 @@
...
+Ordinarily, the "in-process" strategy (which runs the compiler in the Gradle daemon) doesn't support incremental compilation.
+
+With the BTA, the "in-process" strategy now supports incremental compilation. To enable it, add the following property to
+your `gradle.properties` file:
+
+```kotlin
+kotlin.compiler.execution.strategy=in-process
+```
+
+### Share feedback
```
danil-pavlov: This actually could be a What's next since this page doesn't have it yet (plus this way you have fewer sections on the right side for the main text)

Alternatively, this could be a line in the red warning at the top, since we ask for feedback there as well.

docs/topics/build-tools-api.md
sarahhaggarty: It's trying to describe that BTA as a whole is experimental and that BTA mode in KGP is experimental. But it isn't in Maven because it's the default. I could remove the label for KGP and explain it in text if you think it is better?

docs/topics/build-tools-api.md
sarahhaggarty: We have the KGP version like that in our compatibility table: https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin 🤔 

The Kotlin Gradle plugin and the Kotlin compiler can have different versions so Kotlin 2.2.0 alone is unclear. I don't think we should describe the BTA as the actor here because it enables the fact that the Kotlin compiler version can be different from the KGP version. I could separate the last bullet to become two separate bullets for each version, if you think that makes more sense?

docs/topics/build-tools-api.md
sarahhaggarty: I don't think so because the default compiler execution strategy is "daemon" and that has incremental compilation enabled by default.

docs/topics/build-tools-api.md
sarahhaggarty: It's specifically for feedback on the KGP integration though and not the Maven integration so a What's next just for that seems like it would be in the wrong place. 

Unfortunately the warning is about BTA in general not just the Gradle part so I don't think it fits there either. Do you have any other ideas?

docs/topics/build-tools-api.md
danil-pavlov: Yeah, maybe leave the Experimental label for the whole page and add a sentence to the intro

docs/topics/build-tools-api.md
danil-pavlov: Well, since BTA doesn't do much Maven users, I think it's totally fine to have it the general warning

docs/topics/build-tools-api.md
danil-pavlov: > We have the KGP version like that in our compatibility table: https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin 

but that's basically just the kotlin version

I think we can just not mention "Forward-compatible" in the last item because it's already mentioned in the previous list. Plus, it's obvious that past versions are backward-compatible and later versions are forward-compatible

> For example, for 2.2.0, the KGP is compatible with the following compiler versions: 1.9.25, 2.0.x, 2.1.x, 2.2.x and 2.3.x

docs/topics/build-tools-api.md
sarahhaggarty: You're right that the BTA mode isn't used in the Maven plugin for new features yet but it's used by default in Maven. Whereas you have to opt in for Gradle. So I worry that if I follow your suggestion, any Maven users that find issues will report it to the wrong ticket. So I'd prefer to keep it as it is.

docs/topics/build-tools-api.md
danil-pavlov: I still think 4 lines in the right menu is too much. What about `Enable incremental compilation with "in process"` or better `Enable "in process" incremental compilation`?

docs/topics/build-tools-api.md
danil-pavlov: I see. What about dropping the header here? we can just leave a simple paragraph or make a tip out of this. I think 1 sentence doesn't deserve a separate section here

sarahhaggarty: I added the sentence as a note in the introduction.

docs/topics/build-tools-api.md
sarahhaggarty: Shortened to three lines as: `Enable incremental compilation with "in process" strategy`.

PR #4884, opened 2025-06-04
Autotests for main page

solarday:


test/production/main-page-buttons.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Main page buttons', () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto('/');
+        await page.waitForSelector('button.ch2-btn.ch2-btn-primary');
+        await page.click('button.ch2-btn.ch2-btn-primary');
```
nikpachoo, line 7: I can't find this button on the main page. What is it?

nikpachoo: Ah, I see, it's the cookie banner.
Can be replaced by:
```suggestion
   await closeCookiesConsentBanner(context, baseURL);
   await page.goto('/');
```

test/production/main-page-buttons.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        const pageTitle = page.locator('h1').first()
+        await expect(pageTitle).toContainText('Kotlin for AI-powered app development');
+    });
+
+    test('Get started in AI section', async ({ page }) => {
+        const getStartedKoogButton = page.getByTestId('kotlin-plus-ai-block').getByRole('link', { name: 'Get started' });
+        await expect(getStartedKoogButton).toBeVisible();
+        await getStartedKoogButton.click();
+        await expect(page.url()).toContain('/docs.koog.ai/')
+        await expect(page.getByText('Koog is a Kotlin-based framework')).toBeVisible();
```
nikpachoo, line 69: @solarday, the tests check the content of individual pages.
I expect we'll have to update them regularly as those pages change.
But their context doesn't really belong in the main page test.
What do you think about removing the text comparisons and keeping only the URL checks?

nikpachoo (approved):

PR #4886, opened 2025-06-10
Add binary compatibility validation in KGP for Kotlin 2.2.0

sarahhaggarty:
This PR adds information about how to use binary compatibility validation in the KGP for Kotlin 2.2.0.

### DIFF
--- a/docs/topics/gradle/gradle-binary-compatibility-validation.md
+++ b/docs/topics/gradle/gradle-binary-compatibility-validation.md
@@ -1,0 +1,194 @@
+[//]: # (title: Binary compatibility validation in the Kotlin Gradle plugin)
+
+<primary-label ref="experimental-general"/>
+
+Binary compatibility validation helps library authors ensure that users don't break their code when upgrading to newer 
+versions. It's important not only for delivering a smooth upgrade experience, but also for building long-term trust with
+users and encouraging continued adoption of the library.
+
+> Binary compatibility means that the compiled bytecode of two versions of a library can run interchangeably without 
+> needing recompilation.
+> 
+{style="tip"}
+
+Starting with version 2.2.0, the Kotlin Gradle plugin supports binary compatibility validation. When enabled, it generates
+Application Binary Interface (ABI) dumps from the current code and compares them with previous dumps to highlight differences.
+You can review these changes to find any potentially binary-incompatible modifications and take action to address them.
+
+## How to enable
+
+To enable binary compatibility validation, add the following to the `kotlin{}` block in your `build.gradle.kts` file:
+
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+kotlin {
+    @OptIn(org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation::class)
+    abiValidation {
+        // Use the set() function to ensure compatibility with older Gradle versions
+        enabled.set(true)
+    }
+}
+```
+
+</tab>
+<tab title="Groovy" group-key="groovy">
+
+```kotlin
+kotlin {
+    abiValidation {
+        enabled = true
+    }
+}
+```
+
+</tab>
+</tabs>
+
+
+
+If your project has multiple modules where you want to check for binary compatibility, configure each module separately.
+
+## Check for binary compatibility issues
+
+To check for potentially binary incompatible issues after making changes to your code, run the `checkLegacyAbi` Gradle task
+in IntelliJ IDEA or use the following command in your project directory:
+
+```kotlin
+./gradlew checkLegacyAbi
+```
+
+The task compares ABI dumps and prints any detected differences as errors. Check the output carefully to see if you need to
+make changes to your code to preserve binary compatibility.
+
+## Update reference ABI dump
+
+To update the reference ABI dump that Gradle uses to check your latest changes, run the `updateLegacyAbi` task in IntelliJ 
+IDEA or use the following command in your project directory:
+
+```kotlin
+./gradlew updateLegacyAbi
+```
+
+Only update the reference dump when you're confident your changes maintain binary compatibility with the previous version.
+
+## Configure filters
+
+You can define filters to control which classes, properties, and functions the ABI dump includes. Use the `filters {}`
+block to add exclusion and inclusion rules with `excluded {}` and `included {}` blocks respectively.
+
+Gradle includes a declaration in the ABI dump only if it doesn't match any exclusion rule. When inclusion rules are defined,
+the declaration must either match one of them or have at least one member that does.
+
+A rule can be based on:
+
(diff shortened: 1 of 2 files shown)


docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,125 @@
...
+{style = "tip"}
+
+For example:
+
+```kotlin
+kotlin {
+    @OptIn(org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation::class)
+    filters {
+        excluded {
+            byNames.add("com.example.internal.InternalUtils")
```
shanshin: I think it's better to replace with `byNames.add("**.InternalUtils")` to relates description `The `InternalUtils` class.`

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,125 @@
...
+## Check for binary compatibility issues
+
+To check for potentially binary incompatible issues after making changes to your code, run the `checkLegacyAbi` Gradle task
+in IntelliJ IDEA or use the following command in your project directory:
+
+```kotlin
+./gradlew checkLegacyAbi
+```
+
+The task creates an ABI dump with errors that report any differences in your code. Check them carefully to see if you need
```
shanshin: I don't quite understand what that means. Was it meant that if there are discrepancies in the ABI, the differences will be printed?

Because the main purpose of this task is not to create a dump, but to find differences.

sarahhaggarty: Fixed in slack!

shanshin (approved):

3flex:
Are there any migration tips for those moving from https://github.com/Kotlin/binary-compatibility-validator?

Is this a drop-in replacement if moving from BCV 0.17.0 to KGP 2.2.0? Does this new feature use the same output format and location?

shanshin:
> Are there any migration tips for those moving from https://github.com/Kotlin/binary-compatibility-validator?

Since Gradle DSL for BCV in KGP is experimental in the first approximation, we do not give instructions so that existing production projects do not migrate to it. DSL will be changed in the next version.

> Is this a drop-in replacement if moving from BCV 0.17.0 to KGP 2.2.0?

There is no need to replace it now, the experimental DSL is suitable for pet projects to try out the approach and collect initial feedback.
BCV 0.17.0 still works and it can be used along with KGP 2.2.0.

> Does this new feature use the same output format and location?

The format and location match, we are trying to implement a smooth migration from a separate plug-in to the KGP, but so far not all the original features have been migrated.

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+kotlin {
+    @OptIn(org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation::class)
+    abiValidation {
+        // Use the set method to ensure compatibility with older Gradle versions
```
daniCsorbaJB:
```suggestion
        // Use the set() function to ensure compatibility with older Gradle versions
```

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,193 @@
...
+```kotlin
+./gradlew checkLegacyAbi
+```
+
+The task compares ABI dumps and prints any detected differences as errors. Check the output carefully to see if you need to
+make changes to your code to preserve binary compatibility.
+
+## Update reference ABI dump
+
+To update the reference ABI dump that Gradle uses to check your latest changes against, run the `updateLegacyAbi` task in IntelliJ
```
daniCsorbaJB: Hmm 🤔 - For some reason the "against" feels a bit weird to me here- 
Do you think something like this could work? 

To update the reference ABI dump that Gradle uses to validate your latest changes, run the...

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,193 @@
...
+
+```kotlin
+./gradlew updateLegacyAbi
+```
+
+Only update the reference dump when you're confident your changes maintain binary compatibility with the previous version.
+
+## Configure filters
+
+You can define filters to control which declarations (classes, properties, functions) the ABI dump includes. Use the `filters {}`
```
daniCsorbaJB:
```suggestion
You can define filters to control which classes, properties, and functions the ABI dump includes. Use the `filters {}`
```

daniCsorbaJB: If it's an encompassing list, we could drop the parentheses.

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,193 @@
...
+</tabs>
+
+This example:
+
+* Excludes:
+  * The `InternalUtils` class.
+  * Declarations annotated with `@InternalApi`.
+* Includes:
+  * Everything in the `com.example.api` package.
+  * Declarations annotated with the `@PublicApi`.
```
daniCsorbaJB:
```suggestion
  * Declarations annotated with `@PublicApi`.
```

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+Only update the reference dump when you're confident your changes maintain binary compatibility with the previous version.
+
+## Configure filters
+
+You can define filters to control which declarations (classes, properties, functions) the ABI dump includes. Use the `filters {}`
+block to add exclusion and inclusion rules with `excluded {}` and `included {}` blocks respectively.
+
+Gradle includes a declaration in the ABI dump only if it passes **all** exclusion and inclusion rules.
```
daniCsorbaJB: Smallest of nitpicks here. 
Passes all exclusion rules (although IMHO perfectly understandable ) might be misinterpreted by some people?

Also, do you think we should include additional info about the inclusion rule here? 🤔 (not sure if this rule applies here though: if no inclusion rules are defined everything that is not excluded is included) 

A declaration is included in the ABI dump only if it doesn't match any exclusion rule. If inclusion rules are defined, it must either match one of them or have a matching member.

What do you think? 🤔

sarahhaggarty: I see what you mean. What do you think about?

> A declaration is included in the ABI dump only if it doesn't match any exclusion rule. When inclusion rules are defined, the declaration must either match one of them or have at least one member that does.

PR #4892, opened 2025-06-11
feat: doc for llvm-passes

danil-pavlov:


danil-pavlov:
@troelsbjerre Hi Troels! Could you please check the new version? just that I didn't change the meaning while editing

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -153,7 +153,12 @@ This feature is [Experimental](components-stability.md#stability-levels-explaine
 add the `kotlin.incremental.native=true` option to your `gradle.properties` file. If you face any problems,
 create an [issue in YouTrack](https://kotl.in/issue).
 
+### Customize LLVM
```
SvyatoslavScherbina: Btw, this section is located within `## Gradle configuration`, but it is not related to Gradle.

docs/topics/native/native-llvm-passes.md
```diff
@@ -0,0 +1,89 @@
...
+taking an unacceptably long time.
+
+You can customize how Kotlin/Native uses LLVM and adjust the list of optimization passes.
+
+## Examine the build log
+
+Let's take a look at the build log to understand how much compilation time is spent on LLVM optimization passes:
+
+1. Add the `-Xprofile-phases` compiler argument to your build.
+2. Run the `linkRelease*` task. By default, a release binary compilation runs the same LLVM optimization passes that clang
```
SvyatoslavScherbina: Apparently, Gradle hides the LLVM profiling logs from the compiler output. I've submitted it as https://youtrack.jetbrains.com/issue/KT-78641.

To mitigate that, users can extract CLI compiler arguments from Gradle (similarly to how described [here](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/HACKING.md#making-the-command-line-compiler-wait-for-a-debugger)) and run the CLI compiler manually. But that's more cumbersome, so this page becomes even more "advanced". Should we still publish it?

docs/topics/native/native-llvm-passes.md
```diff
@@ -0,0 +1,89 @@
...
+In the log above, the two LLVM optimizations are `ModuleBitcodeOptimization` and `LTOBitcodeOptimization`. The formatted
+tables are the optimizations' output with timing for each pass.
+
+## Customize LLVM optimization passes
+
+If one of the passes above seems unreasonably long, you can skip it. However, this might hurt runtime performance, so
+you should check for changes in the benchmarks' performance afterward.
+
+There is no direct way to disable a given pass. However, you can provide a new list of passes to run by using the
+following compiler options:
```
SvyatoslavScherbina: > There is no direct way to disable a given pass.

Btw, this is tracked in https://youtrack.jetbrains.com/issue/KT-69212. Should we add the link here?

docs/topics/native/native-improving-compilation-time.md
danil-pavlov: My bad, let's put it on the 2nd level

docs/topics/native/native-llvm-passes.md
danil-pavlov: I think since it's already an advanced article we can add complicated instructions, and then simplify it when [KT-78641](https://youtrack.jetbrains.com/issue/KT-78641) is fixed

AFAIU it'll look like this:

1. Add the `-Xprofile-phases` compiler argument to your build.
2. To make Gradle output LLVM profiling details, use the `-Pkotlin.internal.compiler.arguments.log.level=warning` compiler argument as well.
    This extracts the actual command-line arguments passed to the Kotlin/Native compiler.
3. Run the `linkRelease*` Gradle task.
4. Examine the log.

SvyatoslavScherbina: > I think since it's already an advanced article we can add complicated instructions, and then simplify it when [KT-78641](https://youtrack.jetbrains.com/issue/KT-78641) is fixed

Ok, deal.

> AFAIU it'll look like this:

I'm not sure I understand the sequence.
I would put it like that:

First, run the `linkRelease*` Gradle task with `-Pkotlin.internal.compiler.arguments.log.level=warning`. When executing this task, it will print the compiler arguments. No need to wait till it is finished.

The task will print something like
```
> Task :linkReleaseExecutableMacosArm64
Run in-process tool "konanc"
Entry point method = org.jetbrains.kotlin.cli.utilities.MainKt.daemonMain
Classpath = [
        /Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0/konan/lib/kotlin-native-compiler-embeddable.jar
        /Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0/konan/lib/trove4j.jar
]
Arguments = [
        -Xinclude=...
        -library
        /Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0/klib/common/stdlib
        -no-endorsed-libs
        -nostdlib
        ...
]

```

Then, run the CLI compiler with the provided arguments. `/Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0` from the output is the compiler home directory. The compiler itself is `bin/kotlinc-native` in this directory. In other words, the user needs to run
```bash
/Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0/bin/kotlinc-native -Xinclude=... -library /Users/user/.konan/kotlin-native-prebuilt-macos-aarch64-2.2.0/klib/common/stdlib ...
```
If `-Xprofile-phases` is added to this CLI compiler invocation, it will print the desired output.

danil-pavlov: Thank you for the clarification! Please check the new version

PR #4896, opened 2025-06-16
Add migration advice from `kotlinOptions{}` to `compilerOptions{}`

sarahhaggarty:
This PR adds advice how to migrate from `kotlinOptions{}` to `compilerOptions{}` in Kotlin 2.2.0.

### DIFF
--- a/docs/topics/gradle/gradle-compiler-options.md
+++ b/docs/topics/gradle/gradle-compiler-options.md
@@ -16,7 +16,7 @@ in the [Working with command-line compiler](command-line.md) tutorial.
 Kotlin compilers have a number of options for tailoring the compiling process. 
 
 The Gradle DSL allows comprehensive 
-configuration of compiler options. It is available for [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html) and [JVM/Android](#target-the-jvm) projects.
+configuration of compiler options. It is available for [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html#compiler options) and [JVM/Android](#target-the-jvm) projects.
 
 With the Gradle DSL, you can configure compiler options within the build script at three levels: 
 * **[Extension level](#extension-level)**, in the `kotlin {}` block for all targets and shared source sets.
@@ -99,7 +99,7 @@ kotlin {
         val main by compilations.getting {
             compileTaskProvider.configure {
                 compilerOptions {
-
+                    optIn.add("kotlin.RequiresOptIn")
                 }
             }
         }
@@ -136,6 +136,120 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
 </tab>
 </tabs>
 
+### Migrate from `kotlinOptions{}` to `compilerOptions{}` {collapsible="true"}
+
+Before Kotlin 2.2.0, you could configure compiler options using the `kotlinOptions{}` block. Since the `kotlinOptions{}`
+block is deprecated in Kotlin 2.2.0, this section provides guidance and recommendations for migrating your build
+scripts to use the `compilerOptions{}` block instead.
+
+#### Centralize compiler options and use types
+
+Whenever possible, configure compiler options at the [extension level](#extension-level), and override them for specific tasks
+at [compilation unit level](#compilation-unit-level).
+
+To make your configuration safer and more expressive, avoid using raw strings and prefer strongly typed values. For example, if you have:
+
+```kotlin
+plugins {
+    kotlin("jvm") version "2.2.0"
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+    kotlinOptions {
+        jvmTarget = "17"
+        languageVersion = "2.2"
+        apiVersion = "2.2"
+    }
+}
+```
+
+After migration, this becomes:
+
+```kotlin
+plugins {
+    kotlin("jvm") version "2.2.0"
+}
+
+kotlin {
+    // Extension level
+    compilerOptions {
+        jvmTarget = JvmTarget.fromTarget("17")
+        languageVersion = KotlinVersion.fromVersion("2.2")
+        apiVersion = KotlinVersion.fromVersion("2.2")
+    }
+}
+
+// Override at compilation unit level
+tasks.named<KotlinJvmCompile>("compileKotlin"){
+    compilerOptions {
+        apiVersion = KotlinVersion.fromVersion("2.0")
+    }
+}
+```
+
+#### Migrate away from `android.kotlinOptions`
+
+If your build script previously used `android.kotlinOptions`, migrate to `kotlin.compilerOptions` instead. Either at
+the extension level or the target level.
+
+For example, if you have:
+
+```kotlin
+plugins {
+    id("com.android.application")
+    kotlin("android")
+}
+
+android {
+    kotlinOptions {
+        jvmTarget = "17"
+    }
+}
+```
+
+Update it to:
+
+```kotlin
+plugins {
+    id("com.android.application")
+    kotlin("android")
+}
+
+kotlin {
+    androidTarget {
+        // Target level
+        compilerOptions {
+            jvmTarget = JvmTarget.fromTarget("17")
+        }
+    }
+}
+```
+
+#### Migrate `freeCompilerArgs`
+
+* Replace all `+=` operations with `add()` or `addAll()` functions.
+* If you use the `-opt-in` compiler option, check whether a specialized DSL already is available in the [KGP API reference](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/) and use that instead.
+
+For example, if you have:
+
+```kotlin
+kotlinOptions {
+    freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
+    freeCompilerArgs += listOf("-Xcontext-receivers", "-Xinline-classes")
+}
+```
+
+Migrate to:
+
+```kotlin
+kotlin {
+    compilerOptions {
+        optIn.add("kotlin.RequiresOptIn")
+        freeCompilerArgs.addAll(listOf("-Xcontext-receivers", "-Xinline-classes"))
+    }
+}
+```
+
 ## Target the JVM
 
 [As explained before](#how-to-define-options), you can define compiler options for your JVM/Android projects at the extension, target, and compilation unit levels (tasks).
@@ -148,7 +262,6 @@ and searching for `compile*Kotlin` task names in the `Other tasks` group.
 
 Some important details to be aware of:
 
-* The `android.kotlinOptions` and `kotlin.compilerOptions` configuration blocks override each other. The last (lowest) block takes effect.
 * `kotlin.compilerOptions` configures every Kotlin compilation task in the project.
 * You can override the configuration applied by `kotlin.compilerOptions` DSL using the `tasks.named<KotlinJvmCompile>("compileKotlin") { }`
   (or `tasks.withType<KotlinJvmCompile>().configureEach { }`) approach.


docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,120 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
 </tab>
 </tabs>
 
+### Migrate from `kotlinOptions{}` to `compilerOptions{}` {collapsible="true"}
+
+Before Kotlin 2.2.0, you could configure compiler options using the `kotlinOptions{}` block. Since the `kotlinOptions{}`
```
Tapchicoma: `kotlinOptions {}` DSL was deprecated in Kotlin 2.0.0 release. Deprecation level was raised to `ERROR` in Kotlin 2.2.0 release.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,120 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+block is deprecated in Kotlin 2.2.0, this section provides guidance and recommendations for migrating your build
+scripts to use the `compilerOptions{}` block instead.
+
+#### Centralize compiler options and use types
+
+Whenever possible, configure compiler options at the [extension level](#extension-level), and override them for specific tasks
+at [compilation unit level](#compilation-unit-level).
+
+To make your configuration safer and more expressive, avoid using raw strings and prefer strongly typed values. For example, if you have:
```
Tapchicoma: It is not possible to use raw string directly in the new DSL. Maybe better to rephrase here to "avoid using, but migration helpers are available"

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,120 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+            jvmTarget = JvmTarget.fromTarget("17")
+        }
+    }
+}
+```
+
+#### Migrate `freeCompilerArgs`
+
+* Replace all `+=` operations with `add()` or `addAll()` functions.
+* If you use the `-opt-in` compiler option, check whether a specialized DSL already is available in the [KGP API reference](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/) and use that instead.
```
Tapchicoma, line 413: We could mention here as well `-progressive` argument.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,120 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+Update it to:
+
+```kotlin
+plugins {
+    id("com.android.application")
+    kotlin("android")
+}
+
+kotlin {
+    androidTarget {
```
Tapchicoma, line 381: `androidTarget` is not available in `kotlin-android` plugin. So it should be just extension level example

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: I didn't know that. In this case, I think we should be direct about it. What do you think?

> You can't use raw strings in the `compilerOptions {}` block, so convert them to typed values.

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: I tried to simplify this information with this sentence. Technically you can use `kotlinOptions {}` with Kotlin 2.0.0 but you will get warnings. From 2.2.0, it's not possible because it triggers deprecation errors. Do you think there's a benefit to being more explicit? The compat. guide will contain this info and a link to this section.

Tapchicoma: Maybe just change `is deprecated in Kotlin 2.2.0` to `is deprecated since Kotlin 2.0.0`

docs/topics/gradle/gradle-compiler-options.md
Tapchicoma: sounds good to me

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+
+```kotlin
+plugins {
+    kotlin("jvm") version "2.2.0"
+}
+
+kotlin {
+    // Extension level
+    compilerOptions {
+        jvmTarget = JvmTarget.fromTarget("17")
```
Tapchicoma: Below you are using the fully qualified name and here it is not. Better to be consistent and use FQN everywhere for these types.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+</tabs>
+
+After migration, this becomes:
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+plugins {
+    kotlin("jvm") version "2.2.0"
```
Tapchicoma: BTW do we update this version automatically on each new Kotlin release?

sarahhaggarty: Added variables!

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -16,7 +16,7 @@ in the [Working with command-line compiler](command-line.md) tutorial.
 Kotlin compilers have a number of options for tailoring the compiling process. 
 
 The Gradle DSL allows comprehensive 
-configuration of compiler options. It is available for [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html) and [JVM/Android](#target-the-jvm) projects.
+configuration of compiler options. It is available for [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html#compiler options) and [JVM/Android](#target-the-jvm) projects.
```
zamulla:
```suggestion
configuration of compiler options. It is available for [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html#compiler-options) and [JVM/Android](#target-the-jvm) projects.
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
 </tab>
 </tabs>
 
+### Migrate from `kotlinOptions {}` to `compilerOptions {}` {collapsible="true"}
+
+Before Kotlin 2.2.0, you could configure compiler options using the `kotlinOptions {}` block. Since the `kotlinOptions {}`
+block is deprecated since Kotlin 2.0.0, this section provides guidance and recommendations for migrating your build
```
zamulla: Nitpick: the double "since" could be avoided probably.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+scripts to use the `compilerOptions {}` block instead:
+
+* [Centralize compiler options and use types](#centralize-compiler-options-and-use-types)
+* [Migrate away from `android.kotlinOptions`](#migrate-away-from-android-kotlinoptions)
+* [Migrate `freeCompilerArgs`](#migrate-freecompilerargs)
+
+#### Centralize compiler options and use types
+
+Whenever possible, configure compiler options at the [extension level](#extension-level), and override them for specific tasks
+at [compilation unit level](#compilation-unit-level).
```
zamulla:
```suggestion
at the [compilation unit level](#compilation-unit-level).
```

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+        languageVersion = '%languageVersion%'
+        apiVersion = '%apiVersion%'
+    }
+}
+```
+
+</tab>
+</tabs>
+
+After migration, this becomes:
```
zamulla:
```suggestion
After migration, it should be:
```

Just to stress that it doesn't automatically become anything.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+    // Extension level
+    compilerOptions {
+        jvmTarget = JvmTarget.fromTarget("%jvmLTSVersionSupportedByKotlin%")
+        languageVersion = KotlinVersion.fromVersion("%languageVersion%")
+        apiVersion = KotlinVersion.fromVersion("%apiVersion%")
+    }
+}
+
+// Override at compilation unit level
+tasks.named<KotlinJvmCompile>("compileKotlin"){
```
zamulla, line 213: If feel like we should comment on the fact that there is no code corresponding to this in the "before" example. Or I'm having a hard time see that there is, and it's another problem :)

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+
+</tab>
+</tabs>
+
+#### Migrate `freeCompilerArgs`
+
+* Replace all `+=` operations with `add()` or `addAll()` functions.
+* If you use the `-opt-in` compiler option, check whether a specialized DSL already is available in the [KGP API reference](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/) and use that instead.
+* Migrate any use of the `-progressive` compiler option to use the dedicated DSL: `progressiveMode.set(true)`.
+* Migrate any use of the `-Xjvm-default-all` compiler option to [use the dedicated DSL](gradle-compiler-options.md#attributes-specific-to-jvm): `jvmDefault.set()`.
```
zamulla:
```suggestion
* Migrate any use of the `-Xjvm-default` compiler option to [use the dedicated DSL](gradle-compiler-options.md#attributes-specific-to-jvm): `jvmDefault.set()`.
```

Is that right? Or should it be `Xjvm-default=all`?

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -136,6 +136,340 @@ tasks.named('compileKotlin', org.jetbrains.kotlin.gradle.tasks.KotlinCompilation
...
+</tab>
+<tab title="Groovy" group-key="groovy">
+
+```kotlin
+kotlin {
+    compilerOptions {
+        optIn.add("kotlin.RequiresOptIn")
+        freeCompilerArgs.addAll(["-Xcontext-receivers", "-Xinline-classes"])
+        progressiveMode.set(true)
+        jvmDefault.set(org.jetbrains.kotlin.gradle.dsl.JvmDefaultMode.NO_COMPATIBILITY)
```
zamulla: This may be a me problem, but I didn't find the mapping of old options to new options at all transparent :)
Not sure how it can be made easier to understand.

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: I updated the code comment to say:
`// Example of overriding at compilation unit level`
Do you think that makes it clear enough?

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: I added code comments in the "after" block to try and make it clearer what has changed. Does it help?

zamulla (approved):
Awesome, thanks 👍

PR #4903, opened 2025-06-19
Update kotlin-tour-intermediate-lambdas-receiver.md

Geek-Joe:
"receiver type" is right here


sarahhaggarty:
Thanks for your contribution @Geek-Joe !

PR #4904, opened 2025-06-19
update: remove compatiblilty modes

danil-pavlov:


danil-pavlov:
@koshachy let me add a redirect instead of `accepts-web-file-names` because I don't think it's possible to do something with short links

PR #4907, opened 2025-06-20
Update kotlin-tour-intermediate-scope-functions.md

Geek-Joe:
error message is wrong, and String? isn't inferred type in this case


sarahhaggarty (approved):

sarahhaggarty:
The error message is correct. However, it's what is shown in IntelliJ IDEA. The tour is intended to be completed in Kotlin Playground so I've accepted your suggestion to show the message seen in Playground. Thanks for your contribution @Geek-Joe !

PR #4908, opened 2025-06-22
Update kotlin-tour-intermediate-classes-interfaces.md

Geek-Joe:
use inherited function `.toString()`  in the example


sarahhaggarty:
The `.toString()` function is used under the hood when using string templates. So in this case, I won't accept your suggestion. But thank you for reaching out!

Geek-Joe:
It does call `.toString()` implicitly here, but I think examples in tour should be ***intuitive***.  One may not realize this implicit call as a Kotlin **learner**(who this tour targets for), of course you can also add some comments to explain.   
Anyway, thanks for your reply. @sarahhaggarty

PR #4909, opened 2025-06-22
Update kotlin-tour-intermediate-open-special-classes.md

Geek-Joe:
call to the constructor

### DIFF
--- a/docs/topics/tour/kotlin-tour-intermediate-open-special-classes.md
+++ b/docs/topics/tour/kotlin-tour-intermediate-open-special-classes.md
@@ -30,7 +30,7 @@ To create a class that inherits from another, add a colon after your class heade
 of the parent class that you want to inherit from:
 
 ```kotlin
-class Car : Vehicle
+class Car : Vehicle()
 ```
 {validate="false"}
 


docs/topics/tour/kotlin-tour-intermediate-open-special-classes.md
```diff
@@ -30,7 +30,7 @@ To create a class that inherits from another, add a colon after your class heade
 of the parent class that you want to inherit from:
 
 ```kotlin
```
sarahhaggarty: After thinking about it, I'd like to remove this code snippet and bring the example sooner to make it clearer that the constructor is called. Can you apply my below suggestions?

````suggestion
If you can't use interfaces or abstract classes, you can explicitly make a class inheritable by declaring it as **open**.
To do this, use the `open` keyword before your class declaration:

```kotlin
open class Vehicle(val make: String, val model: String)
```

To create a class that inherits from another, add a colon after your class header followed by a call to the constructor
of the parent class that you want to inherit from. In this example, the `Car` class inherits from the `Vehicle` class:

```kotlin
open class Vehicle(val make: String, val model: String)

class Car(make: String, model: String, val numberOfDoors: Int) : Vehicle(make, model)

fun main() {
    // Creates an instance of the Car class
    val car = Car("Toyota", "Corolla", 4)

    // Prints the details of the car
    println("Car Info: Make - ${car.make}, Model - ${car.model}, Number of doors - ${car.numberOfDoors}")
    // Car Info: Make - Toyota, Model - Corolla, Number of doors - 4
}
```
````

sarahhaggarty:
Thanks for your contribution @Geek-Joe !

PR #4910, opened 2025-06-23
Fix incorrect usage of singleOrNull() in null safety example

TarikYarasli:
This PR fixes a logic error in the Kotlin Intermediate Tour's null safety example.

The original code used `singleOrNull()` without a predicate, which incorrectly checked whether the list had only one temperature value.

Replaced it with `singleOrNull { it == 30 }` to correctly check if the temperature 30 occurred exactly once, as described in the comment.

This makes the code behavior match its intention and provides a clearer learning example.


sarahhaggarty:
Hi there!
Thanks for pointing this out! Since the predicate introduces the concept of lambda expressions, I decided to move the `singleOrNull()` function out of this code sample and dedicate a new one to it. It's covered as part of #4938.
Thanks again, and I hope you'll contribute again in the future!

PR #4920, opened 2025-06-24
update: returns feedback

danil-pavlov:
* [KT-78080](https://youtrack.jetbrains.com/issue/KT-78080) Confusing example with non-local returns

### DIFF
--- a/docs/topics/returns.md
+++ b/docs/topics/returns.md
@@ -111,7 +111,8 @@ fun main() {
 
 Note that the use of local returns in the previous three examples is similar to the use of `continue` in regular loops.
 
-There is no direct equivalent for `break`, but it can be simulated by adding another nesting lambda and non-locally returning from it:
+There is no direct equivalent for `break`, but it can be simulated by adding another nesting lambda and non-locally
+returning from it, as long as the lambda is passed to an [inline function](inline-functions.md):
 
 ```kotlin
 //sampleStart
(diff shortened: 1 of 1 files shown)


docs/topics/returns.md
```diff
@@ -111,7 +111,8 @@ fun main() {
 
 Note that the use of local returns in the previous three examples is similar to the use of `continue` in regular loops.
 
-There is no direct equivalent for `break`, but it can be simulated by adding another nesting lambda and non-locally returning from it:
+There is no direct equivalent for `break`, but it can be simulated by adding another nesting lambda and non-locally
+returning from it, as long as the lambda is passed to an [inline function](inline-functions.md):
```
nikitabobko: > but it can be simulated by adding another nesting lambda and non-locally returning from it:

But the lambda we are adding is outer, not nested

Nested is the `forEach` lambda. Outer is the `run` lambda. The difference from the previous example is the addition of forEach lambda.

---

Please note that if you simply update the sentence to say "by adding another outer lambda", the next part of the sentence becomes incorrect:

> as long as the lambda is passed to an [inline function](inline-functions.md):

Because *the lambda* no longer refers to nested lambda, but it refers to the outer lambda, and the outer lambda doesn't have to be inline. Only the nested lambda has to be.

docs/topics/returns.md
```diff
@@ -132,6 +133,8 @@ fun main() {
 ```
 
+Here, both `forEach()` extension and `run` scope function are inline functions from the Kotlin standard library.
```
nikitabobko: It's a technically correct statement, but given that `run` doesn't have to be inline, the statement might be misleading

docs/topics/returns.md
danil-pavlov: Thank you for the clarification! Updated both sentences

nikitabobko (approved):
Thanks!

PR #4925, opened 2025-06-26
kotlin-tour-functions.md : fix filter predicate explanation

julienbraure:
The predicate doesn't loop, .filter() does
The predicate doesn't do the filtering, .filter() does

### DIFF
--- a/docs/topics/tour/kotlin-tour-functions.md
+++ b/docs/topics/tour/kotlin-tour-functions.md
@@ -396,11 +396,10 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-tour-lambda-filter"}
 
-The `.filter()` function accepts a lambda expression as a predicate that will be used on each element of the list.
-The element will be kept only if the predicate returns true:
+The `.filter()` function accepts a lambda expression as a predicate and applies it to each element of the list. The function keeps an element only if the predicate returns `true`:
 
-* `{ x -> x > 0 }` for a given element, returns true if is positive.
-* `{ x -> x < 0 }` for a given element, returns true if is negative.
+* `{ x -> x > 0 }` returns `true` if the element is positive.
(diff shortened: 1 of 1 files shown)


docs/topics/tour/kotlin-tour-functions.md
```diff
@@ -396,10 +396,11 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-tour-lambda-filter"}
 
-The `.filter()` function accepts a lambda expression as a predicate:
+The `.filter()` function accepts a lambda expression as a predicate that will be used on each element of the list.
+The element will be kept only if the predicate returns true:
```
sarahhaggarty:
```suggestion
The `.filter()` function accepts a lambda expression as a predicate and applies it to each element of the list. The function keeps an element only if the predicate returns `true`:
```

docs/topics/tour/kotlin-tour-functions.md
```diff
@@ -396,10 +396,11 @@ fun main() {
...
 
-* `{ x -> x > 0 }` takes each element of the list and returns only those that are positive.
-* `{ x -> x < 0 }` takes each element of the list and returns only those that are negative.
+* `{ x -> x > 0 }` for a given element, returns true if is positive.
+* `{ x -> x < 0 }` for a given element, returns true if is negative.
```
sarahhaggarty:
```suggestion
* `{ x -> x > 0 }` returns `true` if the element is positive.
* `{ x -> x < 0 }` returns `true` if the element is negative.
```

sarahhaggarty (changes requested):
Thanks for your suggestion @julienbraure ! I've suggested some changes to fix grammar errors and use present tense as per our documentation guidelines. Can you apply my suggestions?

julienbraure:
Thanks, suggestions applied

sarahhaggarty:
Thanks for your contribution @julienbraure !

PR #4932, opened 2025-06-27
Add extra information about ES modules inside the js-interop document

JSMonk:


docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -33,8 +34,9 @@ Check the article on [JavaScript Modules](js-modules.md) for more information on
 
 ## Package structure
 
-Kotlin exposes its package structure to JavaScript, so unless you define your declarations in the root package,
-you have to use fully qualified names in JavaScript. For example:
+For most of the module systems, Kotlin exposes its package structure to JavaScript,
```
lppedd: I guess it's worth mentioning which module systems, since Kotlin supports four only (not considering PLAIN).  
Plus, you say ESM is an exception, so CJS, AMD and UMD are the one with the package structure.

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -54,6 +56,15 @@ Or, in the case of using `plain` as a module system setting:
 alert(myModule.my.qualified.packagename.foo());
 ```
 
+However, there is an exception for `ES` modules.
```
lppedd: If you apply my suggestion above, you could reword this one to

> The ECMAScript Modules target instead loses package information, to improve the package' bundle size and to match the typical layout of ESM packages.

JSMonk: Applied your suggestions

AlejandraPedroza (commented):
Thanks for working on this. I made some minor style changes. Can you please check the new version?

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -14,46 +14,58 @@ all declarations are available to JavaScript via the `myModule` object. For exam
...
 
 ```javascript
 alert(myModule.foo());
 ```
 
-This is not applicable when you compile your Kotlin module to JavaScript modules like UMD (which is the default setting
-for both `browser` and `nodejs` targets), CommonJS or AMD. In this case, your declarations will be exposed in the format
-specified by your chosen JavaScript module system. When using UMD or CommonJS, for example, your call site could look
+Calling the function directly like this is not applicable when you compile your Kotlin module to JavaScript modules like [UMD](https://github.com/umdjs/umd) (the default setting
+for both `browser` and `nodejs` targets), [ESM](https://tc39.es/ecma262/#sec-modules), [CommonJS](https://nodejs.org/api/modules.html#modules-commonjs-modules), or [AMD](https://github.com/amdjs/amdjs-api/wiki/AMD).
```
AlejandraPedroza, line 24: I added references to the module systems. Can you check if they're correct?

JSMonk: Yes, they are. Thank you 🙏

PR #4935, opened 2025-06-28
Fixed incorrect description in delegation practice exercise

alanloi:
Updated the exercise documentation on delegation to correctly refer to the `Messenger` interface instead of the `SmartMessenger` class.

Trying to implement the exercise as is will give the following compilation error and lead to user confusion:
```
Delegation is supported only for interfaces.
This type is final, so it cannot be extended.
This type has a constructor, so it must be initialized here.
```


sarahhaggarty:
Great catch! Thanks for your contribution @alanloi !

PR #4936, opened 2025-06-29
Fixed singleOrNull example code snippet that was missing predicate

alanloi:
Updated the singleOrNull code snippet example to include missing predicate that checks for target value of 30.


sarahhaggarty:
Hi there!
Thanks for pointing this out! Since the predicate introduces the concept of lambda expressions, I decided to move the `singleOrNull()` function out of this code sample and dedicate a new one to it. It's covered as part of #4938.
Thanks again, and I hope you'll contribute again in the future!

PR #4947, opened 2025-07-09
update: clarify @JvmExposeBoxed behavior for inherited functions

daniCsorbaJB:
This update adds a note to clarify `@JvmExposeBoxed` behavior for inherited functions.

Related ticket: [KT-78928](https://youtrack.jetbrains.com/issue/KT-78928)

### DIFF
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -640,26 +640,26 @@ MyInt output = ExampleKt.timesTwoBoxed(input);
 To apply this behavior to all inline value classes and the functions that use them within a module, compile it with the `-Xjvm-expose-boxed` option. 
 Compiling with this option has the same effect as if every declaration in the module has the `@JvmExposeBoxed` annotation.
 
-> The `@JvmExposeBoxed` annotation doesn't automatically generate boxed representations for inherited functions.
-> 
-> To generate the necessary representation for an inherited function, override it in the implementing or extending class:
-> 
-> ```kotlin
-> interface Test {
->     fun test(p: UInt): UInt = p
-> }
->
-> // Doesn't generate a boxed representation for the test() function
-> @OptIn(ExperimentalStdlibApi::class)
-> @JvmExposeBoxed
-> class TestClass : Test
-> 
-> // Generates a boxed representation for the test() function
-> @OptIn(ExperimentalStdlibApi::class)
-> @JvmExposeBoxed
-> class TestClass : Test {
->     override fun test(p: UInt): UInt = super.test(p)
-> }
-> ```
-> 
-{style="note"}
+### Inherited functions
+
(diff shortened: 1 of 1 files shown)


docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -639,3 +639,27 @@ MyInt output = ExampleKt.timesTwoBoxed(input);
 
 To apply this behavior to all inline value classes and the functions that use them within a module, compile it with the `-Xjvm-expose-boxed` option. 
 Compiling with this option has the same effect as if every declaration in the module has the `@JvmExposeBoxed` annotation.
+
+> The `@JvmExposeBoxed` annotation doesn't automatically generate boxed representations for inherited functions.
```
sarahhaggarty: Instead of a note, can you make this into a subsection about "Inherited functions"?

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -639,3 +639,27 @@ MyInt output = ExampleKt.timesTwoBoxed(input);
...
+> 
+> To generate the necessary representation for an inherited function, override it in the implementing or extending class:
+> 
+> ```kotlin
```
sarahhaggarty: In the example, can you add names that reflect a real-world example a bit more? Maybe like:
```
interface IdTransformer {
    fun transformId(rawId: UInt): UInt = rawId
}

// Doesn't generate a boxed representation for the transformId() function
@OptIn(ExperimentalStdlibApi::class)
@JvmExposeBoxed
class LightweightTransformer : IdTransformer

// Generates a boxed representation for the transformId() function
@OptIn(ExperimentalStdlibApi::class)
@JvmExposeBoxed
class DefaultTransformer : IdTransformer {
    override fun transformId(rawId: UInt): UInt = super.transformId(rawId)
}
```
And is there a way to avoid using `super` ? Because we don't explain it on this page 🤔

sarahhaggarty (changes requested):
Nice addition! But I have some suggestions for you to consider :)

docs/topics/jvm/java-to-kotlin-interop.md
daniCsorbaJB: sure thing 👍

docs/topics/jvm/java-to-kotlin-interop.md
daniCsorbaJB: I'm not sure - the issue is that if we do it without `super`,  we still have to override the function (probably by implementing the same logic) but I think that way we would recommend a wrong approach to our users. 🤔 (if they just use the same logic in their own codebase they might change the original function, and the overridden function will not be in synch)

Regarding the names — sure thing 👍 I like it 😄

sarahhaggarty (approved):
LGTM! 🚀

PR #4949, opened 2025-07-10
ktl-2512 Server-Side LP: Dynamic Components  

nikpachoo:
This is a draft version of the components we need for the updated lp.


nikpachoo:
The changes were cherry-picked to the feature branch.

PR #4951, opened 2025-07-14
Update tour statement about variables

sarahhaggarty:
This PR updates the tour to explain why we recommend using `val` by default.


saiful148 (commented):
all

saiful148:
New

saiful148:
Upgrade

PR #4952, opened 2025-07-14
Add Maven instructions to releases page

sarahhaggarty:
This PR adds Maven instructions for updating to a new Kotlin release.

### DIFF
--- a/docs/topics/releases.md
+++ b/docs/topics/releases.md
@@ -31,8 +31,11 @@ new features before they are released. See [Early Access Preview](eap.md) for de
 
 ## Update to a new Kotlin version
 
-To upgrade your project to a new release, you need to update your build script file.
-For example, to update to Kotlin %kotlinVersion%, change the version of the Kotlin Gradle plugin in your
+To upgrade your project to a new release, update the Kotlin version in your build system.
+
+### Gradle
+
+To update to Kotlin %kotlinVersion%, change the version of the Kotlin Gradle plugin in your
 `build.gradle(.kts)` file:
 
 <tabs group="build-script">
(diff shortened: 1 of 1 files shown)


docs/topics/releases.md
```diff
@@ -66,10 +72,31 @@ plugins {
...
+> To learn more about how to work with Gradle in your project, see [Configure a Gradle project](gradle-configure-project.md).
+> 
+{style="tip"}
+
+### Maven
+
+To update to Kotlin %kotlinVersion%, change the version in your `pom.xml` file:
+
+```xml
+<properties>
```
Tapchicoma, line 86: I've noticed that our generated by wizard Maven project does not add `<kotlin.version>` in properties.

How about also adding an alternative way here:
or update the plugin version in `<plugins>` block:
```xml
    <build>
        <plugins>
            <plugin>
                <groupId>org.jetbrains.kotlin</groupId>
                <artifactId>kotlin-maven-plugin</artifactId>
                <version>%kotlinVersion%</version>
            </plugin>
        </plugins>
    </build>
```

docs/topics/releases.md
```diff
@@ -31,8 +31,14 @@ new features before they are released. See [Early Access Preview](eap.md) for de
 
 ## Update to a new Kotlin version
 
-To upgrade your project to a new release, you need to update your build script file.
-For example, to update to Kotlin %kotlinVersion%, change the version of the Kotlin Gradle plugin in your
+To upgrade your project to a new release, update the Kotlin version in your build system:
+
+* [Gradle](#gradle)
```
daniCsorbaJB: (suggestion — it's fine as it is)
Do you think we need to list these as a UL? 🤔 It's immediately followed by Gradle and Maven sections, so I think we could omit this part (for example, in [Metadata ](https://kotlinlang.org/docs/metadata-jvm.html#add-the-library-to-your-project).

docs/topics/releases.md
```diff
@@ -66,10 +72,43 @@ plugins {
 </tab>
 </tabs>
 
-If you have projects created with earlier Kotlin versions, change the Kotlin version in your projects and update kotlinx
-libraries if necessary.
+If you have projects created with earlier Kotlin versions, check if you need to also [update the version of any kotlinx
```
daniCsorbaJB: (suggestion)
How about ...check if you also need to? 
I don't see much of a difference, so whichever you think is better.

docs/topics/releases.md
```diff
@@ -66,10 +72,43 @@ plugins {
...
+libraries](gradle-configure-project.md#set-a-dependency-on-a-kotlinx-library).
+
+If you are migrating to a new language release, the Kotlin plugin's migration tools will help you with the migration.
```
daniCsorbaJB: (suggestion)
I know this was already here, but do you think it would make sense like this? (just to avoid saying migration 3 times in one sentence, but if we want to ensure 100% clarity it works — so whichever you think is beetter)

> If you are migrating to a new language release, the Kotlin plugin's migration tools will help you with the process.

docs/topics/releases.md
```diff
@@ -66,10 +72,43 @@ plugins {
...
+    <plugin>
+        <groupId>org.jetbrains.kotlin</groupId>
+        <artifactId>kotlin-maven-plugin</artifactId>
+        <version>%kotlinVersion%</version>
+    </plugin>
+</plugins>
+```
+
+If you have projects created with earlier Kotlin versions, check if you need to also [update the version of any kotlinx
+libraries](maven.md#set-dependencies).
```
daniCsorbaJB, line 104: (suggestion)
This is probably out of scope for this update, but what do you think; could we add information about other libraries to that Maven page aside from the standard library 🤔 

Right now this points to:

> Kotlin has an extensive standard library that can be used in your applications. To use the standard library in your project, add the following dependency to your pom.xml file:

Could we change it to something like:

> Kotlin has an extensive selection of libraries that can be used in your applications. For example, to use the standard library in your project, add the following dependency to your pom.xml file:

sarahhaggarty: Nice idea! Andrey has a PR open on the Maven page so I'll suggest it to him :)

PR #4967, opened 2025-07-21
Add Playwright tests for Community page (overview tab).

solarday:


test/production/community-overview-rest-buttons.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Community page, overview tab, rest buttons', () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto('/community/');
+        await page.waitForSelector('button.ch2-btn.ch2-btn-primary');
```
berezinant, line 6: Should we introduce "tryCloseCookiesBanner" or something here? @zoobestik

I see that existing `closeCookiesConsentBanner` from the /utils might not do exactly what is needed here, but I could refactor it

berezinant (approved):
Tests are cool!

The only question is about closing cookies banner consent, but it's non-blocking

PR #4970, opened 2025-07-24
Playground/ezerskii refactor documentation builds

Quattro8:
- Updated Writerside Builder to use the custom frontend parameter.
- Added an abstract class for building documentation with the updated Writerside Builder.
- Added a new documentation subproject.
- Added Kotlin Reference builds for Coroutines and Kotlin Multiplatform documentation.
- Removed dependency from the WH documentation builds.
- Removed the Reference docs build configuration and its dependencies.


.teamcity/kotlinlang/builds/BuildReferenceDocs.kt
```diff
@@ -1,64 +0,0 @@
-package kotlinlang.builds
```
zoobestik, line 1: @Quattro8 it would be nice to announce changes for techwrighters, is it difficult do it in two iteration? Or keep it as archive project with comment before deletion?

zoobestik (approved):

PR #4976, opened 2025-07-28
docs: fix minor typo in Email inline class explanation

sebaek:
This is a small typo fix in the documentation example.  
The explanation referred to a variable as `email`, while the actual code uses `myEmail`.


sarahhaggarty:
Thanks for your contribution @sebaek !

PR #4979, opened 2025-07-28
Improve Conditions and loops page

sarahhaggarty:
This PR improves the Conditions and loops page by improving the quality of the examples by making them more reflective of the real world, amongst other things.

### DIFF
--- a/docs/topics/control-flow.md
+++ b/docs/topics/control-flow.md
@@ -1,82 +1,96 @@
 [//]: # (title: Conditions and loops)
 
+Kotlin gives you flexible tools to control your program's flow. Use `if`, `when`, and loops to define clear, 
+expressive logic for your conditions.
+
 ## If expression
 
-In Kotlin, `if` is an expression: it returns a value.
-Therefore, there is no ternary operator (`condition ? then : else`) because ordinary `if` works fine in this role.
+To use `if` in Kotlin, add the condition to check within parentheses `()` and the action to take if the result is true
+within curly braces `{}`. You can use `else` and `else if` for additional branches and checks.
+
+You can also write `if` as an expression, which lets you assign its returned value directly to a variable.
+In this form, an `else` branch is required. The `if` expression serves the same purpose as the ternary operator 
+(`condition ? then : else`) found in other languages.
+
+For example:
 
 ```kotlin
 fun main() {
-    val a = 2
-    val b = 3
+    val heightAlice = 160
+    val heightBob = 175
 
     //sampleStart
-    var max = a
-    if (a < b) max = b
+    var taller = heightAlice
+    if (heightAlice < heightBob) taller = heightBob
 
-    // With else
-    if (a > b) {
-      max = a
+    // Uses an else branch
+    if (heightAlice > heightBob) {
+        taller = heightAlice
     } else {
-      max = b
+        taller = heightBob
     }
 
-    // As expression
-    max = if (a > b) a else b
+    // Uses if as an expression
+    taller = if (heightAlice > heightBob) heightAlice else heightBob
 
-    // You can also use `else if` in expressions:
-    val maxLimit = 1
-    val maxOrLimit = if (maxLimit > a) maxLimit else if (a > b) a else b
-  
-    println("max is $max")
-    // max is 3
-    println("maxOrLimit is $maxOrLimit")
-    // maxOrLimit is 3
+    // Uses else if as an expression:
+    val heightLimit = 150
+    val heightOrLimit = if (heightLimit > heightAlice) heightLimit else if (heightAlice > heightBob) heightAlice else heightBob
+
+    println("Taller height is $taller")
+    // Taller height is 175
+    println("Height or limit is $heightOrLimit")
+    // Height or limit is 175
     //sampleEnd
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="if-else-if-kotlin"}
 
-Branches of an `if` expression can be blocks. In this case, the last expression is the value of a block:
-
-```kotlin
-val max = if (a > b) {
-    print("Choose a")
-    a
-} else {
-    print("Choose b")
-    b
-}
-```
-
-If you're using `if` as an expression, for example, for returning its value or
-assigning it to a variable, the `else` branch is mandatory.
-
-## When expressions and statements
-
-`when` is a conditional expression that runs code based on multiple possible values or conditions. It is
-similar to the `switch` statement in Java, C, and similar languages. For example:
+Each branch in an `if` expression can be a block, where the value of the last expression becomes the result:
 
 ```kotlin
 fun main() {
     //sampleStart
-    val x = 2
-    when (x) {
-        1 -> print("x == 1")
-        2 -> print("x == 2")
-        else -> print("x is neither 1 nor 2")
+    val heightAlice = 160
+    val heightBob = 175
+
+    val taller = if (heightAlice > heightBob) {
+        print("Choose Alice\n")
+        heightAlice
+    } else {
+        print("Choose Bob\n")
+        heightBob
     }
-    // x == 2
+
+    println("Taller height is $taller")
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="if-else-blocks-kotlin"}
+
+## When expressions and statements
+
+`when` is a conditional expression that runs code based on multiple possible values or conditions. It's
+similar to the `switch` statement in Java, C, and other languages. `when` evaluates its argument and compares the result
+against each branch in order until one branch condition is satisfied. For example:
+
+```kotlin
+fun main() {
+    //sampleStart
+    val userRole = "Editor"
+    when (userRole) {
+        "Viewer" -> print("User has read-only access")
+        "Editor" -> print("User can edit content")
+        else -> print("User role is not recognized")
+    }
+    // User can edit content
     //sampleEnd
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-conditions-when-statement"}
 
-`when` matches its argument against all branches sequentially until some branch condition is satisfied.
-
-You can use `when` in a few different ways. Firstly, you can use `when` either as an **expression** or as a **statement**.
-As an expression, `when` returns a value for later use in your code. As a statement, `when` completes an action
-without returning anything of further use:
+You can use `when` either as an **expression** or a **statement**. As an expression, `when` returns a value you can use
+later in your code. As a statement, `when` completes an action without returning a result:
 
 <table>
    <tr>
@@ -100,7 +114,7 @@ val text = when (x) {
 <td>
 
 ```kotlin
-// Returns nothing but triggers a 
+// Returns no result but triggers a 
 // print statement
 when (x) {
     1 -> print("x == 1")
@@ -113,9 +127,8 @@ when (x) {
 </tr>
 </table>
 
-Secondly, you can use `when` with or without a subject. Whether you use a subject with `when` or not, your expression or
-statement behaves the same. We recommend using `when` with a subject when possible, as it makes your code easier to read
-and maintain by clearly showing what you're checking.
+Secondly, you can use `when` with or without a subject. The behavior stays the same either way. Using a subject usually
+makes your code more readable and maintainable because it clearly shows what you're checking.
 
 <table>
    <tr>
@@ -140,48 +153,63 @@ when { ... }
 </tr>
 </table>
 
-Depending on how you use `when`, there are different requirements for whether you need to cover all possible cases in your
-branches.
+How you use `when` determines whether you need to cover all possible cases in your branches. Covering all possible cases
+is called being _exhaustive_.
 
-If you use `when` as a statement, you don't have to cover all possible cases. In this example, some cases aren't covered,
-so nothing happens. However, no error occurs:
+### Statements
+
+If you use `when` as a statement, you don't need to cover all possible cases. In this example, some cases aren't covered,
+so no branch is triggered. However, no error occurs:
 
 ```kotlin
 fun main() {
     //sampleStart
-    val x = 3
-    when (x) {
+    val deliveryStatus = "OutForDelivery"
+    when (deliveryStatus) {
         // Not all cases are covered
-        1 -> print("x == 1")
-        2 -> print("x == 2")
+        "Pending" -> print("Your order is being prepared")
+        "Shipped" -> print("Your order is on the way")
     }
     //sampleEnd
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-when-statement"}
 
-In a `when` statement, the values of individual branches are ignored. Just like with `if`, each branch can be a block, 
-and its value is the value of the last expression in the block.
+Just like with `if`, each branch can be a block, and its value is the value of the last expression in the block.
 
-If you use `when` as an expression, you have to cover all possible cases. In other words, it must be _exhaustive_.
-The value of the first matching branch becomes the value of the overall expression. If you don't cover all cases, 
-the compiler throws an error.
+### Expressions
+
+If you use `when` as an expression, you **must** cover all possible cases. The value of the first matching branch becomes
+the value of the overall expression. If you don't cover all cases, the compiler throws an error.
 
 If your `when` expression has a subject, you can use an `else` branch to make sure that all possible cases are covered, but
 it isn't mandatory. For example, if your subject is a `Boolean`, [`enum` class](enum-classes.md), [`sealed` class](sealed-classes.md),
 or one of their nullable counterparts, you can cover all cases without an `else` branch:
 
 ```kotlin
+import kotlin.random.Random
+//sampleStart
 enum class Bit {
     ZERO, ONE
 }
 
-val numericValue = when (getRandomBit()) {
-    // No else branch is needed because all cases are covered
-    Bit.ZERO -> 0
-    Bit.ONE -> 1
+fun getRandomBit(): Bit {
+    return if (Random.nextBoolean()) Bit.ONE else Bit.ZERO
+}
+
+fun main() {
+    val numericValue = when (getRandomBit()) {
+        // No else branch is needed because all cases are covered
+        Bit.ZERO -> 0
+        Bit.ONE -> 1
+    }
+
+    println("Random bit as number: $numericValue")
+    // Random bit as number: 0
+    //sampleEnd
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-when-expression-subject"}
 
 > To simplify `when` expressions and reduce repetition, try out context-sensitive resolution (currently in preview).
 > This feature allows you to omit the type name when using enum entries or sealed class members in `when` expressions if the expected type is known.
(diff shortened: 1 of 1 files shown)


docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
 ## If expression
 
-In Kotlin, `if` is an expression: it returns a value.
-Therefore, there is no ternary operator (`condition ? then : else`) because ordinary `if` works fine in this role.
+In Kotlin, there is no ternary operator (`condition ? then : else`) because `if` serves the same purpose. Add the condition
+to check within parentheses `()` and the action to take if the result is true within curly braces `{}`. You can use `else`
+and `else if` for additional branches and checks. 
+
+You can also write `if` as an expression, which lets you assign its returned value directly to a variable.
+If you use `if` as an expression, you must have an `else` branch.
```
serras: The way this paragraph is written feels a bit strange: it starts with what `if` is not, and the ternary part is actually related to the fact that `if` is an expression.

As a suggestion, I would move the ternary operator part to the second paragraph:

```
You can also write `if` as an expression, which lets you assign its returned value directly to a variable.
If you use `if` as an expression, you must have an `else` branch.
This fits the same purpose as the ternary operator (`condition ? then : else`) in other languages.
```

docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
-If you're using `if` as an expression, for example, for returning its value or
-assigning it to a variable, the `else` branch is mandatory.
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="if-else-blocks-kotlin"}
 
 ## When expressions and statements
 
-`when` is a conditional expression that runs code based on multiple possible values or conditions. It is
-similar to the `switch` statement in Java, C, and similar languages. For example:
+`when` is a conditional expression that runs code based on multiple possible values or conditions. It's
+similar to the `switch` statement in Java, C, and similar languages. `when` evaluates its argument against each branch
```
serras: The word "evaluates" here is a bit dangerous, since this phrase could be read as if the argument was run multiple times, each per branch. Maybe "`when` evaluates its argument and compares the result against each branch ..."?

docs/topics/control-flow.md
```diff
@@ -143,17 +156,19 @@ when { ... }
 Depending on how you use `when`, there are different requirements for whether you need to cover all possible cases in your
```
serras: Maybe the term "exhaustive" should be introduced here?

docs/topics/control-flow.md
```diff
@@ -172,106 +189,177 @@ it isn't mandatory. For example, if your subject is a `Boolean`, [`enum` class](
...
->
-{style="warning"}
+Guard conditions allow you to include more than one condition to the branches of a `when` expression or statement, making complex 
+control flow more explicit and concise. You can use guard conditions in:
 
-Guard conditions allow you to include 
-more than one condition to the branches of a `when` expression, making complex control flow more explicit and concise.
-You can use guard conditions in `when` expressions or statements with a subject.
+* `when` expressions.
+* `when` statements **with a subject**.
```
serras: Whether `when` is an expression or a statement does not play a role here. The only requirement to using guard is to have a `when` with a subject.

docs/topics/control-flow.md
sarahhaggarty: OK, how does this revised version sound to you?

```
To use `if` in Kotlin, add the condition to check within parentheses `()` and the action to take if the result is true within curly braces `{}`. You can use `else` and `else if` for additional branches and checks.

You can also write `if` as an expression, which lets you assign its returned value directly to a variable.
In this form, an `else` branch is required. The `if ` expression serves the same purpose as the ternary operator (`condition ? then : else`) found in other languages.
```

serras: Sounds good 👍

serras (approved):

docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
+
 ## If expression
 
-In Kotlin, `if` is an expression: it returns a value.
-Therefore, there is no ternary operator (`condition ? then : else`) because ordinary `if` works fine in this role.
+To use `if` in Kotlin, add the condition to check within parentheses `()` and the action to take if the result is true
+within curly braces `{}`. You can use `else` and `else if` for additional branches and checks.
+
+You can also write `if` as an expression, which lets you assign its returned value directly to a variable.
+In this form, an `else` branch is required. The `if ` expression serves the same purpose as the ternary operator
```
daniCsorbaJB:
```suggestion
In this form, an `else` branch is required. The `if` expression serves the same purpose as the ternary operator 
```

docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
 
-    // You can also use `else if` in expressions:
-    val maxLimit = 1
-    val maxOrLimit = if (maxLimit > a) maxLimit else if (a > b) a else b
-  
-    println("max is $max")
-    // max is 3
-    println("maxOrLimit is $maxOrLimit")
-    // maxOrLimit is 3
+    // With `else if` as an expression:
```
daniCsorbaJB:
```suggestion
    // With else if as an expression:
```

docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
+    val heightAlice = 160
+    val heightBob = 175
 
     //sampleStart
-    var max = a
-    if (a < b) max = b
+    var taller = heightAlice
+    if (heightAlice < heightBob) taller = heightBob
 
     // With else
```
daniCsorbaJB: (suggestion) 
What is your opinion on writing the code comments like: 
// Uses an else branch

// Uses if as an expression

or 

// Example: else branch

// Example: if expression

To avoid possible syntactic collision (As + expression someone might misread it as a keyword)

docs/topics/control-flow.md
daniCsorbaJB: Since we didn't use backtics for most of the comments I'd suggest avoiding it for all instances (alternatively we could add it to all of them)

docs/topics/control-flow.md
```diff
@@ -1,82 +1,96 @@
...
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-conditions-when-statement"}
 
-`when` matches its argument against all branches sequentially until some branch condition is satisfied.
-
-You can use `when` in a few different ways. Firstly, you can use `when` either as an **expression** or as a **statement**.
-As an expression, `when` returns a value for later use in your code. As a statement, `when` completes an action
-without returning anything of further use:
+You can use `when` either as an **expression** or a **statement**.  As an expression, `when` returns a value you can use
```
daniCsorbaJB:
```suggestion
You can use `when` either as an **expression** or a **statement**. As an expression, `when` returns a value you can use
```

docs/topics/control-flow.md
```diff
@@ -113,9 +127,8 @@ when (x) {
 </tr>
```
daniCsorbaJB, line 127: In the table the comment that says: 
// Returns nothing but triggers a print statement - do you think we could rephrase that like
// Doesn't return a result but triggers a print statement? 

Since Nothing is a type we could clarify this to ensure it's not misunderstood

docs/topics/control-flow.md
```diff
@@ -141,19 +154,21 @@ when { ... }
 </table>
 
 Depending on how you use `when`, there are different requirements for whether you need to cover all possible cases in your
```
daniCsorbaJB: Hmm what do you think of something like: 

The requirements for covering all possible cases depend on how you use `when`. If you're required to account for every possible case, the `when` is considered _exhaustive_.

docs/topics/control-flow.md
```diff
@@ -141,19 +154,21 @@ when { ... }
...
-branches.
+branches. In other words, whether it needs to be _exhaustive_ or not.
+
+### Statements
 
-If you use `when` as a statement, you don't have to cover all possible cases. In this example, some cases aren't covered,
-so nothing happens. However, no error occurs:
+If you use `when` as a statement, you don't need to cover all possible cases. In this example, some cases aren't covered,
+so no code is run. However, no error occurs:
```
daniCsorbaJB:
```suggestion
so no code in the `when` statement is executed. However, no error occurs:
```

docs/topics/control-flow.md
```diff
@@ -163,25 +178,39 @@ fun main() {
 In a `when` statement, the values of individual branches are ignored. Just like with `if`, each branch can be a block,
```
daniCsorbaJB: (not sure if this is accurate, but it might be misleading to say they are ignored)
In a `when` statement, the values of individual branches are evaluated but discarded. 

(also it might be nice to connect the terminology used from the Expressions section)

docs/topics/control-flow.md
```diff
@@ -163,25 +178,39 @@ fun main() {
...
+
+fun main() {
+    val numericValue = when (getRandomBit()) {
+        // No else branch is needed because all cases are covered
+        Bit.ZERO -> 0
+        Bit.ONE -> 1
+    }
+
+    println("Random bit as number: $numericValue")
+    // Random bit as number: 0
```
daniCsorbaJB, line 208: (nit)
// Random bit as number: 0 or 1

docs/topics/control-flow.md
```diff
@@ -283,25 +372,57 @@ sealed interface Animal {
 
 fun feedAnimal(animal: Animal) {
     when (animal) {
-        // Branch with only primary condition. Calls `feedDog()` when `animal` is `Dog`
+        // Branch with only primary condition. 
+        // Calls `feedDog()` when `animal` is `Dog`
```
daniCsorbaJB: same comment as before with the backtics in comments
// Calls the feedDog() function when animal is Dog

docs/topics/control-flow.md
```diff
@@ -283,25 +372,57 @@ sealed interface Animal {
...
+0, 1 -> print("x == 0 or x == 1")
+```
 
-If you use guard conditions in `when` statements without an `else` branch, and none of the conditions matches, none of the branches is executed. 
+In a single `when` expression or statement, you can combine branches with and without guard conditions. 
+The code in a branch with a guard condition runs only if both the primary condition and the guard condition evaluate to `true`.
+If the primary condition doesn't match, the guard condition isn't evaluated. 
+
+Since `when` statements don't need to cover all cases, using guard conditions in `when` statements without an 
+`else` branch means that if no conditions match, nothing happens.
```
daniCsorbaJB:
```suggestion
`else` branch means that if no conditions match, nothing is executed. 
```

docs/topics/control-flow.md
```diff
@@ -283,25 +372,57 @@ sealed interface Animal {
 
 fun feedAnimal(animal: Animal) {
     when (animal) {
-        // Branch with only primary condition. Calls `feedDog()` when `animal` is `Dog`
+        // Branch with only primary condition. 
+        // Calls `feedDog()` when `animal` is `Dog`
         is Animal.Dog -> feedDog()
-        // Branch with both primary and guard conditions. Calls `feedCat()` when `animal` is `Cat` and not `mouseHunter`
+        // Branch with both primary and guard conditions.
```
daniCsorbaJB:
```suggestion
        // Branch with both primary and guard conditions
```

docs/topics/control-flow.md
```diff
@@ -283,25 +372,57 @@ sealed interface Animal {
 
 fun feedAnimal(animal: Animal) {
     when (animal) {
-        // Branch with only primary condition. Calls `feedDog()` when `animal` is `Dog`
+        // Branch with only primary condition.
```
daniCsorbaJB:
```suggestion
        // Branch with only primary condition
```

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
-You can use guard conditions in any `when` expression or statement with a subject, except the case when you have multiple conditions separated by a comma.
-For example, `0, 1 -> print("x == 0 or x == 1")`.
+We would appreciate your feedback on guard conditions in [YouTrack](https://youtrack.jetbrains.com/issue/KT-71140/Guard-conditions-in-when-expressions-feedback).
 
 ## For loops
 
-The `for` loop iterates through anything that provides an iterator. This is equivalent to the `foreach` loop in languages like C#.
+Use the `for` loop to iterate through a [collection](collections-overview.md), [array](arrays.md), or [range](ranges.md).
+
 The syntax of `for` is the following:
```
daniCsorbaJB: Do you think we could remove this line and just place the colon at the end of the previous line? 🤔

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
 
 ```kotlin
 for (item in collection) print(item)
 ```
 
-The body of `for` can be a block.
+The body of `for` can be a block with curly braces `{}`.
```
daniCsorbaJB:
```suggestion
The body of a `for` loop can be a block with curly braces `{}`.
```

daniCsorbaJB: (suggestion)
Just for clarity but whichever you prefer

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
-//sampleEnd
+    // Wake up
+    // Brush teeth
+    // Make coffee
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-for-loop-array"}
+
+Alternatively, you can use the [`.withIndex`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/with-index.html) library function:
```
daniCsorbaJB:
```suggestion
Alternatively, you can use the [`.withIndex()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/with-index.html) function from the Kotlin standard library:
```

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-for-loop-array-index"}
+
+### Iterators
+
+The `for` loop iterates through anything that provides an [iterator](iterators.md). Collections provide iterators by 
+default, whereas ranges and arrays are compiled into index-based loops.
+
+You can create your own iterators. Iterators must have:
```
daniCsorbaJB: This list looks a bit strange 🤔 

Do you think we could rephrase this so the first list item (which is essentially the only item in the list with two subitems) to look something like: 

You can create your own iterators. To do so, implement an `iterator()` function that returns an `Iterator<T>` object, which has:

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
+    // Reading page 3
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-for-loop-inherit-iterator"}
+
+> Learn more about [interfaces](interfaces.md) and [inheritance](inheritance.md).
+> 
+{style="tip"}
+
+Alternatively, you can create the functions from scratch. In this case, add the `operator` keywords to the functions:
```
daniCsorbaJB:
```suggestion
Alternatively, you can create the functions from scratch. In this case, add the `operator` keyword to the functions:
```

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
+    // Reading page 2
+    // Reading page 3
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-for-loop-iterator-from-scratch"}
 
 ## While loops
 
 `while` and `do-while` loops process their body continuously while their condition is satisfied.
```
daniCsorbaJB: (suggestion)
Do you think we could use "runs the code in its body" instead of "processes their body". (sounds a bit more natural to me)  

or perhaps evaluates the code in its body? 🤔 (if we want to stick to the Language specification)

docs/topics/control-flow.md
```diff
@@ -316,112 +437,243 @@ when (animal) {
...
 The difference between them is the condition checking time:
+
 * `while` checks the condition and, if it's satisfied, processes the body and then returns to the condition check.
```
daniCsorbaJB: same here for the list items

daniCsorbaJB (changes requested):
Thank you for the update - big improvement with the examples! 💯 
I left a couple of suggestions

docs/topics/control-flow.md
sarahhaggarty: Hmm I prefer the original to be honest. What makes you suggest an alternative?

docs/topics/control-flow.md
sarahhaggarty: What about "no branch is triggered"?

docs/topics/control-flow.md
sarahhaggarty: What about just removing that entire first sentence?

docs/topics/control-flow.md
sarahhaggarty: OK, what about:

> You can create your own iterators by providing a member or extension function called `iterator()` that returns an `Iterator<>`. The `iterator()` function must have a `next()` function and a `hasNext()` function that returns a `Boolean`.

docs/topics/control-flow.md
sarahhaggarty: I think the list format makes the information easier to absorb and digest. Do you have any alternative suggestions?

docs/topics/control-flow.md
daniCsorbaJB: 🤔 
My thought was to shift the focus a bit here, in the original I see the logic as: 
If X (Depending on how you use `when`), then there are requirements about whether Y needs to happen (you need to cover all possible cases in your branches)

Then we introduce the concept of _exhaustiveness_ as almost an afterthought: 
In other words, this is called Z. (whether it needs to be _exhaustive_ or not.)

So my thought process was to try to break up the information here a bit using this logic:
The requirements depend on X. If Y is required, we call it Z.

This way, the sentence shifts the focus to introducing exhaustiveness as a concept.

On the wording itself I have no strong thoughts - for example, cover every possible case in your branches might be better than simply using account for.

docs/topics/control-flow.md
daniCsorbaJB: I think that works as well 👍  — more concise as well

docs/topics/control-flow.md
daniCsorbaJB: that also works! 👍  might be even better - the second sentence already explains whats happening here.

docs/topics/control-flow.md
daniCsorbaJB: oh sorry, I meant this part here (not to remove the list): 

> Do you think we could use "runs the code in its body" instead of "processes their body". (sounds a bit more natural to me)
> 
> or perhaps evaluates the code in its body? 🤔 (if we want to stick to the Language specification)

docs/topics/control-flow.md
sarahhaggarty: I see! I think it makes sense to keep the format of when as the focus. What if I make it less formal about "requirements" and more active, like this?

> How you use `when` determines whether you need to cover all possible cases in your branches. Covering all possible cases is called being _exhaustive_.

daniCsorbaJB: Nice! 💯 I like it! 👍

daniCsorbaJB (approved):
Great work! — LGTM! 🚀

PR #4981, opened 2025-07-29
fix: add a test selector to the footer

nikpachoo:
Tested locally. 
Screenshot 2025-07-29 at 13 34 31

And the second test: one failure, but that’s expected (we don’t have Google tags locally, so the Opt-Out button didn’t work).
Screenshot 2025-07-29 at 13 43 18

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -252,7 +252,9 @@ function Index() {
             </ThemeProvider>
 
             <ThemeProvider theme={'dark'}>
-                <GlobalFooter />
+                <div data-testid={'footer'}>
+                    <GlobalFooter />
+                </div>
             </ThemeProvider>
 
             <Script src={'https://cdn.optimizely.com/js/26633200186.js'} strategy={"beforeInteractive"}/>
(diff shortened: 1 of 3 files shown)


test/production/footer-kotlin-ecosystem-buttons.spec.ts
```diff
@@ -100,7 +100,7 @@ test.describe('Footer kotlin ecosystem buttons', () => {
 
     // Click on the JetBrains logo button in footer.
     test('JetBrains logo button should navigate to JetBrains homepage', async ({ page, context }) => {
-        const jetBrainsLink = page.getByRole('link', { name: 'JetBrains' }).nth(3);
+        const jetBrainsLink = page.getByTestId('footer').getByRole('link', { name: 'jetbrains.com link' });
```
nikpachoo, line 103: The logo link has been renamed to match its label fully.

pages/index.tsx
```diff
@@ -252,7 +252,9 @@ function Index() {
             </ThemeProvider>
 
             <ThemeProvider theme={'dark'}>
-                <GlobalFooter />
+                <div data-testid={'footer'}>
```
zoobestik, line 255: It's really unnecessary wrapper for this cases and in future it would be nice pass `data-testid` for to components of ui-kit.

nikpachoo: Agree. Unfortunately, the component doesn't support this prop right now, will add it later.

PR #4982, opened 2025-07-29
Autotests for navbar buttons.

solarday:


test/production/navbar_buttons.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Navbar buttons', () => {
```
nikpachoo, line 3: I suggest renaming the test: this part of navigation we call "header".
Also, there is a test in the master branch, [api navigation](https://github.com/JetBrains/kotlin-web-site/blob/master/test/production/api-navigation.spec.ts). What do you think about unifying them into one file and calling it header.spec.ts? I would expect to have all the header tests related to it in one place.

test/production/navbar_buttons.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    test.beforeEach(async ({ page }) => {
+        await page.goto('/');
+        await page.waitForSelector('button.ch2-btn.ch2-btn-primary');
+        await page.click('button.ch2-btn.ch2-btn-primary');
+    });
+
+    test('Kotlin logo opens the homepage', async ({ page }) => {
+        const kotlinLogo = page.locator('.ktl-logo-large-module_link_AOGas');
```
nikpachoo, line 11: The locators should be change to not use classes and especially classes with hashes like "_AOGas". They may change after an update, and we'll be forced to update the tests.
My suggestion: we need to add data-testid for the header elements.

nikpachoo:
@solarday, i changed the status of PR, until we add the data-selectors to the head

PR #4983, opened 2025-07-29
Paphos KUG

narko:
Adding the Paphos KUG as described in https://github.com/narko/kotlin-web-site?tab=readme-ov-file#kotlin-user-group

### DIFF
--- a/data/user-groups.yml
+++ b/data/user-groups.yml
@@ -222,6 +222,12 @@
     position:
       lat: 48.1351253
       lng: 11.5819806
+  - name: Paphos KUG
+    country: Cyprus
+    url: https://www.meetup.com/pafos-coders-designers/
+    position:
+      lat: 34.7745909
+      lng: 32.4190263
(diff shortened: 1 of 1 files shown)


narko:
@meilalina 🙏

data/user-groups.yml
```diff
@@ -222,6 +222,12 @@
     position:
       lat: 48.1351253
       lng: 11.5819806
+  - name: Paphos KUG
```
meilalina: Hi @narko! I don’t think it’s necessary to list the group as Paphos KUG since that’s not its actual name. I believe it’s totally fine to include the group under the original name and it won’t confuse visitors. The group's topic is fairly general, so it’s understandable that they occasionally organize Kotlin meetups.

jellymonsale:
> Adding the Paphos KUG as described in https://github.com/narko/kotlin-web-site?tab=readme-ov-file#kotlin-user-group

PR #4988, opened 2025-07-31
Replace `StringBuilder` example in intermediate tour: Lambda expressions with receiver

sarahhaggarty:
This PR replaces the `StringBuilder` example in the Lambda expressions with receiver chapter of the intermediate tour with an example that explicitly uses a lambda expression with receiver.

Approved by Alejandro Serrano Mena

### DIFF
--- a/docs/topics/tour/kotlin-tour-intermediate-lambdas-receiver.md
+++ b/docs/topics/tour/kotlin-tour-intermediate-lambdas-receiver.md
@@ -20,15 +20,15 @@ can help you create a domain-specific language.
 ## Lambda expressions with receiver
 
 In the beginner tour, you learned how to use [lambda expressions](kotlin-tour-functions.md#lambda-expressions). Lambda expressions can also have a receiver. 
-In this case, lambda expressions can access any member functions or properties of the receiver object without having
-to explicitly specify the receiver object each time. Without these additional references, your code is easier to read and maintain.
+In this case, lambda expressions can access any member functions or properties of the receiver without having
+to explicitly specify the receiver each time. Without these additional references, your code is easier to read and maintain.
 
 > Lambda expressions with receiver are also known as function literals with receiver.
 >
 {style="tip"}
 
 The syntax for a lambda expression with receiver is different when you define the function type. First, write the receiver
-type that you want to extend. Next, put a `.` and then complete the rest of your function type definition. For example:
+that you want to extend. Next, put a `.` and then complete the rest of your function type definition. For example:
 
 ```kotlin
 MutableList<Int>.() -> Unit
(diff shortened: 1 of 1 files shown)


docs/topics/tour/kotlin-tour-intermediate-lambdas-receiver.md
```diff
@@ -20,50 +20,64 @@ can help you create a domain-specific language.
...
-* The function type of the lambda expression has no function parameters `()` and has no return value `Unit`.
-* The lambda expression calls the `append()` member function from the `StringBuilder` class and uses the string `"Hello!"` as the function parameter.
-* An instance of the `StringBuilder` class is created.
-* The lambda expression assigned to `appendText` is called on the `stringBuilder` instance.
-* The `stringBuilder` instance is converted to string with the `toString()` function and printed via the `println()` function.
+* The `Canvas` class has two functions that simulate drawing a circle or a square.
+* The `render()` function takes a `block` parameter and returns an instance of the `Canvas` class.
+* The `block` parameter is a lambda expression with receiver, where the `Canvas` class is the receiver.
+* The `render()` function creates an instance of the `Canvas` class and calls the `block()` lambda expression on the `canvas` instance, using it as the receiver.
+* The `main()` function calls the `render()` function with a lambda expression that becomes the `block` parameter.
```
daniCsorbaJB:
```suggestion
* The `main()` function calls the `render()` function with a lambda expression, which is passed to the `block` parameter as an argument.
```

daniCsorbaJB: Perhaps this is just a nitpick, and it might be better not to complicate it, but if I understand it correctly, it doesn't become the `block` parameter; rather, the lambda becomes the value of the parameter, not the parameter itself.

docs/topics/tour/kotlin-tour-intermediate-lambdas-receiver.md
```diff
@@ -20,50 +20,64 @@ can help you create a domain-specific language.
...
+* Inside the lambda passed to the `render()` function, the program calls the `drawCircle()` and `drawSquare()` functions on an instance of the `Canvas` class.
+
+  Because the `drawCircle()` and `drawSquare()` functions are called in the lambda expression with receiver, they can be called
```
daniCsorbaJB, line 79:
```suggestion
  Because the `drawCircle()` and `drawSquare()` functions are called in the lambda expression with a receiver, they can be called
```

daniCsorbaJB: Do we need an article here? 🤔

daniCsorbaJB (approved):
Just two minor suggestions, otherwise it looks good to me! 🚀

docs/topics/tour/kotlin-tour-intermediate-lambdas-receiver.md
sarahhaggarty: It's a good question! From what I can see, it's a technical term so an article isn't necessary :)

PR #5000, opened 2025-08-11
Improve Properties page

sarahhaggarty:
This PR improves the Properties page by improving the quality of the examples by making them more reflective of the real world, amongst other things.

### DIFF
--- a/docs/topics/properties.md
+++ b/docs/topics/properties.md
@@ -1,64 +1,132 @@
 [//]: # (title: Properties)
 
+In Kotlin, properties let you store and manage data without writing functions to access or change the data.
+You can use properties in [classes](classes.md), [interfaces](interfaces.md), [objects](object-declarations.md), [companion objects](object-declarations.md#companion-objects),
+and even outside these structures as top-level properties.
+
+Every property has a name, a type, and an automatically generated `get()` function called a getter. You can use the getter
+to read the property's value. If the property is mutable, it also has a `set()` function called a setter, which allows 
+you to change the property's value.
+
+> Getters and setters are called _accessors_.
+> 
+{style="tip"}
+
 ## Declaring properties
 
-Properties in Kotlin classes can be declared either as mutable, using the `var` keyword, or as read-only, using the `val` keyword.
+Properties can be mutable (`var`) or read-only (`val`).
+You can declare them as a top-level property in a `.kt` file. Think of a top-level property as a global variable
+that belongs to a package:
+
+```kotlin
+// File: Constants.kt
+package my.app
+
+val pi = 3.14159
+var counter = 0
+```
+
+You can also declare properties inside a class, interface, or object:
+
+```kotlin
+// Class with properties
+class Address {
+    var name: String = "Holmes, Sherlock"
+    var street: String = "Baker"
+    var city: String = "London"
+}
+
+// Interface with a property
+interface ContactInfo {
+    val email: String
+}
+
+// Object with properties
+object Company {
+    var name: String = "Detective Inc."
+    val country: String = "UK"
+}
+
+// Class implementing the interface
+class PersonContact : ContactInfo {
+    override val email: String = "sherlock@example.com"
+}
+```
+
+To use a property, refer to it by its name:
 
 ```kotlin
 class Address {
     var name: String = "Holmes, Sherlock"
     var street: String = "Baker"
     var city: String = "London"
-    var state: String? = null
-    var zip: String = "123456"
 }
-```
 
-To use a property, simply refer to it by its name:
+interface ContactInfo {
+    val email: String
+}
 
-```kotlin
+object Company {
+    var name: String = "Detective Inc."
+    val country: String = "UK"
+}
+
+class PersonContact : ContactInfo {
+    override val email: String = "sherlock@example.com"
+}
+
+//sampleStart
 fun copyAddress(address: Address): Address {
-    val result = Address() // there's no 'new' keyword in Kotlin
-    result.name = address.name // accessors are called
+    val result = Address()
+    // Accesses properties in the result instance
+    result.name = address.name
     result.street = address.street
-    // ...
+    result.city = address.city
     return result
 }
+
+fun main() {
+    val sherlockAddress = Address()
+    val copy = copyAddress(sherlockAddress)
+    // Accesses properties in the copy instance
+    println("Copied address: ${copy.name}, ${copy.street}, ${copy.city}")
+    // Copied address: Holmes, Sherlock, Baker, London
+
+    // Accesses properties in the Company object
+    println("Company: ${Company.name} in ${Company.country}")
+    // Company: Detective Inc. in UK
+    
+    val contact = PersonContact()
+    // Access properties in the contact instance
+    println("Email: ${contact.email}")
+    // Email: sherlock@email.com
+}
+//sampleEnd
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-access-properties"}
 
-## Getters and setters
+In Kotlin, we recommend initializing properties when you declare them to keep your code safe and easy to read. However, 
+you can [initialize them later](#late-initialized-properties-and-variables) in special cases.
 
-The full syntax for declaring a property is as follows:
+Declaring the property type is optional if the compiler can infer it from the initializer or the getter's return type:
 
 ```kotlin
-var <propertyName>[: <PropertyType>] [= <property_initializer>]
-    [<getter>]
-    [<setter>]
+var initialized = 1 // The inferred type is Int
+var allByDefault    // ERROR: Property must be initialized.
 ```
+{validate="false"}
 
-The initializer, getter, and setter are optional. The property type is optional if it can be inferred from the initializer
-or the getter's return type, as shown below:
+## Custom getters and setters
 
-```kotlin
-var initialized = 1 // has type Int, default getter and setter
-// var allByDefault // ERROR: explicit initializer required, default getter and setter implied
-```
+By default, Kotlin automatically generates getters and setters. You can define your own custom accessors when
+you need extra logic, such as validation, formatting, or calculations based on other properties.
 
-The full syntax of a read-only property declaration differs from a mutable one in two ways: it starts with `val` instead
-of `var` and does not allow a setter:
-
-```kotlin
-val simple: Int? // has type Int, default getter, must be initialized in constructor
-val inferredType = 1 // has type Int and a default getter
-```
-
-You can define custom accessors for a property. If you define a custom getter, it will be called every time you access
-the property (this way you can implement a computed property). Here's an example of a custom getter:
+A custom getter runs every time the property is accessed:
 
 ```kotlin
 //sampleStart
 class Rectangle(val width: Int, val height: Int) {
-    val area: Int // property type is optional since it can be inferred from the getter's return type
+    val area: Int
         get() = this.width * this.height
 }
 //sampleEnd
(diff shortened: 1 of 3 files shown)


docs/topics/properties.md
```diff
@@ -1,64 +1,128 @@
...
+
+// Object with properties
+object Company {
+    var name: String = "Detective Inc."
+    val country: String = "UK"
+}
+
+// Class implementing the interface
+class PersonContact : ContactInfo {
+    override val email: String = "sherlock@detective.com"
```
nikitabobko: I understand that `detective.com` is just an example. Unfortunately, it's the real accessible domain name with a site hosted on it. I'd restrain from endorsing any 3rd party sites, except for JB/Kotlin sites or ones that are known to be illustrative examples (such as example.com)

docs/topics/properties.md
```diff
@@ -1,64 +1,128 @@
...
+    val email: String
+}
+
+object Company {
+    var name: String = "Detective Inc."
+    val country: String = "UK"
+}
+
+class PersonContact : ContactInfo {
+    override val email: String = "sherlock@detective.com"
```
nikitabobko: The same comment about `detective.com` as above

docs/topics/properties.md
```diff
@@ -1,64 +1,128 @@
...
+    println("Email: ${contact.email}")
+    // Email: sherlock@detective.com
+}
+//sampleEnd
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-access-properties"}
 
-The initializer, getter, and setter are optional. The property type is optional if it can be inferred from the initializer
-or the getter's return type, as shown below:
+You don't have to initialize a property immediately. It can be [initialized later](#late-initialized-properties-and-variables), like the `email` property in the `ContactInfo` interface.
```
nikitabobko: > You don't have to initialize a property immediately. It can be [initialized later](#late-initialized-properties-and-variables), like the `email` property in the `ContactInfo` interface.

These two sentences are misleading. There are several issues.

1. Generally, in Kotlin, properties should be initialized immediately at the place they are declare or inside the `init` block. It's more safe, and Kotlin endorses safety and clarity in this place. Unless, it's a property without any backing fields (property without backing fields is a property that has explicit getter and setter AND the getter and the setter don't refer to backing field via `field`)

2. We reference `lateinit` documentation but `email` property in the `ContractInfo` interface is not a `lateinit` property. 

3. And there is a small terminology problem. `email` property in the `ContractInfo` interface has nothing to do with deferred initialization. We don't have to initialize `email` property in the `ContractInfo` because `ContractInfo` is an interface and doesn't require to contain any implementation (initialization is a special case of implementation). In the interface actual implementation (`PersonContact`), the property has to be immediately initialized as usual.

docs/topics/properties.md
```diff
@@ -1,64 +1,128 @@
...
+Declaring the property type is optional if it can be inferred from the initializer or the getter's return type:
 
 ```kotlin
-var initialized = 1 // has type Int, default getter and setter
-// var allByDefault // ERROR: explicit initializer required, default getter and setter implied
+var initialized = 1 // The inferred type is Int
+var allByDefault    // ERROR: This variable must either have an explicit type or be initialized.
```
nikitabobko: > This variable must either have an explicit type or be initialized.

Incorrect. The variable must be initialized. Explicit type is not enough

Initialization is not required only in the case when the property has explicit getter and setter AND the getter and the setter don't refer to the backing field

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+
+class Service {
+    var dependency: String = "Default Service"
+        // Annotate the getter
+        @Inject get 
+}
+
+fun main() {
+    val service = Service()
+    println(service.dependency)
```
nikitabobko, line 228: We could improve the example by demonstrating that annotations are observable:

```
println(service::dependency.getter.annotations) // [@Inject()]
println(service::dependency.setter.annotations) // []
```

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-annotate-getter"}
+
+> To learn more about annotations, see [Annotations](annotations.md).
+>
+{style="tip"}
+
+### Backing fields
+
+In Kotlin, accessors use backing fields to hold the property's value in memory. Backing fields are useful
+when you want to add extra logic to your getter and setter without causing an infinite loop. Or if you want to trigger
```
nikitabobko: > without causing an infinite loop

The sentence is misleading. No infinite loop in Kotlin can be fixed with `field` keyword.

My guess is that you think that `field` is itself a property which has its own getter and setter. And one gets into an infinite loop by calling getters or setters indefinitely. But that is incorrect. `field` is called "backing field" not "backing property" exactly because of that problem.

Fields are inexpressible in Kotlin language. Fields are dummy data holders which can't have setters/getters.

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-field"}
+
 ### Backing properties
 
-If you want to do something that does not fit into this _implicit backing field_ scheme, you can always fall back to having
-a _backing property_:
+In some cases, you might want more flexibility than using a backing field can provide. For example, if you have an API
+where you want to be able to modify the property internally but not externally. In such cases, you can use a _backing property_.
```
nikitabobko: I'd emphasize that _backing property_ is a code pattern. _backing field_ is a language concept

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+
+For example, you have a `ShoppingCart` class with an `items` property that contains everything inside the shopping cart.
+You want to make the `items` property read-only outside the class, but you want to allow one "approved" way for the user to modify
+the `items` property directly. In your code, you create a backing property called `_items` which is private, and you also create a normal property
+called `items` which is public and initializes to the value of the backing property.
```
nikitabobko:
```suggestion
called `items` which is public and delegates to the value of the backing property.
```

"initializes" may be technically incorrect in some contexts. For example, in the example below you don't initialize `items` with `_items`, but the `items` property delegates to to `_items` backing property

"delegates" is a more general term that includes "initialization"

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+    println(cart.items) 
+    // [Banana]
+}
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-property"}
+
+The user can only add items to the cart by calling the `addItem()` function, but they can still access the `items` property
+to see what's inside the cart.
 
-> On the JVM: Access to private properties with default getters and setters is optimized to avoid function call overhead.
```
nikitabobko, line 145: This note about the JVM is actually good. It demonstrates that the compiler optimizes common and trivial cases so that people should not be afraid introducing backing properties. Personally, I'd keep it

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+> Name backing properties with a leading underscore to avoid name clashes with other properties.
```
nikitabobko:
```suggestion
> It's a common pattern to name backing properties with a leading underscore.
```

Otherwise, the tip looks like an advice on how to avoid name clashes (which is trivial and doesn't require any explanation)

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
+In this example, the backing property `_celsius` is accessed by both the `celsius` and `fahrenheit` properties. So there 
+is one source of truth but two public views.
 
 ## Compile-time constants
 
-If the value of a read-only property is known at compile time, mark it as a _compile time constant_ using the `const` modifier.
-Such a property needs to fulfill the following requirements:
+If the value of a read-only property is known at compile time, mark it as a _compile-time constant_ using the `const` modifier.
+Compile-time constants are inlined at compile time, so every reference is replaced with its actual value. They are accessed
+more efficiently as no getter is called and no memory is allocated for the property.
```
nikitabobko: > and no memory is allocated for the property

Incorrect. `const val` properties still generate a backing field => they allocate memory

docs/topics/properties.md
```diff
@@ -67,156 +131,353 @@ fun main() {
...
 ```kotlin
 const val SUBSYSTEM_DEPRECATED: String = "This subsystem is deprecated"
 
-@Deprecated(SUBSYSTEM_DEPRECATED) fun foo() { ... }
+@Deprecated(SUBSYSTEM_DEPRECATED) fun processLegacyOrders() { ... }
 ```
 
 ## Late-initialized properties and variables
 
 Normally, properties declared as having a non-nullable type must be initialized in the constructor.
```
nikitabobko: Properties with nullable type must be initialized in the constructor as well. In the context of initialization, nullability doesn't matter.

docs/topics/properties.md
sarahhaggarty: Very good point! Updated.

docs/topics/properties.md
sarahhaggarty: Unfortunately Kotlin playground doesn't support reflection so adding this line would break the runnable sample. Do you think it's important enough to make the code sample not runnable?

docs/topics/properties.md
sarahhaggarty: I removed the part about the infinite loop 👍

docs/topics/properties.md
sarahhaggarty: Added!

docs/topics/properties.md
sarahhaggarty: Is there another word we can use? I'm concerned that the reader might get confused between how we use "delegates" here and what it means in delegated properties.

docs/topics/properties.md
sarahhaggarty: I added your suggestion but used active voice.

docs/topics/properties.md
sarahhaggarty: Added it back into the doc.

docs/topics/properties.md
sarahhaggarty: Removed!

docs/topics/properties.md
sarahhaggarty: OK. Corrected it not to be dependent on nullability.

docs/topics/properties.md
sarahhaggarty: This code comment is a word for word quote from the error reported in Kotlin playground: https://pl.kotl.in/RnerEYo_b
I'll put it underneath the code to try and make this more obvious.

docs/topics/properties.md
nikitabobko: I didn't know that, thank you. It only affects the "run" button near each code snippet in the docs, right?

I'd say that docs come first, possibility to run samples in the docs comes second. That's why I personally ok with the fact that the "run" button doesn't work.

But you have the final say here, so if the "run" button is a showstopper, let's keep the sample as it is.

docs/topics/properties.md
sarahhaggarty: OK, so let's remove the link to the `email` property because that's just wrong. What do you think of this text as an alternative?

> In Kotlin, we recommend initializing properties when you declare them to keep your code safe and easy to read. However, it's possible to [initialize them later](#late-initialized-properties-and-variables) for special cases.
>
> Declaring the property type is optional if it can be inferred from the initializer or the getter's return type:

nikitabobko: Looks good, I like the recommendation style, as I think it's really a good recommendation, and people generally tend to initialize their properties in the constructor when it's possible

Thank you

docs/topics/properties.md
nikitabobko: Ah, ooh, I see. It's about local variables 🤦 

If you put the property on the top-level, or inside a class - initialization is mandatory, and the compiler error message is different.

The whole document I have been so confused by the fact that we say that initialization is not a requirement. For local properties (which I'd rather call local variables) - yes, it's not a requirement. According to my experience, people rarely call local variables - properties. When people say properties, they often mean top-level properties or properties inside a class. And local properties are more commonly called "local variables"

docs/topics/properties.md
nikitabobko: I'm sorry to nitpick here, but I think that the new version still could be improved.

```
> We recommend naming backing properties with a leading underscore.
```

```
> It's a common pattern to name backing properties with a leading underscore.
```

The first one is a recommendation, and as a pedantic programmer who reads the documentation, I instantly question myself on why would they recommend doing that? But the second wording doesn't raise any of these questions - we just state the fact that it's a common pattern.

And is my initial suggestion a passive voice? I think it's technically an active voice. If it were passive, it'd look like `Backing properties are commonly named with a leading underscore`

docs/topics/properties.md
nikitabobko: Yeah, I thought that you would raise this question 🙂 

Unfortunately, this code pattern is commonly known as _delegation_ across variety of programming languages (Kotlin included). Thus, `items` _delegates_ to `_items`. No other word really comes to my mind

BTW, the delegation language feature could be used here to implement the delegation code pattern:

```
class ShoppingCart {
    // Backing property
    private val _items = mutableListOf<String>()

    // Public read-only view
    val items: List<String> by ::_items

    fun addItem(item: String) {
        _items.add(item)
    }

    fun removeItem(item: String) {
        _items.remove(item)
    }
}
```

Even if the delegation code pattern is confused in the sentence with the delegation language feature, the main idea remains understandable

docs/topics/properties.md
sarahhaggarty: I see now where I made a mistake in generating the error message. Thanks for pointing this out!

OK, so based on your feedback I've adjusted the text to be about top-level properties OR properties within a class/interface/object. I've also changed the text to say that immediate initialization is necessary, but you can initialize later if you use `lateinit`. And I updated the error message to the correct one. How does it look now for you?

docs/topics/properties.md
sarahhaggarty: Right! We try to make samples runnable by default so that readers can adjust the code sample if they want and they also help with reader engagement in general. But it's OK not to have it for every sample. I made the sample not runnable and explained that we're using reflection to show what annotations are present on the getter and setter.

docs/topics/properties.md
sarahhaggarty: OK, I've added your suggested change :)

docs/topics/properties.md
sarahhaggarty: Never be sorry to nitpick! I'd rather spend time getting it just right :)
You're right that the original is not passive voice but the actor isn't explicit (some group of people follow this pattern) so the tone seems detached.
What do you think about this as an alternative?
> We recommend following the common pattern of naming backing properties with a leading underscore.

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
...
-The initializer, getter, and setter are optional. The property type is optional if it can be inferred from the initializer
-or the getter's return type, as shown below:
+Declaring the property type is optional if it can be inferred from the initializer or the getter's return type:
 
 ```kotlin
-var initialized = 1 // has type Int, default getter and setter
-// var allByDefault // ERROR: explicit initializer required, default getter and setter implied
+var initialized = 1 // The inferred type is Int
+var allByDefault    
+// ERROR: Property must be initialized.
```
nikitabobko: I'd consider joining these two lines. But it's up to you

docs/topics/properties.md
nikitabobko: Looks good, thank you

docs/topics/properties.md
nikitabobko: Did you push the change?

docs/topics/properties.md
nikitabobko: The final alternative looks good, thank you

docs/topics/properties.md
nikitabobko: Did you push the change?

docs/topics/properties.md
nikitabobko: Just to point out: the version in the document differs from what you wrote here. I'm fine with both versions, but still wanted to point out in case you forgot to push the change

nikitabobko (approved):

docs/topics/properties.md
sarahhaggarty: Fixed!

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
 [//]: # (title: Properties)
 
+In Kotlin, properties let you store and manage data without writing functions to access or change the data.
+You can use properties in [classes](classes.md), [interfaces](interfaces.md), [objects](object-declarations.md), [companion objects](object-declarations.md#companion-objects),
+and even outside these structures as top-level properties.
+
+Every property has a name, a type, and an automatically generated `get()` function called a getter, which allows you to read the property's value.
```
daniCsorbaJB:
```suggestion
Every property has a name, a type, and an automatically generated `get()` function called a getter. You can use the getter to read the property's value.
```

daniCsorbaJB: (suggestion) 
just to break up the two repeating "which allows you to phrases."

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
...
     var name: String = "Holmes, Sherlock"
     var street: String = "Baker"
     var city: String = "London"
-    var state: String? = null
-    var zip: String = "123456"
+}
+
+// Interface with a property
+interface ContactInfo {
+    val email: String  // Abstract property, must be implemented
```
daniCsorbaJB: What do you think about moving this comment up before declaring the val email? 🤔

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
...
+class PersonContact : ContactInfo {
+    override val email: String = "sherlock@example.com"
+}
+
+//sampleStart
 fun copyAddress(address: Address): Address {
-    val result = Address() // there's no 'new' keyword in Kotlin
-    result.name = address.name // accessors are called
+    val result = Address()
+    // Access properties in the Address class
```
daniCsorbaJB: Are we accessing the properties of the Address class or the instance of the Address class here? 🤔

daniCsorbaJB: Should we use the third person for the comments or is there a reason for imperative here? 🤔

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
...
-    [<getter>]
-    [<setter>]
+fun main() {
+    val sherlockAddress = Address()
+    val copy = copyAddress(sherlockAddress)
+    // Access properties in the copy instance
+    println("Copied address: ${copy.name}, ${copy.street}, ${copy.city}")
+    // Copied address: Holmes, Sherlock, Baker, London
+
+    // Access properties in the object
```
daniCsorbaJB:
```suggestion
    // Access properties in the Company object
```

daniCsorbaJB: (suggestion)
Since we specifically refer to each other type.

docs/topics/properties.md
```diff
@@ -1,64 +1,131 @@
...
 
-The full syntax of a read-only property declaration differs from a mutable one in two ways: it starts with `val` instead
-of `var` and does not allow a setter:
+## Custom getters and setters
 
-```kotlin
-val simple: Int? // has type Int, default getter, must be initialized in constructor
-val inferredType = 1 // has type Int and a default getter
-```
+By default, Kotlin generates getters and setters behind the scenes. You can define your own custom accessors when
```
daniCsorbaJB:
```suggestion
By default, Kotlin automatically generates getters and setters. You can define your own custom accessors when
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 ```kotlin
-var stringRepresentation: String
-    get() = this.toString()
-    set(value) {
-        setDataFromString(value) // parses the string and assigns values to other properties
-    }
+class Point(var x: Int, var y: Int) {
+    var coordinates: String
+        get() = "$x,$y"
+        set(value) {
```
daniCsorbaJB, line 153: (very light suggestion - feel free to ignore)
Perhaps we could add a code comment here: 
// You can change the name of the setter, but by convention, it's value

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-custom-setter"}
 
-By convention, the name of the setter parameter is `value`, but you can choose a different name if you prefer.
+### Changing visibility or adding annotations
 
-If you need to annotate an accessor or change its visibility, but you don't want to change the default implementation,
-you can define the accessor without defining its body:
+You can change accessor visibility or add [annotations](annotations.md) without replacing the default implementation. To make your code concise,
+in Kotlin you don't have to make these changes within a body.
```
daniCsorbaJB: (suggestion)
What do you think about moving In Kotlin to the beginning of the first sentence here. 
then we could follow up with 

In Kotlin, you can change accessor visibility or add [annotations](annotations.md) without replacing the default implementation.
You don't have to make these changes within a body, which makes your code more concise.

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
-            field = value
-            // counter = value // ERROR StackOverflow: Using actual name 'counter' would make setter recursive
-    }
+// Inject annotation
+@Target(AnnotationTarget.PROPERTY_GETTER)
+annotation class Inject
+
+class Service {
+    var dependency: String = "Default Service"
+        // Annotate the getter
```
daniCsorbaJB:
```suggestion
        // Annotates the getter
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+Or if you want to annotate the getter but keep the default implementation:
 
 ```kotlin
-var counter = 0 // the initializer assigns the backing field directly
-    set(value) {
-        if (value >= 0)
-            field = value
-            // counter = value // ERROR StackOverflow: Using actual name 'counter' would make setter recursive
-    }
+// Inject annotation
```
daniCsorbaJB:
```suggestion
// Defines an annotation that can be applied to a property getter
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-custom-setter"}
 
-By convention, the name of the setter parameter is `value`, but you can choose a different name if you prefer.
+### Changing visibility or adding annotations
 
-If you need to annotate an accessor or change its visibility, but you don't want to change the default implementation,
-you can define the accessor without defining its body:
+You can change accessor visibility or add [annotations](annotations.md) without replacing the default implementation. To make your code concise,
+in Kotlin you don't have to make these changes within a body.
+
```
daniCsorbaJB, line 176: (suggestion)
What is your opinion on first describing the syntax for these examples? I'm thinking of something like: 
To change accessor visibility, place the modifier, like `private` before the `get` or `set` keyword:

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+{validate="false"}
 
-The `field` identifier can only be used in the accessors of the property.
+This example uses [reflection](reflection.md) to show which annotations are present on the getter and setter.
 
-A backing field will be generated for a property if it uses the default implementation of at least one of the accessors,
-or if a custom accessor references it through the `field` identifier.
+### Backing fields
+
+In Kotlin, accessors use backing fields to hold the property's value in memory. Backing fields are useful
```
daniCsorbaJB:
```suggestion
In Kotlin, accessors use backing fields to store the property's value in memory. Backing fields are useful
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+when you want to add extra logic to your getter and setter, or when you want to trigger an additional action whenever the property
+changes.
+
+You can't declare backing fields directly. Kotlin generates them only when necessary. You can reference the backing field
+in accessors using the `field` keyword.
+
+Kotlin only generates backing fields if you use the default getter or setter, or if you use ` field` in at least one custom accessor.
 
-For example, there would be no backing field in the following case:
+For example, there is no backing field in this case because it uses a custom getter without the `field` keyword:
```
daniCsorbaJB:
```suggestion
For example, this property has no backing field because it uses a custom getter without the `field` keyword:
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 
 ```kotlin
 val isEmpty: Boolean
     get() = this.size == 0
 ```
 
+In this example, there is a backing field because the setter uses the `field` keyword:
```
daniCsorbaJB:
```suggestion
In this example, the property has a backing field because the setter uses the `field` keyword:
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 
-The `field` identifier can only be used in the accessors of the property.
+This example uses [reflection](reflection.md) to show which annotations are present on the getter and setter.
 
-A backing field will be generated for a property if it uses the default implementation of at least one of the accessors,
-or if a custom accessor references it through the `field` identifier.
+### Backing fields
+
+In Kotlin, accessors use backing fields to hold the property's value in memory. Backing fields are useful
+when you want to add extra logic to your getter and setter, or when you want to trigger an additional action whenever the property
```
daniCsorbaJB:
```suggestion
when you want to add extra logic to a getter or a setter, or when you want to trigger an additional action whenever the property
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 ```
 
+In this example, there is a backing field because the setter uses the `field` keyword:
+
+```kotlin
+class Scoreboard {
+    var score: Int = 0
+        set(value) {
+            field = value
+            // Additional action of printing a log message
```
daniCsorbaJB:
```suggestion
            // Adds additional behavior to the setter by logging the updated value
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+    // Score updated to 20
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-field"}
+
 ### Backing properties
 
-If you want to do something that does not fit into this _implicit backing field_ scheme, you can always fall back to having
-a _backing property_:
+Sometimes you might want more flexibility than using a backing field can provide. For example, if you have an API
```
daniCsorbaJB:
```suggestion
Sometimes you might need more flexibility than a backing field can provide. For example, if you have an API
```

daniCsorbaJB: (suggestion)
What do you think about adding a link to the backing field section here? 🤔

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+where you want to be able to modify the property internally but not externally. In such cases, you can use a coding pattern
+called a _backing property_.
+
+For example, you have a `ShoppingCart` class with an `items` property that contains everything in the shopping cart.
```
daniCsorbaJB:
```suggestion
In the following example, the `ShoppingCart` class has an `items` property that represents everything in a shopping cart.
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+You want the `items` property to be read-only outside the class but still allow one "approved" way for the user to modify
+the `items` property directly. In your code, you create a private backing property called `_items` and a public property
```
daniCsorbaJB:
```suggestion
the `items` property directly. To achieve this, you can define a private backing property called `_items` and a public property
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+    
+    cart.removeItem("Apple")
+    println(cart.items) 
+    // [Banana]
+}
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-property"}
 
-> On the JVM: Access to private properties with default getters and setters is optimized to avoid function call overhead.
+The user can only add items to the cart through the `addItem()` function, but can still access the `items` property
```
daniCsorbaJB:
```suggestion
In this example, the user can only add items to the cart through the `addItem()` function, but can still access the `items` property
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+to see what's inside.
+
+> We recommend following the common pattern of naming backing properties with a leading underscore.
```
daniCsorbaJB: (suggestion)
Would something like this work? 🤔 
Use a leading underscore when naming backing properties to follow Kotlin naming conventions.

(to shift it from a recommendation to explain why to do it? )

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 >
-{style="note"}
+{style="tip"}
+
+On the JVM, the compiler optimizes access to private properties with default getters and setters to avoid function call overhead.
```
daniCsorbaJB:
```suggestion
On the JVM, the compiler optimizes access to private properties with default accessors to avoid function call overhead.
```

daniCsorbaJB: (light suggestion)
Just to make it more concise

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+
+    var fahrenheit: Double
+        get() = _celsius * 9 / 5 + 32
+        set(value) { _celsius = (value - 32) * 5 / 9 }
+}
+
+fun main() {
+    val temp = Temperature()
+    temp.celsius = 25.0
+    println("${temp.celsius}°C = ${temp.fahrenheit}°F") // 25.0°C = 77.0°F
```
daniCsorbaJB:
```suggestion
    println("${temp.celsius}°C = ${temp.fahrenheit}°F") 
    // 25.0°C = 77.0°F
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+
+    temp.fahrenheit = 212.0
+    println("${temp.celsius}°C = ${temp.fahrenheit}°F") // 100.0°C = 212.0°F
```
daniCsorbaJB:
```suggestion
    println("${temp.celsius}°C = ${temp.fahrenheit}°F") 
    // 100.0°C = 212.0°F
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-property-multiple-properties"}
+
+In this example, the `_celsius` backing property is accessed by both the `celsius` and `fahrenheit` properties. This setup
+provides one source of truth with two public views.
```
daniCsorbaJB:
```suggestion
provides a single source of truth with two public views.
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 
 ## Compile-time constants
 
-If the value of a read-only property is known at compile time, mark it as a _compile time constant_ using the `const` modifier.
-Such a property needs to fulfill the following requirements:
+If the value of a read-only property is known at compile time, mark it as a _compile-time constant_ using the `const` modifier.
+Compile-time constants are inlined at compile time, so each reference is replaced with its actual value. They are accessed
+more efficiently because no getter is called.
```
daniCsorbaJB:
```suggestion
more efficiently because no getter is called:
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
-* It cannot be a custom getter
+* They must be either a top-level property, or a member of an [`object` declaration](object-declarations.md#object-declarations-overview) or a [companion object](object-declarations.md#companion-objects).
+* They must be initialized with a value of type `String` or a [primitive type](basic-types.md).
+* They can't have a custom getter.
 
-The compiler will inline usages of the constant, replacing the reference to the constant with its actual value. However, the field will not be removed and therefore can be interacted with using [reflection](reflection.md).
+Compile-time constants still have a backing field, so you can interact with them using [reflection](reflection.md).
 
-Such properties can also be used in annotations:
+These properties have the added benefit that you can use them in annotations:
```
daniCsorbaJB:
```suggestion
You can also use these properties in annotations:
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 ```
 
 ## Late-initialized properties and variables
 
-Normally, properties declared as having a non-nullable type must be initialized in the constructor.
-However, it is often the case that doing so is not convenient. For example, properties can be initialized through dependency
-injection, or in the setup method of a unit test. In these cases, you cannot supply a non-nullable initializer in the constructor,
-but you still want to avoid null checks when referencing the property inside the body of a class.
+Normally, you must initialize properties in the constructor.
+However, in some cases, doing so isn't convenient. For example, you might initialize properties through dependency
```
daniCsorbaJB:
```suggestion
However, this isn't always convenient. For example, you might initialize properties through dependency
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+injection or inside the setup method of a unit test.
 
-To handle such cases, you can mark the property with the `lateinit` modifier:
+To handle these situations, you can mark the property with the `lateinit` modifier:
```
daniCsorbaJB:
```suggestion
To handle these situations, mark the property with the `lateinit` modifier:
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
 
     @SetUp fun setup() {
-        subject = TestSubject()
+        orderService = OrderService()
     }
 
-    @Test fun test() {
-        subject.method()  // dereference directly
+    @Test fun processesOrderSuccessfully() {
+        // Dereference directly without checking for null
```
daniCsorbaJB:
```suggestion
        // Calls orderService without checking for null or initialization
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+For class properties:
 
-### Checking whether a lateinit var is initialized
+* You can't declare them in the primary constructor.
+* They must not have a custom getter or setter.
 
-To check whether a `lateinit var` has already been initialized, use `.isInitialized` on the [reference to that property](reflection.md#property-references):
+In all cases, the property or variable must be non-nullable and must not be a [primitive type](basic-types.md).
+
+If you access a `lateinit` property before initializing it, Kotlin throws a specific exception that identifies the property
```
daniCsorbaJB:
```suggestion
If you access a `lateinit` property before initializing it, Kotlin throws a specific exception that identifies the uninitialized property
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+being accessed:
 
 ```kotlin
-if (foo::bar.isInitialized) {
-    println(foo.bar)
+class ReportGenerator {
+    lateinit var report: String
+
+    fun printReport() {
+        // Access before initialization
```
daniCsorbaJB:
```suggestion
        // Throws an exception if accessed before initialization
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+    generator.printReport()
+    // Exception in thread "main" kotlin.UninitializedPropertyAccessException: lateinit property report has not been initialized
+    //   at ReportGenerator.getReport (File.kt:2) 
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-lateinit-property" validate="false"}
 
-This check is only available for properties that are lexically accessible when declared in the same type, in one of the
-outer types, or at top level in the same file.
+To check whether a `lateinit var` has already been initialized, use the [`isInitialized`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-lazy/is-initialized.html)
```
daniCsorbaJB:
```suggestion
To check whether a `lateinit var` has already been initialized, use the [`isInitialized`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/is-initialized.html#)
```

daniCsorbaJB: I might be wrong - but that seems like the function 🤔 so this should be the property used here. (not sure)

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+property on the [reference to that property](reflection.md#property-references):
+
+```kotlin
+class WeatherStation {
+    lateinit var latestReading: String
+
+    fun printReading() {
+        // Check whether the property is initialized
```
daniCsorbaJB:
```suggestion
        // Checks whether the property is initialized
```

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+        // Access before initialization
+        println(report)
+    }
+}
+
+fun main() {
+    val generator = ReportGenerator()
+    generator.printReport()
+    // Exception in thread "main" kotlin.UninitializedPropertyAccessException: lateinit property report has not been initialized
+    //   at ReportGenerator.getReport (File.kt:2)
```
daniCsorbaJB:
```suggestion
```

daniCsorbaJB: Since we don't list the rest of the stack trace elements I think we could omit that part as well

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
-See [Overriding properties](inheritance.md#overriding-properties)
+See [Overriding properties](inheritance.md#overriding-properties).
 
 ## Delegated properties
 
-The most common kind of property simply reads from (and maybe writes to) a backing field, but custom getters and setters
-allow you to use properties so one can implement any sort of behavior of a property.
-Somewhere in between the simplicity of the first kind and variety of the second, there are common patterns for what properties
-can do. A few examples: lazy values, reading from a map by a given key, accessing a database, notifying a listener on access.
+The most common kind of property simply reads from (and possibly writes to) a backing field using the default getter and setter, while
```
daniCsorbaJB: (Suggestion)
By default, the getter returns the value from the backing field, and the setter updates the backing field. You use custom accessors to define different behaviors.

docs/topics/properties.md
```diff
@@ -67,156 +134,366 @@ fun main() {
...
+custom getters and setters let you define any kind of property behavior.
+
+Between these two, there are common patterns for how properties work. Examples include:
```
daniCsorbaJB: I'm not sure I understand what do we mean with these two here? 🤔 getters and setters | default vs custom accessors? 
Could you please clarify?

daniCsorbaJB: hmm 🤔 I think I'm confused — what are we trying to say here? 🤔 

If we want to introduce delegated properties, what do you think about a logic like: 
1. Default getter and setter reads or writes. 
2. contrast with custom accessors
3. Introduce the use case, with something like You might want to reuse some common property behaviors, like (list examples)
4. Kotlin supports these patterns through delegated properties, allowing the user to reuse implementation of such behaviors in a consistent way --> link to delegated properties.

(or we could remove the entire 1-2 part and just keep 3-4 "There some common property behaviors with custom accessors" --> "Kotlin supports these through delegated properties")

docs/topics/properties.md
sarahhaggarty: Actually, that's not supposed to be there. Removed!

docs/topics/properties.md
sarahhaggarty: Good point! Fixed!

docs/topics/properties.md
sarahhaggarty: Thinking about it, what about removing "automatically" as well since it is "by default"? 🤔

docs/topics/properties.md
sarahhaggarty: I think I'd prefer not to duplicate the text since we already mention it just before the code sample.

docs/topics/properties.md
sarahhaggarty: I went with a slightly different sentence. Please check it and let me know what you think :)

docs/topics/properties.md
sarahhaggarty: @nikitabobko since we discussed this part previously (https://github.com/JetBrains/kotlin-web-site/pull/5000#discussion_r2272618092), can you check that you are OK with Daniel's suggestion?

nikitabobko: Daniel suggestion sounds good to me.

docs/topics/properties.md
sarahhaggarty: Good catch! Thanks!

docs/topics/properties.md
sarahhaggarty: Yeah. I tried to retain the original text but I agree that it doesn't make much sense. I decided to go with just the 3-4 part. Please take a look at the new section and let me know what you think :)

docs/topics/properties.md
sarahhaggarty: Nice idea! Added.

docs/topics/properties.md
sarahhaggarty: I went with something similar to what you suggest. Let me know what you think :)

docs/topics/properties.md
daniCsorbaJB: It works with both, I think. With automatically it's a bit more emphasized but not necessarily needed.

docs/topics/properties.md
daniCsorbaJB: I like it 👍

docs/topics/properties.md
daniCsorbaJB: I think it works (a bit less explicit I prefer the more explicit one a bit but this works well! 👍 )  😄

docs/topics/properties.md
daniCsorbaJB: It's a great improvement — I really like it, clear - concise - easy to follow 💯 
I have two tiny comments:
at the last line about implementing in libraries, could we perhaps add that theoretically users could take these from existing libraries as well? 🤔

Something like... 
You can implement these behaviors in libraries yourself or use existing delegates provided by external libraries. 
___
I feel like we could add the information about reuse to this part instead since that's when it becomes most applicable:
As a rough draft: 
Delegating the accessor behavior keeps the property's accessor logic in one place, which makes it reusable.

sarahhaggarty: Nice suggestions. Added! :)

PR #5003, opened 2025-08-12
disable slack notifications if project contains playground in its name

Quattro8:
Update to disable slack notifications for playground projects

### DIFF
--- a/.teamcity/tests/buildTypes/E2EProductionTest.kt
+++ b/.teamcity/tests/buildTypes/E2EProductionTest.kt
@@ -1,7 +1,7 @@
 package tests.buildTypes
 
+import common.extensions.isProjectPlayground
 import jetbrains.buildServer.configs.kotlin.BuildType
-import jetbrains.buildServer.configs.kotlin.DslContext
 import jetbrains.buildServer.configs.kotlin.buildFeatures.notifications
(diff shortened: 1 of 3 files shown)


.teamcity/tests/buildTypes/E2EProductionTest.kt
```diff
@@ -1,56 +1,58 @@
...
-        connection = "PROJECT_EXT_486"
-        sendTo = "#kotlin-web-site-e2e-tests"
-        messageFormat = simpleMessageFormat()
-      }
-      buildFailedToStart = true
-      buildFailed = true
-      buildProbablyHanging = true
+    features {
+        notifications {
+            enabled = !DslContext.projectName.lowercase().contains("playground")
```
zoobestik: 1. Could we do it helper `isProjectPlayground()`?
2. I prefer use id - `DslContext.projectId.value.contains("Kotlin_KotlinSites_Playground")`, but up to you.

zoobestik (changes requested):

.teamcity/tests/buildTypes/E2EProductionTest.kt
Quattro8: Hi @zoobestik, thanks for idea, I've implemented it. Could you check this PR again?

zoobestik (approved):

PR #5017, opened 2025-08-14
KTL-2944 Add webteam youtube-playlist

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-2944/Add-webteam-youtube-playlist

20-minute-adventure-rick-morty

### How does it work
By default WEBTEAM_UI_NPM_TOKEN is expected.
External users can set `export USE_FALLBACK_FOR_INTERNAL_PACKAGES=1` to install dependencies and run the website locally.
In this case, a fallback UI component will be used.

### How we can use webteam-ui
No limits, we just need to to the following things:
- add a package to the `transpiledPackages`section in the next.js config.
- provide a fallback for a case when webteam-ui is not available.


blocks/server-side/additional-materials/additional-materials.tsx
```diff
@@ -5,7 +5,8 @@ import cn from 'classnames';
 import styles from './additional-materials.module.css';
 import { useTextStyles } from '@rescui/typography';
 
-import YoutubePlayer from '@jetbrains/kotlin-web-site-ui/out/components/youtube-player';
+// @ts-ignore
+import YoutubePlaylist from '@webteam/youtube-playlist/lib';
```
nikpachoo, line 9: I don't know why, but types resolve doesn't work with it.

next.config.js
```diff
@@ -44,6 +59,17 @@ const nextConfig = {
...
+        const useFallbackForInternalPackages = process.env.USE_FALLBACK_FOR_INTERNAL_PACKAGES === '1';
+
+        config.resolve.alias = config.resolve.alias || {};
+
+        if (useFallbackForInternalPackages) {
+            config.resolve.alias['@webteam'] = path.resolve(
+                __dirname,
+                'components/webteam-fallback'
+            );
+        }
```
nikpachoo, line 71: this is the place, where all the magic happens.
We just resolve another module in case the env variable is set.

scripts/install-internal.js
```diff
@@ -0,0 +1,41 @@
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+const flagFile = path.join(process.cwd(), '.peer-deps-installed');
+
+if (fs.existsSync(flagFile)) {
+    console.log('Peer dependencies already installed, skipping...');
+    fs.rmSync(flagFile);
+    process.exit(0);
+}
```
nikpachoo: Why this flagFile is needed
The install-internal.js script runs after `yarn` or `npm` install. Since this script also installs dependencies, it would trigger itself again, causing an infinite loop. To prevent this, we use a flagFile.
I’m fine with implementing it differently.
Another option could be to call install-internal.js in the build scripts after yarn. I tried to avoid aditional steps, but this solution is not optimal either.

.npmrc
```diff
@@ -1 +1,3 @@
-save-exact = true
\ No newline at end of file
+save-exact = true
+
+@webteam:registry=https://packages.jetbrains.team/npm/p/wt/npm/
```
nikpachoo: This registry will be used only if a dependency from the namespace is installing.

components/webteam-fallback/youtube-playlist/lib/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+const YoutubePlaylist: FC<YoutubePlaylistProps> = ({ playlistId, playlistTitle }) => {
+    return (
+        <YoutubePlayer
+            mode={1}
+            id={playlistId}
+        />
+    );
+};
+
+export default YoutubePlaylist;
```
nikpachoo, line 19: Fallback component. 
Initially I wanted to use a very simple one, only with a text, but then decided to use the player.
Not sure that this is needed.

next.config.js
```diff
@@ -13,6 +14,20 @@ let transpiledPackages = [
...
+        '@webteam/ui-contexts',
+        '@webteam/colors',
+        '@webteam/toggle',
+        '@webteam/youtube-player',
+        '@webteam/use-fetch',
+        '@webteam/use-async-data',
+        '@webteam/bem-cn-fast',
+        '@webteam/list'
+    )
+}
```
nikpachoo, line 29: This is not a workaround, it is expected to be here, we do the same with the RescUI components above.

next.config.js
```diff
@@ -13,6 +14,20 @@ let transpiledPackages = [
...
+
 const withTranspile = nextTranspileModules(transpiledPackages);
```
krutilov, line 31: We can probably keep it for now, but I found out that `next-transpile-modules` is deprecated see https://github.com/martpie/next-transpile-modules/releases/tag/the-end

PR #5023, opened 2025-08-18
feat(KTL-2844): Experiment #1: Make server-related materials more visible

zoobestik:


zoobestik:
@berezinant, i've fixed issue with disable

PR #5024, opened 2025-08-18
feat: block parameter names

danil-pavlov:
This PR updates Obj-C/Swift interop page for the 2.2.20 release, following [KT-77488](https://youtrack.jetbrains.com/issue/KT-77488) [ObjCExport] Add explicit ObjCBlock parameter name in objc export

### DIFF
--- a/docs/topics/native/native-objc-interop.md
+++ b/docs/topics/native/native-objc-interop.md
@@ -543,9 +543,9 @@ foo {
 
 #### Explicit parameter names in Objective-C block types
 
-You can add explicit parameter names to Kotlin's function types for exported Objective-C headers. Xcode's autocompletion
-suggests calling such functions with no parameter names in the Objective-C block, and the generated block triggers Clang
-warnings.
+You can add explicit parameter names to Kotlin's function types for exported Objective-C headers. Without them,
+Xcode's autocompletion suggests calling Objective-C functions with no parameter names in the Objective-C block,
+and the generated block triggers Clang warnings.
 
 To enable explicit parameter names, add the following [binary option](native-binary-options.md) to your `gradle.properties` file:
 


docs/topics/native/native-objc-interop.md
```diff
@@ -534,13 +534,44 @@ func foo(block: (KotlinInt) -> KotlinUnit)
...
+```swift
 foo {
     bar($0 as! Int32)
     return KotlinUnit()
 }
 ```
 
+#### Explicit names in block types
+
+You can add explicit names to Kotlin's block types for exported Objective-C headers. Without them, Xcode's autocompletion
```
SvyatoslavScherbina: Replace "block" with "function".

nagatsuka-shuuya: I guess calling generated ObjC from ObjC in XCode? As said below it doesn't affect calling from Swift, right?

nagatsuka-shuuya: Thank you for making it more clear

SvyatoslavScherbina: I meant only in this sentence, not everywhere. Sorry for the confusion. Other changes should be reverted.

Explanation: there are no "block types" in Kotlin. Kotlin has function types.
On the other hand, there are block types in Objective-C.

Another example:
> Without them, Xcode's autocompletion suggests calling such functions with no names in the block, and the generated block triggers Clang warnings.

became
> Without them, Xcode's autocompletion suggests calling such functions without parameter names, and the generated function triggers Clang warnings.

The first one was clear and correct. In the second one, it is not clear what is meant by "calling without parameter names".

danil-pavlov: Thank you for the clarification! I reverted the changes and added some details to make it clearer

SvyatoslavScherbina (approved):

docs/topics/native/native-objc-interop.md
```diff
@@ -534,13 +534,46 @@ func foo(block: (KotlinInt) -> KotlinUnit)
...
 foo {
     bar($0 as! Int32)
     return KotlinUnit()
 }
 ```
 
+#### Explicit parameter names in Objective-C block types
+
+You can add explicit parameter names to Kotlin's function types for exported Objective-C headers. Xcode's autocompletion
+suggests calling such functions with no parameter names in the Objective-C block, and the generated block triggers Clang
```
sarahhaggarty: Can you explain what "such functions" means here?

docs/topics/native/native-objc-interop.md
```diff
@@ -534,13 +534,46 @@ func foo(block: (KotlinInt) -> KotlinUnit)
...
+```
+
+For example, for the following Kotlin code:
+
+```kotlin
+// Kotlin:
+fun greetUser(block: (name: String) -> Unit) = block("John")
+```
+
+Kotlin forwards the parameter names from Kotlin function types to Objective-C block types, so Xcode uses them in suggestions:
```
sarahhaggarty:
```suggestion
Kotlin forwards the parameter names from Kotlin function types to Objective-C block types, allowing Xcode to use them in suggestions:
```

sarahhaggarty (commented):
Nice! There's only one minor part I'm not sure I understand fully.

PR #5026, opened 2025-08-19
Update intermediate tour delegation section

sarahhaggarty:
This PR fixes the code samples in the delegation section of the intermediate tour and replaces them with a better example.

### DIFF
--- a/docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
+++ b/docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
@@ -155,7 +155,7 @@ only support single inheritance. If you need to inherit from multiple sources, c
 Interfaces are similar to classes, but they have some differences:
 
 * You can't create an instance of an interface. They don't have a constructor or header.
-* Their functions and properties are implicitly inheritable by default. In Kotlin, we say that they are "open".
+* Their functions and properties are implicitly inheritable by default. In Kotlin, we say that they are "open."
 * You don't need to mark their functions as `abstract` if you don't give them an implementation.
 
 Similar to abstract classes, you use interfaces to define a set of functions and properties that classes can inherit and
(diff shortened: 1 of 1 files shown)


nikitabobko (approved):
Looks good, thank you. I have zero comments :)

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
```diff
@@ -269,87 +269,185 @@ For more information about interfaces and interface inheritance, see [Interfaces
...
+        return "PenTool(color=$color)"
     }
-   override val color = null
 }
 ```
 
-If you wanted to create a child class of the `Circle` class which had the same behavior **except** for the value of the
-`color` property, you still need to add implementations for each member function of the `Circle` class:
+If you want to create a class similar to the `PenTool` class which has the same behavior **except** for the value of the
+`color` property, you can. First, create a new class that expects an object that implements the `DrawingTool` interface as a parameter,
```
koshachy: If you want to create a class like `PenTool` with the same behavior but a different color `value`, you can create one.

Or we'd like to mention the **except** explicitly?

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
```diff
@@ -269,87 +269,185 @@ For more information about interfaces and interface inheritance, see [Interfaces
...
 
-You can see that if you have a large number of member functions in the `Drawable` interface, the amount of boilerplate
-code in the `RedCircle` class can be very large. However, there is an alternative.
+You can see that if you have a large number of member functions in the `DrawingTool` interface, the amount of boilerplate
+code in the `CanvasSession` class can be very large. However, there is an alternative.
 
-In Kotlin, you can use delegation to delegate the interface implementation to an instance of a class. For example,
-you can create an instance of the `Circle` class and delegate the implementations of the member functions of the `Circle`
-class to this instance. To do this, use the `by` keyword. For example:
+In Kotlin, you can use delegation to delegate the interface implementation to an instance of a class. To do this, use
```
koshachy: IMO, too verbose.

In Kotlin, you can delegate the interface implementation to another class instance using the `by` keyword.

Will it work?

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
```diff
@@ -269,87 +269,185 @@ For more information about interfaces and interface inheritance, see [Interfaces
...
+In Kotlin, you can use delegation to delegate the interface implementation to an instance of a class. To do this, use 
+the `by` keyword. For example:
 
 ```kotlin
-class RedCircle(param: Circle) : Drawable by param
+class CanvasSession(val tool: DrawingTool) : DrawingTool by tool
 ```
 
-Here, `param` is the name of the instance of the `Circle` class that the implementations of member functions are delegated to.
+Here, `tool` is the name of the instance of the `PenTool` class that the implementations of member functions are delegated to.
```
koshachy: From my perspective, it is really hard to read because of too many "of"s

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
sarahhaggarty: What if I change the sentence to not start with "if"? Something like:

> You want to create a class like `PenTool` with the same behavior but a different `color` value. One approach is to create a new class that expects an object implementing the `DrawingTool` interface as a parameter, like a `PenTool` class instance. Then, inside the class, you can override the `color` property.

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
sarahhaggarty: Is this better?

> Here, `tool` is the name of the `PenTool` class instance where the implementations of member functions are delegated to.

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
koshachy: I like your version, thanks!

docs/topics/tour/kotlin-tour-intermediate-classes-interfaces.md
koshachy: WFM, thanks!

PR #5027, opened 2025-08-19
feat: kdoc export by default

danil-pavlov:


docs/topics/native/native-objc-interop.md
```diff
@@ -150,22 +111,36 @@ After that, the Objective-C header will contain a corresponding comment:
 + (void)printSumA:(int32_t)a b:(int32_t)b __attribute__((swift_name("printSum(a:b:)")));
 ```
 
-You'll be able to see comments on classes and methods in autocompletion, for example, in Xcode. If you go to the
-definition of functions (in the `.h` file), you'll see comments on `@param`, `@return`, and so on.
+KDoc comments are embedded into klibs and extracted from klibs into the produced Apple frameworks.
+As a result, comments on classes and methods appear during autocompletion, for example in Xcode.
+If you go to the definition of functions (in the `.h` file), you'll see comments on `@param`, `@return`, and so on.
```
daniCsorbaJB: Do we need the parentheses here?
If you go to the definition of functions in the `.h` file, you'll

daniCsorbaJB:
```suggestion
If you go to the definition of functions (in the `.h` file), you'll see comments on `@param`, `@return`, and similar tags.
```

daniCsorbaJB: again just a suggestion - I think it works as it is, but perhaps a bit clearer? 🤔

docs/topics/native/native-objc-interop.md
```diff
@@ -150,22 +111,36 @@ After that, the Objective-C header will contain a corresponding comment:
...
-> Opt-in is required (see the details below), and you should use it only for evaluation purposes.
-> We would appreciate your feedback on it in [YouTrack](https://youtrack.jetbrains.com/issue/KT-38600).
->
-{style="warning"}
-
-* Dependency documentation is not exported unless it is compiled with `-Xexport-kdoc` itself. The feature is Experimental,
-  so libraries compiled with this option might be incompatible with other compiler versions.
+* Dependency documentation is not exported unless it is compiled with the `-Xexport-kdoc` option. Libraries compiled with
+  this compiler option might be incompatible with other compiler versions.
 * KDoc comments are mostly exported as is. Many KDoc features, for example `@property`, are not supported.
```
daniCsorbaJB:
```suggestion
* KDoc comments are mostly exported as is, but many KDoc block tags, such as `@property`, are not supported.
```

daniCsorbaJB: (suggestion)
not part of the update but perhaps a bit more natural this way

docs/topics/native/native-objc-interop.md
```diff
@@ -150,22 +111,36 @@ After that, the Objective-C header will contain a corresponding comment:
...
 
+If necessary, you can disable the export of Kdoc comments from klibs to the produced Apple frameworks in the `binaries {}`
```
daniCsorbaJB:
```suggestion
If necessary, you can disable the export of KDoc comments from klibs to the produced Apple frameworks in the `binaries {}`
```

daniCsorbaJB (approved):
Extremely small comments, please check - otherwise looks good to me! 🚀

PR #5031, opened 2025-08-20
Add compiler execution strategies to Maven page

sarahhaggarty:
This PR adds information about the different compiler execution strategies that are available with Maven after Kotlin 2.2.20 is released.


docs/topics/maven.md
```diff
@@ -212,6 +212,26 @@ making sure that the `kotlin` plugin comes before the `maven-compiler-plugin` in
...
 ```
 
+## Configure Kotlin compiler execution strategy
+
+The _Kotlin compiler execution strategy_ defines where the Kotlin compiler runs. There are two available strategies:
+
+| Strategy       | Where the Kotlin compiler is executed      |
+|----------------|--------------------------------------------|
+| Daemon         | Inside its own daemon process              |
+| In process     | Inside the Maven daemon process            |
```
ALikhachev: "Maven daemon process" is not entirely correct here. I'd suggest just saying "Maven process". Maven becomes a daemon only if explicitly used with https://maven.apache.org/tools/mvnd.html

ALikhachev (changes requested):

docs/topics/maven.md
sarahhaggarty: Fixed!

docs/topics/maven.md
```diff
@@ -212,6 +212,26 @@ making sure that the `kotlin` plugin comes before the `maven-compiler-plugin` in
 </build>
 ```
 
+## Configure Kotlin compiler execution strategy
+
+The _Kotlin compiler execution strategy_ defines where the Kotlin compiler runs. There are two available strategies:
+
+| Strategy   | Where the Kotlin compiler is executed |
+|------------|---------------------------------------|
+| Daemon     | Inside its own daemon process         |
```
daniCsorbaJB:
```suggestion
| Kotlin daemon     | Inside its own daemon process         |
```

daniCsorbaJB: just to make it consistent

(suggestion)
What do you think about writing it as Kotlin daemon (default) ? 🤔

docs/topics/maven.md
```diff
@@ -212,6 +212,26 @@ making sure that the `kotlin` plugin comes before the `maven-compiler-plugin` in
...
+| In process | Inside the Maven process              |
+
+By default, the Kotlin daemon is used. However, you can switch to the "in process" strategy by setting the following
```
daniCsorbaJB:
```suggestion
By default, the Kotlin daemon is used. You can switch to the "in process" strategy by setting the following
```

docs/topics/build-tools-api.md
```diff
@@ -100,9 +100,8 @@ kotlin.compiler.execution.strategy=in-process
 
 ## Integration with Maven
 
-From Kotlin 2.2.0, the BTA is enabled by default in the [`kotlin-maven-plugin`](maven.md).
+The BTA enables the [`kotlin-maven-plugin`](maven.md) to support the Kotlin daemon, which is used by default as the
```
daniCsorbaJB:
```suggestion
The BTA enables the [`kotlin-maven-plugin`](maven.md) to support the Kotlin daemon, which is the default
```

docs/topics/build-tools-api.md
```diff
@@ -100,9 +100,8 @@ kotlin.compiler.execution.strategy=in-process
...
+[compiler execution strategy](maven.md#configure-kotlin-compiler-execution-strategy). The `kotlin-maven-plugin` uses BTA by default,
```
daniCsorbaJB, line 104: Hmm 🤔 Here do we mean there is no need to configure anything to use the BTA? or if you use BTA you don't have to configure anything, because it automates all processes? 🤔

docs/topics/maven.md
```diff
@@ -212,6 +212,26 @@ making sure that the `kotlin` plugin comes before the `maven-compiler-plugin` in
...
+property in your `pom.xml` file:
+
+```xml
+<properties>
+    <kotlin.compiler.daemon>false</kotlin.compiler.daemon>
+</properties>
+```
+
+If you use the "daemon" strategy, you still need to explicitly configure incremental compilation.
```
daniCsorbaJB: (suggestion)
Do we need the "daemon" part here (specifically the quotes I mean:
🤔 What do you think would it make sense with a description like this? 

If you use the Kotlin daemon, you must still explicitly configure/enable incremental compilation.

Or do you have to set it for both the Daemon and in process? 🤔

daniCsorbaJB (commented):
Thank you for the update 💯 
I left a couple of comments/suggestions please let me know what you think

docs/topics/maven.md
sarahhaggarty: For both. I updated the text to reflect this.

docs/topics/build-tools-api.md
sarahhaggarty: Yes. You don't need to opt in to use the BTA, like you do with Gradle :)

docs/topics/maven.md
daniCsorbaJB: nice! Thank you, looks good! 💯

daniCsorbaJB (approved):
Thank you for the updates — LGTM 🚀

PR #5034, opened 2025-08-22
Add informative note for the JVM binary DSL regarding `runJvm`

bitspittle:
Recently, I updated a very old KMP project of mine over to latest Kotlin versions (2.2.10 at the time of writing this message) and was informed that the application plugin was no longer compatible, and that I should migrate over to the new JVM binary DSL via https://kotl.in/jvm-binaries-dsl

Confusingly, after doing that, I kept getting an error:
```
Execution failed for task ':myproject:jvmRun'.
> No main class specified and classpath is not an executable jar.
```
It turns out `jvmRun` (which is what I've used for years) is now obsolete, and I thought it would be nicer if the docs made sure to call that out (especially as these docs are linked to directly from the warning issued by the Kotlin Multiplatform Gradle plugin). Especially, this isn't intuitive, as otherwise, all other tasks associated with KMP targets are prefixed by their platform (e.g. nativeBuild, jsBrowserDevelopmentRun, etc.)

### DIFF
--- a/docs/topics/whatsnew/whatsnew2120.md
+++ b/docs/topics/whatsnew/whatsnew2120.md
@@ -117,6 +117,13 @@ kotlin {
 In this example, Gradle's [Distribution](https://docs.gradle.org/current/userguide/distribution_plugin.html#distribution_plugin)
 plugin is applied on the first `executable {}` block.
 
+> After configuring your JVM binary information, be sure to use the `runJvm` task and *not* the legacy `jvmRun` task.
+>
+> You may also wish to see [KT-75240](https://youtrack.jetbrains.com/issue/KT-75240) which requests that the behavoir of the two
+> tasks are better aligned.
+>
+{style="note"}
+
 If you run into any issues, report them in our [issue tracker](https://kotl.in/issue) or let us know in our [public Slack channel](https://kotlinlang.slack.com/archives/C19FD9681).
 
 ## Kotlin/Native


bitspittle:
I am not a technical writer, so this is just an initial proposal, but you are of course welcome to modify it or even reject it as you see fit. This is also my first PR against the Kotlin website so apologies if I didn't conform to some style guide rules.

**The part that would have REALLY helped save me time would have been seeing the new `runJvm` task called out. I realize now it was hinted at inside the code sample, but that was easy to miss.**

When I reached out to the Slack about this issue (https://kotlinlang.slack.com/archives/C19FD9681/p1755846829808169), someone pointed out KT-75240, which I thought was also useful, so I included it in my note, but you could also choose to delete it if you think it's better without it.

sarahhaggarty:
Hi @bitspittle! 
Thanks for your suggestion and for bringing this to our attention. I think it would be better to include this information on one of our Gradle pages, but it's still unclear what the actual difference between these two tasks is.
So I'm going to close this PR for now, but I've created a ticket ([KT-80886](https://youtrack.jetbrains.com/issue/KT-80886/Docs-Document-the-difference-between-jvmRun-and-runJvm-tasks)) to investigate and document it. You can follow along there.
Thanks again for your contribution, and stay tuned!

bitspittle:
That's fine but, just to reiterate, I didn't simply read this section because I was randomly going through "what's new" articles. A gradle error linked me straight to it.

So if you clarify the difference somewhere else, I wouldn't find it. 

Of course, if you document this behavior clearly somewhere else AND change the error generated by the Kotlin multiplatform plugin, that would work too. 

I'll note this on the youtrack issue as well.

sarahhaggarty:
Thanks for pointing that out @bitspittle! I will definitely take it into account when working on the documentation update. If you can add this to the YouTrack issue as well, that would be great.

PR #5040, opened 2025-08-25
Update kotlin-tour-null-safety.md

nasif:
Updated the example snippet for Nullable types
Here notNull is a string type, so argument passed must be a string and length is 9.


sarahhaggarty:
Hi @nasif !
In this example, `neverNull` is already assigned the string "This can't be null" earlier in the code. This gives it a `String` type with a length of 18. I hope this makes the example clearer for you! Let me know if you have any further questions.

PR #5041, opened 2025-08-26
fix: update kotlin-tour-classes.md

slowhigh:
Clarified copy() behavior and added example for non-recursive deep copy.

### DIFF
--- a/docs/topics/tour/kotlin-tour-classes.md
+++ b/docs/topics/tour/kotlin-tour-classes.md
@@ -247,6 +247,26 @@ fun main() {
 Creating a copy of an instance is safer than modifying the original instance because any code that relies on the
 original instance isn't affected by the copy and what you do with it.
 
+> The copy() function creates a **deep copy** of the instance at the top level, but it does **not** perform a **recursive deep copy**.
+> If a property is a reference type (like a list or another object), both the original and the copied instance will share that reference.
+> For example:
+> ```kotlin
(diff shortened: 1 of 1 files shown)


docs/topics/tour/kotlin-tour-classes.md
```diff
@@ -247,6 +247,26 @@ fun main() {
 Creating a copy of an instance is safer than modifying the original instance because any code that relies on the
 original instance isn't affected by the copy and what you do with it.
 
+> The copy() function creates a **deep copy** of the instance at the top level, but it does **not** perform a **recursive deep copy**.
+> If a property is a reference type (like a list or another object), both the original and the copied instance will share that reference.
```
serras:
```suggestion
> The copy() function creates a **shallow copy** of the instance, that is, it doesn't copy its components recursively. As a result, references to instances are shared.
> For example, if one of the properties is a mutable list, changes performed through the "original" value are visible through the copy, and vice versa.
```

slowhigh: Thanks for the review.

sarahhaggarty:
Hi @slowhigh!
Thanks for your suggestion! Your addition is useful, but I think it's better suited to the page on data classes. To save you some time, I created PR (#5076) to add your content there with a bit of polish. Thanks again for your contribution, and I hope you'll contribute again soon!

PR #5054, opened 2025-09-01
feat: add information about receiver-context parameter ambiguity

daniCsorbaJB:
This PR adds updates for [KT-78866](https://youtrack.jetbrains.com/issue/KT-78866) Warn implicit receiver shadowed by context parameter

### DIFF
--- a/docs/topics/type-safe-builders.md
+++ b/docs/topics/type-safe-builders.md
@@ -175,7 +175,7 @@ In the last section you can read through the full definition of this package.
 ## Scope control: @DslMarker
 
 When using DSLs, one might have come across the problem that too many functions can be called in the context. 
-You can call methods of every available implicit receiver inside a lambda and therefore get an inconsistent result, 
+You can call methods of every available [implicit receiver](lambdas.md#function-literals-with-receiver) inside a lambda and therefore get an inconsistent result, 
 like the tag `head` inside another `head`: 
 
 ```kotlin
@@ -275,7 +275,7 @@ html {
 
 Only the nearest receiver's members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
 
-When a member of an implicit receiver and a declaration from a [context parameter](context-parameters.md) are both in a scope with the same name,
+When both a member of an implicit receiver and a declaration from a [context parameter](context-parameters.md) are in a scope with the same name,
 the compiler reports a warning because the implicit receiver is shadowed by the context parameter.
 To resolve this, use a `this` qualifier to explicitly call the receiver, or use `contextOf<T>()` to call the context declaration:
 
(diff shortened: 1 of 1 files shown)


docs/topics/type-safe-builders.md
```diff
@@ -275,6 +275,36 @@ html {
 
 Only the nearest receiver's members and extensions are accessible within a lambda, preventing unintended interactions between nested scopes.
 
+When a member of an implicit receiver and a declaration from a [context parameter](context-parameters.md) are both in a scope with the same name,
```
AlejandraPedroza:
```suggestion
When both a member of an implicit receiver and a declaration from a [context parameter](context-parameters.md) are in a scope with the same name,
```

AlejandraPedroza: Do we have link for expanding info on implicit receiver as we do for context parameter?

docs/topics/type-safe-builders.md
```diff
@@ -275,6 +275,36 @@ html {
...
+
+// Declares a top-level function with the same name available through a context parameter
+context(tag: HtmlTag)
+fun setAttribute(name: String, value: String) { tag.setAttribute(name, value) }
+
+fun test(head: HtmlTag, extraInfo: HtmlTag) {
+    with(head) {
+        // Introduces a context value of the same type in an inner scope
+        context(extraInfo) {
+            // Reports a warning: uses an implicit receiver shadowed by a context parameter.
```
AlejandraPedroza:
```suggestion
            // Reports a warning: uses an implicit receiver shadowed by a context parameter
```

AlejandraPedroza:
```suggestion
            // Reports a warning: uses an implicit receiver shadowed by a context parameter
```

docs/topics/type-safe-builders.md
```diff
@@ -275,6 +275,36 @@ html {
...
+When a member of an implicit receiver and a declaration from a [context parameter](context-parameters.md) are both in a scope with the same name,
+the compiler reports a warning because the implicit receiver is shadowed by the context parameter.
+To resolve this, use a `this` qualifier to explicitly call the receiver, or use `contextOf<T>()` to call the context declaration:
+
+```kotlin
+interface HtmlTag {
+    fun setAttribute(name: String, value: String)
+}
+
+// Declares a top-level function with the same name available through a context parameter
```
AlejandraPedroza: It seems a bit hard to read. What's actually "available", the top-level function or the name?
To make it clearer, does the following work?: "// Declares a top-level function with the same name, which is available through a context parameter"

docs/topics/type-safe-builders.md
daniCsorbaJB: hmm 🤔 there are many potential options for it. We could link to [implicit this](https://kotlinlang.org/docs/this-expressions.html?utm_source=chatgpt.com#implicit-this) which is the closest in my opinion

What do you think?

docs/topics/type-safe-builders.md
daniCsorbaJB: I like it I think it works 👍 
I'll split into two lines though to make the output readable without scrolling

AlejandraPedroza (approved):
Nice job. LGTM!

PR #5067, opened 2025-09-09
Fixing irrelevant auto-complete is now top priority (KTIJ-32954)

LouisCAD:
https://youtrack.jetbrains.com/issue/KTIJ-32954/

### DIFF
--- a/docs/topics/roadmap.md
+++ b/docs/topics/roadmap.md
@@ -13,7 +13,20 @@
 
 Welcome to the Kotlin roadmap! Get a sneak peek into the priorities of the JetBrains Team.
 
-## Key priorities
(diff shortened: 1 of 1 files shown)


koshachy:
Hi @LouisCAD,

Thank you for your contribution! Unfortunately, we can't accept this PR. We don't change key directions in the roadmap because of a tooling bug. The Kotlin roadmap covers many subsystems, and prioritizing one of them based only on user opinion would be unfair to other teams.
I understand you may be unhappy with some aspects of working with Kotlin, but updating the public roadmap is not the right channel for this feedback. Please avoid using documentation for this purpose in the future.

Cheers!

PR #5080, opened 2025-09-11
feat(ktl-2781): added data for case studies page

berezinant:


data/case-studies/case-studies.yml
```diff
@@ -0,0 +1,523 @@
+items:
```
nikpachoo, line 1: @berezinant, could you please explain the difference beetwen  case-studies.yml and _case-study.example.yaml? I'd expect to have one data file not two

nikpachoo (commented):
Would be great to add a section to readme.md describing how to add a case study.

data/case-studies/case-studies.yml
berezinant: There is one data file, case-studies.yml is the actual data file, and _case-study.example.yaml is the example featuring and commenting all possible fields and types of content.

As for the README.md I'll sure add there some notes on how to edit case studies

nikpachoo:
this is a parf of #5088

PR #5087, opened 2025-09-25
Spring CrudRepository: remove that id should be var

Ironeer:
All the code snippets show `val id` and during manual tests that worked. Therefore the removed sentence is misleading.


danil-pavlov:
@Ironeer, you're absolutely right, thank you for your contribution! I'm sorry it took so long to get to the PR
Due to conflicts, closing in favor of #5381

PR #5090, opened 2025-10-01
Expand HttpClient explanation in Extensions chapter of the Tour

sarahhaggarty:
This PR adds additional explanation in the Extensions chapter of the Kotlin tour, to explain in more detail how the `HttpClient` class works with extension functions.

### DIFF
--- a/docs/topics/tour/kotlin-tour-intermediate-extension-functions.md
+++ b/docs/topics/tour/kotlin-tour-intermediate-extension-functions.md
@@ -86,8 +86,9 @@ fun HttpClient.post(url: String): HttpResponse = request("POST", url, emptyMap()
 ```
 {validate="false"}
 
-These `.get()` and `.post()` functions call the `request()` function with the correct HTTP method, so you don't have to.
-They streamline your code and make it easier to understand:
+These `.get()` and `.post()` functions extend the `HttpClient` class. They can directly use the `request()` function from the `HttpClient` class
+because they're called on an instance of the `HttpClient` class as the receiver. You can use these extension functions to
+call the `request()` function with the appropriate HTTP method, which simplifies your code and makes it easier to understand:
 
 ```kotlin
 class HttpClient {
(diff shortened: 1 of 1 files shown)


serras (approved):

docs/topics/tour/kotlin-tour-intermediate-extension-functions.md
```diff
@@ -86,8 +86,10 @@ fun HttpClient.post(url: String): HttpResponse = request("POST", url, emptyMap()
 ```
 {validate="false"}
 
-These `.get()` and `.post()` functions call the `request()` function with the correct HTTP method, so you don't have to.
-They streamline your code and make it easier to understand:
+These `.get()` and `.post()` functions extend the `HttpClient` class. Because they're called on an instance of the `HttpClient` class
+as the receiver, they can directly use the `request()` function from the `HttpClient` class. These extension functions handle
+the common cases for you by calling the `request()` function with the appropriate HTTP method. They simplify your code
```
zamulla: I would suggest going second person here, because the current version may imply too much agency for functions: "You can use these extension functions to [conveniently?] call the `request()` function with the appropriate HTTP method. This way you simplify your code and make it easier to understand."

docs/topics/tour/kotlin-tour-intermediate-extension-functions.md
```diff
@@ -86,8 +86,10 @@ fun HttpClient.post(url: String): HttpResponse = request("POST", url, emptyMap()
 ```
 {validate="false"}
 
-These `.get()` and `.post()` functions call the `request()` function with the correct HTTP method, so you don't have to.
-They streamline your code and make it easier to understand:
+These `.get()` and `.post()` functions extend the `HttpClient` class. Because they're called on an instance of the `HttpClient` class
```
zamulla:
```suggestion
These `.get()` and `.post()` functions extend the `HttpClient` class. Since they're called on an instance of the `HttpClient` class
```

This feels nitpicky, "because" is just something that goes after a premise for me.

Another thing to consider is flipping the sentence around: "They can directly use ... because they are called on an instance ...", to put the new/non-obvious information at the end.

zamulla (approved):
Nothing major, just a couple of phrasing suggestions to consider.

PR #5099, opened 2025-10-07
Improve Extensions page

sarahhaggarty:
This PR improves the examples and overall structure of the Extensions page.

### DIFF
--- a/docs/topics/extensions.md
+++ b/docs/topics/extensions.md
@@ -1,55 +1,191 @@
 [//]: # (title: Extensions)
 
-Kotlin provides the ability to extend a class or an interface with new functionality
-without having to inherit from the class or use design patterns such as _Decorator_.
-This is done via special declarations called _extensions_.
+Kotlin _extensions_ let you extend a class or an interface with new functionality without using inheritance or
+design patterns like _Decorator_. They are useful when working with third-party libraries you can't modify 
+directly. Once created, you call these extensions as if they were members of the original class or interface.
 
-For example, you can write new functions for a class or an interface from a third-party library that you can't modify.
-Such functions can be called in the usual way, as if they were methods of the original class.
-This mechanism is called an _extension function_. There are also _extension properties_ that let you define
-new properties for existing classes.
+The most common forms of extensions are [_extension functions_](#extension-functions) and [_extension properties_](#extension-properties).
+
+Importantly, extensions don't modify the classes or interfaces they extend. When you define an extension, you don't add new members.
+You make new functions callable or new properties accessible using the same syntax.
+
+## Receivers
+
+Extensions are always called on a receiver. The receiver has to have the same type as the class or interface being extended.
+To use an extension, prefix it with the receiver followed by a `.` and the function or property name.
+
+For example, the [`.appendLine()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/or-empty.html) extension function from the standard library extends the `StringBuilder` class.
+So in this case, the receiver is a `StringBuilder` instance, and the _receiver type_ is `StringBuilder`:
+
+```kotlin
+fun main() { 
+//sampleStart
+    // builder is an instance of StringBuilder
+    val builder = StringBuilder()
+        // Calls .appendLine() extension function on builder
+        .appendLine("Hello")
+        .appendLine()
+        .appendLine("World")
+    println(builder.toString())
+    // Hello
+    //
+    // World
+}
+//sampleEnd
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-stringbuilder"}
 
 ## Extension functions
 
-To declare an extension function, prefix its name with a _receiver type_, which refers to the type being extended.
-The following adds a `swap` function to `MutableList<Int>`:
+Before creating your own extension functions, see if what you are looking for is already available in the Kotlin [standard library](https://kotlinlang.org/api/core/kotlin-stdlib/).
+The standard library provides many useful extension functions for:
+
+* Operating on collections: [`.map()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/map.html), [`.filter()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/filter.html), [`.reduce()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/reduce.html), [`.fold()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/fold.html), [`.groupBy()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/group-by.html).
+* Converting to strings: [`.joinToString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/join-to-string.html).
+* Working with null values: [`.filterNotNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/filter-not-null.html).
+
+To create your own extension function, prefix its name with a receiver type followed by a `.`. In this example, the `.truncate()`
+function extends the `String` class, so the receiver type is `String`:
 
 ```kotlin
-fun MutableList<Int>.swap(index1: Int, index2: Int) {
-    val tmp = this[index1] // 'this' corresponds to the list
-    this[index1] = this[index2]
-    this[index2] = tmp
+fun String.truncate(maxLength: Int): String {
+    return if (this.length <= maxLength) this else take(maxLength - 3) + "..."
+}
+
+fun main() {
+    val shortUsername = "KotlinFan42"
+    val longUsername = "JetBrainsLoverForever"
+
+    println("Short username: ${shortUsername.truncate(15)}") 
+    // KotlinFan42
+    println("Long username:  ${longUsername.truncate(15)}")
+    // JetBrainsLov...
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-truncate"}
 
-The `this` keyword inside an extension function corresponds to the receiver object (the one that is passed before the dot).
-Now, you can call such a function on any `MutableList<Int>`:
+The `.truncate()` function truncates any string that it's called on by the number in the `maxLength` argument and adds an ellipsis `...`.
+If the string is shorter than `maxLength`, the function returns the original string.
 
-```kotlin
-val list = mutableListOf(1, 2, 3)
-list.swap(0, 2) // 'this' inside 'swap()' will hold the value of 'list'
-```
-
-This function makes sense for any `MutableList<T>`, and you can make it generic:
+In this example, the `.displayInfo()` function extends the `User` interface:
 
 ```kotlin
-fun <T> MutableList<T>.swap(index1: Int, index2: Int) {
-    val tmp = this[index1] // 'this' corresponds to the list
-    this[index1] = this[index2]
-    this[index2] = tmp
+interface User {
+    val name: String
+    val email: String
+}
+
+fun User.displayInfo(): String = "User(name=$name, email=$email)"
+
+// Inherits from and implements the properties of the User interface
+class RegularUser(override val name: String, override val email: String) : User
+
+fun main() {
+    val user = RegularUser("Alice", "alice@example.com")
+    println(user.displayInfo()) 
+    // User(name=Alice, email=alice@example.com)
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-interface"}
 
-You need to declare the generic type parameter before the function name to make it available in the receiver type expression.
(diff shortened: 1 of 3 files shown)


docs/topics/extensions.md
```diff
@@ -1,55 +1,195 @@
...
+Kotlin _extensions_ let you extend a class or an interface with new functionality without using inheritance or
+design patterns like _Decorator_. They are useful when working with third-party libraries you can't modify 
+directly. Once created, you call these extensions as if they were members of the original class or interface.
 
-For example, you can write new functions for a class or an interface from a third-party library that you can't modify.
-Such functions can be called in the usual way, as if they were methods of the original class.
-This mechanism is called an _extension function_. There are also _extension properties_ that let you define
-new properties for existing classes.
+The most common forms of extensions are _extension functions_ and [_extension properties_](#extension-properties), which
+are available in [companion objects](#companion-object-extensions) as well as classes and interfaces.
```
serras: Citing companion objects at this point is quite strange, since extending it is usually considered an advanced pattern in Kotlin.

```suggestion
The most common forms of extensions are _extension functions_ and [_extension properties_](#extension-properties). You can extend any Kotlin type, regardless of being defined as a class, interface, or object.
```

docs/topics/extensions.md
```diff
@@ -1,55 +1,195 @@
...
+Extensions don't modify the classes or interfaces they extend. By defining an extension, you aren't inserting new members,
+only making new functions callable or new properties accessible via special syntax.
+
+## Receivers
+
+Extensions are always called on a receiver. The receiver has to have the same type as the class or interface being extended.
+To use an extension, prefix it with the receiver followed by a `.` and the function or property name.
+
+For example, the [`.orEmpty()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/or-empty.html) extension function
+from the standard library extends the `String?` class:
```
serras: Nitpick: `String?` is a _type_, not a class (`String` is a both a class and a type, though).

docs/topics/extensions.md
```diff
@@ -1,55 +1,195 @@
...
 For more information about generics, see [generic functions](generics.md).
 
-## Extensions are resolved _statically_
+### Nullable receivers
 
-Extensions do not actually modify the classes they extend. By defining an extension, you are not inserting new members into
-a class, only making new functions callable with the dot-notation on variables of this type.
+You can define extension functions with a nullable receiver type, which allows you to call them on a variable
+even if its value is null. When the receiver is `null`, `this` is also `null`. To avoid compiler errors, we recommend checking
+`this == null` check inside the function body when defining an extension with a nullable receiver.
```
serras:
```suggestion
even if its value is null. When the receiver is `null`, `this` is also `null`. 
You are responsible for handling nullability correctly within the function, for example by using `this == null` checks, or the safe call and Elvis operators.
```

I think that the text there is too prescriptive on what to do with `null`s, but there are many different ways to handle them.

docs/topics/extensions.md
```diff
@@ -60,143 +200,206 @@ fun main() {
...
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-member-function-overload"}
+
+In this example, since an `Int` is passed to the `.printFunctionType()` function, the compiler chooses the extension
+function because it matches the signature. The compiler ignores the member function, which takes no arguments.
 
-## Nullable receiver
+## Anonymous extension functions
```
serras:
```suggestion
### Anonymous extension functions
```

Should this be a sub-section of "Extension functions"?

docs/topics/extensions.md
```diff
@@ -60,143 +200,206 @@ fun main() {
...
 ```kotlin
-class MyClass {
-    companion object { }  // will be called "Companion"
-}
+data class User(val firstName: String, val lastName: String)
 
-fun MyClass.Companion.printCompanion() { println("companion") }
+// An extension property to get a username-style email handle
+val User.emailUsername: String
+get() = "${firstName.lowercase()}.${lastName.lowercase()}"
```
serras:
```suggestion
    get() = "${firstName.lowercase()}.${lastName.lowercase()}"
```

docs/topics/extensions.md
```diff
@@ -60,143 +200,206 @@ fun main() {
...
+// Error: Initializers are not allowed for extension properties
+
+// Compiles successfully
+val houseNumbers = mutableMapOf<House, Int>()
+var House.number: Int
+get() = houseNumbers[this] ?: 1
+set(value) {
+    println("Setting house number for ${this.streetName} to $value")
+    houseNumbers[this] = value
+}
```
serras:
```suggestion
    get() = houseNumbers[this] ?: 1
    set(value) {
        println("Setting house number for ${this.streetName} to $value")
        houseNumbers[this] = value
    }
```

docs/topics/extensions.md
```diff
@@ -206,81 +409,198 @@ class Host(val hostname: String) {
...
+    println("Cleaned: '$cleaned'")
+    // Cleaned: 'user@example.com'
+    println("Looks like an email: ${cleaned.contains("@") && cleaned.contains(".")}") 
+    // Looks like an email: true
 }
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-visibility-top-level"}
+
+And if an extension is declared outside its receiver type, it can't access the receiver's `private` or `protected` members:
```
serras, line 555: Maybe add that `internal` members may only be accessed if you're in the same module?

docs/topics/extensions.md
sarahhaggarty: Fair enough! I'll remove this paragraph 👍

docs/topics/extensions.md
sarahhaggarty: I opened it up a bit based on your suggestion! Please check if this is clearer.

docs/topics/extensions.md
sarahhaggarty: Added!

docs/topics/extensions.md
sarahhaggarty: So is it correct to say that `.orEmpty()` extends the `String` class?

serras: Not really. The `orEmpty` function just extends the `String?` type.

Why this distinction about extensions extending _types_ and not classes is important for two reasons:
- You can use nullable types, generics, everything you want there.
- If the class has type parameters, they need to be given (because otherwise they are not a type).

For the second point, this means that you can extend `List<Int>` or `<A> List<A>` generically, but you cannot write `List.extensionFunction`, because `List` is not a (well-formed) type.

docs/topics/extensions.md
```diff
@@ -1,55 +1,194 @@
...
+The standard library provides many useful extension functions for:
+
+* Operating on collections: [`.map()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/map.html), [`.filter()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/filter.html), [`.reduce()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/reduce.html), [`.fold()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/fold.html), [`.groupBy()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/group-by.html).
+* Converting to strings: [`.joinToString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/join-to-string.html).
+* Working with null values: [`.filterNotNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/filter-not-null.html).
+
+To create your own extension function, prefix its name with a receiver type followed by a `.`. In this example, the `.truncate()`
+function extends the `String` class so the receiver type is `String`:
```
danil-pavlov:
```suggestion
function extends the `String` class, so the receiver type is `String`:
```

docs/topics/extensions.md
```diff
@@ -1,55 +1,194 @@
...
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-nullable-receiver"}
 
-Extensions do not actually modify the classes they extend. By defining an extension, you are not inserting new members into
-a class, only making new functions callable with the dot-notation on variables of this type.
+### Extension or member functions?
 
-Extension functions are dispatched _statically_. So which extension function is called is already known at compile time
-based on the receiver type. For example:
+Since calling extension functions and member functions uses the same notation, how does the compiler know which one to use?
```
danil-pavlov:
``` suggestion
Since extension and member function calls have the same notation, how does the compiler know which one to use?
```

docs/topics/extensions.md
```diff
@@ -60,143 +199,206 @@ fun main() {
...
-## Companion object extensions
+## Extension properties
 
-If a class has a [companion object](object-declarations.md#companion-objects) defined, you can also define extension
-functions and properties for the companion object. Just like regular members of the companion object,
-they can be called using only the class name as the qualifier:
+Kotlin supports extension properties, which are useful for performing data transformations or creating UI display helpers
+without cluttering the class you're working with. 
+
+To create an extension property, write the name of the class that you want to extend followed by a `.` and the name of your property.
```
danil-pavlov:
```suggestion
To create an extension property, write the name of the class that you want to extend, followed by a `.` and the name of your property.
```

docs/topics/extensions.md
```diff
@@ -1,55 +1,194 @@
...
+To use an extension, prefix it with the receiver followed by a `.` and the function or property name.
+
+For example, the [`.orEmpty()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/or-empty.html) extension function
+from the standard library extends the `String?` class:
+
+```kotlin
+fun String?.orEmpty(): String
+```
+
+So the receiver is a `String?` instance and the _receiver type_ is `String?`:
```
danil-pavlov:
```suggestion
So in this case, the receiver is a `String?` instance, and the _receiver type_ is `String?`:
```

docs/topics/extensions.md
```diff
@@ -1,55 +1,194 @@
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-extension-function-mostvoted"}
+
+The `.mostVoted()` function iterates through the key-value pairs of the map it's called on and uses the [`maxByOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/max-by-or-null.html)
+function to return the key of the pair containing the highest value. If the map is empty, the `maxByOrNull()` function
+returns `null`. The `mostVoted()` function uses a safe call `?.` to only access the `key` property when the `maxByOrNull()` function
+returns a non-null value.
+
+### Generic extension functions
+
+You can also create generic extension functions. Declare the generic type parameter before the function name
```
danil-pavlov:
```suggestion
You can also create generic extension functions. To do that, declare the generic type parameter before the function name
```

docs/topics/extensions.md
```diff
@@ -1,55 +1,194 @@
...
+directly. Once created, you call these extensions as if they were members of the original class or interface.
 
-For example, you can write new functions for a class or an interface from a third-party library that you can't modify.
-Such functions can be called in the usual way, as if they were methods of the original class.
-This mechanism is called an _extension function_. There are also _extension properties_ that let you define
-new properties for existing classes.
+The most common forms of extensions are [_extension functions_](#extension-functions) and [_extension properties_](#extension-properties).
+
+Importantly, extensions don't modify the classes or interfaces they extend. By defining an extension, you aren't inserting new members,
+only making new functions callable or new properties accessible via special syntax.
```
danil-pavlov:
```suggestion
only making new functions callable or new properties accessible using special syntax.
```

docs/topics/extensions.md
sarahhaggarty: I rephrased the sentences to "To create generic extension functions, declare...". Is this an acceptable alternative?

docs/topics/extensions.md
```diff
@@ -1,55 +1,191 @@
...
+directly. Once created, you call these extensions as if they were members of the original class or interface.
 
-For example, you can write new functions for a class or an interface from a third-party library that you can't modify.
-Such functions can be called in the usual way, as if they were methods of the original class.
-This mechanism is called an _extension function_. There are also _extension properties_ that let you define
-new properties for existing classes.
+The most common forms of extensions are [_extension functions_](#extension-functions) and [_extension properties_](#extension-properties).
+
+Importantly, extensions don't modify the classes or interfaces they extend. By defining an extension, you aren't inserting new members,
+only making new functions callable or new properties accessible using special syntax.
```
serras:
```suggestion
only making new functions callable or new properties accessible using the same syntax as regular members.
```

serras (approved):

docs/topics/extensions.md
sarahhaggarty: I replaced the example so this isn't a problem anymore.

docs/topics/extensions.md
sarahhaggarty: Applied with a little polish!

PR #5461, opened 2026-03-26
Fix KT-85276: Update Gradle + JPMS example

not-a-hack-er:
Fixes KT-85276

Updated the Gradle + JPMS example to ensure compatibility with the Gradle configuration cache.

- Replaced direct sourceSets output usage with a FileCollection
- Removed outdated Gradle < 7.0 configuration

### DIFF
--- a/docs/topics/gradle/gradle-configure-project.md
+++ b/docs/topics/gradle/gradle-configure-project.md
@@ -525,17 +525,11 @@ tasks.named("compileJava", JavaCompile::class.java) {
 
 ```groovy
 // Add the following three lines if you use a Gradle version less than 7.0
-java {
-    modularity.inferModulePath = true
-}
-
-tasks.named("compileJava", JavaCompile.class) {
-    options.compilerArgumentProviders.add(new CommandLineArgumentProvider() {
-        @Override
-        Iterable<String> asArguments() {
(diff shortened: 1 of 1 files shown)


antohaby:
Hi @not-a-hack-er ,

Thanks for willing to contribute to Kotlin!

The change seemed to be a direct copy paste from https://youtrack.jetbrains.com/issue/KT-85276
While this is OK, but it should be done properly.

Kotlin docs has at least two sections "groovy" and "kotlin script". please ensure you update both sections.

You also need to provide a sample gradle projects that prove that snippet actually work.
We don't have publicly available test infrastructure for that, so please do the following:
1. create two small gradle projects, one with groovy, another with kts
2. use 1-to-1 matching snippets from kotlin docs that you're proposing. wrap them with comments `//sampleStart` and `//sampleEnd`
3. run them and provide logs and other evidences that these script work. Pay attention to configuration cache.

Then I will be able to review your change properly. 

Thanks

antohaby (changes requested):
Requested changes in the comment above.

sarahhaggarty:
Hi @not-a-hack-er !
Do you have some time to complete @antohaby request? Please let us know. If we don't hear anything from you in the next two weeks, I'll have to close the PR.

PR #5484, opened 2026-04-09
Fix the 'package' property of the definition file.

bennyhuo:
There should only be a 'package' property in the definition file while we should use 'packageName' in the gradle file.


SvyatoslavScherbina (approved):

danil-pavlov (approved):
Nice catch! Thanks @bennyhuo

PR #5485, opened 2026-04-09
update: String page

aepanchi:


docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+> A `String` index starts at zero. 
+> 
+{style="note"}
+
+If you try to access an index outside the valid range, Kotlin throws an exception:
```
haas-dtv: Do we want to specify which exception is thrown?

docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+## Extract parts of a string
+
+To extract part of a `String`, use substring functions. 
+
+Substring functions return a new `String` that contains the selected part of the original text.
```
haas-dtv: I'd leave that out, or add something like "As `String` is immutable, ...", else it's redundant.

docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
+ ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Modify `String`
+
+> Strings are immutable. Functions that change text return a new `String`.
+> 
+{style="note"}
+
+To create a modified copy of a `String`, you can use `replace()`, `uppercase()`, and `lowercase()`:
```
haas-dtv: This is a bit weirdly worded, as it sounds like these are the only methods that would create a modified version of a `String`.

Maybe describe these 3 methods as examples?

docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
+    val first = "Hello"
+    val second = "Kotlin"
+    val message = first + ", " + second
+    println(message) // Hello, Kotlin
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+This creates a new `String` that contains the combined text of both operands.
```
haas-dtv: Are we sure this always creates a new `String`, or could the compiler maybe optimize stuff here, like `foo + ""` returning `foo`, or the constant expression `"Hello "+ "World"` returning an already interned `String` on the JVM?

docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
+    val language = "Kotlin"
+    println(language.length) // 6
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Concatenate strings
+
+To concatenate strings, use the `+` operator:
```
serras: I'm not sure this is an important topic, in most cases people use string interpolation. The code you've written would be definitely more idiomatic as 

```kotlin
val message = "$first, $second"
```

docs/topics/strings.md
```diff
@@ -186,93 +213,264 @@ val KClass<*>.jsonSchema : String
...
-```
-{interpolate-variables="false" kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+ ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Modify `String`
+
+> Strings are immutable. Functions that change text return a new `String`.
+> 
+{style="note"}
```
serras: The combination of this title + the note reads quite weird: are strings modifiable or not? I would suggest something like "String replacement" as title, and telling that `uppercase` and `lowercase` are about replacing lowercase by uppercase, or vice versa.

serras (approved):
The content seems OK. I'm missing some link or discussion of regular expressions, which are very related to strings (especially once we start talking about matching, splitting, or replacing).

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
 [//]: # (title: Strings)
+[//]: # (description: Learn how to work with strings in Kotlin, including string literals, string templates, multiline strings, and common text operations.)
 
-Strings in Kotlin are represented by the type [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/). 
+<show-structure depth="1"/>
 
-> On the JVM, an object of `String` type in UTF-16 encoding uses approximately 2 bytes per character.
-> 
-{style="note"}
+The [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/) type represents a sequence of
```
fzhinkin: Should it be `https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/` instead?

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
...
+val name: String = "Kotlin"
+val message = "Hello, world!" // Kotlin infers String
 ```
 
-To concatenate strings, use the `+` operator. This also works for concatenating strings with values of other types, as long
-as the first element in the expression is a string:
+### Nullable values
 
+To declare a nullable value, use `String?`:
```
fzhinkin: Any nullable value will be declared the same way. Is there a reason to showcase it for strings in particular?

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
...
 ```
 
-> In most cases using [string templates](#string-templates) or [multiline strings](#multiline-strings) is preferable to string concatenation.
-> 
-{style="note"}
+### Escape sequences
 
-## String literals
+Double-quoted strings support [escape sequences](characters.md#escape-sequences) such as `\n` or `\t`:
```
fzhinkin:
```suggestion
Double-quoted string literals support [escape sequences](characters.md#escape-sequences) such as `\n` or `\t`: 
```

I don't have any strong opinion here, but for me "double-quoted" is about literals, not string objects.

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
...
-
+    println(text)
 //sampleEnd
 }
 ```
 
-or an expression in curly braces:
+To control indentation more explicitly, use 
+[`trimMargin()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-margin.html). This function
```
fzhinkin: `https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html`?

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
...
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+> You can also combine strings with the `+` operator. However, string templates are usually easier 
+> to read and more idiomatic.
+>
+{style="tip"}
+
+In multiline strings, template expressions can contain double-quoted strings without escaping:
```
fzhinkin: It works for double-quoted string literals too:
```kotlin
 println("${"test".uppercase()}")
```
https://pl.kotl.in/6h9mIXaYY

docs/topics/strings.md
```diff
@@ -186,93 +222,270 @@ val KClass<*>.jsonSchema : String
...
+  
+    println(domain.substringBefore(".")) // kotlinlang
+    println(domain.substringAfter("."))  // org
+//sampleEnd
+}
 ```
 
-Here, the `$$$` prefix allows the string to include `$` and `$$` without requiring the `${'$'}` construct for escaping.
+  > Use delimiter-based functions to separate structured text, such as file names, paths, or key-value pairs.
```
fzhinkin: I don't think it's worth recommending string operations for file paths and names, it's a common cause of various bugs.

docs/topics/strings.md
```diff
@@ -186,93 +222,270 @@ val KClass<*>.jsonSchema : String
...
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+The start index is **inclusive**, and the end index is **exclusive**.
+
+### Extract around a delimiter
+
+Use [`substringBefore()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/substring-before.html)
```
fzhinkin: There are many different functions to parse a string, is there a reason these two functions have a dedicated section in the doc?

docs/topics/strings.md
```diff
@@ -186,93 +222,270 @@ val KClass<*>.jsonSchema : String
...
+
+* To build a `String` step by step, use 
+  [`buildString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/build-string.html):
+
+  ```kotlin 
+  val result = buildString {
+      append("Hours: ")
+      append("12")
+      append(", minutes: ")
+      append("30")
```
fzhinkin: This particular string does not require a builder or a template to construct. `buildString` is handy when building a string require some complex logic (like multiple conditions affecting the resulting string).

docs/topics/strings.md
```diff
@@ -1,177 +1,213 @@
...
-    val str = "abcd" 
-//sampleStart
-    for (c in str) {
-        println(c)
-    }
-//sampleEnd
-}
-```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+To declare a `String`, enclose the value in double quotes (`""`).
```
ilya-g: Strictly speaking, this is a string literal.

To declare a string one writes `val s: String = ...` and the quoted sequence of characters can be used as a literal value for that string.

docs/topics/strings.md
aepanchi: Addressed

docs/topics/strings.md
aepanchi: fixed

docs/topics/strings.md
aepanchi: true, thank you! fixed

docs/topics/strings.md
aepanchi: I believe we should mention nullable values on this page, since it's an overview one. However, I agree that it should be a separate section. What do you think about adding this example to the code block above (lines 20-21) with the corresponding comment?

docs/topics/strings.md
aepanchi: fair enough, fixed

docs/topics/strings.md
aepanchi: thanks! added to the example

docs/topics/strings.md
aepanchi: removed

docs/topics/strings.md
aepanchi: Agree; however, that example is used to showcase the logic behind `buildString`, since particular use cases are too specific for such a page. The goal of this sample is to highlight that this function exists, rather than explain the concept in detail. What do you think?

docs/topics/strings.md
aepanchi: It seems to be the easiest ones to comprehend for the novice users (the target audience of this page) and to show the basic functionality. Are there other functions that, in your opinion, can replace these two?

docs/topics/strings.md
fzhinkin: I totally agree that examples should be concise and easy to grasp without any additional domain- or context-specific knowledge. However, it's not the most efficient or elegant way to construct the "Hours: 12, minutes: 30" string. And we should  probably discourage users from constructing string literals that way. At the same time, the string builder itself has a particular practical niche, and it would be nice to showcase it instead.

docs/topics/strings.md
```diff
@@ -186,93 +223,267 @@ val KClass<*>.jsonSchema : String
...
 
-In addition, you can assign the format string from a variable. This can be useful when the format string changes, 
-for example, in localization cases that depend on the user locale.
+  ```kotlin
+  val text = String.format("Hello, %s", "Kotlin")
+  ```
 
-Be careful when using the `String.format()` function because it can be easy to mismatch the number or position of the 
-arguments with their corresponding placeholders.
+  > Use `String.format()` only when you specifically need formatter-style specifiers on the JVM.
```
fzhinkin: Does it make sense to add a small hint on when one may need it? For example, when a template is used to format a localized message.

docs/topics/strings.md
```diff
@@ -1,177 +1,214 @@
...
+  ```kotlin
+  fun main() {
+  //sampleStart
+      val text = "abc"
+      println("The length of $text is ${text.length}")
+      // // The length of abc is 3
+      val language = "Kotlin"
+      println("${language}Lang")
+      // KotlinLang
+  // sampleEnd
```
sarahhaggarty:
```suggestion
  //sampleEnd
```

docs/topics/read-standard-input.md
```diff
@@ -14,6 +14,23 @@ println(readln())
 // Hi, Kotlin!
 ```
 
+You can also read multiline input as one string.
+For that join it with newline characters using 
+[`joinToString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/join-to-string.html):
```
sarahhaggarty:
```suggestion
Use the [`joinToString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/join-to-string.html) function with newline characters to combine the lines:
```
As an alternative to the "for that" construction. WDYT?

docs/topics/read-standard-input.md
```diff
@@ -14,6 +14,23 @@ println(readln())
...
+```kotlin
+val input = generateSequence(::readlnOrNull)
+  .joinToString("\n")
+```
+
+> This code does not work correctly with `readln()`, because the function throws an
+> exception at end of input instead of returning `null`.
+> 
+> To read input until end-of-file, use
+> [`readlnOrNull()`](#handle-standard-input-safely).
```
sarahhaggarty: I would actually reverse this note to give the advice and then the reason why. Maybe something like:

```suggestion
> Use [`readlnOrNull()`](#handle-standard-input-safely) instead of `readln()` to read input until end-of-file without throwing an exception.
```
WDYT?

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+The [`String`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/) type represents a sequence of
+[characters](characters.md). 
 
-> On the JVM, an object of `String` type in UTF-16 encoding uses approximately 2 bytes per character.
-> 
-{style="note"}
+Use `String` for text values, such as words, sentences, messages, or structured text. 
 
-Generally, a string value is a sequence of characters in double quotes (`"`):
+The `String` type is immutable.
```
sarahhaggarty:
```suggestion
The `String` type is immutable. After an object with `String` type is created, its contents stay the same for the rest of its lifetime. Any operation that appears to modify the string actually creates a new string.
```
Could we expand on this a little to explain what impact it has? WDYT?

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 
 ```kotlin
-val s = "Hello, world!\n"
+val message = "Hello,\nworld!"
+val quote = "Kotlin says, \"Hi\"."
 ```
 
-Escaping is done in the conventional way, with a backslash (`\`).  
-See [Characters](characters.md) page for the list of supported escape sequences.
+Use escape sequences for special characters that are difficult to write directly in source code or have a
```
sarahhaggarty:
```suggestion
Use escape sequences for special characters that are difficult to write directly in source code or that have a
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+special meaning.
 
 ### Multiline strings
 
-_Multiline strings_ can contain newlines and arbitrary text. It is delimited by a triple quote (`"""`),
-contains no escaping and can contain newlines and any other characters:
+To store text that consists of multiple lines or contains quotes that you do not want to escape,
```
sarahhaggarty:
```suggestion
To store text that consists of multiple lines or contains quotes that you don't want to escape,
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
-    """
+Hello,
+Kotlin
+"""
+
+val quote = """Kotlin says, "Hi"."""
 ```
 
-To remove leading whitespace from multiline strings, use the [`trimMargin()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-margin.html) function:
+> Multiline strings do not support escape sequences.
```
sarahhaggarty:
```suggestion
> Multiline strings don't support escape sequences. 
```
It's OK to use contractions :)

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 
-To remove leading whitespace from multiline strings, use the [`trimMargin()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-margin.html) function:
+> Multiline strings do not support escape sequences. 
+> Characters such as `\n` and `\t` are treated as regular text. 
+>
+{style="note"}
+
+Multiline strings preserve **line breaks** and **indentations** as written in the source code. 
+
+You can use this when you want the runtime value to match the text layout in your file.
```
sarahhaggarty:
```suggestion
This behavior is useful when you want the runtime value to match the text layout in your file.
```
I think this version makes "this" more explicit. WDYT? And how about combining it in a paragraph with the previous sentence?

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+
+In the following example, the spaces before each line are part
+of the resulting `String`:
```
sarahhaggarty:
```suggestion
of the resulting string:
```
Since we're not talking about the `String` type here.

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 
-By default, a pipe symbol `|` is used as margin prefix, but you can choose another character and pass it as a parameter, like `trimMargin(">")`.
-
-## String templates
-
-String literals may contain _template expressions_ – pieces of code that are evaluated and whose results are concatenated into a string.
-When a template expression is processed, Kotlin automatically calls the `.toString()` function on the expression's result
-to convert it into a string. A template expression starts with a dollar sign (`$`) and consists of either a variable name:
+To remove common leading indentation, use 
+[`trimIndent()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-indent.html). It detects the common
```
sarahhaggarty:
```suggestion
the [`trimIndent()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-indent.html) function. It detects the common
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
-
+    println(text)
 //sampleEnd
 }
 ```
 
-or an expression in curly braces:
+To control indentation more explicitly, use 
+[`trimMargin()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html). This function
```
sarahhaggarty:
```suggestion
the [`trimMargin()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html) function. This function 
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 //sampleEnd
 }
 ```
 
-You can use templates both in multiline and escaped strings. However, multiline strings don't support backslash escaping. 
-To insert the dollar sign `$` in a multiline string
-before any symbol allowed at the beginning of an [identifier](https://kotlinlang.org/grammar/#identifiers),
-use the following syntax:
+By default, `trimMargin()` uses a pipe symbol (`|`) as the margin prefix, but you can pass another character
```
sarahhaggarty:
```suggestion
By default, the `trimMargin()` function uses a pipe symbol (`|`) as the margin prefix, but you can pass another character 
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
 [//]: # (title: Strings)
+[//]: # (description: Learn how to work with strings in Kotlin, including string literals, string templates, multiline strings, and common text operations.)
 
-Strings in Kotlin are represented by the type [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/). 
+The [`String`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/) type represents a sequence of
+[characters](characters.md).
```
sarahhaggarty, line 6:
```suggestion
```
Can you combine the first two sentences into a paragraph? Currently it looks a little like it should be a list with three separate sentences at the beginning 🤔

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+[//]: # (description: Learn how to work with strings in Kotlin, including string literals, string templates, multiline strings, and common text operations.)
 
-Strings in Kotlin are represented by the type [`String`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-string/). 
+The [`String`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/) type represents a sequence of
+[characters](characters.md). 
 
-> On the JVM, an object of `String` type in UTF-16 encoding uses approximately 2 bytes per character.
-> 
-{style="note"}
+Use `String` for text values, such as words, sentences, messages, or structured text.
```
sarahhaggarty:
```suggestion
Use the `String` type for text values, such as words, sentences, messages, or structured text. 
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
-    val str = "abcd" 
-//sampleStart
-    for (c in str) {
-        println(c)
-    }
-//sampleEnd
-}
-```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+To declare a `String` literal, enclose the value in double quotes (`""`).
```
sarahhaggarty:
```suggestion
```
Again, I think you can combine these two into a paragraph.

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+val name: String = "Kotlin"
+val message = "Hello, world!" // Kotlin infers String
 ```
-
-To concatenate strings, use the `+` operator. This also works for concatenating strings with values of other types, as long
-as the first element in the expression is a string:
 
+To declare a nullable value, use `String?`:
 ```kotlin
```
sarahhaggarty:
```suggestion

```kotlin
```
Generally, we like to add a newline before code samples. Just to be on the safe side with writerside and it makes it a little easier to read :) Can you check the rest of the samples in the doc?

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+val quote = """Kotlin says, "Hi"."""
 ```
 
-To remove leading whitespace from multiline strings, use the [`trimMargin()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-margin.html) function:
+> Multiline strings do not support escape sequences. 
+> Characters such as `\n` and `\t` are treated as regular text. 
+>
+{style="note"}
+
+Multiline strings preserve **line breaks** and **indentations** as written in the source code.
```
sarahhaggarty: I'm not sure the emphasis is needed in this sentence. Would you mind removing it?

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
-    // Letters: [a, b, c, d, e]
-
+    println(text)
 //sampleEnd
 }
 ```
 
-or an expression in curly braces:
+To control indentation more explicitly, use
```
sarahhaggarty:
```suggestion
To control indentation removal more explicitly, use 
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-You can use templates both in multiline and escaped strings. However, multiline strings don't support backslash escaping. 
-To insert the dollar sign `$` in a multiline string
-before any symbol allowed at the beginning of an [identifier](https://kotlinlang.org/grammar/#identifiers),
-use the following syntax:
+By default, `trimMargin()` uses a pipe symbol (`|`) as the margin prefix, but you can pass another character 
+as a parameter. For example: `trimMargin(">")`.
+
+> When you process a `String` with functions like `trimIndent()` or `trimMargin()`, the resulting `String`
```
sarahhaggarty:
```suggestion
> When you process a string with functions like `trimIndent()` or `trimMargin()`, the resulting string
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
 """
 ```
 
-> To avoid `${'$'}` sequences in strings, you can use the Experimental [multi-dollar string interpolation feature](#multi-dollar-string-interpolation).
->
-{style="note"}
+### Nullable values in string templates
+
+If an interpolated expression or variable evaluates to `null`, the Kotlin compiler 
+inserts the text `null` into the resulting `String`:
```
sarahhaggarty:
```suggestion
inserts the text `null` into the resulting string:
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
     """
 ```
 
-Here, the `$$` prefix specifies that two consecutive dollar signs are required to trigger string interpolation. 
-Single dollar signs remain as literal characters.
 
-You can adjust how many dollar signs trigger interpolation.
-For example, using three consecutive dollar signs (`$$$`) allows `$` and `$$` to remain as literals 
-while enabling interpolation with `$$$`:
+> If you use single-dollar string interpolation, multi-dollar string interpolation does not affect your code.
```
sarahhaggarty:
```suggestion
> If you use single-dollar string interpolation, multi-dollar string interpolation doesn't affect your code.
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+>
```
sarahhaggarty:
```suggestion
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
-      "currency": "$",
-      "enteredAmount": "42.45 $$",
-      "$$serviceField": "none",
-      "product": "$$$productName"
-    }
-    """
+## Basic string operations
+
+Kotlin provides a range of operations for working with strings.
+
```
sarahhaggarty, line 210:
```suggestion
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+This section introduces some of the most commonly used operations.
+
+> Learn more about all available functions in the
+> [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/).
+
+Since the `String` type is immutable, you cannot change its contents. 
+String operations return new strings with modifications applied and do not update
+the original variable.
```
sarahhaggarty: Since I suggest to add this text at the beginning, if you accept that then this duplicated part can be removed :)

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
-println(requestedData)
-//{
-//    "currency": "$",
-//    "enteredAmount": "42.45 $$",
-//    "$$serviceField": "none",
-//    "product": "carrot"
-//}
+### Get string length
+
+To get the number of characters in a `String`, use the
```
sarahhaggarty:
```suggestion
To get the number of characters in a string, use the
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
 ```
 
-Here, the `$$$` prefix allows the string to include `$` and `$$` without requiring the `${'$'}` construct for escaping.
+### Access characters by index
 
-Multi-dollar string interpolation doesn't affect existing code that uses single-dollar string interpolation. 
-You can continue using a single `$`
-as before and apply multi-dollar signs when you need to handle literal dollar signs in strings.
+To access an individual character in a `String`,
```
sarahhaggarty:
```suggestion
To access an individual character in a string, 
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+    println(language[0]) // K
+    println(language[5]) // n
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> String formatting with the `String.format()` function is only available in Kotlin/JVM.
->
+> A `String` index starts at zero.
```
sarahhaggarty:
```suggestion
> A string index starts at zero. 
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> For a detailed understanding and an extensive list of format specifiers, see [Java's Class Formatter documentation](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html#summary).
->
-{style="note"}
+### Extract parts of a string
+
+To extract part of a `String`, use
+[`substring()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/substring.html) functions
+or [`subSequence()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/sub-sequence.html).
```
sarahhaggarty:
```suggestion
or the [`subSequence()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/sub-sequence.html) function. 
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> For a detailed understanding and an extensive list of format specifiers, see [Java's Class Formatter documentation](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html#summary).
->
-{style="note"}
+### Extract parts of a string
+
+To extract part of a `String`, use
```
sarahhaggarty:
```suggestion
To extract part of a string, use
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+[`substring()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/substring.html) functions
+or [`subSequence()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/sub-sequence.html). 
 
-Let's look at an example:
+Substring functions return a new `String` that contains the selected part of the original text. 
+The original `String` is not modified.
```
sarahhaggarty:
```suggestion
The `substring()` functions return a new string that contains the selected part of the original text without modifying the original string.
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+    // Checks if the string ends with ".org"
+    println(domain.endsWith(".org"))  // true
+//sampleEnd
+}
+ ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Split strings
+
+* To divide a `String` into parts around a delimiter, use
```
sarahhaggarty:
```suggestion
* To divide a string into parts around a delimiter, use 
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+    println(domain.endsWith(".org"))  // true
+//sampleEnd
+}
+ ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Split strings
+
+* To divide a `String` into parts around a delimiter, use 
+  [`split()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/split.html):
```
sarahhaggarty:
```suggestion
  the [`split()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/split.html) function:
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+  fun main() { 
+  //sampleStart
+      val numbers = "one, two, three"
+      println(numbers.split(", ")) // [one, two, three]
+  //sampleEnd
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+* To split a `String` into individual lines, use
```
sarahhaggarty:
```suggestion
* To split a string into individual lines, use 
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+  //sampleStart
+      val numbers = "one, two, three"
+      println(numbers.split(", ")) // [one, two, three]
+  //sampleEnd
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+* To split a `String` into individual lines, use 
+  [`lines()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/lines.html):
```
sarahhaggarty:
```suggestion
  the [`lines()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/lines.html) function:
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-> String formatting with the `String.format()` function is only available in Kotlin/JVM.
->
+> A `String` index starts at zero. 
+> 
 {style="note"}
```
sarahhaggarty: Can you make this a tip?

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
-to reference the same argument multiple times within the format string in different formats.
+```kotlin 
+fun main (){
+    //sampleStart
+    val language = "Kotlin"
+    println(language[20]) // Exception
+    //sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
```
sarahhaggarty, line 270:
```suggestion
{kotlin-runnable="true" kotlin-min-compiler-version="1.3" validate="false"}
```
To stop the verify samples GH action from getting upset :)

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
-(indicated by `%`) for a given argument, followed by format specifiers.
-Format specifiers are formatting instructions for the respective argument, consisting of flags, width, precision, and 
-conversion type. Collectively, format specifiers shape the output's formatting. Common format specifiers include 
-`%d` for integers, `%f` for floating-point numbers, and `%s` for strings. You can also use the `argument_index$` syntax 
-to reference the same argument multiple times within the format string in different formats.
+```kotlin 
+fun main (){
+    //sampleStart
+    val language = "Kotlin"
+    println(language[20]) // Exception
```
sarahhaggarty:
```suggestion
    println(language[20]) 
    // Exception
```
We try to put results in code comments below the relevant code.

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+    // Checks if the string starts with "kotlin"
+    println(domain.startsWith("kotlin")) // true
+    // Checks if the string ends with ".org"
+    println(domain.endsWith(".org"))  // true
+//sampleEnd
+}
+ ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Split strings
```
sarahhaggarty, line 393: Can you remove the bullet points in this section? Since there are only two topics I think it is fine without them :)

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+> 
+{style="note"}
+
+
+## String templates
+
+String templates let you embed variables and expressions directly inside a `String` literal. 
+This process is called _interpolation_. 
+
+You can use string templates in both regular and multiline strings.
```
sarahhaggarty:
```suggestion
You can use string templates in both regular and multiline strings:
```

docs/topics/strings.md
```diff
@@ -1,177 +1,210 @@
...
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+* To insert an expression, or to place a variable directly next to other text, use `${}`:
+
+  ```kotlin
+  fun main() {
+  //sampleStart
+      val text = "abc"
+      println("The length of $text is ${text.length}")
+      // // The length of abc is 3
```
sarahhaggarty:
```suggestion
      // The length of abc is 3
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+### Get string length
+
+To get the number of characters in a `String`, use the
+[`length`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/length.html) property:
+
+```kotlin 
+fun main (){
+    //sampleStart
+    val language = "Kotlin"
+    println(language.length) // 6
```
sarahhaggarty:
```suggestion
    println(language.length) 
    // 6
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+use the indexing operator (`[]`): 
 
-## String formatting
+```kotlin 
+fun main (){
+    //sampleStart
+    val language = "Kotlin"
+    
+    println(language[0]) // K
+    println(language[5]) // n
```
sarahhaggarty:
```suggestion
    println(language[0]) 
    // K
    println(language[5]) 
    // n
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
-    val negativeNumberInParentheses = String.format("%(d means %1\$d", -31416)
-    println(negativeNumberInParentheses)
-    //(31416) means -31416
-//sampleEnd    
+fun main() {
+//sampleStart    
+    val text = "Kotlin"
+    println(text.substring(1))    // otlin
+    println(text.substring(1, 5)) // otli
+    println(text.subSequence(1, 5)) // otli
```
sarahhaggarty:
```suggestion
    println(text.substring(1))    
    // otlin
    println(text.substring(1, 5)) 
    // otli
    println(text.subSequence(1, 5)) 
    // otli
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+fun main() { 
+//sampleStart
+    val domain = "kotlinlang.org"
+    
+    // Checks if the string contains "."
+    println(domain.contains(".")) // true
+    // Checks if the string starts with "kotlin"
+    println(domain.startsWith("kotlin")) // true
+    // Checks if the string ends with ".org"
+    println(domain.endsWith(".org"))  // true
```
sarahhaggarty:
```suggestion
    // Checks if the string contains "."
    println(domain.contains(".")) 
    // true
    
    // Checks if the string starts with "kotlin"
    println(domain.startsWith("kotlin")) 
    // true
    
    // Checks if the string ends with ".org"
    println(domain.endsWith(".org"))  
    // true
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+### Split strings
+
+* To divide a `String` into parts around a delimiter, use 
+  [`split()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/split.html):
+
+  ```kotlin
+  fun main() { 
+  //sampleStart
+      val numbers = "one, two, three"
+      println(numbers.split(", ")) // [one, two, three]
```
sarahhaggarty:
```suggestion
      println(numbers.split(", ")) 
      // [one, two, three]
```

docs/topics/strings.md
```diff
@@ -186,93 +219,142 @@ val KClass<*>.jsonSchema : String
...
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+* To split a `String` into individual lines, use 
+  [`lines()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/lines.html):
+
+    ```kotlin
+    fun main() { 
+    //sampleStart
+        val numbers = "one\ntwo\nthree"
+        println(numbers.lines()) // [one, two, three]
```
sarahhaggarty:
```suggestion
        println(numbers.lines()) 
        // [one, two, three]
```

sarahhaggarty (changes requested):
Thanks for your hard work on this! 🚀 I like the new structure. Please take a look at my comments. It's mostly style tweaks.

sarahhaggarty (changes requested):
Thanks for your hard work on this! 🚀 I like the new structure. Please take a look at my comments. It's mostly style tweaks.

docs/topics/strings.md
```diff
@@ -1,278 +1,536 @@
...
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-Multi-dollar string interpolation simplifies this
-by letting you treat dollar signs as literal characters in both single-line and multiline strings.
-For example:
+You can also compare strings lexicographically (character by character) with the [`compareTo()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-string/compare-to.html)
+function. It scans both strings until it finds the first differing pair of characters and returns the
+difference between their Unicode code points. If strings are equal, the function returns `0`.
```
ilya-g: The specific values returned by `compareTo` is an implementation detail and users should not rely on them, only on their relation to zero (== 0, < 0, > 0).

docs/topics/strings.md
```diff
@@ -1,278 +1,536 @@
...
+The `compareTo()` function also allows you to ignore case differences
```
ilya-g: The same is true for comparing strings for equality using `string.equals(other, ignoreCase = true)` function.

docs/topics/strings.md
```diff
@@ -1,278 +1,536 @@
...
+Use [`StringBuilder`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/-string-builder/) when you need the buffer as an explicit value. For example, to pass
+it into another function or build it in separate steps:
 
-In addition, you can assign the format string from a variable. This can be useful when the format string changes, 
-for example, in localization cases that depend on the user locale.
+```kotlin
+fun main() {
+//sampleStart
+    // Creates a new empty StringBuilder
+    val sb = StringBuilder()
```
ilya-g: The sample isn't much different from `buildString` and I would rather use the latter for this task. 
A StringBuilder can be used instead of `buildString`, for example, if one wants to start with an existing content and modify it in-place, e.g. 
```
val sb = StringBuilder(text)
// do something with the builder populated with text, maybe not only 'append'
```

docs/topics/strings.md
```diff
@@ -1,278 +1,533 @@
...
+fun main() {
+//sampleStart    
+    println("abc".compareTo("abd") < 0)
+    // true
+    
+    println("abc".compareTo("ABC") > 0)
+    // true
+    
+    // Pass true to ignore case differences
+    println("abc".compareTo("ABC", true) == 0)
```
ilya-g:
```suggestion
    println("abc".compareTo("ABC", ignoreCase = true) == 0)
```
More explicit what the second argument means

docs/topics/strings.md
```diff
@@ -1,278 +1,533 @@
...
+val text = number.toString()
+```
+
+In string templates and string concatenation, Kotlin converts values to strings automatically.
+
+To convert a string to another type, use the corresponding parsing functions:
+
+* For integer values: [`toByte()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-byte.html), [`toShort()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-short.html), [`toInt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-int.html), [`toLong()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-long.html)
+* For floating-point values: [`toDouble()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-double.html), [`toFloat()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-float.html)
+* For booleans: [`toBoolean()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-boolean.html)
```
ilya-g: Maybe also https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-boolean-strict.html because `toBoolean` has its pitfalls

docs/topics/strings.md
```diff
@@ -1,278 +1,533 @@
...
+
+These functions return a value of the requested type if the string has a valid format.
+If the input may be invalid, use the `OrNull` variants. These functions return `null` 
+instead of throwing an exception making them safe for user input or data that you don't
+fully control:
+
+```kotlin
+val toInt = "10".toInt()
+val toBoolean = "true".toBoolean()
```
ilya-g: `toBooleanStrict` has `OrNull` variant, so we can use `toBooleanStrict` and `toBooleanStrictOrNull` in this example

ilya-g (changes requested):

docs/topics/strings.md
```diff
@@ -526,7 +526,7 @@ fully control:
 
 ```kotlin
 val toInt = "10".toInt()
-val toBoolean = "true".toBoolean()
+val toBoolean = "true".toBooleanStrictOrNull()
```
ilya-g: I meant here `toBooleanStrict` and later `toBooleanStrictOrNull` with some invalid string value for boolean producing null in the result.

docs/topics/strings.md
```diff
@@ -526,7 +526,7 @@ fully control:
...
 
 // 1000000000000 exceeds maximum value of Int
 val toIntInvalid= "1000000000000".toIntOrNull()
```
ilya-g: Minor: code formatting

ilya-g (approved):

serras (approved):

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
+val text = """
+Hello,
+Kotlin
+"""
+
+val quote = """Kotlin says, "Hi"."""
+```
+
+> Multiline strings don't support escape sequences.
+> Kotlin treats such characters as a regular text.
```
sarahhaggarty:
```suggestion
> Kotlin treats these characters as regular text.
```
"These" sounds more natural and regular text doesn't need an article since it's plural.

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
 ```
 
-Escaping is done in the conventional way, with a backslash (`\`).  
-See [Characters](characters.md) page for the list of supported escape sequences.
+### Multi-dollar string interpolation
 
-### Multiline strings
+In regular string templates, a single dollar sign (`$`) starts interpolation. If you need
+to include literal dollar signs in a `String` value, use **multi-dollar string interpolation**.
```
sarahhaggarty:
```suggestion
to include literal dollar signs in a string, use **multi-dollar string interpolation**.
```

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
+
+Multi-dollar string interpolation allows you to specify how many consecutive dollar signs are required
+to trigger interpolation. Dollar signs below that number are treated as literal characters.
 
-_Multiline strings_ can contain newlines and arbitrary text. It is delimited by a triple quote (`"""`),
-contains no escaping and can contain newlines and any other characters:
+For example, when you use `$$` before a `String` literal,
```
sarahhaggarty:
```suggestion
For example, when you use `$$` before a string literal,
```

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
 
-Here, the `$$` prefix specifies that two consecutive dollar signs are required to trigger string interpolation. 
-Single dollar signs remain as literal characters.
+### Work with string content
 
-You can adjust how many dollar signs trigger interpolation.
-For example, using three consecutive dollar signs (`$$$`) allows `$` and `$$` to remain as literals 
-while enabling interpolation with `$$$`:
+If you want to change the content of a string, create a modified copy of it
+with functions like [`replace()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/replace.html), [`trim()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim.html),
```
sarahhaggarty: Nitpick. Can you put `trim()` first so that the order is reflected in the code sample?

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-Here, the `$$` prefix specifies that two consecutive dollar signs are required to trigger string interpolation. 
-Single dollar signs remain as literal characters.
+### Work with string content
 
-You can adjust how many dollar signs trigger interpolation.
-For example, using three consecutive dollar signs (`$$$`) allows `$` and `$$` to remain as literals 
-while enabling interpolation with `$$$`:
+If you want to change the content of a string, create a modified copy of it
```
sarahhaggarty, line 343: Can you double-check that these aren't extension functions? Otherwise they should have  `.` in the name.

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
+val toInt = "10".toInt() // 10
 
-In addition, you can assign the format string from a variable. This can be useful when the format string changes, 
-for example, in localization cases that depend on the user locale.
+// 1000000000000 exceeds maximum value of Int
+val toIntInvalid = "1000000000000".toIntOrNull()
 
-Be careful when using the `String.format()` function because it can be easy to mismatch the number or position of the 
-arguments with their corresponding placeholders.
+val toBoolean = "true".toBooleanStrict() //true
```
sarahhaggarty:
```suggestion
val toBoolean = "true".toBooleanStrict() // true
```

sarahhaggarty (changes requested):
Nice one! Just a few more suggestions from me and I'd like you to double-check something.

docs/topics/strings.md
```diff
@@ -1,278 +1,535 @@
...
+>
+{style="note"}
 
-The `String.format()` function provides similar functionality to string templates. However, the 
-`String.format()` function is more versatile because there are more formatting options available. 
+## String conversion
 
-In addition, you can assign the format string from a variable. This can be useful when the format string changes, 
-for example, in localization cases that depend on the user locale.
+Often you may use strings to represent values of other types, such as numbers, Boolean values, or identifiers from the input.
```
sarahhaggarty:
```suggestion
Often you may use strings to represent values of other types, such as numbers, `Boolean` values, or identifiers from the input.
```

sarahhaggarty (approved):
Nice work! LGTM! 🚀

PR #5491, opened 2026-04-15
Update tour welcome page

sarahhaggarty:
This PR updates the Kotlin tour welcome page to:
* Shorten the introduction
* Place the tour buttons side by side
* Move the contents into parallel lists with icons


docs/images/tour/start-intermediate-tour.svg
koshachy, line 1: Do we still need this file? 🤔

koshachy (approved):
Niiiice!
It looks that you forgot to delete one file (or did I miss something?).

docs/images/tour/start-intermediate-tour.svg
sarahhaggarty: We still need it for the last page of the beginner tour. But it doesn't need to be shortened anymore so I reverted this change. Thanks for pointing it out!

PR #5493, opened 2026-04-16
update: Boolean page

aepanchi:
Update https://kotlinlang.org/docs/booleans.html

### DIFF
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -76,7 +76,7 @@ You can use them to invert a `Boolean` value or combine multiple `Boolean` value
 
 The NOT operator inverts a `Boolean` value. 
 
-To use it, place the `!` operator before a `Boolean` value:
+To use NOT, place the `!` operator before a `Boolean` value:
 
 
 ```kotlin
@@ -88,7 +88,7 @@ val isOff = !isOn // isOff is false
 
 The AND operator returns `true` only if both operands are `true`. 
 
-To use it, place the `&&` operator between operands:
+To use logical AND, place the `&&` operator between operands:
 
 ```kotlin
 val a = false && false // false
@@ -106,7 +106,7 @@ val d = true && true  // true
 
 The OR operator returns `true` if at least one operand is `true`.
 
-To use it, place the `||` operator between operands:
+To use logical OR, place the `||` operator between operands:
 
 ```kotlin
 val a = false || false // false
@@ -122,9 +122,9 @@ val d = true || true   // true
 
 ### Exclusive OR (XOR)
 
-The XOR operation returns `true` if the operands have different values.
+The exclusive OR (XOR) operation returns `true` if the operands have different values.
 
-To use it, write `xor` between operands:
+To use XOR, write `xor` between operands:
 
 ```kotlin
 val a = false xor false // false
@@ -148,7 +148,7 @@ operations with lower precedence.
 For the `Boolean` operations described in this section, the precedence order is as follows:
 
 1. `!`
-2. `xor`
+2. `xor` (and other infix functions)
 3. `&&`
 4. `||`
 
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -115,7 +115,9 @@ val d = true xor true   // false
 ## Operator precedence
 
 Logical operators follow precedence rules. They determine how to group variables in
-the absence of parentheses:
+the absence of parentheses. 
+
+The subsequence of execution is: 
 
 1. `!`
 2. `xor`
--- a/docs/topics/booleans.md
+++ b/docs/topics/booleans.md
@@ -1,45 +1,198 @@
 [//]: # (title: Booleans)
+[//]: # (description: Learn how to use Boolean values in Kotlin, including declaration, logical operators, and conditions.)
 
-The type `Boolean` represents boolean objects that can have two values: `true` and `false`.
-`Boolean` has a [nullable](null-safety.md) counterpart declared as `Boolean?`.
+<show-structure depth="1"/>
 
-> On the JVM, booleans stored as the primitive `boolean` type typically use 8 bits.
+The [`Boolean`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/) type represents logical values.
+
+A `Boolean` has one of two values: `true` or `false`.
+
+Use `Boolean` values in comparisons, conditions, loops, and functions that answer yes-or-no questions.
+
+
+## Declare `Boolean`
+
+To declare a `Boolean`, assign it `true` or `false`.
+
+You can specify the `Boolean` type explicitly or let Kotlin infer it from the value:
+
+```kotlin
+val isTrue: Boolean = true
+val isFalse = false // Kotlin infers Boolean
+```
+
+If a value can be `null`, use `Boolean?`:
+
+```kotlin
+val isEnabled: Boolean? = null
+```
+
+> You cannot assign an integer value to a `Boolean` variable.
+> In Kotlin, `0` and `1` are not `Boolean` values.
 >
 {style="note"}
 
-Built-in operations on booleans include:
+## Produce `Boolean` values
 
-* `||` – disjunction (logical _OR_)
-* `&&` – conjunction (logical _AND_)
-* `!` – negation (logical _NOT_)
+You can use comparison expressions and functions to produce `Boolean` values:
 
-For example:
+```kotlin
+val number = 10
+val isPositive = number > 0 // true
+
+val language = "Kotlin"
+val isEmpty = language.isEmpty() // false
+```
+
+You can use the results in conditions and other expressions.
+
+>Learn more about `Boolean` functions in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/).
+>
+{style="tip"}
+
+## `Boolean` operators
+
+Use logical operators to combine or invert `Boolean` values.
+
+### Negation (NOT)
+
+Negation inverts a `Boolean` value.
+
+Use the `!` operator:
+
+```kotlin
+val isOn = true
+val isOff = !isOn // isOff is false
+```
+
+### Logical AND
+
+AND returns `true` only if both operands are `true`.
+
+Use the `&&` operator:
+
+```kotlin
+val a = false && false // false
+val b = false && true // false
+val c = true && false // false
+val d = true && true  // true
+```
+
+> If the first operand is `false`, the `&&` operator does not evaluate the second operand.
+>
+{style="tip"}
+
+### Logical OR
+
+OR returns `true` if at least one operand is `true`.
+
+Use the `||` operator:
+
+```kotlin
+val a = false || false // false
+val b = false || true  // true
+val c = true || false  // true
+val d = true || true   // true
+```
+
+> If the first operand is `true`, the `||` operator does not evaluate the second operand.
+>
+{style="tip"}
+
+### Exclusive OR (XOR)
+
+XOR returns `true` if the operands have different values.
+
+Use the `xor` infix function:
+
+```kotlin
+val a = false xor false // false
+val b = false xor true  // true
+val c = true xor false  // true
+val d = true xor true   // false
+```
+
+## Operator precedence
+
+Logical operators follow precedence rules. They determine how to group variables in
+the absence of parentheses:
+
+1. `!`
+2. `xor`
+3. `&&`
+4. `||`
+
+In the following example, the compiler evaluates `&&` before `||`:
 
 ```kotlin
 fun main() {
 //sampleStart
-    val myTrue: Boolean = true
-    val myFalse: Boolean = false
-    val boolNull: Boolean? = null
-
-    println(myTrue || myFalse)
-    // true
-    println(myTrue && myFalse)
-    // false
-    println(!myTrue)
-    // false
-    println(boolNull)
-    // null
-//sampleEnd
+    val result = true || false && false
+    println(result) // true
+//sampleEnd    
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-The `||` and `&&` operators work lazily, which means:
-
-* If the first operand is `true`, the `||` operator does not evaluate the second operand.
-* If the first operand is `false`, the `&&` operator does not evaluate the second operand.
-
-> On the JVM, nullable references to boolean objects are boxed in Java classes, just like with [numbers](numbers.md#boxing-and-caching-numbers-on-the-java-virtual-machine).
+> Use parentheses to make evaluation order explicit.
 >
-{style="note"}
+{style="tip"}
+
+## `Boolean` in conditions
+
+You can use `Boolean` expressions to control program flow with `if`, `when`, and loops.
+
+### `if` expressions
+
+```kotlin
+fun main() {
+//sampleStart
+    val number = 4
+    val isEven = number % 2 == 0
+    
+    if (isEven) {
+        println("The number is even.")
+    } else {
+        println("The number is odd.")
+    }
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+> Since the condition already has the `Boolean` type, you do not need to compare it to `true` or `false`.
+>
+{style="tip"}
+
+### `when` expressions
+
+```kotlin
+fun main() {
+//sampleStart
+    val number = 3
+
+    when {
+        number > 0 -> println("The number is positive.")
+        number < 0 -> println("The number is negative.")
+        else -> println("The number is zero.")
+    }
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### Loops
+
+```kotlin
+fun main() {
+//sampleStart
+    var isCalculating = true
+    
+    while (isCalculating) {
+        println("Calculating...")
+        isCalculating = false
+    }
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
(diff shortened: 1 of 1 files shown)


docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
-The type `Boolean` represents boolean objects that can have two values: `true` and `false`.
-`Boolean` has a [nullable](null-safety.md) counterpart declared as `Boolean?`.
+<show-structure depth="1"/>
 
-> On the JVM, booleans stored as the primitive `boolean` type typically use 8 bits.
+The [`Boolean`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/) type represents logical values.
+
+A `Boolean` has one of two values: `true` or `false`.
+
+Use `Boolean` values in comparisons, conditions, loops, and functions that answer yes-or-no questions.
```
e2e4b6b7: - Use `Boolean` in comparisons is kinda: `true > false`
- Use `Boolean` in loops is kinda for (...) { true }

To be fully correct, it probably should be "in functions that answer yes-or-no questions and conditions for while, if or when expressions"

But probably my concern is too formal and the current sentence gives required intuition

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+
+
+## Declare `Boolean`
+
+To declare a `Boolean`, assign it `true` or `false`.
```
e2e4b6b7: Probably in both cases "declare a `Boolean` variable" would be better. As "declare a `Boolean`" sounds like to declare a type itself

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
 
-* `||` – disjunction (logical _OR_)
-* `&&` – conjunction (logical _AND_)
-* `!` – negation (logical _NOT_)
+### Negation (NOT)
 
-For example:
+Negation inverts a `Boolean` value.
+
+Use the `!` operator:
```
e2e4b6b7: Sounds too unrelated to the previous sentence. Should it be something like "In Kotlin negation is (introduced/used/...) as the `!` operator" (same for the following operators)

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+
+> If the first operand is `true`, the `||` operator does not evaluate the second operand.
+>
+{style="tip"}
+
+### Exclusive OR (XOR)
+
+XOR returns `true` if the operands have different values.
+
+Use the `xor` infix function:
```
e2e4b6b7: The same as with operators, but I guess there should be also a bit of accent on the fact that there is no dedicated operator for XOR, but there is an infix function instead.

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+```kotlin
+val a = false xor false // false
+val b = false xor true  // true
+val c = true xor false  // true
+val d = true xor true   // false
+```
+
+## Operator precedence
+
+Logical operators follow precedence rules. They determine how to group variables in
```
e2e4b6b7: "They" - logical operators or precedence rules? (Probably in English there is a rule for proper binding of pronouns, but it is still not that clear)

I guess "variables" is a bit incorrect here, should be either operands, expressions.

Probably, it should be something like: 
```
In the presence of several operators in a single expression, evaluation order is determined by the precedence rules: Operators with the higher precedence evaluated before ones with lower precedence. The precedence for the mentioned `Boolean` operators is as follows:

1. `||`
2. `&&`
3. `xor` and other infix functions
4. `!`
```

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-The `||` and `&&` operators work lazily, which means:
+> Use parentheses to make evaluation order explicit.
+>
+{style="tip"}
+
+## `Boolean` in conditions
+
+You can use `Boolean` expressions to control program flow with `if`, `when`, and loops.
```
e2e4b6b7: "and loops" -> "and `while` loops" as `for` loops have nothing with booleans

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+val number = 10
+val isPositive = number > 0 // true
+
+val language = "Kotlin"
+val isEmpty = language.isEmpty() // false
+```
+
+You can use the results in conditions and other expressions.
+
+>Learn more about `Boolean` functions in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/).
```
e2e4b6b7: There are no producing functions in this reference, so probably this note should be in the end of section "### Exclusive OR (XOR)"

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+val c = true xor false  // true
+val d = true xor true   // false
+```
+
+## Operator precedence
+
+Logical operators follow precedence rules. They determine how to group variables in
+the absence of parentheses. 
+
+The subsequence of execution is:
```
e2e4b6b7: I do not understand this sentence at all...

docs/topics/booleans.md
```diff
@@ -1,45 +1,200 @@
...
+
+You can use the results in conditions and other expressions.
+
+>Learn more about `Boolean` functions in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/).
+>
+{style="tip"}
+
+## `Boolean` operators
+
+Use logical operators to combine or invert `Boolean` values.
```
e2e4b6b7: 1. "logical" seems incorrect here, programmer uses operators and infix functions, while they behave as logical operators
2. Here and in many other places, I do not fully agree with the "Use ..." form. Would "You can use ...", "`Boolean` values can be ...", or something similar fit better in this context?

e2e4b6b7 (approved):

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
 [//]: # (title: Booleans)
+[//]: # (description: Learn how to use Boolean values in Kotlin, including declaration, logical operators, and conditions.)
 
-The type `Boolean` represents boolean objects that can have two values: `true` and `false`.
-`Boolean` has a [nullable](null-safety.md) counterpart declared as `Boolean?`.
+<show-structure depth="1"/>
 
-> On the JVM, booleans stored as the primitive `boolean` type typically use 8 bits.
+The [`Boolean`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/) type represents logical values.
```
zamulla:
```suggestion
The [`Boolean`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/) type represents logical values (`true` and `false`).
```

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+A `Boolean` has one of two values: `true` or `false`.
```
zamulla: I'd wrap this into the previous sentence, also because this gets repeated in the next section.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+Use `Boolean` values in functions that answer yes-or-no questions, and in the
+`while`, `if`, and `when` conditions.
+
+
+## Declare a `Boolean` variable
+
+To declare a `Boolean` variable, assign it `true` or `false`.
```
zamulla, line 15: I understand the progression between this and the next section, but I don't know if it's efficient to go so granular.

The useful information about declaring is that you can make Boolean nullable, which we can add to the intro, otherwise this can all be in the "Produce Boolean values" section. I'd argue that inference is also "producing" a value, for example.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+```kotlin
+val number = 10
+val isPositive = number > 0 // true
+
+val language = "Kotlin"
+val isEmpty = language.isEmpty() // false
+```
+
+You can use the results in conditions and other expressions.
```
zamulla: An example here?

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+The NOT operator inverts a `Boolean` value. 
+
+To use it, place the `!` operator before a `Boolean` value:
 
-For example:
+
+```kotlin
+val isOn = true
+val isOff = !isOn // isOff is false
```
zamulla, line 84: You do use runnable samples, but we can try to make more of them runnable — to let people immediately see how even the simple code works, and also to track interest in particular samples.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+```
+
+### Logical AND
+
+The AND operator returns `true` only if both operands are `true`. 
+
+To use it, place the `&&` operator between operands:
```
zamulla: Here and elsewhere, it seems like we can make one paragraph out of these two sentences. Paragraph breaks take a lot of space on the page, so if it's all the same let's make fewer of them :)

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+```kotlin
+val a = false && false // false
+val b = false && true // false
+val c = true && false // false
+val d = true && true  // true
+```
+
+> If the first operand is `false`, the `&&` operator does not evaluate the second operand.
```
zamulla: A runnable sample here, too? Not sure I would have made this a tip — that's a pretty important thing to know about operators.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+To use it, write `xor` between operands:
+
+```kotlin
+val a = false xor false // false
+val b = false xor true  // true
+val c = true xor false  // true
+val d = true xor true   // false
+```
+> `xor` is an infix function, not an operator.
```
zamulla: Link? https://kotlinlang.org/docs/functions.html#infix-notation

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+The XOR operation returns `true` if the operands have different values.
+
+To use it, write `xor` between operands:
+
+```kotlin
+val a = false xor false // false
+val b = false xor true  // true
+val c = true xor false  // true
+val d = true xor true   // false
+```
```
zamulla, line 134: There should be an empty line after the code block.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+> `xor` is an infix function, not an operator. 
+> 
+> Learn more about `Boolean` functions in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/).
+>
+{style="note"} 
+
+## Operator precedence
+
+If an expression contains multiple logical operations and no parentheses to specify the evaluation order, 
+Kotlin applies precedence rules. That means, operations with higher precedence are evaluated before
```
zamulla:
```suggestion
Kotlin applies precedence rules. Operations with higher precedence are evaluated before 
```

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
-//sampleEnd
+    val result = true || false && false
+    println(result) // true
+//sampleEnd    
 }
 ```
 
-The `||` and `&&` operators work lazily, which means:
+> Use parentheses to make evaluation order explicit.
```
zamulla: Runnable sample? Maybe exactly like the previous one, only with parentheses?

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+>
+{style="tip"}
+
+## `Boolean` in conditions
+
+You can use `Boolean` expressions to control program flow with `if`, `when`, and `while`.
```
zamulla: This may read a bit ambiguous, as if Boolean expressions control the program flow.
Maybe something like "`if`, `when` and `while` evaluate Boolean expressions to direct program flow".

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+        println("The number is even.")
+    } else {
+        println("The number is odd.")
+    }
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+> Since the condition already has the `Boolean` type, you do not need to compare it to `true` or `false`.
```
zamulla: Not sure if this should be an admonition at all, a simple comment.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+```kotlin
+val a = false || false // false
+val b = false || true  // true
+val c = true || false  // true
+val d = true || true   // true
+```
+
+> If the first operand is `true`, the `||` operator does not evaluate the second operand.
+>
+{style="tip"}
```
zamulla: Again, not sure if it's a tip — misunderstanding evaluation flow is a common beginner mistake.

Also, should we mention `or()` to highlight that you can make Kotlin evaluate both? Same for `and()`.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+
+You can use the results in conditions and other expressions.
+
+## `Boolean` operations
 
-* `||` – disjunction (logical _OR_)
-* `&&` – conjunction (logical _AND_)
-* `!` – negation (logical _NOT_)
+Kotlin provides operators and infix functions for working with `Boolean` values. 
+You can use them to invert a `Boolean` value or combine multiple `Boolean` values into a single result.
```
zamulla, line 73: This might be a nitpick, but: "You can..." can often be omitted in favor of a simple imperative: "Use them to invert a Boolean value..." is more to the point. But I don't think there's a hard rule here.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+> Since the condition already has the `Boolean` type, you do not need to compare it to `true` or `false`.
 >
-{style="note"}
+{style="tip"}
+
+### `when` expressions
+
```
zamulla, line 207: Should we link the docs for `when`? https://kotlinlang.org/docs/control-flow.html#when-expressions-and-statements
At least in ToC it comes after Types.

docs/topics/booleans.md
```diff
@@ -1,45 +1,205 @@
...
+        number < 0 -> println("The number is negative.")
+        else -> println("The number is zero.")
+    }
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+### `while` loops
+
```
zamulla, line 224: Similarly, should we have a link? https://kotlinlang.org/docs/control-flow.html#while-loops

zamulla (commented):
Thanks! Left some suggestions on clarity and Adding More Stuff.

docs/topics/booleans.md
aepanchi: I think this section works well for structuring the content on the page, so it's not just one big block of text. Plus it aligns this one with other pages of the types section.

docs/topics/booleans.md
aepanchi: added

docs/topics/booleans.md
aepanchi: Made applicable samples runnable, but I don't think this one should be it. And I believe we may not be particularly interested in the usage data of such samples and use heatmap instead?

docs/topics/booleans.md
aepanchi: Agree, but these two sentences have different purposes. So I would keep it as is for a quicker distinguishing between what and how. Plus, I'm not sure someone would read the whole page :D 
We can track the scroll page rate and update the page in the future if needed.

docs/topics/booleans.md
aepanchi: I'm not sure how to showcase it with the example tbh. 
Good point about the tip, though! Do you think it should be a note or just a plain text?

docs/topics/booleans.md
aepanchi: thanks, added

docs/topics/booleans.md
aepanchi: added

docs/topics/booleans.md
aepanchi: fixed

docs/topics/booleans.md
aepanchi: done

docs/topics/booleans.md
aepanchi: Yes, makes sense to add it, thanks!

docs/topics/booleans.md
aepanchi: Agree, but with that particular case, I feel that omitting "you can" may lead to the conclusion that it is all that the functions do,

zamulla: Yeah, that's exactly the impression that current wording makes for me, so I guess it's in the eye of the beholder.

docs/topics/booleans.md
zamulla: Heatmaps are less reliable than GA click events, hence the focus on explicit interactions.

Generally I try to follow the "don't trust me, run the code for yourself" logic: if there's anything that may not be evident from the code, it's probably a good idea to be able to run it. But for this sample in particular, I don't have strong feelings either way.

docs/topics/booleans.md
zamulla: Here I specifically noticed using "it" to refer to the previous sentence, which is in another paragraph and makes the structure awkward to read — I still need to read both of them to get the gist. If you're sure these are separate paragraphs, this sentence should be "To apply logical AND..." or something like that.

As for the first point, not a lot of people are going to read the whole thing, but we can make it so everyone scrolls less :)

zamulla (approved):
Thanks! Only nitpicks remain, so I'm putting the green check here 👍

PR #5503, opened 2026-04-22
upd: Numbers page

aepanchi:
Updates https://kotlinlang.org/docs/numbers.html


docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
-val one = 1 // Int
-val threeBillion = 3000000000 // Long
-val oneLong = 1L // Long
-val oneByte: Byte = 1
-```
+Kotlin supports the following literal forms for integer values:
 
-## Floating-point types
+* Decimals: `123`
+* `Long`, with the capital `L` suffix: `123L`
```
e2e4b6b7: I guess this line should not be here. Suffixes are orthogonal to forms and mentioned later

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
-* Decimals: `123`
-* Longs, ending with the capital `L`: `123L`
-* Hexadecimals: `0x0F`
-* Binaries: `0b00001011`
+**Declare floating-point values**
 
-> Octal literals are not supported in Kotlin.
->
-{style="note"}
+To declare a floating-point literal, include a decimal point (`.`) or use an exponent notation.
```
e2e4b6b7: I guess it would be useful to demonstrate an exponent notation

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+For example, the following code converts an `Int` value to `Double`:
 
 ```kotlin
 fun main() {
 //sampleStart
-    val x = 5 / 2
-    println(x == 2.5) 
-    // Operator '==' cannot be applied to 'Int' and 'Double'
+    val intValue: Int = 1
+    val doubleValue= intValue.toDouble()
```
e2e4b6b7:
```suggestion
    val doubleValue = intValue.toDouble()
```

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+{kotlin-runnable="true" validate="false"}
 
-You can override these operators in custom number classes.
-See [Operator overloading](operator-overloading.md) for details.
+All number types support conversions to other numeric types.
+To convert a number to another type, use an explicit conversion function:
 
-### Division of integers
+* `toByte()` (deprecated for [Float](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-float/to-byte.html) and [Double](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-double/to-byte.html))
+* `toShort()` (deprecated for [Float](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-float/to-short.html) and [Double](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-double/to-short.html))
```
e2e4b6b7: Probably, it is an unnecessary overcomplication. They will be able to figure that out in that super rare scenario where they need it and the fix is described in the deprecation note.

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3" validate="false"}
+
+### Mixed numeric expressions
 
-To return a division result with the fractional part, explicitly convert one of the arguments to a floating-point type:
+Kotlin does not support implicit conversion for assignments or function arguments. 
+However, you can combine different numeric types in arithmetic expressions. In such cases, 
+Kotlin infers a result type based on the operand types,
```
e2e4b6b7: Probably it would be too detailed to describe rules, but could we provide a link to somewhere with it? (considering there is a reference to this section "The result type depends on the types of the operands. Learn more in [](#mixed-numeric-expressions).", it seems it should be either explained or referenced to somewhere else)

Also it is not actually an inference, so the following might be better: "Kotlin determines a result type ..."

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+val intNumber: Int = 1
+val longNumber: Long = 1000
+val result: Int = intNumber + longNumber 
+// Error: Initializer type mismatch
+```
+
+> This behavior applies to expressions only. It does not replace explicit conversions
+> when a specific target type is required.
+>
+{style="note"}
```
e2e4b6b7: I do not really understand what does this mean... Probably it should be something like: 
```
This behavior applies to arithmetic expressions only. As was already mentioned, Kotlin does not support implicit conversion for assignments or function arguments.
```

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+    println(x and 0x000FF000) // 0
 //sampleEnd
 }
 ```
 
-The complete list of bitwise operations:
+Bitwise operations include:
 
 * `shl(bits)` – signed shift left
```
e2e4b6b7: What is `bits`? It is not even the name of the argument in the declaration. Probably it is better to have just `shl` or use that actual name of the argument (`shl(bitCount)` for this case). The same for other operations.

docs/topics/numbers.md
```diff
@@ -330,49 +348,89 @@ The complete list of bitwise operations:
...
-### Floating-point numbers comparison
+## Floating-point number comparison
 
-The operations on floating-point numbers discussed in this section are:
+In Kotlin, floating-point comparison depends on the static type of the operands.
 
-* Equality checks: `a == b` and `a != b`
-* Comparison operators: `a < b`, `a > b`, `a <= b`, `a >= b`
-* Range instantiation and range checks: `a..b`, `x in a..b`, `x !in a..b`
+When the operands are statically known to be `Float` or `Double` (including their nullable counterparts),
```
e2e4b6b7: "nullable counterparts"? AFAIU, it is not possible to make arithmetic over nullable versions. So probably it is better to just remove the text in parenthesis

docs/topics/numbers.md
```diff
@@ -330,49 +348,89 @@ The complete list of bitwise operations:
...
     println(listOf(Double.NaN) == listOf(Double.NaN)) // true
 
     // Operand statically typed as floating-point number
-    println(0.0 == -0.0)                              // true
+    println(0.0 == -0.0) // true
     
     // Operand NOT statically typed as floating-point number
-    // So -0.0 is less than 0.0
-    println(listOf(0.0) == listOf(-0.0))              // false
+    println(listOf(0.0) == listOf(-0.0)) // false
```
e2e4b6b7: Not sure if this example is easy to understand. It assumes that reader already understand how generics in Kotlin works (about erasure). Would not it better to demonstrate it with the following:
```
fun generalizedEquals(a: Any, b: Any) {
    return a == b
}

fun main() {
    ...
    println(generalizedEquals(Double.NaN, Double.NaN))
    ...
}
```

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 ```
-{validate="false"}
 
-To explicitly specify the `Float` type for a value, add the suffix `f` or `F`.
-If a value provided in this way contains more than 7 decimal digits, it is rounded:
+> When you declare a numeric type explicitly, the compiler checks that the value
+> fits in the range of that type.
+>
+> When you do not specify a numeric type, Kotlin infers `Int` if the 
+> value fits in the `Int` range. Otherwise, Kotlin infers `Long`.
```
e2e4b6b7: The structure of this part seem a bit off. Firstly it mentions "let Kotlin infer it", but explains how it is inferred only at the end. And "To declare X value, specify the type explicitly" is repeated separately for Long and Short/Byte.

Also it does not mention about inference for the following case:
```
fun foo(b: Byte)

foo(42) <- inferred to Byte, not Int
```

Would not it be better to make it as follows:
- To declare values of any numeric type, you can specify the type explicitly
- Specifically for Long, there is a suffix
- When you do not specify a numeric type, Kotlin trying to infer type from the context (example with `foo(b: Byte)`)
- If there is no information in the context, Kotlin infers `Int` if the value fits in the `Int` range. Otherwise, Kotlin infers `Long`.

The main idea is that it might be useful to mention this example of inference. On the other hand, it might be overcomplication for this chapter, so I would leave it on you to decide.

docs/topics/numbers.md
```diff
@@ -12,315 +39,306 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 ```kotlin
-val oneMillion = 1_000_000
-val creditCardNumber = 1234_5678_9012_3456L
-val socialSecurityNumber = 999_99_9999L
-val hexBytes = 0xFF_EC_DE_5E
-val bytes = 0b11010010_01101001_10010100_10010010
-val bigFractional = 1_234_567.7182818284
+val eFloat = 2.7182818284f // Float   
+// Actual value is 2.7182817
 ```
```
e2e4b6b7, line 131: Probably it would be better with the following structure: 
- To declare a floating-point literal, include a decimal point (`.`) or use an exponent notation. 
- Example (with example for exponent notation)
- By default, Kotlin infers floating-point literals as `Double` + To declare a `Float`, add the `f` or `F` suffix
- Example

e2e4b6b7 (changes requested):

docs/topics/numbers.md
aepanchi: removed, thanks!

docs/topics/numbers.md
aepanchi: Thanks, I agree with most of your points and will address them. However, yes, I don't believe the inference example is helpful here — it can overcomplicate things. So I would suggest focusing on the declaration part only and just mention inference as a note (what we have rn)

docs/topics/numbers.md
aepanchi: Makes sense, thanks! Removed.

docs/topics/numbers.md
aepanchi: tbh, I don't see it being documented anywhere (or close enough to reference from this page).

docs/topics/numbers.md
aepanchi: rewrote

docs/topics/numbers.md
aepanchi: thanks, fixed.

docs/topics/numbers.md
aepanchi: makes sense, updated.

docs/topics/numbers.md
aepanchi: Thanks! I decided to drop the part in () completely to avoid confusion. I have a small question though: should we make a note that inv() is used like `a.inv()`, not as infix like other operations?

docs/topics/numbers.md
e2e4b6b7: Yes, probably it would be too detailed to explain it here, so it is fine

docs/topics/numbers.md
e2e4b6b7: > should we make a note that inv() is used like a.inv(), not as infix like other operations

IMO, it is not necessary, as intuition about existence of it is enough. It is not that complicated to figure out the form, considering different signature.

On the other hand, what do you think about the following presentation:
- `a shl b` - ...
- `a shl b` - ...
- `a.inv()` - ...

It is simultaneously a demonstration of form, and avoids misleading `shl()` which might confuse reader about why there are no arguments

e2e4b6b7 (approved):

docs/topics/numbers.md
```diff
@@ -1,9 +1,36 @@
 [//]: # (title: Numbers)
+[//]: # (description: Learn how to use numbers in Kotlin, including numeric types, literals, conversions, arithmetic operations, overflow, and JVM-specific behavior.)
+
+
```
zamulla: Extra empty line :)

docs/topics/numbers.md
```diff
@@ -1,9 +1,36 @@
...
+The Kotlin number types represent:
+* Integer values ([Byte](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-byte/),
+  [Short](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-short/),
+  [Int](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-int/),
+  and [Long](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-long/))
+* Floating-point values ([Float](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-float/)
+  and [Double](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-double/))
+
+Use numbers to store and process numeric data, for example, in arithmetic, counters, measurements,
```
zamulla: Probably "number types", not "numbers"? I would be amused if someone started to explain to me what numbers are for :)

docs/topics/numbers.md
```diff
@@ -1,9 +1,36 @@
...
+and other calculations.
+
+## Choose a number type
```
zamulla, line 15: The paragraph above seems to me to go well with the list below, so I'm not sure a separate section is the best approach here. Also it can probably be cut down to something like:

> Use number types to store <...> other calculations. For example:
> * Use `Int` for whole numbers
> * ...

docs/topics/numbers.md
```diff
@@ -1,9 +1,36 @@
...
+In most cases, you can refer to the following rules to determine the
+correct number type for your task:
+
+* Use `Int` for whole numbers.
+* Use `Long` for whole numbers outside the `Int` range.
+* Use `Double` for decimal numbers.
+* Use `Float` when lower precision is acceptable or required.
+* Use `Byte` and `Short` when an API or data format requires them.
+
+> Kotlin also provides [](unsigned-integer-types.md).
```
zamulla: Since we just explained what `Int` is for, probably worth it to say a couple of words how unsigned `Int` is different, if only for consistency.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
-> See [](unsigned-integer-types.md).
-> 
-{style="tip"}
 
-When you initialize a variable with no explicit type specification, the compiler automatically infers the type with the 
-smallest range enough to represent the value starting from `Int`. If it doesn't exceed the range of `Int`, the type is `Int`.
-If it does exceed that range, the type is `Long`. To specify the `Long` value explicitly, append the suffix `L` to the value.
-To use the `Byte` or `Short` type, specify it explicitly in the declaration. 
-Explicit type specification triggers the compiler to check that the value doesn't exceed the range of the specified type.
+**Declare integer values**
```
zamulla: Not sure why this header is in bold instead of ###

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+val oneLong = 1L
+```
+
+> When you declare a numeric type explicitly, the compiler checks that the value
+> fits in the range of that type.
+>
+> When you do not specify a numeric type, Kotlin infers `Int` if the
+> value fits in the `Int` range. Otherwise, Kotlin infers `Long`.
+>
+{style="note"}
```
zamulla: Is that a sidenote? To me this is a pretty straightforward explanation of what's happening when I declare stuff.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 | Type	    | Size (bits) | Significant bits | Exponent bits | Decimal digits |
 |----------|-------------|------------------|---------------|----------------|
 | `Float`	 | 32          | 24               | 8             | 6-7            |
 | `Double` | 64          | 53               | 11            | 15-16          |    
 
-You can initialize `Double` and `Float` variables only with numbers that have a fractional part.
-Separate the fractional part from the integer part by a period (`.`)
 
-For variables initialized with fractional numbers, the compiler infers the `Double` type:
+**Declare floating-point values**
```
zamulla: Again, should be a proper heading IMO.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+
+To declare a floating-point literal, include a decimal point (`.`) or use an exponent notation:
 
 ```kotlin
-val pi = 3.14          // Double
+val pi = 3.14
+val avogadro = 6.02214076e23
```
zamulla:
```suggestion
val avogadro = 6.02214076e23
```

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+    printDouble(xFloat) // Error: argument type mismatch
 //sampleEnd
 }
 ```
+{kotlin-runnable="true" validate="false"}
 
-You can override these operators in custom number classes.
-See [Operator overloading](operator-overloading.md) for details.
+All number types support conversions to other numeric types.
```
zamulla: The question I immediately have is whether number types are the same as numeric types — when referring to the same concept, I'd use the same term.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+To convert a number to another type, use an explicit conversion function:
 
-### Division of integers
+* `toByte()`
+* `toShort()`
+* `toInt()`
+* `toLong()`
+* `toFloat()`
+* `toDouble()`
```
zamulla, line 235: I know that the list has been here already, but I can't decide whether I like it. It takes a lot of space on the page and communicates basically the same thing as the sentence before it. Maybe we should put these into parentheses at the end of the sentence?

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+```kotlin
+val intNumber: Int = 1
+val longNumber: Long = 1000
+val result: Int = intNumber + longNumber 
+// Error: Initializer type mismatch
+```
+
+> This behavior applies to arithmetic expressions only.
+>
+{style="note"}
```
zamulla: I am not sure what this note is about, what else this behavior can apply to. Maybe this part would make more sense — and not require a note —inside the "Arithmetic operations﻿" section? Since the whole "Mixed numeric expressions" part seems to be about arithmetic expressions.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 
-Kotlin provides a set of _bitwise operations_ on integer numbers. They operate on the binary level directly with
-bits of the numbers' representation.
-Bitwise operations are represented by functions that can be called in infix form. They can be applied only to `Int` and `Long`:
+
+## Bitwise operations
+
+Kotlin provides _bitwise operations_ for `Int` and `Long`. These operations are represented by functions that can
+be called in [infix form](functions.md#infix-notation):
```
zamulla: I would suggest simply going "...by a set of [infix functions]():" because "infix form" is not actually defined in functions.md.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+  println(Double.NaN == Double.NaN) // false
+  println(0.0 == -0.0) // true
+
+  // Operands used through a non-floating-point static type
+  println(generalizedEquals(Double.NaN, Double.NaN)) // true
+  println(generalizedEquals(0.0, -0.0)) // false
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-numbers-floating-comp"}
```
zamulla, line 428: The sample uses two-space indent instead of four.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
-    val l = 1L + 3       // Long + Int => Long
-    println(l is Long)   // true
-//sampleEnd
+    println(Double.MAX_VALUE * 2) // Infinity
+//sampleEnd    
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.5"}
-
-### Reasoning against implicit conversions
```
zamulla, line 221: I would argue that this section, while more technical, is small enough to be left inside the "Type conversion" section for those wondering why implicit conversions are not supported. It's not like we have a more in-depth explanation of this elsewhere — probably because here's where it usually matters.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 ```
-{validate="false"}
 
-To explicitly specify the `Float` type for a value, add the suffix `f` or `F`.
-If a value provided in this way contains more than 7 decimal digits, it is rounded:
+> Kotlin rounds a `Float` literal that contains more precision than `Float` can store.
+>
+{style="note"}
+
+If a value can be absent, use nullable types:
```
zamulla, line 137: I don't know whether this is useful for numbers specifically. I think this is one of the things Kotlin beginners bump against on every step, so talking about this explicitly may not be too useful.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 | `Float`	 | 32          | 24               | 8             | 6-7            |
 | `Double` | 64          | 53               | 11            | 15-16          |    
 
-You can initialize `Double` and `Float` variables only with numbers that have a fractional part.
-Separate the fractional part from the integer part by a period (`.`)
 
-For variables initialized with fractional numbers, the compiler infers the `Double` type:
+**Declare floating-point values**
+
+To declare a floating-point literal, include a decimal point (`.`) or use an exponent notation:
```
zamulla:
```suggestion
To declare a floating-point literal, include a decimal point (`.`) or use the exponential notation:
```

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+{style="note"}
+
+To declare a numeric value, specify the type explicitly: 
 
 ```kotlin
-val one = 1 // Int
-val threeBillion = 3000000000 // Long
-val oneLong = 1L // Long
+val one: Int = 1
+val oneBillion: Long = 1_000_000_000 // Use underscores to improve readability
```
zamulla: I'd add the examples of non-decimals with underscores here, they are even less apparent or intuitive.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+        print(x) 
+    }
+
+    val x = 1.0
+    val xInt = 1    
+    val xFloat = 1.0f
+
+    printDouble(x) // OK
+    printDouble(xInt) // Error: argument type mismatch
+    printDouble(xFloat) // Error: argument type mismatch
```
zamulla, line 221: Maybe also tuck in the `val one: Double = 1` example here — it's also strictly speaking a type mismatch, but may be not apparent.

docs/topics/numbers.md
```diff
@@ -12,367 +39,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
+
+```kotlin
+fun main() {
+//sampleStart
+    val score: Int = 10000
+    val savedScore: Int? = score
+    val displayedScore: Int? = score
+
+    println(savedScore == displayedScore) // true
+    println(savedScore === displayedScore) // false
```
zamulla, line 467: That's nitpicky, but I'd put the bad example before the good one. It emulates making a mistake: the fact that === should not be used is a "surprise", and the correct solution is then revealed. When we show the correct one first and then show a mistake, the mistake doesn't feel relevant.

zamulla (commented):
Thanks, it does read more streamlined, I think!

There are a couple of pieces you cut I'd suggest keeping, and a question about whether "Mixed numeric expressions" should be where it is, but the rest is polishing suggestions.

docs/topics/numbers.md
aepanchi: I want to keep an intro as short and crisp as possible :) Plus I believe it is better for the SEO

docs/topics/numbers.md
aepanchi: Wouldn't it overwhelm a reader? People who need such a page probably will need some time before needing such types. And for people experienced enough to need it, this note should be okay.

docs/topics/numbers.md
aepanchi: I didn't want it to be included in the TOC, but could not use <structure depth> since other titles should be displayed

docs/topics/numbers.md
aepanchi: I wanted to divide this theoretical information from the applicable one (how to declare). If I keep it as plain text, I feel like it could be lost and ignored due to the surrounding code blocks.

docs/topics/numbers.md
aepanchi: Good catch, thanks!

docs/topics/numbers.md
aepanchi: Agree that it takes too much space, but I feel like it acts as a divider between why and how, making it easier to find the how part if you don't care/already know about the why.

docs/topics/numbers.md
aepanchi: I would like to keep it for consistency with other Types pages, tbh

docs/topics/numbers.md
aepanchi: I think **exponential notation** sounds less idiomatic? But now I also feel like the article should be omitted here

docs/topics/numbers.md
aepanchi: done

docs/topics/numbers.md
aepanchi: I don't want to make this example too big. WDYT about replacing float with a double?

docs/topics/numbers.md
aepanchi: That's a good point! Thanks

docs/topics/numbers.md
aepanchi: good catch, thanks:)

docs/topics/numbers.md
aepanchi: Rewrote, but clearly stated `inv()` since it is not an infix function.

docs/topics/numbers.md
aepanchi: Yes, I feel like that note can be confusing. Thanks!

docs/topics/numbers.md
aepanchi: I tired to focus on why we prefer explicit conversion over implicit one. A "glass half full" kinda logic :D 
And find the example in this section to be tricky for novice users. tbh

docs/topics/numbers.md
zamulla: The point is in the fact that cross-type assignment _also_ fails, which is not obvious; the fact that the function call fails we explicitly mention before the sample. I wouldn't say that a single line is too big of a price for that.

docs/topics/numbers.md
zamulla: Why explicit is good / why implicit is bad is kind of sides of the same coin, but I think we should describe both sides to show the coin as a whole. We can even hide it under the cut to not push it on people who just want to get through the basics.

My point is also that this page is not solely for novice users — experienced developers who are looking for "why numbers in Kotlin work the way they work" also deserve some attention.

That said, this conversation may be indeed too academic to remain in the docs at all, so it's your call.

docs/topics/numbers.md
zamulla: I would like to remove it from all Types pages :) but I don't have a strong argument, let them bump some more.

docs/topics/numbers.md
zamulla: An admonition should be formatted as such if it's actually a break from the rest of the text. Here "...specify the type explicitly: <...> When you declare a numeric type explicitly" is a clear sequence for me, that's why it looks strange that it's visually broken up. Admonition is metainformation; using it for visual effect tends to lead to warning fatigue, when people just start skipping everything that's not the main body.

To help it not get lost we can add a tailored example here. If it would get lost, it's because we have a clear "theory → immediate example" pattern on the page, and here it's not working.

And, last but not least, I wouldn't say it's purely theoretical :) one needs to know how inference works to handle it properly, so to me this is straightforwardly useful.

docs/topics/numbers.md
zamulla: As far as I remember, unsigned ints are separate primarily because they are in Beta still, not because there's something overwhelming about them. They are unsigned, big deal :)

The previous version was also vague, maybe we should address this head on: "Kotlin also provides unsigned integer types, but they are still considered a Beta feature." — that definitely would deter those, who shouldn't waste time on digging into this.

docs/topics/numbers.md
zamulla: Why not though? I think you have a good structure, but right now it's obscured — I very much enjoy that the boring table and actual declaration examples are clearly separate. Showing this in the ToC simply supports this.

More to the point, bold type has its own place — it's emphasis, on UI elements or parts of a sentence. Using it as a header muddies the users' general perception of our markup.

docs/topics/numbers.md
```diff
@@ -12,367 +38,381 @@ For integer numbers, there are four types with different sizes and value ranges:
...
 //sampleEnd
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.5"}
-
-### Reasoning against implicit conversions
 
-Kotlin doesn't support implicit conversions because they can lead to unexpected behavior.
+### Data overflow
```
zamulla: I just thought that data overflow can also occur with negation and conversion, not only arithmetic. Should this be its own subject lower on the page? In the "Type conversion" section we only show integer to floating point, but converting between integers also has a side effect.

zamulla (commented):
Thanks!

There are a couple of threads where I still would like to get on the same page, although none are too important.

docs/topics/numbers.md
aepanchi: Good point, you are right! Updated the tip as suggested.

docs/topics/numbers.md
aepanchi: Fair enough, fixed :)

docs/topics/numbers.md
aepanchi: Fixed, added examples as well.

docs/topics/numbers.md
aepanchi: Yeah, I feel like it can be turned into a good blog post we can reference from here

docs/topics/numbers.md
aepanchi: added

docs/topics/numbers.md
aepanchi: rewrote

zamulla (approved):
🙌

PR #5508, opened 2026-04-28
Annotations in metadata enabled by default in Kotlin 2.4.0

sarahhaggarty:
This PR updates the docs to explain that annotations in metadata are enabled by default from Kotlin 2.4.0.

### DIFF
--- a/docs/topics/metadata-jvm.md
+++ b/docs/topics/metadata-jvm.md
@@ -177,32 +177,19 @@ fun main() {
 ```
 
 ### Write and read annotations in metadata
-<primary-label ref="experimental-general"/>
 
-You can store annotations in Kotlin metadata and access them using the `kotlin-metadata-jvm` library.
-This removes the need to match annotations by signature, making access more reliable for overloaded declarations.
+Kotlin stores annotations in both bytecode and Kotlin metadata. If you use the `kotlin-metadata-jvm` library to read or
+write annotations, you work with their metadata representation.
 
-To make annotations available in the metadata of your compiled files, add the following compiler option:
+> Kotlin stores annotations in Kotlin metadata starting with Kotlin 2.4.0. If you inspect class files compiled with earlier versions,
+> the annotations aren't present in the metadata.
+>
+{style="note"}
 
-```kotlin
--Xannotations-in-metadata
-```
+When you change the annotations in metadata, make sure they stay consistent with the annotations stored in bytecode.
+If they get out of sync, tools that rely on reflection or bytecode analysis may report different results than tools that read Kotlin metadata.
 
-Alternatively, add it to the `compilerOptions {}` block of your Gradle build file:
-
-```kotlin
-// build.gradle.kts
-kotlin {
-    compilerOptions {
-        freeCompilerArgs.add("-Xannotations-in-metadata")
-    }
-}
-```
-
-When you enable this option, the Kotlin compiler writes annotations into metadata alongside the JVM bytecode,
-making them accessible to the `kotlin-metadata-jvm` library.
-
-The library provides the following APIs for accessing annotations:
+The `kotlin-metadata-jvm` library provides the following APIs for accessing annotations:
 
 * `KmClass.annotations`
 * `KmFunction.annotations`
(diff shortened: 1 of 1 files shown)


udalov (changes requested):
To someone not familiar with the history or the reasons why we stored annotations in bytecode only, and then started to store them in metadata as well, this page might now read a bit strange. For example, the part about "This removes the need ...". Also, the part about this API being experimental, which it isn't anymore.

I think this section should be rewritten, and the main part should be about the problem of keeping the annotations consistent between bytecode and metadata if you use kotlin-metadata-jvm to modify annotations. And we should leave the note somewhere, that annotations are stored in metadata only since Kotlin 2.4.0.

sarahhaggarty:
> To someone not familiar with the history or the reasons why we stored annotations in bytecode only, and then started to store them in metadata as well, this page might now read a bit strange. For example, the part about "This removes the need ...". Also, the part about this API being experimental, which it isn't anymore.
> 
> I think this section should be rewritten, and the main part should be about the problem of keeping the annotations consistent between bytecode and metadata if you use kotlin-metadata-jvm to modify annotations. And we should leave the note somewhere, that annotations are stored in metadata only since Kotlin 2.4.0.

Thanks @udalov! I rewrote the section as you suggested. What do you think? 
Our docs are written to reflect the latest Stable version of Kotlin, which will be 2.4.0 when this is published. So I won't add this detail here. It will be present in the What's new for Kotlin 2.4.0 though.

docs/topics/metadata-jvm.md
udalov: Let's remove this, they're no longer experimental

udalov:
> Our docs are written to reflect the latest Stable version of Kotlin, which will be 2.4.0 when this is published. So I won't add this detail here. It will be present in the What's new for Kotlin 2.4.0 though.

You might use kotlin-metadata-jvm 2.4.0 to inspect class files compiled by Kotlin version earlier than 2.4.0, which lacked annotations in metadata. We have to support those class files indefinitely, so IMHO it's a pretty significant detail for users of kotlin-metadata-jvm.

udalov (changes requested):

sarahhaggarty:
> > Our docs are written to reflect the latest Stable version of Kotlin, which will be 2.4.0 when this is published. So I won't add this detail here. It will be present in the What's new for Kotlin 2.4.0 though.
> 
> You might use kotlin-metadata-jvm 2.4.0 to inspect class files compiled by Kotlin version earlier than 2.4.0, which lacked annotations in metadata. We have to support those class files indefinitely, so IMHO it's a pretty significant detail for users of kotlin-metadata-jvm.

I understand. I've made an exception and added a note to the text. Please let me know what you think!

udalov (approved):

docs/topics/metadata-jvm.md
```diff
@@ -177,32 +177,19 @@ fun main() {
 ```
 
 ### Write and read annotations in metadata
-<primary-label ref="experimental-general"/>
 
-You can store annotations in Kotlin metadata and access them using the `kotlin-metadata-jvm` library.
-This removes the need to match annotations by signature, making access more reliable for overloaded declarations.
+Kotlin stores annotations in both bytecode and Kotlin metadata. If you use the `kotlin-metadata-jvm` library to read or
+write annotations, you work with the metadata representation.
```
daniCsorbaJB:
```suggestion
write annotations, you work with their metadata representation.
```

docs/topics/metadata-jvm.md
```diff
@@ -177,32 +177,19 @@ fun main() {
...
-```kotlin
-// build.gradle.kts
-kotlin {
-    compilerOptions {
-        freeCompilerArgs.add("-Xannotations-in-metadata")
-    }
-}
-```
+> Kotlin stores annotations in Kotlin metadata starting with Kotlin 2.4.0. If you inspect class files compiled with earlier versions,
+> the annotations aren't present.
```
daniCsorbaJB:
```suggestion
> the annotations aren't present in the metadata.
```

daniCsorbaJB: not sure if needed just for clarity perhaps? WDYT? 🤔

docs/topics/metadata-jvm.md
```diff
@@ -177,32 +177,19 @@ fun main() {
...
+>
+{style="note"}
 
-When you enable this option, the Kotlin compiler writes annotations into metadata alongside the JVM bytecode,
-making them accessible to the `kotlin-metadata-jvm` library.
+When you change the annotations in metadata, make sure they stay consistent with the annotations stored in bytecode.
+If they get out of sync, tools that rely on reflection or bytecode analysis may see different results than tools that read Kotlin metadata.
```
daniCsorbaJB:
```suggestion
If they get out of sync, tools that rely on reflection or bytecode analysis may report different results than tools that read Kotlin metadata.
```

daniCsorbaJB: see vs report not sure which is more accurate so whichever you think fits more

daniCsorbaJB (approved):
Thank you for the update — looks good to me, I only left a couple of suggestions 🚀

PR #5510, opened 2026-04-28
Binary compatibility validation updates for Kotlin 2.4.0

sarahhaggarty:
This PR updates some DSL for binary compatibility as well as adds a feature for `maven-publish` plugin artifact inclusion that's coming in Kotlin 2.4.0.

### DIFF
--- a/docs/topics/gradle/gradle-binary-compatibility-validation.md
+++ b/docs/topics/gradle/gradle-binary-compatibility-validation.md
@@ -11,13 +11,14 @@ users and encouraging continued adoption of the library.
 > 
 {style="tip"}
 
-The Kotlin Gradle plugin includes support for binary compatibility validation. When enabled, it generates
+The Kotlin Gradle plugin includes support for binary compatibility validation. The plugin generates
 Application Binary Interface (ABI) dumps from the current code and compares them with previous dumps to highlight differences.
 You can review these changes to find any potentially binary-incompatible modifications and take action to address them.
 
 ## How to enable
 
-To enable binary compatibility validation, add the following to the `kotlin{}` block in your `build.gradle.kts` file:
+To enable binary compatibility validation, in your `build.gradle.kts` file, add an `abiValidation {}` block. If you have
+no custom configuration, you can use the `abiValidation()` function instead:
 
 <tabs group="build-script">
 <tab title="Kotlin" group-key="kotlin">
@@ -25,10 +26,7 @@ To enable binary compatibility validation, add the following to the `kotlin{}` b
 ```kotlin
 kotlin {
     @OptIn(org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation::class)
-    abiValidation {
-        // Use the set() function to ensure compatibility with older Gradle versions
-        enabled.set(true)
-    }
+    abiValidation()
 }
 ```
 
@@ -37,16 +35,14 @@ kotlin {
 
 ```groovy
 kotlin {
-    abiValidation {
-        enabled = true
-    }
+    abiValidation()
 }
 ```
 
 </tab>
 </tabs>
 
-If your project has multiple modules where you want to check for binary compatibility, configure each module separately.
+The KGP creates the necessary Gradle tasks. If your project has multiple modules where you want to check for binary compatibility, configure each module separately.
 
 ## Check for binary compatibility issues
 
@@ -166,9 +162,7 @@ To disable this behavior, add the following to your `build.gradle.kts` file:
 kotlin {
     @OptIn(org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation::class)
     abiValidation {
-        klib {
-            keepUnsupportedTargets.set(false)
-        }
+        keepLocallyUnsupportedTargets.set(false)
     }
 }
 ```
@@ -179,9 +173,7 @@ kotlin {
 ```groovy
 kotlin {
     abiValidation {
-        klib {
-            keepUnsupportedTargets = false
-        }
+        keepLocallyUnsupportedTargets = false
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/topics/gradle/gradle-binary-compatibility-validation.md
shanshin: `enable` property now is deprecated in [this task](https://youtrack.jetbrains.com/issue/KT-83486).

You must specify that ABI validations is enabled if `kotlin { abiValidation() }` or `kotlin { abiValidation { ... } }` has been called at least once in the build script.

And also it's Gradle tasks are created only if ABI validation is enabled.
As a result, accessors to tasks are no longer available.

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -191,3 +187,38 @@ kotlin {
...
 If a target is unsupported and inference is disabled, the `checkKotlinAbi` task fails because it can't generate a complete
 ABI dump. This behavior may be useful if you'd prefer the task to fail rather than risk missing a binary-incompatible change.
+
+## Include publications from the `maven-publish` plugin
+
+By default, binary compatibility validation uses Kotlin compilation outputs. If your project publishes artifacts with
+the [`maven-publish` plugin](https://docs.gradle.org/current/userguide/publishing_maven.html), this approach means the ABI dumps may not reflect the actual published artifacts. For example,
+this can happen when post-processing steps such as relocation modify the artifact.
+
+To make sure the ABI dumps accurately reflect the artifacts published by the `maven-publish` plugin, add the following to
```
shanshin, line 193: It is worth pointing out somewhere that this approach does not work for Android projects and android target in KMP, because the usual jar file is not published for them.

sarahhaggarty: I added a note at the bottom.

docs/topics/gradle/gradle-binary-compatibility-validation.md
sarahhaggarty: Updated!

shanshin (approved):

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -11,24 +11,22 @@ users and encouraging continued adoption of the library.
 > 
 {style="tip"}
 
-The Kotlin Gradle plugin includes support for binary compatibility validation. When enabled, it generates
+The Kotlin Gradle plugin includes support for binary compatibility validation. It generates
```
zamulla: I think this still is opaque as to what "it" is. Maybe I'd phrase it more directly, like "The KGP can validate binary compatibility. To do that, it generates..."

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -191,3 +183,43 @@ kotlin {
 
 If a target is unsupported and inference is disabled, the `checkKotlinAbi` task fails because it can't generate a complete
 ABI dump. This behavior may be useful if you'd prefer the task to fail rather than risk missing a binary-incompatible change.
+
+## Include publications from the `maven-publish` plugin
+
+By default, binary compatibility validation uses Kotlin compilation outputs. If your project publishes artifacts with
+the [`maven-publish` plugin](https://docs.gradle.org/current/userguide/publishing_maven.html), this approach means the ABI dumps may not reflect the actual published artifacts. For example,
+this can happen when post-processing steps such as relocation modify the artifact.
```
zamulla: I feel that this paragraph is precise, but can be unclear, if that makes sense. In my head I laid it out as follows in the end:
1. The validation uses outputs by default.
2. This may not always work.
3. For example, the maven-publish plugin can change artifacts in post-processing, which invalidates ABI dumps.

docs/topics/gradle/gradle-binary-compatibility-validation.md
```diff
@@ -191,3 +183,43 @@ kotlin {
...
+}
+```
+
+</tab>
+</tabs>
+
+> Because Kotlin/Android projects and multiplatform projects with an Android target don't publish JAR files, this feature
+> doesn't apply to them.
+> 
+{style="note"}
```
zamulla: Should this be a warning as opposed to a note? I think the feature not working at all in major use cases qualifies this for "[difficult to complete the task](https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit?tab=t.0#heading=h.i1qtnyywcdla)" :)

zamulla (approved):
All good 👍 

Left a couple of polishing suggestions.

docs/topics/gradle/gradle-binary-compatibility-validation.md
sarahhaggarty: I can change it to "The plugin generates"? The KGP doesn't validate directly. You have to do that yourself by reviewing the generated diffs.

docs/topics/gradle/gradle-binary-compatibility-validation.md
sarahhaggarty: Fair enough!

docs/topics/gradle/gradle-binary-compatibility-validation.md
zamulla: "The plugin" works for me 👌

docs/topics/gradle/gradle-binary-compatibility-validation.md
sarahhaggarty: How does this paragraph read as an alternative?
By default, binary compatibility validation uses Kotlin compilation outputs to generate ABI dumps. As a result, the generated ABI dumps might not reflect the final published artifacts. For example, when you use the [`maven-publish` plugin](https://docs.gradle.org/current/userguide/publishing_maven.html),
post-processing steps such as relocation can modify artifacts after compilation.

zamulla: Noice 👍 

Maybe "Because of that / for that reason" instead of "as a result"? Just reads smoother to me personally, feel free to ignore :)

PR #5518, opened 2026-05-04
chore(ktl-4396): added ai cases

berezinant:
For the demo new cases are the first four, but actually they were reordered
image


nikpachoo (commented):
@berezinant, I don't see the new category in the filter. Is it expected?

Image

nikpachoo (approved):

PR #5520, opened 2026-05-06
update: add info for the new returnsResultOf contract

daniCsorbaJB:
This PR adds information for https://youtrack.jetbrains.com/issue/KT-84190

### DIFF
--- a/docs/topics/compiler/compiler-reference.md
+++ b/docs/topics/compiler/compiler-reference.md
@@ -240,7 +240,7 @@ Allows using the `holdsIn` keyword in contracts to assume that a boolean conditi
 
 #### -Xallow-returns-result-of
 
-Allows using the `returnsResultOf()` contract to improve unused result checks for higher-order functions.
+Allows using the `returnsResultOf()` contract to enable the unused return value checker to distinguish between ignorable and meaningful unused results from higher-order functions.
 
 ### -Xreturn-value-checker
 <primary-label ref="experimental-general"/>
--- a/docs/topics/compiler/compiler-reference.md
+++ b/docs/topics/compiler/compiler-reference.md
@@ -238,6 +238,10 @@ Allows using the `returnsNotNull()` function in contracts to assume a non-null r
 
 Allows using the `holdsIn` keyword in contracts to assume that a boolean condition is `true` inside a lambda.
 
+#### -Xallow-returns-result-of
+
+Allows using the `returnsResultOf()` contract to improve unused result checks for higher-order functions.
+
 ### -Xreturn-value-checker
 <primary-label ref="experimental-general"/>
 
--- a/docs/topics/compiler/compiler-reference.md
+++ b/docs/topics/compiler/compiler-reference.md
@@ -240,7 +240,7 @@ Allows using the `holdsIn` keyword in contracts to assume that a boolean conditi
 
 #### -Xallow-returns-result-of
 
-Allows using the `returnsResultOf()` contract to enable the unused return value checker to distinguish between ignorable and meaningful unused results from higher-order functions.
+Allows using the `returnsResultOf()` contract so that the unused return value checker can distinguish between results that can be ignored and meaningful results from higher-order functions.
 
 ### -Xreturn-value-checker
 <primary-label ref="experimental-general"/>
--- a/docs/topics/unused-return-value-checker.md
+++ b/docs/topics/unused-return-value-checker.md
@@ -41,6 +41,10 @@ fun formatGreeting(name: String): String {
 }
 ```
 
+> You can also use the [experimental `returnsResultOf()` contract](whatsnew24.md#improve-unused-result-checks-for-higher-order-functions) to enable the checker to distinguish between meaningful and ignorable unused results from higher-order functions that return the result of a lambda.
+>
+{style="note"}
+
 ## Configure the unused return value checker
 
 You can control how the compiler reports ignored results with the `-Xreturn-value-checker` compiler option.
--- a/docs/topics/unused-return-value-checker.md
+++ b/docs/topics/unused-return-value-checker.md
@@ -185,7 +185,7 @@ fun check(g: Greeter) {
 ## Check for unused results in higher-order functions
 
 Some higher-order functions, such as the `.let()` extension function, return the result of a lambda.
-To check for unused results in these functions, add the [Experimental](components-stability.md#stability-levels-explained) `returnsResultOf()` contract to the function's `contract` block.
+To check for unused results for these functions, add the [Experimental](components-stability.md#stability-levels-explained) `returnsResultOf()` contract to the function's `contract` block.
 
 > Kotlin contracts are Experimental. To opt in, add the `@OptIn(ExperimentalContracts::class)` annotation when declaring a function with a contract.
 >
@@ -206,7 +206,7 @@ inline fun <T, R> T.myLet(block: (T) -> R): R {
 }
 ```
 
-You can then use a function with this contract, such as `.myLet()`, to check the lambda result for unused results:
+You can then use a function with this contract, such as `.myLet()`, to check if the lambda result is used:
 
 ```kotlin
 fun handleNullablePackageName(s: String?, sb: StringBuilder) {
--- a/docs/topics/unused-return-value-checker.md
+++ b/docs/topics/unused-return-value-checker.md
@@ -41,10 +41,6 @@ fun formatGreeting(name: String): String {
 }
 ```
 
-> You can also use the [experimental `returnsResultOf()` contract](whatsnew24.md#improve-unused-result-checks-for-higher-order-functions) to enable the checker to distinguish between meaningful and ignorable unused results from higher-order functions that return the result of a lambda.
->
-{style="note"}
-
 ## Configure the unused return value checker
 
 You can control how the compiler reports ignored results with the `-Xreturn-value-checker` compiler option.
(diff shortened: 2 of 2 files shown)


docs/topics/unused-return-value-checker.md
```diff
@@ -41,6 +41,10 @@ fun formatGreeting(name: String): String {
 }
 ```
 
+> You can also use the [experimental `returnsResultOf()` contract](whatsnew24.md#improve-unused-result-checks-for-higher-order-functions) to enable the checker to distinguish between meaningful and ignorable unused results from higher-order functions that return the result of a lambda.
```
sandwwraith: This note looks very out of context. Why didn't you add a whole paragraph, like in what's new?

```
## Return value checker and higher order functions

To better convey semantic of higher-order functions, Kotlin has`returnsResultOf()` contract to improve the [unused return value checker](unused-return-value-checker.md).

This contract enables the checker to distinguish between ignorable and meaningful unused results from higher-order functions that return the result of a lambda, such as the `.let()` extension function.

> Kotlin contracts are [Experimental](components-stability.md#stability-levels-explained). To opt > in, add the `@OptIn(ExperimentalContracts::class)` annotation when declaring a function with > a contract.
>
{style=”warning”}

To use this feature, add `returnsResultOf()` to the function's `contract` block:

```kotlin
import kotlin.contracts.ExperimentalContracts
import kotlin.contracts.contract

@OptIn(ExperimentalContracts::class)
inline fun <T, R> T.myLet(block: (T) -> R): R {
    contract {
        returnsResultOf(block)
    }
    return block(this)
}
```

Here's an example that uses a custom `myLet()` function with a nullable value:

```kotlin
fun handleNullablePackageName(s: String?, sb: StringBuilder) {
    // The checker doesn't report a warning because append() has an ignorable return value
    s?.myLet { sb.append(it) }

    // The checker reports a warning because the returned String is unused
    s?.myLet { "kotlin.$it" }
    // Unused return value.
}
```

```

**Review by sandwwraith — CHANGES_REQUESTED** [review]

File: docs/topics/unused-return-value-checker.md (hunk 1 above)
Comment (daniCsorbaJB) [reply docs/topics/unused-return-value-checker.md]: as discussed on Slack let's add it as a separate section then but cover the compiler option as well.

File: docs/topics/unused-return-value-checker.md [hunk 2]
```diff
@@ -182,6 +182,81 @@ fun check(g: Greeter) {
...
+@OptIn(ExperimentalContracts::class)
+inline fun <T, R> T.myLet(block: (T) -> R): R {
+    contract {
+        returnsResultOf(block)
+    }
+    return block(this)
+}
+```
+
+You can then use a function with this contract, such as `.myLet()`, to check the lambda result for unused results:
```
sandwwraith:
```suggestion
You can then use a function with this contract, such as `.myLet()`, to check if the lambda result is properly used:
```

result...results sounds too repetetive

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,81 @@ fun check(g: Greeter) {
 }
 ```
 
+## Check for unused results in higher-order functions
+
+Some higher-order functions, such as the `.let()` extension function, return the result of a lambda.
+To check for unused results in these functions, add the [Experimental](components-stability.md#stability-levels-explained) `returnsResultOf()` contract to the function's `contract` block.
```
sandwwraith: I think "in" is slightly wrong: in `s?.myLet { "kotlin.$it" }` we report `myLet` is unused, so we check "outside" the lambda function. Maybe just say "for these functions"? I don't know better preposition.

sandwwraith (changes requested):

docs/topics/unused-return-value-checker.md
daniCsorbaJB: sure that version works well as well 👍 — I'd suggest removing the word *properly* unless we explain what properly means here, it adds noise

docs/topics/unused-return-value-checker.md
daniCsorbaJB: 🤔 That is a great point, thank you. for these functions works perfectly imho 💯

sandwwraith:
Why was this force pushed? I see a lot of irrelevant commits

daniCsorbaJB:
> Why was this force pushed? I see a lot of irrelevant commits

Should be fixed now

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,81 @@ fun check(g: Greeter) {
...
+You can then use a function with this contract, such as `.myLet()`, to check if the lambda result is used:
+
+```kotlin
+fun handleNullablePackageName(s: String?, sb: StringBuilder) {
+    // The checker doesn't report a warning because append() has an ignorable return value
+    s?.myLet { sb.append(it) }
+
+    // The checker reports a warning because the returned String is unused
+    s?.myLet { "kotlin.$it" }
+    // Unused return value.
```
sandwwraith: Obsolete comment?

daniCsorbaJB: We can remove one of them, although one of them (above the line) states what happens, and the other (below the line) shows the output. Or am I missing something? Was something changed here? 🤔

sandwwraith: I didn't have a clue what it meant until you explained it to me. Do we usually do this with warnings? I think I've never seen this approach before (including this very same document)

daniCsorbaJB: I put it there for clarity, but it's not necessary. If you think it creates confusion, let's remove the output part. 👍

sandwwraith: The fact that this is the only place throughout the document where you did that indeed creates confusion

sandwwraith (approved):

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
 }
 ```
 
+## Check for unused results in higher-order functions
+
+Some higher-order functions, such as the `.let()` extension function, return the result of a lambda.
```
sarahhaggarty:
```suggestion
Some higher-order functions, such as the `let` scope function, return the result of a lambda.
```
You mean the scope function, right?

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+To check for unused results for these functions, add the [Experimental](components-stability.md#stability-levels-explained) `returnsResultOf()` contract to the function's `contract` block.
```
sarahhaggarty:
```suggestion
To check for unused lambda results of higher-order functions, add the [Experimental](components-stability.md#stability-levels-explained) `returnsResultOf()` contract to the function's contract.
```
In the past, we've just referred to it as "contract" so I think we can continue doing the same. Also, what do you think about this for a beginning instead?

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+{style="warning"}
+
+Here's an example:
+
+```kotlin
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+
+@OptIn(ExperimentalContracts::class)
+inline fun <T, R> T.myLet(block: (T) -> R): R {
```
sarahhaggarty, line 201: Can we use a better name for the `.myLet()` function? Like `transform()`?

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+        returnsResultOf(block)
+    }
+    return block(this)
+}
+```
+
+You can then use a function with this contract, such as `.myLet()`, to check if the lambda result is used:
+
+```kotlin
+fun handleNullablePackageName(s: String?, sb: StringBuilder) {
```
sarahhaggarty: Can you use better names for the arguments? Like maybe `packageName` and `builder`?

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+    // The checker doesn't report a warning because append() has an ignorable return value
+    s?.myLet { sb.append(it) }
+
+    // The checker reports a warning because the returned String is unused
+    s?.myLet { "kotlin.$it" }
+}
+```
+
+The `returnsResultOf()` contract requires a separate compiler option to opt in, and using it produces pre-release binaries that Kotlin compiler versions earlier than 2.4.0 can't read.
```
sarahhaggarty:
```suggestion
The `returnsResultOf()` contract requires a separate compiler option to opt in. Be aware that using it produces pre-release binaries that Kotlin compiler versions earlier than 2.4.0 can't read.
```
I think this is important enough to become a warning.

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+    }
+    return block(this)
+}
+```
+
+You can then use a function with this contract, such as `.myLet()`, to check if the lambda result is used:
+
+```kotlin
+fun handleNullablePackageName(s: String?, sb: StringBuilder) {
+    // The checker doesn't report a warning because append() has an ignorable return value
```
sarahhaggarty:
```suggestion
    // The checker doesn't report a warning because the return value of the append() function can be ignored
```
Ignorable isn't a common word so how about this instead?

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+    s?.myLet { sb.append(it) }
+
+    // The checker reports a warning because the returned String is unused
```
sarahhaggarty:
```suggestion
    // The checker reports a warning because the returned string is unused
```

docs/topics/compiler/compiler-reference.md
```diff
@@ -238,6 +238,10 @@ Allows using the `returnsNotNull()` function in contracts to assume a non-null r
 
 Allows using the `holdsIn` keyword in contracts to assume that a boolean condition is `true` inside a lambda.
 
+#### -Xallow-returns-result-of
+
+Allows using the `returnsResultOf()` contract to enable the unused return value checker to distinguish between ignorable and meaningful unused results from higher-order functions.
```
sarahhaggarty:
```suggestion
Allows using the `returnsResultOf()` contract so that the unused return value checker can distinguish between results that can be ignored and and results that should be used in higher-order functions.
```
To avoid "ignorable".

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,80 @@ fun check(g: Greeter) {
...
+    s?.myLet { "kotlin.$it" }
+}
+```
+
+The `returnsResultOf()` contract requires a separate compiler option to opt in, and using it produces pre-release binaries that Kotlin compiler versions earlier than 2.4.0 can't read.
+
+To enable it in your project, add the following compiler option to your build file:
```
sarahhaggarty:
```suggestion
To opt in for your project, add the following compiler option to your build file:
```

sarahhaggarty (changes requested):
Thanks for adding this! I made some suggestions, please take a look 🙏 Also, it would be great to use some better names in the example!

docs/topics/unused-return-value-checker.md
daniCsorbaJB: I like the start (bit less connected to the previous sentence, but feels more natural) — For the contract part, I have no strong opinion, tbh I find the original a bit clearer, and it avoids repeating the word contract: add contract to the function's contract.

docs/topics/unused-return-value-checker.md
daniCsorbaJB: Hmm 🤔 I think something like `transform()` would be misleading, and this way it contrasts with the previous example, the let() scope function. we could perhaps call it `customLet()`, but I think `.myLet()` works here.

docs/topics/compiler/compiler-reference.md
daniCsorbaJB: I like the beginning, but I prefer the ending of the original. the doc already uses the word meaningful result (should be used feels be a bit misleading - it could be a false positive)

> Allows using the `returnsResultOf()` contract so that the unused return value checker can distinguish between results that can be ignored and meaningful results from higher-order functions.

docs/topics/unused-return-value-checker.md
daniCsorbaJB: sure 👍 good idea 🙏

docs/topics/unused-return-value-checker.md
sarahhaggarty: I mainly want to avoid saying block because we currently use that for parts of the Gradle build file 🤔

docs/topics/unused-return-value-checker.md
sarahhaggarty: `customLet()` would be fine :) `myLet()` sounds more like a placeholder than a meaningful name that gives some idea what the purpose of the function is.

docs/topics/compiler/compiler-reference.md
sarahhaggarty: Your suggestion works for me!

docs/topics/unused-return-value-checker.md
```diff
@@ -182,6 +182,83 @@ fun check(g: Greeter) {
 }
 ```
 
+## Check for unused results in higher-order functions
+
+Some higher-order functions, such as the `let()` scope function, return the result of a lambda.
```
sarahhaggarty:
```suggestion
Some higher-order functions, such as the `let` scope function, return the result of a lambda.
```
We don't use parentheses when talking about scope functions :)

sarahhaggarty (approved):
Just one last suggestion! :)

PR #5521, opened 2026-05-06
update: add information for new map nullability functions

daniCsorbaJB:
This PR updates the Map-specific operations page with new map nullability fallback functions.
Related ticket is here: [KT-67337](https://youtrack.jetbrains.com/issue/KT-67337)

### DIFF
--- a/docs/topics/map-operations.md
+++ b/docs/topics/map-operations.md
@@ -11,25 +11,44 @@ For retrieving a value from a map, you must provide its key as an argument of th
 The shorthand `[key]` syntax is also supported. If the given key is not found, it returns `null`.
 There is also the function [`getValue()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-value.html)
 which has slightly different behavior: it throws an exception if the key is not found in the map.
-Additionally, you have two more options to handle the key absence: 
+Additionally, you have more options to handle missing keys and nullable values:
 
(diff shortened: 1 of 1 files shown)


docs/topics/map-operations.md
```diff
@@ -11,25 +11,44 @@ For retrieving a value from a map, you must provide its key as an argument of th
 The shorthand `[key]` syntax is also supported. If the given key is not found, it returns `null`.
 There is also the function [`getValue()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-value.html)
 which has slightly different behavior: it throws an exception if the key is not found in the map.
-Additionally, you have two more options to handle the key absence: 
+Additionally, you have more options to handle missing keys and nullable values:
```
ilya-g: I would extract the new variants into a separate list.
So the first list is for the key absence in general, and the second - specifically for maps with nullable values.
The reason is that we would like to discourage using `getOrElse` for maps with nullable values in future and suggest using one of the new variants instead.

ilya-g (changes requested):

docs/topics/map-operations.md
daniCsorbaJB: good point 👍  — I updated it and phrased it to show that users should prefer these if it has nullable values.

ilya-g (approved):

koshachy (approved):
LGTM!

PR #5526, opened 2026-05-07
feat: stabilization of features and add explicit context arguments

daniCsorbaJB:
This PR adds information about explicit context arguments to context parameters.
Related issue: [KT-84188](https://youtrack.jetbrains.com/issue/KT-84188)

It also updates the relevant pages of the Stabilization of `all` meta-target and context parameters.

### DIFF
--- a/docs/topics/context-parameters.md
+++ b/docs/topics/context-parameters.md
@@ -1,7 +1,5 @@
 [//]: # (title: Context parameters)
 
-<primary-label ref="experimental-general"/>
-
 > Context parameters replace an older experimental feature called [context receivers](whatsnew1620.md#prototype-of-context-receivers-for-kotlin-jvm).
 > You can find their main differences in the [design document for context parameters](https://github.com/Kotlin/KEEP/blob/master/proposals/context-parameters.md#summary-of-changes-from-the-previous-proposal).
 > To migrate from context receivers to context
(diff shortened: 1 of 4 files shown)


serras (approved):

docs/topics/context-parameters.md
```diff
@@ -86,39 +103,82 @@ fun main() {
 }
 ```
 
-#### Restrictions
+### Pass context arguments explicitly
 
-Context parameters are in continuous improvement, and some of the current restrictions include:
+When overloads differ only by context parameters, a call can become ambiguous if multiple matching context values are available.
+You can resolve this ambiguity by passing an explicit context argument at the call site.
```
koshachy: To resolve the ambiguity, pass...

koshachy: Does it work?

koshachy: In this case you can add ":" after the sentence without stating "here's an example" explicitly.

docs/topics/context-parameters.md
```diff
@@ -86,39 +103,82 @@ fun main() {
...
 
-```Bash
--Xcontext-parameters
+    // Selects the overload with the SmsSender context parameter
+    sendNotification(smsSender = defaultSmsSender)
+}
 ```
 
-Or add it to the `compilerOptions {}` block of your Gradle build file:
+You can also use explicit context arguments instead of the `context()` function to reduce nesting and make some calls easier to read. If you need to use the same context arguments in multiple calls, use the `context()` function instead.
```
koshachy: Are these two options do the opposite?
It is a bit indirect now, what if we change the meaning "To reduce nesting and make some (do we need this word?) call easier to read, ... . Otherwise, ...."

koshachy (changes requested):
See my comments pls

docs/topics/context-parameters.md
daniCsorbaJB: I think it works, I like it 👍 — let's go with that

docs/topics/context-parameters.md
daniCsorbaJB: 🤔 I think a sentence with Otherwise... would make it a bit indirect (or if we say Otherwise, if .... — it feels a bit clunky)

the word some is just there to anchor the user that it's not always the right choice - WDYT about something like this instead to show the contrast a bit more: 

You can also use explicit context arguments to reduce nesting in some function calls:

* For a single call, use explicit context arguments to make the call easier to read.
* If multiple calls use the same context arguments, use the `context()` function.

koshachy (approved):
LGTM

PR #5530, opened 2026-05-11
update: add information about isSorted APIs to ordering page

daniCsorbaJB:
This PR adds information about new APIs to check whether elements are already sorted without sorting them again or creating your own helper functions.

Releated issue is here: [KT-78499](https://youtrack.jetbrains.com/issue/KT-78499)

### DIFF
--- a/docs/topics/collection-ordering.md
+++ b/docs/topics/collection-ordering.md
@@ -180,6 +180,10 @@ You can use the following extension functions to check whether elements already 
 These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
 They return `false` and stop checking as soon as they find an out-of-order pair.
 
+For collections and sequences, the result depends on the order in which elements are iterated.
+When checking `Double` and `Float` values, these functions treat `NaN` as greater than any other value and `-0.0` as less than `0.0`.
+Additionally, the `.isSortedBy()` and `.isSortedByDescending()` functions treat `null` selector results as less than any non-null value.
+
 Here's an example of checking sorted order with the `.isSorted()` and `.isSortedBy()` functions:
 
 ```kotlin
--- a/docs/topics/collection-ordering.md
+++ b/docs/topics/collection-ordering.md
@@ -167,6 +167,40 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
+## Check sorted order
+
+You can use the following extension functions to check whether elements already follow a specified order:
+
+* `.isSorted()`
+* `.isSortedDescending()`
+* `.isSortedWith(comparator)`
+* `.isSortedBy(selector)`
+* `.isSortedByDescending(selector)`
+
+These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
+They return `false` and stop checking as soon as they find an out-of-order pair.
+
+Here's an example of checking sorted order with the `.isSorted()` and `.isSortedBy()` functions:
+
+```kotlin
+data class User(val name: String, val age: Int)
+
+fun main() {
+    val numbers = listOf(1, 2, 3, 4)
+    println(numbers.isSorted())
+    // true
+
+    val users = listOf(
+        User("Alice", 24),
+        User("Bob", 31),
+        User("Charlie", 29),
+    )
+    println(users.isSortedBy(User::age))
+    // false
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.4"}
+
 ## Reverse order
 
 You can retrieve the collection in the reversed order using the [`reversed()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/reversed.html) function. 
--- a/docs/topics/collection-ordering.md
+++ b/docs/topics/collection-ordering.md
@@ -180,14 +180,15 @@ You can use the following extension functions to check whether elements already 
 These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
 They return `false` and stop checking as soon as they find an out-of-order pair.
 
-For collections without a guaranteed iteration order, such as `HashSet`, and sequences that don't produce elements in a consistent order, the result may vary across calls.
-For ordered collections, such as `List`, repeated calls on the same elements return the same result.
+For collections without a guaranteed iteration order, such as `HashSet`, the result may vary across calls.
+The same applies to sequences that don't produce elements in a consistent order.
+To get the same result across calls, use these functions only on collections with a guaranteed iteration order, such as `List`.
 
 When checking `Double` and `Float` values, these functions treat `NaN` as greater than any other value and `-0.0` as less than `0.0`.
 Additionally, the `.isSortedBy()` and `.isSortedByDescending()` functions treat `null` selector results as less than any non-null value.
 
-When you call these functions on a `Sequence`, the operation is terminal.
-It consumes the sequence to produce a `Boolean` instead of returning another sequence.
+When you call these functions on a sequence, the operation is terminal.
+It consumes the sequence to produce a `Boolean` value instead of returning another sequence.
 
 > These sorted-order functions are also available for arrays, primitive arrays, and unsigned arrays.
 > Unsigned arrays and operations on them are [Experimental](components-stability.md#stability-levels-explained) and require opt-in with the `@ExperimentalUnsignedTypes` annotation.
--- a/docs/topics/collection-ordering.md
+++ b/docs/topics/collection-ordering.md
@@ -94,9 +94,9 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-Since sorting by multiple criteria is a common scenario, the Kotlin standard library provides the [`thenBy()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.comparisons/then-by.html) function that you can use to add a secondary sorting rule.
+Since sorting by multiple criteria is a common scenario, the Kotlin standard library provides the [`.thenBy()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.comparisons/then-by.html) function that you can use to add a secondary sorting rule.
 
-For example, you can combine `compareBy()` with `thenBy()` to sort strings by their length first and alphabetically second, just like in the previous example:
+For example, you can combine `compareBy()` with `.thenBy()` to sort strings by their length first and alphabetically second, just like in the previous example:
 
 ```kotlin
 fun main() {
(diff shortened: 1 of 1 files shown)


docs/topics/collection-ordering.md
```diff
@@ -167,6 +167,44 @@ fun main() {
...
 
+## Check sorted order
+
+You can use the following extension functions to check whether elements already follow a specified order:
+
+* `.isSorted()`
+* `.isSortedDescending()`
+* `.isSortedWith(comparator)`
+* `.isSortedBy(selector)`
+* `.isSortedByDescending(selector)`
```
DmitryNekrasov, line 178: Two style notes to match the rest of the page:

1. **Leading dots.** Other inline function references on this page use `sorted()`, `sortedDescending()`, `sortedWith()`, etc., without a leading dot. Suggest dropping the `.` here and in the prose mentions below.

2. **API reference links.** Every function name elsewhere on the page is wrapped in a markdown link to its API page (e.g. [`sorted()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/sorted.html)). Worth adding the same for `isSorted`, `isSortedDescending`, `isSortedWith`, `isSortedBy`, and `isSortedByDescending` once their 2.4 API ref pages are published (the stdlib 2.4 pages aren't live yet, so I won't guess at the exact URLs — you'll know the convention better than me when they go up).

docs/topics/collection-ordering.md
```diff
@@ -167,6 +167,44 @@ fun main() {
...
+
+These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
+They return `false` and stop checking as soon as they find an out-of-order pair.
+
+For collections and sequences, the result depends on the order in which elements are iterated.
```
DmitryNekrasov: This is a bit too broad. For `List` and most `Sequence` implementations the iteration order is deterministic, so this currently reads as a warning when nothing is actually wrong. The KDoc concern is specifically *unstable* iteration order — unordered collections like `HashSet`. Suggested rewording:

> For collections without a guaranteed iteration order (such as `HashSet`), the result may be inconsistent across calls. For ordered collections like `List`, the result is stable.

docs/topics/collection-ordering.md
```diff
@@ -167,6 +167,44 @@ fun main() {
...
+When checking `Double` and `Float` values, these functions treat `NaN` as greater than any other value and `-0.0` as less than `0.0`.
+Additionally, the `.isSortedBy()` and `.isSortedByDescending()` functions treat `null` selector results as less than any non-null value.
```
DmitryNekrasov, line 188: Two additions worth considering near these notes:

1. **`Sequence` is terminal.** The `Sequence<T>` overloads carry an explicit "The operation is _terminal_." note in their KDoc — worth surfacing, since terminal vs intermediate is something `Sequence` users actively pay attention to.

2. **Arrays are supported too.** The section is currently scoped to "collections and sequences", but the same family ships on `Array<T>`, all primitive arrays (`IntArray`, `FloatArray`, …), and unsigned array types (`UIntArray`, etc., behind `@ExperimentalUnsignedTypes`). Even if the Ordering page is collection-focused, a one-line aside ("These functions are also available on arrays and primitive arrays.") would head off the question.

DmitryNekrasov (commented):
Hey Daniel — thanks for putting this together. Overall a solid first pass: the API family is correctly enumerated, you caught the right edge cases (`NaN`/`-0.0`, `null` selector handling), and placement between "Custom orders" and "Reverse order" is the right call.

Inline comments below — mostly style alignment with the rest of the page (leading dots, API links, sample markers) plus a couple of small technical nuances (the iteration-order caveat is a bit too broad; `Sequence` overloads are terminal; arrays are supported too).

Separately, on KT-73111 (`UInt`/`ULong` → `BigInteger`) — fully agree with your call. The Unsigned integer types page is too introductory, the Numbers page intentionally omits `Int.toBigInteger()` / `Long.toBigInteger()` for the same reason, and the API ref + What's new entry is the right level of coverage. A dedicated "Working with `BigInteger` / `BigDecimal` on the JVM" topic would be the natural home if we ever want a fuller picture, but that's well out of scope for 2.4.0.

Thanks again for syncing on this!

docs/topics/collection-ordering.md
```diff
@@ -167,6 +167,44 @@ fun main() {
...
+
+Here's an example of checking sorted order with the `.isSorted()` and `.isSortedBy()` functions:
+
+```kotlin
+data class User(val name: String, val age: Int)
+
+fun main() {
```
DmitryNekrasov, line 203: Missing `//sampleStart` / `//sampleEnd` markers — every other runnable example on this page uses them to highlight the relevant region. As written, the `data class User` declaration is shown as part of the focus area.

docs/topics/collection-ordering.md
```diff
@@ -167,6 +167,44 @@ fun main() {
...
+    println(numbers.isSorted())
+    // true
+
+    val users = listOf(
+        User("Alice", 24),
+        User("Bob", 31),
+        User("Charlie", 29),
+    )
+    println(users.isSortedBy(User::age))
+    // false
```
DmitryNekrasov, line 215: The example exercises 2 of the 5 functions, and only the `false` outcome for `isSortedBy`. If you wanted to round it out without bloating the snippet, one extra line showing `isSortedWith(comparator)` (ties back to the "Custom orders" section directly above) or `isSortedDescending()` (to complement the ascending case) would help. Something like:

```kotlin
val descending = listOf(4, 3, 2, 1)
println(descending.isSortedDescending())
// true
```

docs/topics/collection-ordering.md
daniCsorbaJB: 1. Good point, the docs use leading dots for extension functions — the issue is that the original functions like `sortedBy()` didn't follow this convention, but it would be more correct to change those to have the leading dots instead. (`.sortedBy()`) I'll add all of these so it's aligned with the new ones which should be correct. 
I'll leave the compare functions because they are either top-level or member functions

2. you are right 👍 💯  — I'll add the links after the API reference is available.

docs/topics/collection-ordering.md
daniCsorbaJB: thank you! 👍 very nice I slightly rephrased but your version is good: 

> For collections without a guaranteed iteration order, such as `HashSet`, and sequences that don't produce elements in a consistent order, the result may vary across calls. For ordered collections, such as `List`, repeated calls on the same elements return the same result.

docs/topics/collection-ordering.md
daniCsorbaJB: both are good points thank you! 

I'm thinking of adding the first as an extra paragraph, and second as a note since it's a bit of an extra information: 

When you call these functions on a `Sequence`, the operation is terminal.
It consumes the sequence to produce a `Boolean` instead of returning another sequence.

> These sorted-order functions are also available for arrays, primitive arrays, and unsigned arrays.
> Unsigned arrays and operations on them are [Experimental](components-stability.md#stability-levels-explained) and require opt-in with the `@ExperimentalUnsignedTypes` annotation.
> 
{style="note"}

docs/topics/collection-ordering.md
daniCsorbaJB: I don't think it's needed, but it doesn't hurt either, and you are right it's better for consistency — I added it. 👍

docs/topics/collection-ordering.md
daniCsorbaJB: really nice addition — I added it 👍

DmitryNekrasov (approved):
Thanks Daniel — all five threads are addressed, and your rephrasing of the iteration-order caveat reads cleaner than my draft. Approving. API ref links can land in a follow-up once the 2.4 stdlib pages publish.

docs/topics/collection-ordering.md
```diff
@@ -167,12 +167,66 @@ fun main() {
...
+* `.isSorted()`
+* `.isSortedDescending()`
+* `.isSortedWith(comparator)`
+* `.isSortedBy(selector)`
+* `.isSortedByDescending(selector)`
+
+These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
+They return `false` and stop checking as soon as they find an out-of-order pair.
+
+For collections without a guaranteed iteration order, such as `HashSet`, and sequences that don't produce elements in a consistent order, the result may vary across calls.
```
sarahhaggarty:
```suggestion
For collections and sequences that don't have a guaranteed order, such as `HashSet`, results may vary across calls.
```
This is a little complicated. Can you shorten it to something like this?

docs/topics/collection-ordering.md
```diff
@@ -167,12 +167,66 @@ fun main() {
...
+For ordered collections, such as `List`, repeated calls on the same elements return the same result.
 
-You can retrieve the collection in the reversed order using the [`reversed()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/reversed.html) function. 
+When checking `Double` and `Float` values, these functions treat `NaN` as greater than any other value and `-0.0` as less than `0.0`.
+Additionally, the `.isSortedBy()` and `.isSortedByDescending()` functions treat `null` selector results as less than any non-null value.
+
+When you call these functions on a `Sequence`, the operation is terminal.
```
sarahhaggarty:
```suggestion
When you call these functions on a sequence, the operation is terminal.
```

docs/topics/collection-ordering.md
```diff
@@ -167,12 +167,66 @@ fun main() {
...
+It consumes the sequence to produce a `Boolean` instead of returning another sequence.
```
sarahhaggarty:
```suggestion
It consumes the sequence to produce a `Boolean` value instead of returning another sequence.
```

docs/topics/collection-ordering.md
```diff
@@ -167,12 +167,66 @@ fun main() {
...
+* `.isSortedDescending()`
+* `.isSortedWith(comparator)`
+* `.isSortedBy(selector)`
+* `.isSortedByDescending(selector)`
+
+These extension functions return `true` if the elements are in the specified order or if there are fewer than two elements.
+They return `false` and stop checking as soon as they find an out-of-order pair.
+
+For collections without a guaranteed iteration order, such as `HashSet`, and sequences that don't produce elements in a consistent order, the result may vary across calls.
+For ordered collections, such as `List`, repeated calls on the same elements return the same result.
```
sarahhaggarty: Can you adjust this paragraph to make it more advisory? That these functions should only be called on collections/sequences that have a guaranteed order?

sarahhaggarty (approved):
It's OK to merge as-is but I suggested some small tweaks to improve it even further :)

docs/topics/collection-ordering.md
daniCsorbaJB: Hmm 🤔 , that’s a tough one. 

My concern with the simpler version is that it muddies the distinction a little: HashSet is an example of a collection without a guaranteed iteration order, not an example of a sequence. (which feels unclear to me in the second one) For sequences, the issue is whether they produce elements in a consistent order, which isn’t quite the same as a collection having a guaranteed iteration order.

Maybe we can keep the more precise structure, but make it a bit more parallel to make it parse a bit easier (although it makes the sentence even longer 😞 ):

For collections without a guaranteed iteration order, such as HashSet, and for sequences that don't produce elements in a consistent order, the result may vary across calls.

docs/topics/collection-ordering.md
daniCsorbaJB: good idea 👍 

To get the same result across calls, use these functions only on collections with a guaranteed iteration order, such as `List`.

docs/topics/collection-ordering.md
sarahhaggarty: I really struggle to parse something that long 😞 Maybe we could swap it around?

> Use these functions when the order stays consistent across calls. For example, with ordered collections like lists. 
> `HashSet` collections and sequences in general don't guarantee iteration order.

WDYT?

daniCsorbaJB: As discussed on Slack let's go with: 

> For collections without a guaranteed iteration order, such as `HashSet`, the result may vary across calls.
The same applies to sequences that don't produce elements in a consistent order.
To get the same result across calls, use these functions only on collections with a guaranteed iteration order, such as `List`.

PR #5537, opened 2026-05-12
feat: New page for UUIDs

aepanchi:


docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+  You can also generate a version 7 UUID for a specific moment in time,
+  using the [`Uuid.generateV7NonMonotonicAt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7-non-monotonic-at.html) function.
+
+> Learn more about the generation algorithm in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/).
+>
+{style="tip"}
+
+## Parse UUIDs
+
+Some UUIDs can be first declared as strings.
```
fzhinkin: can be represented as strings?

fzhinkin (approved):
Great document, thank you!

docs/topics/uuids.md
aepanchi: thanks, good catch, updated!

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
+[//]: # (title: UUIDs)
+[//]: # (description: Learn how to use UUIDs in Kotlin, including creating, parsing, formatting, serializing, and working with UUID values across multiplatform and JVM code.)
+
+The [`Uuid`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/)
+class represents universally unique identifiers (UUIDs),
+also known as Globally Unique Identifiers (GUIDs).
```
daniCsorbaJB, line 6:
```suggestion
also known as globally unique identifiers (GUIDs).
```

daniCsorbaJB: since we didn't capitalize it above

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+Use the `Uuid` class to work with UUID values. Unlike plain strings, a dedicated UUID
+type makes your code more explicit and prevents accidental use of invalid values.
+
+The Kotlin standard library provides the `Uuid` class in the `kotlin.uuid` package:
+
+```kotlin
+import kotlin.uuid.Uuid
+```
+## Generate UUIDs
```
daniCsorbaJB:
```suggestion

## Generate UUIDs
```

daniCsorbaJB: missing line

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+A `Uuid` is a 128-bit value used to uniquely identify an entity without
+relying on a central system that assigns IDs. This makes UUIDs useful
+in distributed applications, databases, client-generated records,
+or [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform/get-started.html) applications.
+
+Use the `Uuid` class to work with UUID values. Unlike plain strings, a dedicated UUID
+type makes your code more explicit and prevents accidental use of invalid values.
+
+The Kotlin standard library provides the `Uuid` class in the `kotlin.uuid` package:
```
daniCsorbaJB: this feels a bit unclear WDYT about: To use UUIDs in your project, import the `Uuid` class from the `kotlin.uuid` package:

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+```kotlin
+import kotlin.uuid.Uuid
+```
+## Generate UUIDs
+
+To create a new random UUID, use the [`Uuid.random()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/random.html) function.
+It generates a version 4 UUID from random values.
```
daniCsorbaJB: I think we could make this a bit more concise, right now I feel like it repeats the same idea twice: 

To generate a random version 4 UUID for regular identifiers, such as user or database IDs, use the [`Uuid.random()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/random.html) function:

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+Use the `Uuid.random()` function for regular identifiers, such as user or database IDs.
+
+```kotlin
+import kotlin.uuid.Uuid
+fun main() {
```
daniCsorbaJB, line 30:
```suggestion

fun main() {
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+//sampleStart    
+    val id = Uuid.random()
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
```
daniCsorbaJB:
```suggestion
{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
```

daniCsorbaJB: afaik this only works from 2.0.0 and up but I would omit it because some of these APIs only work 2.3.0 and up.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+If your application require a specific version of UUIDs,
```
daniCsorbaJB:
```suggestion
If your application requires a specific version of UUIDs,
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+  fun main() {
+  //sampleStart    
+      val id = Uuid.generateV7()
+      println(id)
+  //sampleEnd    
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+  You can also generate a version 7 UUID for a specific moment in time,
```
daniCsorbaJB:
```suggestion
  You can also generate a version 7 UUID for a specific moment in time
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+  using the [`Uuid.generateV7NonMonotonicAt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7-non-monotonic-at.html) function.
```
daniCsorbaJB:
```suggestion
  with the [`Uuid.generateV7NonMonotonicAt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7-non-monotonic-at.html) function.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+type makes your code more explicit and prevents accidental use of invalid values.
+
+The Kotlin standard library provides the `Uuid` class in the `kotlin.uuid` package:
+
+```kotlin
+import kotlin.uuid.Uuid
+```
+## Generate UUIDs
+
+To create a new random UUID, use the [`Uuid.random()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/random.html) function.
```
daniCsorbaJB:
```suggestion
To generate a new random UUID, use the [`Uuid.random()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/random.html) function.
```

daniCsorbaJB: to make it consistent with the title

docs/topics/uuids.md
daniCsorbaJB: WDYT about making this list a bit less crowded we could perhaps introduce this part with something like.

> If your application requires a specific UUID version, you can generate UUIDs with version-specific functions.

Then we could follow it up with 
Use the `Uuid.generateV4()` function to generate a version 4 UUID:

[example]

This function makes the UUID version explicit, but otherwise works the same as `Uuid.random()`.

To generate version 7 UUID... and so on

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+you can use the following functions:
+
+* The [`Uuid.generateV4()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v4.html) function creates the same type of UUID as the `Uuid.random()` function
+  but explicitly states that the value is a version 4 UUID.
+
+  Use it when version 4 UUIDs are specifically required.
+
+  ```kotlin
+  import kotlin.uuid.Uuid
+  fun main() {
```
daniCsorbaJB: same comments here, please add a line after import and change the version - here it's only 2.3 afaik but I suggest omitting it.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+      println(id)
+  //sampleEnd    
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+  You can also generate a version 7 UUID for a specific moment in time,
+  using the [`Uuid.generateV7NonMonotonicAt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7-non-monotonic-at.html) function.
+
+> Learn more about the generation algorithm in the [API Reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/).
```
daniCsorbaJB: I think this tip is a bit narrow if we are linking to the companion object. 
If we want to provide more information, I suggest something like

 > For the complete list of `Uuid` companion functions, see the [`Uuid.Companion` API reference](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/).

But even that's a bit too narrow because then we don't talk about properties, so I would suggest omitting it, we already linked to the relevant parts above.

docs/topics/uuids.md
daniCsorbaJB: I think we could add an example here as well. 

Also I suggest adding information about `Uuid.generateV7NonMonotonicAt()` not guaranteeing monotonic ordering, so multiple UUIDs created for the same timestamp might not be sequential.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+>
+{style="tip"}
+
+## Parse UUIDs
+
+Some UUIDs can be first represented as strings.
+For example, when you receive a UUID from a URL parameter or database record.
```
daniCsorbaJB: I think we could make this a bit more compact: 

UUID values are often represented as strings, such as in URL parameters or database records.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+To convert a `String` value to a `Uuid` value,
+use the [`Uuid.parse()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse.htm) function:
+
+```kotlin
+val id = Uuid.parse("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
```
daniCsorbaJB: could we make this into a runnable example as well? 🤔 — it feels weird that the runnable example is the one that isn't suppoed to work.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    val id = Uuid.parse("10")
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" validate="false"}
+
+If your application accepts only one representation, use:
+
+* The [`Uuid.parseHexDash()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash.html) function to parse from the hex-and dash string representation.
```
daniCsorbaJB:
```suggestion
* The [`Uuid.parseHexDash()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash.html) function to parse from the hex-and-dash string representation.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+fun main() { 
+//sampleStart    
+    val id = Uuid.parse("10")
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" validate="false"}
+
+If your application accepts only one representation, use:
```
daniCsorbaJB:
```suggestion
If your application accepts only one representation, use the format-specific functions:
```

daniCsorbaJB: this way we could make the list a bit more concise: 

* [`Uuid.parseHexDash()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash.html) for the hex-and-dash string representation.
* [`Uuid.parseHex()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex.html) for the hexadecimal string representation without dashes.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+* The [`Uuid.parseHexDash()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash.html) function to parse from the hex-and dash string representation.
+* The [`Uuid.parseHex()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex.html) function to parse from the hexadecimal string representation without dashes.
+
+For example:
+
+```kotlin
```
daniCsorbaJB, line 117: we could also make this runnable WDYT?

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+or [`Uuid.parseHexOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-or-null.html)
+functions to handle invalid values safely:
+
+```kotlin
+fun parseId (input: String): Uuid? {
+    return Uuid.parseOrNull(input)
+}
+```
+
+> Use the `Uuid.parseOrNull()`, `Uuid.parseHexDashOrNull()`, and `Uuid.parseHexOrNull()` functions
```
daniCsorbaJB: I think we can explain this in flow text instead of making it a tip see my comment above

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+* The [`Uuid.parseHex()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex.html) function to parse from the hexadecimal string representation without dashes.
+
+For example:
+
+```kotlin
+val standard = Uuid.parseHexDash("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
+val compact = Uuid.parseHex("de2bc56cea734f3c8a375a46fdb2d79a")
+```
+
+If you have UUIDs from external input, use the
```
daniCsorbaJB: I think we can refine that and fold the tip into it. WDYT:

If your application handles UUID strings from external sources and must handle invalid input without throwing exceptions, use `Uuid.parseOrNull()`, `Uuid.parseHexDashOrNull()`, or `Uuid.parseHexOrNull()`. These functions return `null` when the input is invalid:

[Example]

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+[`Uuid.parseOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-or-null.html), [`Uuid.parseHexDashOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash-or-null.html),
+or [`Uuid.parseHexOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-or-null.html)
+functions to handle invalid values safely:
+
+```kotlin
+fun parseId (input: String): Uuid? {
```
daniCsorbaJB: we could also make this runnable 🏃

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    return Uuid.parseOrNull(input)
+}
+```
+
+> Use the `Uuid.parseOrNull()`, `Uuid.parseHexDashOrNull()`, and `Uuid.parseHexOrNull()` functions
+> when invalid input is part of your normal application workflow.
+>
+{style="tip"}
+
+## Convert UUIDs
```
daniCsorbaJB: Let's make the title a bit more specific WDYT? if we are only converting to strings something like : 

Convert UUIDs to strings

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+You can convert a `Uuid` value to a `String` value using the following functions:
```
daniCsorbaJB:
```suggestion
You can convert a `Uuid` value to a `String` value with the following functions:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+For example:
+
+```kotlin
+import kotlin.uuid.Uuid
+fun main() {
+//sampleStart    
+    val id = Uuid.parse("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
+    
+    println(id.toString()) // de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a
```
daniCsorbaJB:
```suggestion
    println(id.toString())
    // de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a
```

daniCsorbaJB: If the code outputs a result , reflect this in a comment on the next line.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+> Use the `Uuid.parseOrNull()`, `Uuid.parseHexDashOrNull()`, and `Uuid.parseHexOrNull()` functions
+> when invalid input is part of your normal application workflow.
+>
+{style="tip"}
+
+## Convert UUIDs
+
+You can convert a `Uuid` value to a `String` value using the following functions:
+
+* [`.toString()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-string.html) for the standard string representation
```
daniCsorbaJB: afaik these are member functions not extension functions so we don't need the `.˙ in `toString()` here and for the rest in this list

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+## Use UUIDs on the JVM
+
+When you work on the JVM, you can interoperate your Kotlin code with Java libraries
+that use the `java.util.UUID` class.
+
+The `java.util.UUID` and `kotlin.unit.UUID` classes represent the same kind of values,
+but they are two distinct types. You need to convert values explicitly to pass UUIDs between Kotlin and Java:
+
+* Convert a Java UUID to Kotlin using the [`.toKotlinUuid()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/to-kotlin-uuid.html) function:
```
daniCsorbaJB:
```suggestion
* Convert a Java UUID to Kotlin with the [`.toKotlinUuid()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/to-kotlin-uuid.html) extension function:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    println(id.toHexDashString()) // de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a
+    println(id.toHexString()) // de2bc56cea734f3c8a375a46fdb2d79a 
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+## Compare UUIDs
+
+You can perform equality checks on the `Uuid` values using the `==` operator.
```
daniCsorbaJB: 🤔 this one parses a bit mechanical WDYT something like: You can check whether `Uuid` values are equal using the `==` operator.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+Kotlin evaluates values according to the UUID value, not to the textual representation.
```
daniCsorbaJB: here I think we could make the connection clearer by saying something along the lins of: 
Kotlin compares values according to the UUID value, not the textual representation. 

Kotlin evaluates values according to the UUID value feels a bit more distant compared to compares

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+For example, two values in the different forms are equal if they represent the same 128-bit value:
```
daniCsorbaJB:
```suggestion
For example, two values in different forms are equal if they represent the same 128-bit value:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    val first = Uuid.parse("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
+    val second = Uuid.parse("de2bc56cea734f3c8a375a46fdb2d79a")
+
+    println(first == second) // true 
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+This functionality makes the `Uuid` comparison safer than comparing UUIDs as strings.
```
daniCsorbaJB:
```suggestion
This makes the `Uuid` comparison safer than comparing UUIDs as strings.
```

daniCsorbaJB: do we need this part? 🤔

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+String comparison fails if you write the same value in different formats.
```
daniCsorbaJB: saying it fails is a bit strange here 🤔  The comparison doesn't fail - it returns `false`. 
so wdyt: String comparison treats the same value in different formats as different values.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+`Uuid` comparison checks the actual identifier value.
```
daniCsorbaJB, line 195: Hmm 🤔 reading these sentences together, WDYT about merging these 3 sentences into one:

This makes `Uuid` comparison more reliable than string comparison, which treats different string formats as different values.

docs/topics/uuids.md
daniCsorbaJB: I feel like the word "safer" is a bit misleading here. It could probably give you a false negative, but wouldn't crash anything afaik. perhaps reliable? 🤔 WDYT?

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+You can also sort UUIDs using the `Comparable<Uuid>` interface. In that case,
```
daniCsorbaJB: I might be missing a connection here 🤔 Could you please explain how `Comparable` connects to sorting in relation to UUIDs? 
Do we mean that because `Uuid` implements `Comparable<Uuid>`, the user can sort UUIDs with standard collection functions such as `sorted()`.

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    val second = Uuid.generateV7()
+
+    val sorted = listOf(first, second).sorted()
+    println(sorted) 
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+## Work with bytes
```
daniCsorbaJB:
```suggestion
## Work with UUID binary representations
```

daniCsorbaJB: WDYT would that work? 🤔

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+    val bytes = id.toByteArray()
+    val original = Uuid.fromByteArray(bytes)
+  
+    println(id)
+    
+    println(bytes)
+    println(original)
+
+    println(id == original) // true
```
daniCsorbaJB:
```suggestion
    println(id == original)
    // true
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+You can also create a UUID from two 64-bit values. Since Kotlin does not provide a built-in 128-bit integer type,
```
daniCsorbaJB:
```suggestion
You can also create a UUID from two 64-bit values. Since Kotlin doesn't provide a built-in 128-bit integer type,
```

daniCsorbaJB: I found this transition really difficult to follow. 
Do you think we could start the section to introduce that these are basically two ways to do this: 

Some APIs and storage formats represent UUIDs as bytes or as two 64-bit values. representing bytes this way --- add use cases
Then we can follow with .toByteArray() example. 

Then transition to this one. with the explanation that Kotlin doesn't provide a built-in 128-bit integer type

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+the API represents a UUID as two `Long` values using:
+* The `mostSignificantBits` parameter for the first 64 bits of a UUID.
+* The `leastSignificantBits` parameter for the last 64 bits of a UUID.
+
+If another system gives you a UUID in the two-part format,
```
daniCsorbaJB: does it have to be another system ? 🤔 Could something like this stand on its own?

To create a `Uuid` value from two `Long` values, use the [`Uuid.fromLongs()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/from-longs.html) function:

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    println(leastSignificantBits)
+    }
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+## Serialize UUIDs
+
+Kotlin supports serialization for the `Uuid` values.
```
daniCsorbaJB:
```suggestion
Kotlin supports serialization for `Uuid` values.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+Use it to store or transfer a UUID value outside Kotlin code, for example, in JSON APIs or configuration files.
+
+To serialize a `Uuid` value, represent it as a `String` value unless otherwise required by your system.
```
daniCsorbaJB:
```suggestion
To serialize a `Uuid` value, represent it as a string unless your application requires another format.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+The [`kotlinx.serialization`](https://kotlinlang.org/docs/serialization.html) functionality uses the standard hex-and-dash format:
```
daniCsorbaJB:
```suggestion
The [`kotlinx.serialization`](https://kotlinlang.org/docs/serialization.html) library uses the hex-and-dash format:
```

daniCsorbaJB: I wouldn't call this a functionality, it's a library/libraries (or a plugin depending on the context) 

regarding standard: is there a reason to call it standard hex-and-dash? 🤔

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+    )
+
+    println(Json.encodeToString(user))
+    // {"id":"de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a","name":"Kotlin"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+## Use UUIDs on the JVM
```
daniCsorbaJB: Hmm 🤔 I feel like the section focuses more on the interaction between java.util.UUID and Kotlin UUIDs than on the JVM in my opinion, WDYT about a title like ## Use UUIDs with Java APIs or similar

docs/topics/uuids.md
```diff
@@ -0,0 +1,334 @@
...
+
+When you work on the JVM, you can interoperate your Kotlin code with Java libraries
```
daniCsorbaJB: Do you think we could make this a bit more concise? 

Someting like: 

Java represents UUIDs with the `java.util.UUID` class. On the JVM, Java APIs may accept or return this type.

Although `java.util.UUID` and `kotlin.uuid.Uuid` both represent UUIDs, they are separate classes/types. 
(we could even add the note from below here)

To convert Java UUID to Kotlin, use X..

daniCsorbaJB (changes requested):
Thank you for the nice addition! 💯
I left a couple of comments. Please let me know if anything is unclear.

daniCsorbaJB (changes requested):
Thank you for the nice addition! 💯
I left a couple of comments. Please let me know if anything is unclear.

docs/topics/uuids.md
aepanchi: Nice catch! But prefer to capitalize the first instance instead.

docs/topics/uuids.md
aepanchi: Yes, sounds much better!

docs/topics/uuids.md
aepanchi: Nice catch! Will apply where possible.

docs/topics/uuids.md
aepanchi: The use case for this function is quite niche tbh. I don't think it's worth documenting it here - people who may need it, would be able to do so using API ref.

docs/topics/uuids.md
aepanchi: fair enough!

docs/topics/uuids.md
aepanchi: Converted to align, but I'm not sure if it is needed, since the output looks just like the content we have in (). It does not demonstrate the change from string to a Uuid.

docs/topics/uuids.md
aepanchi: nice catch!

docs/topics/uuids.md
aepanchi: Exactly, the connection is:

Uuid implements Comparable<Uuid>, which means UUID values have a defined natural order. Because of that, collection functions that rely on natural ordering, such as sorted(), can sort a collection of UUIDs directly.

Rephrased to make clearer.

docs/topics/uuids.md
aepanchi: Restructured. WDYT?

docs/topics/uuids.md
daniCsorbaJB: also works 👍  thank you

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+  //sampleEnd    
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="2.3"}
+
+  This function makes the UUID version explicit, but otherwise works
+  the same as the `Uuid.random()` function.
+
+* Use the [`Uuid.generateV7()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7.html) function to generate a version 7 UUID that contains
+  a time-based component and random data:
```
daniCsorbaJB: What do we mean here by random data? 🤔 — I assume the part that makes it unique, but it wouldn't be a UUID without it 🤔 (I might be missing something so please correct me if I'm wrong)

I think this would be clearer if we would say something along the lines:  

Use the [`Uuid.generateV7()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7.html) function to generate a version 7 UUID that includes a timestamp for when it was generated:

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+    val id = Uuid.random()
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+You can also generate a specific version of UUIDs:
+
+* Use the [`Uuid.generateV4()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v4.html) function to generate a version 4 UUID:
```
daniCsorbaJB: hmm 🤔 this list is a bit crowded — do you think we could restructure it to work without it?

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+  You can also generate a version 7 UUID for a specific moment in time
+  with the [`Uuid.generateV7NonMonotonicAt()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7-non-monotonic-at.html) function.
+
+## Parse UUIDs
+
+UUID values are often represented as strings,
+such as in URL parameters or database records.
+
+To convert a `String` value to a `Uuid` value,
+use the [`Uuid.parse()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse.htm) function:
```
daniCsorbaJB:
```suggestion
use the [`Uuid.parse()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse.html) function:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+
+  ```kotlin
+  import kotlin.uuid.toJavaUuid
+  
+  val javaId: java.util.UUID = kotlinId.toJavaUuid()
+  ```
+
+These functions allow you to represent your UUID values using `Uuid` at JVM interoperability boundaries.
+
+> The `java.util.UUID` and `kotlin.unit.UUID` classes are comparable, but the ordering may differ.
```
daniCsorbaJB:
```suggestion
> The `java.util.UUID` and `kotlin.unit.Uuid` classes are comparable, but the ordering may differ.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+    // {"id":"de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a","name":"Kotlin"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3" validate="true"}
+
+## Use UUIDs with Java APIs
+
+To represent UUIDs, Java uses `java.util.UUID` class. On the JVM, Java APIs may
+accept or return this type. Although `java.util.UUID` and `kotlin.unit.UUID` both represent UUIDs,
```
daniCsorbaJB:
```suggestion
accept or return this type. Although `java.util.UUID` and `kotlin.unit.Uuid` both represent UUIDs,
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+//sampleStart    
+    val id = Uuid.parse("10")
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0" validate="false"}
+
+If your application accepts only one representation, use
+the format specific-functions:
```
daniCsorbaJB:
```suggestion
the format-specific functions:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+    val second = Uuid.parse("de2bc56cea734f3c8a375a46fdb2d79a")
+
+    println(first == second) 
+    // true 
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3" validate="true"}
+
+This makes the `Uuid` comparison more reliable than string comparing,
```
daniCsorbaJB:
```suggestion
This makes the `Uuid` comparison more reliable than comparing strings,
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+    // true
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+You can also represent the same 128-bit UUID value as two `Long` values. 
+This is useful because Kotlin does not provide a built-in 128-bit integer type. 
+The two `Long` values store the UUID in two parts:
+* The `mostSignificantBits` parameter for the first 64 bits of a UUID.
```
daniCsorbaJB, line 255:
```suggestion

* The `mostSignificantBits` parameter for the first 64 bits of a UUID.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3" validate="true"}
+
+If you have UUIDs from external sources and must handle invalid input safely,
+use [`Uuid.parseOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-or-null.html), [`Uuid.parseHexDashOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash-or-null.html),
+or [`Uuid.parseHexOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-or-null.html). 
+These functions return `null` if the input is invalid:
+
+```kotlin
+fun parseId (input: String): Uuid? {
```
daniCsorbaJB:
```suggestion
fun parseId(input: String): Uuid? { 
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+import kotlin.uuid.Uuid
+
+fun main() {
+//sampleStart
+    val id = Uuid.parse("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0" validate="true"}
```
daniCsorbaJB: As a general note: validate="true" isn't necessary

docs/topics/uuids.md
```diff
@@ -0,0 +1,367 @@
...
+  fun main() {
+  //sampleStart    
+      val id = Uuid.generateV7()
+      println(id)
+  //sampleEnd    
+  }
+  ```
+  {kotlin-runnable="true" kotlin-min-compiler-version="2.3"}
+
+  Use the `Uuid.generateV7()` function when you need to sort UUIDs by creation time.
```
daniCsorbaJB: Hmm 🤔 I'm a bit worried that this is an overpromise, so WDYT a bit softer version.

The `Uuid.generateV7()` function generates UUIDs that you can sort by their timestamp. or something similar (also avoids repeating Use)

even it gets a bit repetitive I suggest  to differentiate between generate and create words: generate something that the compiler creates - create is something the user creates 🤔

daniCsorbaJB (changes requested):
Thank you for the changes💯 — I left a couple of additiona comments!

docs/topics/uuids.md
aepanchi: I removed comparison with the random function, I don't think it's that important tbh.

docs/topics/uuids.md
aepanchi: Yes, good point, fixed.

docs/topics/uuids.md
aepanchi: Moved this to the part above the code block

docs/topics/uuids.md
aepanchi: Good catch!

docs/topics/uuids.md
aepanchi: removed

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+    // {"id":"de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a","name":"Kotlin"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3"}
+
+## Use UUIDs with Java APIs
+
+To represent UUIDs, Java uses `java.util.UUID` class. On the JVM, Java APIs may
+accept or return this type. Although `java.util.UUID` and `kotlin.unit.Uuid` both represent UUIDs,
```
daniCsorbaJB:
```suggestion
accept or return this type. Although `java.util.UUID` and `kotlin.uuid.Uuid` both represent UUIDs,
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+
+  ```kotlin
+  import kotlin.uuid.toJavaUuid
+  
+  val javaId: java.util.UUID = kotlinId.toJavaUuid()
+  ```
+
+These functions allow you to represent your UUID values using `Uuid` at JVM interoperability boundaries.
+
+> The `java.util.UUID` and `kotlin.unit.Uuid` classes are comparable, but the ordering may differ.
```
daniCsorbaJB:
```suggestion
> The `java.util.UUID` and `kotlin.uuid.Uuid` classes are comparable, but the ordering may differ.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+
+Some APIs, storage formats, and binary protocols do not represent UUIDs as strings. 
+Instead, they store the 128-bit UUID value as either:
+
+* A 16-byte array 
+* Two 64-bit values
+
+Use these representations when you need to exchange UUIDs with systems that expect binary UUID data.
+
+To convert a UUID to and from 16-byte representation, use the [`.toByteArray()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-byte-array.html)
```
daniCsorbaJB:
```suggestion
To convert a UUID to and from a 16-byte representation, use the [`.toByteArray()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-byte-array.html)
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+
+    println(id == original) 
+    // true
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+You can also represent the same 128-bit UUID value as two `Long` values. 
+This is useful because Kotlin does not provide a built-in 128-bit integer type.
```
daniCsorbaJB:
```suggestion
This is useful because Kotlin doesn't provide a built-in 128-bit integer type. 
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+        leastSignificantBits = -6288180676521310383.toLong()
+    )
+    println(id) 
+    // bf6ac971-52fd-4aad-a8bb-e4fdac78c751
+//sampleEnd  
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+To extract two parts from an existing `Uuid` value, use the [`Uuid.toLongs()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-longs.html) function:
```
daniCsorbaJB:
```suggestion
To extract the two parts from an existing `Uuid` value, use the [`Uuid.toLongs()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-longs.html) function:
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+    println(Json.encodeToString(user))
+    // {"id":"de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a","name":"Kotlin"}
+}
+//sampleEnd
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3"}
+
+## Use UUIDs with Java APIs
+
+To represent UUIDs, Java uses `java.util.UUID` class. On the JVM, Java APIs may
```
daniCsorbaJB:
```suggestion
Java uses the `java.util.UUID` class to represent UUIDs. On the JVM, Java APIs may
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+//sampleStart
+    val id = Uuid.parse("de2bc56c-ea73-4f3c-8a37-5a46fdb2d79a")
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+The `Uuid.parse()` function accepts both the standard hex-and-dash
+format and hexadecimal format without dashes.
```
daniCsorbaJB:
```suggestion
format and the hexadecimal format without dashes.
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+
+These functions allow you to represent your UUID values using `Uuid` at JVM interoperability boundaries.
+
+> The `java.util.UUID` and `kotlin.unit.Uuid` classes are comparable, but the ordering may differ.
+> Make sure to check your code that depends on UUID ordering before migrating from Java API to Kotlin API.
+>
+{style="note"}
+
+Kotlin also provides support for working with Java buffers.
+Use JVM-specific functions to work with UUIDs from a `ByteBuffer`:
```
daniCsorbaJB:
```suggestion
Use JVM-specific functions to work with UUIDs in a `ByteBuffer`:
```

daniCsorbaJB: nit: from works for get but for put it parses slightly wrong

docs/topics/uuids.md
```diff
@@ -0,0 +1,363 @@
...
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+To extract two parts from an existing `Uuid` value, use the [`Uuid.toLongs()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-longs.html) function:
+
+```kotlin
+import kotlin.uuid.Uuid
+
+fun main() {
+//sampleStart 
+    val id = Uuid.random().toLongs { mostSignificantBits, leastSignificantBits ->
```
daniCsorbaJB: I think we should use something like this: 

```

import kotlin.uuid.Uuid

fun main() {
//sampleStart 
    val id = Uuid.random()

    id.toLongs { mostSignificantBits, leastSignificantBits ->
        println(mostSignificantBits)
        println(leastSignificantBits)
    }
//sampleEnd  
}
```

The issue with the current snippet is that it assigns the result of `toLongs()` to `id`, but the lambda returns `Unit`, so it's not a Uuid.

daniCsorbaJB (changes requested):
Thank you for the ongoing updates. The page is shaping up well. 💯 

I have a few additional suggestions.

docs/topics/uuids.md
daniCsorbaJB: hmm 🤔 — I think it should be otherwise it might lead to some confusion what's the difference between the two.

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+fun main() {
+//sampleStart    
+    val id = Uuid.random()
+    println(id)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.0"}
+
+You can also generate a specific version of UUIDs with the following functions:
```
daniCsorbaJB:
```suggestion
You can also generate specific versions of UUIDs with the following functions:
```

daniCsorbaJB: or You can also generate UUIDs of specific versions

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+
+* The [`Uuid.generateV4()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v4.html) function generates the same type of UUID as the `Uuid.random()` function 
+  but explicitly states that the value is a version 4 UUID.
+*  The [`Uuid.generateV7()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7.html) function generates a version 7 UUID with a timestamp
```
daniCsorbaJB:
```suggestion
* The [`Uuid.generateV7()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/generate-v7.html) function generates a version 7 UUID with a timestamp
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+   that you can use for UUIDs sorting.
```
daniCsorbaJB:
```suggestion
   that you can use for sorting UUIDs.
```

daniCsorbaJB: or for UUID sorting

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+  
+fun main() { 
+//sampleStart    
+    val idVersion4 = Uuid.generateV4()
+    println(idVersion4)
+    
+    val idVersion7 = Uuid.generateV7()
+    println(idVersion7)
+
+   val timestamp = Instant.fromEpochMilliseconds(1757440583000L)
```
daniCsorbaJB: fix indendation in this lines (only 3 spaces instead of 4)

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+
+Here's an example that generates version-specific UUIDs:
+
+```kotlin
+import kotlin.time.Instant
+import kotlin.uuid.Uuid
+  
+fun main() { 
+//sampleStart    
+    val idVersion4 = Uuid.generateV4()
```
daniCsorbaJB, line 56: Could you please add code comments here before each relevant line like val idVersion4 = ...? 

// Generates a version 4 UUID
// Generates a version 7 UUID
// Generates a version 7 UUID for the specified timestamp

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+    println(standard)
+    println(compact)
+//sampleEnd    
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="2.3"}
+
+If you have UUIDs from external sources and must handle invalid input safely,
+use [`Uuid.parseOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-or-null.html), [`Uuid.parseHexDashOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-dash-or-null.html),
+or [`Uuid.parseHexOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-or-null.html).
```
daniCsorbaJB:
```suggestion
or [`Uuid.parseHexOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/-companion/parse-hex-or-null.html).
```

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+
+Some APIs, storage formats, and binary protocols do not represent UUIDs as strings. 
+Instead, they store the 128-bit UUID value as either:
+
+* A 16-byte array 
+* Two 64-bit values
+
+Use these representations when you need to exchange UUIDs with systems that expect binary UUID data.
+
+To convert a UUID to and from a 16-byte representation, use the [`.toByteArray()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.uuid/-uuid/to-byte-array.html)
```
daniCsorbaJB: (stlye suggestion) 
To convert between a UUID and a 16-byte representation — this way we can avoid to and from

docs/topics/uuids.md
```diff
@@ -0,0 +1,362 @@
...
+  ```kotlin
+  import kotlin.uuid.toJavaUuid
+  
+  val javaId: java.util.UUID = kotlinId.toJavaUuid()
+  ```
+
+These functions allow you to represent your UUID values using `Uuid` at JVM interoperability boundaries.
+
+> The `java.util.UUID` and `kotlin.uuid.Uuid` classes are comparable, but the ordering may differ.
+> Make sure to check your code that depends on UUID ordering before migrating from Java API to Kotlin API.
```
daniCsorbaJB: 🤔 
If we mean specifically between these two : `java.util.UUID` and `kotlin.uuid.Uuid` 
let's add the artice:   Make sure to check your code that depends on UUID ordering before migrating from the Java API to the Kotlin API.

daniCsorbaJB (approved):
Just a couple more comments from me, otherwise LGTM — thank you 🚀

PR #5545, opened 2026-05-18
Add explicit backing fields content

sarahhaggarty:
This PR adds documentation about explicit backing fields, which becomes stable in Kotlin 2.4.0.

### DIFF
--- a/docs/topics/properties.md
+++ b/docs/topics/properties.md
@@ -237,25 +237,21 @@ fun main() {
 
 This example uses [reflection](reflection.md) to show which annotations are present on the getter and setter.
 
-### Backing fields
+## Backing fields
 
-In Kotlin, accessors use backing fields to store the property's value in memory. Backing fields are useful
-when you want to add extra logic to a getter or setter, or when you want to trigger an additional action whenever the property
-changes.
+The compiler automatically generates backing fields for properties when a value needs to be stored in memory.
 
-You can't declare backing fields directly. Kotlin generates them only when necessary. You can reference the backing field
-in accessors using the `field` keyword.
-
-Kotlin only generates backing fields if you use the default getter or setter, or if you use ` field` in at least one custom accessor.
-
-For example, the `isEmpty` property has no backing field because it uses a custom getter without the `field` keyword:
+For example, the compiler creates a backing field when you use the default `get()` and `set()` functions because they
+read and write the stored value:
 
 ```kotlin
-val isEmpty: Boolean
-    get() = this.size == 0
+var count = 0
 ```
 
-In this example, the `score` property has a backing field because the setter uses the `field` keyword:
+You can access backing fields by using the `field` keyword in a [custom `get()` or `set()` function](#custom-getters-and-setters).
+For example, you can add extra logic to a getter or setter, or trigger an additional action when a property changes.
+
+In this example, the `score` property uses the backing field inside the `set()` function so that updating the value also triggers a log event:
 
 ```kotlin
 class Scoreboard {
@@ -277,32 +273,35 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-field"}
 
-### Backing properties
+Backing fields aren't created by default for all properties because they might not need them. For example, the `isEmpty`
+property doesn't have a backing field because the value is calculated from the `size` property each time you access it:
 
-Sometimes you might need more flexibility than using a [backing field](#backing-fields) can provide. For example, if you have an API
-where you want to be able to modify the property internally but not externally. In such cases, you can use a coding pattern
-called a _backing property_.
+```kotlin
+val isEmpty: Boolean
+    get() = this.size == 0
+```
+
+### Explicit backing fields
+
+Sometimes you might need more flexibility. For example, if you have an API where you want to be able to modify the property
+internally but not externally. In such cases, you can use an _explicit backing field_.
 
 In the following example, the `ShoppingCart` class has an `items` property that represents everything in the shopping cart.
-You want the `items` property to be read-only outside the class but still allow one "approved" way for the user to modify
-the `items` property directly. To achieve this, you can define a private backing property called `_items` and a public property
-called `items` that delegates to the backing property's value.
+The class exposes the `items` property as a read-only list of strings, but internally it stores the data in a mutable list
+with an explicit backing field:
 
 ```kotlin
 class ShoppingCart {
-    // Backing property
-    private val _items = mutableListOf<String>()
-
-    // Public read-only view
+    // Public read-only view with explicit backing field
     val items: List<String>
-        get() = _items
-
+        field = mutableListOf()
+    
     fun addItem(item: String) {
-        _items.add(item)
+        items.add(item)
     }
 
     fun removeItem(item: String) {
-        _items.remove(item)
+        items.remove(item)
     }
 }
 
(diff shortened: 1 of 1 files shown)


docs/topics/properties.md
```diff
@@ -319,48 +314,134 @@ fun main() {
     // [Banana]
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="kotlin-backing-property"}
+{kotlin-runnable="true" kotlin-min-compiler-version="2.4" id="kotlin-explicit-backing-field"}
 
-In this example, the user can only add items to the cart through the `addItem()` function, but can still access the 
+In this example, the user can only add and remove items from the cart through the `addItem()` and `removeItem()` functions, but can still access the
```
e2e4b6b7: nit

> In this example, the user

I do not really understand what “user” means here. Wouldn’t it be better to say “API user”, “external code/user/caller”?

docs/topics/properties.md
```diff
@@ -277,32 +277,27 @@ fun main() {
...
 class ShoppingCart {
-    // Backing property
-    private val _items = mutableListOf<String>()
-
-    // Public read-only view
+    // Public read-only view with explicit backing field
     val items: List<String>
-        get() = _items
-
+        field = mutableListOf()
```
e2e4b6b7, line 297: I would say this needs a bit more explanation:

* That in this example, the type of `field` is inferred as `MutableList<String>`
* (optionally) That this type could also be written explicitly
* That inside the class the property access is smart-casted or just typed as a type of `field`, while outside it uses the public type

This behavior is not very intuitive, so it should be explained in at least one paragraph after the code snippet.

docs/topics/properties.md
```diff
@@ -319,48 +314,134 @@ fun main() {
...
 >
 {style="tip"}
 
-On the JVM, the compiler optimizes access to private properties with default accessors to avoid function call overhead.
+In this example, the `UserDirectory` class has a read-only `users` property that lists every user in the directory. The 
+`_users` variable is the private backing property containing the real list. The getter for the public `users` property
+sorts the entries before returning them.
 
-Backing properties are also useful when you want more than one public property to share a state. For example:
+Alternatively, if you need your backing property to have `protected` visibility so that it can be accessed by subclasses:
```
e2e4b6b7: I do not think that last two examples worth mentioning. They were not mentioned before and they are quite rare. They are also quite intuitive to construct if you know the pattern. So, in my opinion the first and the most common example with modification explaining the pattern and naming convention is enough.

e2e4b6b7 (changes requested):

docs/topics/properties.md
sarahhaggarty: I will remove them.

docs/topics/properties.md
sarahhaggarty: Good point! Added "API user".

docs/topics/properties.md
sarahhaggarty: I added some more explanation as you suggested. What do you think?

docs/topics/properties.md
```diff
@@ -316,8 +316,19 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="2.4" id="kotlin-explicit-backing-field"}
 
-In this example, the user can only add and remove items from the cart through the `addItem()` and `removeItem()` functions, but can still access the
-`items` property to see what's inside.
+In this example, the compiler infers the type of the backing field from the `mutableListOf()` function: `MutableList<String>`.
```
e2e4b6b7: > from the `mutableListOf()` function

It infers from the call to the function, not from the function. So something like "from the `mutableListOf()` call" or "from the call to `mutableListOf` function" would be better

e2e4b6b7 (approved):

docs/topics/properties.md
e2e4b6b7: Looks good

docs/topics/properties.md
sarahhaggarty: Corrected!

docs/topics/properties.md
```diff
@@ -237,18 +237,18 @@ fun main() {
...
-changes.
+when you want to add extra logic to a getter or setter, or trigger an additional action when a property changes.
 
-You can't declare backing fields directly. Kotlin generates them only when necessary. You can reference the backing field
-in accessors using the `field` keyword.
+You can reference the backing field in accessors using the `field` keyword but only if they exist. Kotlin generates backing
+fields automatically in two scenarios:
 
-Kotlin only generates backing fields if you use the default getter or setter, or if you use ` field` in at least one custom accessor.
+* You use the default `get()` and `set()` functions.
```
zamulla: I was confused here as to why Kotlin generates backing fields here, or at least why is it important to me. If I only have default accessors, I have no way of using the backing fields, right?

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
-    println("${temp.celsius}°C = ${temp.fahrenheit}°F") 
-    // 100.0°C = 212.0°F
+    val directory = UserDirectory()
+
+    directory.addUser("alex")
+    println(directory.users)
+    // sarah
+    // mike
+    // emma
+    // alex
```
zamulla: If the comment is showing the result of the `println()`, then it's not in line with actual Playground output, which is `[alex, emma, mike, sarah]`

docs/topics/properties.md
```diff
@@ -237,18 +237,18 @@ fun main() {
...
+## Backing fields
 
 In Kotlin, accessors use backing fields to store the property's value in memory. Backing fields are useful
-when you want to add extra logic to a getter or setter, or when you want to trigger an additional action whenever the property
-changes.
+when you want to add extra logic to a getter or setter, or trigger an additional action when a property changes.
 
-You can't declare backing fields directly. Kotlin generates them only when necessary. You can reference the backing field
-in accessors using the `field` keyword.
+You can reference the backing field in accessors using the `field` keyword but only if they exist. Kotlin generates backing
```
zamulla: I'm not sure I understand the logic here. The backing field can be referenced only if accessors exist? Don't they always exist?

docs/topics/properties.md
zamulla:
```suggestion
* You use the default `get()` and `set()` accessors. 
```

Since that's what we usually call them?

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
+#### Limitations
 
-```kotlin
-class Temperature {
-    // Backing property storing temperature in Celsius
-    private var _celsius: Double = 0.0
+Explicit backing fields have some limitations. Properties can have explicit backing fields if they:
+
+* Don't have a custom getter.
+* Are a read-only variable `val`.
```
zamulla:
```suggestion
* Are read-only (`val`).
```

Since we never actually said that properties are variables, I feel like it's better to stay away from this association.

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
+uses the public property type `List<String>`, so API users can only see what's inside.
 
-Backing properties are also useful when you want more than one public property to share a state. For example:
+#### Limitations
 
-```kotlin
-class Temperature {
-    // Backing property storing temperature in Celsius
-    private var _celsius: Double = 0.0
+Explicit backing fields have some limitations. Properties can have explicit backing fields if they:
```
zamulla:
```suggestion
Explicit backing fields have some limitations. Properties can have explicit backing fields only if they:
```

Since the list below shows all conditions rather than a list of positive examples.

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
+* Don't have a custom getter.
+* Are a read-only variable `val`.
+* Aren't `open`.
+* Aren't a [delegated property](delegated-properties.md).
+* Aren't [compile-time constants](#compile-time-constants).
 
-    var celsius: Double
-        get() = _celsius
-        set(value) { _celsius = value }
+In addition, the backing field type must be a subtype of the property's type and have [`private` visibility](visibility-modifiers.md).
```
zamulla, line 348: I see the problem with adding this on to the list, but maybe something like "if they: <...> Declare the backing field as a subtype of the property's type with the `private` modifier."

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
 
-    var fahrenheit: Double
-        get() = _celsius * 9 / 5 + 32
-        set(value) { _celsius = (value - 32) * 5 / 9 }
+You can work around these restrictions by using backing properties instead.
+
+### Backing properties
+
+If explicit backing fields don't support your use case, you can try using a coding pattern called a _backing property_.
```
zamulla:
```suggestion
If explicit backing fields don't fit your use case, you can try using a coding pattern called a _backing property_.
```

"Support" reads more binary to me, from what I see with backing fields there's a gray area of suitability.

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
-{style="tip"}
+```kotlin
+val items: List<String>
+    // Explicit backing field with explicit type
+    field: MutableList<String> = mutableListOf()
+```
+{validate="false"}
 
-On the JVM, the compiler optimizes access to private properties with default accessors to avoid function call overhead.
+Inside the `ShoppingCart` class of the example, the compiler smart casts the `items` property to the `MutableList<String>` type, so the
```
zamulla:
```suggestion
In the example of the `ShoppingCart` class, the compiler smart casts the `items` property to the `MutableList<String>` type, so the
```

docs/topics/properties.md
```diff
@@ -319,48 +314,78 @@ fun main() {
...
+class can add and remove items from the cart through the `add()` and `remove()` functions. Outside the class, the compiler
+uses the public property type `List<String>`, so API users can only see what's inside.
```
zamulla: My implicit assumption here is to complete the sentence with "...but not what's outside" :)
Maybe something like "...can only read the `items` list." is less ambiguous?

zamulla (changes requested):
Thanks! 

My request is mostly about the phrasing in the beginning of the "Backing fields" section rather than explicit backing fields themselves; generally the suggestions are superficial.

docs/topics/properties.md
sarahhaggarty: I rewrote this section to explain the purpose of backing fields, when they are generated, and what you can use them for. Hope it reads much better now :)

docs/topics/properties.md
sarahhaggarty: This should be fixed by the rewrite.

sarahhaggarty: I decided to stick with functions as I think it's clearer :)

docs/topics/properties.md
sarahhaggarty: I'd like to keep it separate because the bullet list is conditions for the property to be able to have an explicit backing field. Then the paragraph after explains the conditions for the explicit backing field to be valid.

docs/topics/properties.md
sarahhaggarty: Thanks!

zamulla (approved):
Thanks!

PR #5546, opened 2026-05-18
Use-site defaulting for annotations in Kotlin 2.4.0

sarahhaggarty:
This PR updates the content about use-site defaulting for annotations as a result of changes coming in Kotlin 2.4.0.

### DIFF
--- a/docs/topics/annotations.md
+++ b/docs/topics/annotations.md
@@ -199,13 +199,14 @@ The full list of supported use-site targets is:
 
 ### Defaults when no use-site targets are specified
 
-If you don't specify a use-site target, the target is chosen according to the `@Target` annotation of the annotation
-being used.
-If there are multiple applicable targets, the first applicable target from the following list is used:
+If you don't specify a use-site target, the compiler chooses the target according to the `@Target` annotation of the annotation
+you use. If there are multiple applicable targets, the compiler chooses one or more of them in the following order:
 
-* `param`
-* `property`
-* `field`
+* The constructor parameter target (`param`).
+* The property target (`property`).
+* The field target (`field`), if it's applicable and the property target (`property`) isn't.
+
+If none of `param`, `property`, or `field` are applicable, the annotation is invalid, and you need to specify a use-site target explicitly.
 
 Let's use the [`@Email` annotation from Jakarta Bean Validation](https://jakarta.ee/specifications/bean-validation/3.0/apidocs/jakarta/validation/constraints/email):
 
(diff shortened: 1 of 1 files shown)


serras (approved):

docs/topics/annotations.md
```diff
@@ -199,13 +199,14 @@ The full list of supported use-site targets is:
...
-If there are multiple applicable targets, the first applicable target from the following list is used:
+If you don't specify a use-site target, the compiler chooses the target according to the `@Target` annotation of the annotation
+you use. If there are multiple applicable targets, the compiler chooses one or more of them in the following order:
 
-* `param`
-* `property`
-* `field`
+* The constructor parameter target (`param`), if applicable.
+* The property target (`property`), if applicable.
+* The field target (`field`), if applicable and the property target (`property`) isn't.
```
zamulla: "If applicable" here seems redundant with the context set, but I can see how it may sound less precise without it.

docs/topics/annotations.md
```diff
@@ -199,13 +199,14 @@ The full list of supported use-site targets is:
...
+
+If none of `param`, `property`, or `field` are applicable, the annotation is invalid.
```
zamulla: Should we tack on a call to action here? "...the annotation is invalid, and you need to specify a use-site target explicitly."

docs/topics/annotations.md
```diff
@@ -245,32 +226,9 @@ data class User(val username: String,
...
-
-    ```kotlin
-    // build.gradle.kts
-    kotlin {
-        compilerOptions {
-            freeCompilerArgs.add("-Xannotation-default-target=first-only")
-        }
-    }
-    ```
+In this example, the `@Email` annotation applies to both constructor parameters and fields. Because the `email`
```
zamulla:
```suggestion
In this example, the `@Email` annotation applies to a constructor parameter and a field. Because the `email` 
```

docs/topics/annotations.md
```diff
@@ -245,32 +226,9 @@ data class User(val username: String,
...
-    // build.gradle.kts
-    kotlin {
-        compilerOptions {
-            freeCompilerArgs.add("-Xannotation-default-target=first-only")
-        }
-    }
-    ```
+In this example, the `@Email` annotation applies to both constructor parameters and fields. Because the `email` 
+property is declared in the primary constructor and has no custom getter or setter, the compiler applies the annotation to both the `param` and `field` targets. 
+The `secondaryEmail` property isn't a constructor parameter and also has no custom getter or setter, so only the `field` target applies.
```
zamulla: I'd be tempted to make it a two-item list, to clearly separate reasoning for one and the other:

> ...applies to both:
> * The `email` is declared...
> * The `secondaryEmail` property...

zamulla (approved):
👍 Left a couple of suggestions, but nothing major.

PR #5565, opened 2026-05-26
feat: new annotation for exporting interfaces

danil-pavlov:


docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -129,12 +129,62 @@ Currently, the `@JsExport` annotation is the only way to make your functions vis
...
+    export interface DataProcessor {
+        process(data: string): number;
+    }
+    ```
+
+For Kotlin Multiplatform projects, the general rules are:
+
+* Both `expect` and `actual` interface declarations must be annotated with `@JsNoRuntime`. The only exception is
+  `external` implementations in platform-specific code on the `actual` side.
+* Using `external` interface declarations in common code on the `expect` side is not recommended. Instead, use regular
```
JSMonk: It's `forbidden` to use `expect external`

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -129,12 +129,62 @@ Currently, the `@JsExport` annotation is the only way to make your functions vis
...
+For Kotlin Multiplatform projects, the general rules are:
+
+* Both `expect` and `actual` interface declarations must be annotated with `@JsNoRuntime`. The only exception is
+  `external` implementations in platform-specific code on the `actual` side that require no annotation.
+* Using `external` interface declarations in common code on the `expect` side is prohibited. Instead, use regular 
+  interfaces annotated with `@JsNoRuntime`.
+
+Exporting Kotlin interfaces with `@JsNoRuntime` has some restrictions. The annotation is not allowed with:
+
+* `external` interfaces as they already behave as if they have `@JsNoRuntime` by default. Adding it will result in a compiler warning.
```
sarahhaggarty:
```suggestion
* `external` interfaces as they already behave as if they have `@JsNoRuntime` by default. Adding it results in a compiler warning.
```
Future > Present tense.

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -129,12 +129,62 @@ Currently, the `@JsExport` annotation is the only way to make your functions vis
...
+* `is` and `as` type checks.
+* Class references that use the [`::class` syntax](js-reflection.md).
+* Interfaces that are passed as reified type argument.
```
sarahhaggarty: What about linking to this text? https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters I know it's in inline functions, but I think it could be helpful if they really have no idea what it is.

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -129,12 +129,62 @@ Currently, the `@JsExport` annotation is the only way to make your functions vis
...
+    ```
+
+For Kotlin Multiplatform projects, the general rules are:
+
+* Both `expect` and `actual` interface declarations must be annotated with `@JsNoRuntime`. The only exception is
+  `external` implementations in platform-specific code on the `actual` side that require no annotation.
+* Using `external` interface declarations in common code on the `expect` side is prohibited. Instead, use regular 
+  interfaces annotated with `@JsNoRuntime`.
+
+Exporting Kotlin interfaces with `@JsNoRuntime` has some restrictions. The annotation is not allowed with:
```
sarahhaggarty:
```suggestion
Exporting Kotlin interfaces with `@JsNoRuntime` has some restrictions. The annotation isn't allowed with:
```

sarahhaggarty (approved):
Looks good! Only very minor comments.

PR #5567, opened 2026-05-26
feat: ES2015 features for js inlining

danil-pavlov:


docs/topics/js/js-interop.md
```diff
@@ -30,12 +38,25 @@ fun jsTypeOf(o: Any): String {
...
 
-Note that invoking `js()` returns a result of type [`dynamic`](dynamic-type.md), which provides no type safety at compile time.
+    const nums = [1, 2, 3];
+    const sum = add(...nums);
+
+    const a = [1, 2, 3];
+    const b = [...a, 4, 5, 6];
+
+    return { sum, b: b };
+""")
```
seclerp: I feel like this example is too synthetic. For example, the above example for `typeof` actually makes sense to use and it highlights why `js()` in general can be useful. Can we please change this one to a shorter, concise one?

What I mean is something like:

```kotlin
val sum = js("(...nums) => nums.reduce((a, b) => a + b, 0)");
println(sum(1, 2, 3, 4))
```

Of course, it is not filled with everything, but it provides an idea, I suppose.

seclerp (changes requested):

docs/topics/js/js-interop.md
```diff
@@ -30,12 +38,18 @@ fun jsTypeOf(o: Any): String {
...
 
-> As the JavaScript code is parsed by the Kotlin compiler, not all ECMAScript features might be supported.
-> In this case, you can encounter compilation errors.
-> 
-{style="note"}
+Instead, for example for the spread operator, use:
 
-Note that invoking `js()` returns a result of type [`dynamic`](dynamic-type.md), which provides no type safety at compile time.
+```kotlin
+fun runSpreadExample() {
```
seclerp:
```suggestion
fun runSumExample() {
```

Since there is no "spread" anymore - `...nums` is a similar, but another feature called "rest parameters". But in general this impl looks more like it's about sum.

seclerp (changes requested):

docs/topics/js/js-interop.md
```diff
@@ -10,15 +10,23 @@ the surrounding tooling.
...
 ```
 
+JavaScript code inlining has full support for [ES2015 features](js-project-setup.md#support-for-es2015-features), including:
+
+* Lambdas ([arrow functions](whatsnew21.md#support-for-generating-es2015-arrow-functions))
+* ES classes
+* Template strings
+* Spread and rest operators
+* `const` and `let` variable declarations
+* Generators
```
sarahhaggarty: Nitpick: Unless there's a hidden logic to the order that I'm not seeing, how about alphabetizing the list?

docs/topics/js/js-interop.md
```diff
@@ -10,15 +10,23 @@ the surrounding tooling.
...
+
 Because the parameter of `js` is parsed at compile time and translated to JavaScript code "as-is", it is required to be
```
sarahhaggarty:
```suggestion
Because the parameter of `js` is parsed at compile time and translated to JavaScript code "as-is", it must be
```

docs/topics/js/js-interop.md
```diff
@@ -10,15 +10,23 @@ the surrounding tooling.
...
 a string constant. So, the following code is incorrect:
```
sarahhaggarty, line 31: Can you add a little explanation that it's incorrect because the string concatenation turns it into a runtime expression? It took me a couple of minutes to figure this bit out and I think it could help other readers.

docs/topics/js/js-interop.md
```diff
@@ -30,12 +38,18 @@ fun jsTypeOf(o: Any): String {
 fun getTypeof() = "typeof"
 ```
 
-> As the JavaScript code is parsed by the Kotlin compiler, not all ECMAScript features might be supported.
-> In this case, you can encounter compilation errors.
-> 
-{style="note"}
+Instead, for example for the rest operator, use:
```
sarahhaggarty:
```suggestion
Instead, you can pass strings using the rest operator. For example:
```

docs/topics/js/js-interop.md
```diff
@@ -30,12 +38,18 @@ fun jsTypeOf(o: Any): String {
...
 
-Note that invoking `js()` returns a result of type [`dynamic`](dynamic-type.md), which provides no type safety at compile time.
+```kotlin
+fun runSumExample() {
+    val sum = js("(...nums) => nums.reduce((a, b) => a + b, 0)")
+    println(sum(1, 2, 3, 4))
+}
+```
+
+> Invoking `js()` returns a result of type [`dynamic`](dynamic-type.md), which provides no type safety at compile time.
```
sarahhaggarty:
```suggestion
> Invoking `js()` returns a result of [`dynamic`](dynamic-type.md) type, which doesn't provide type safety at compile time.
```

sarahhaggarty (approved):
I made some small suggestions to try and help the text flow a bit better. Please take a look!

docs/topics/js/js-interop.md
```diff
@@ -30,12 +38,18 @@ fun jsTypeOf(o: Any): String {
...
+>
+{style="note"}
 
 ## external modifier
```
sarahhaggarty:
```suggestion
## `external` modifier
```
Since we have inline formatting in headers now!

seclerp (approved):

docs/topics/js/js-interop.md
danil-pavlov: Added a comment to the code sample

docs/topics/js/js-interop.md
danil-pavlov: I'd focus here more on inlining the mentioned features than passing strings

PR #5580, opened 2026-06-01
[Docs] Add missing dependency note to CrudRepository tutorial

ZeyadAbdullah679:
## Problem
The "Use Spring Data CrudRepository for database access" tutorial instructs
readers to import `@Id` and `@Table` without first ensuring
`spring-boot-starter-data-jdbc` is on the classpath. Without this starter,
both annotations and the CrudRepository interface are unresolvable, causing
confusing build errors with no clear indication of the cause.

## Fix
Added a note block before Step 1 with the correct `dependencies {}` block,
explaining which starter is required and why (it transitively pulls
spring-data-commons and spring-data-relational).


koshachy (commented):
Hey,

Thanks for the PR.

This is already clearly mentioned in the first steps, where it should be added to `build.gradle.kts`: [Create a Spring Boot project with Kotlin﻿](https://kotlinlang.org/docs/jvm-create-project-with-spring-boot.html#create-a-spring-boot-project).

Since this tutorial expects the previous steps to be completed first, I'd like to reject the PR.

Thank you again.

PR #5583, opened 2026-06-01
update: ksp-incremental.md

budindepunk:
[KEDOC-153](https://youtrack.jetbrains.com/issue/KEDOC-153/Update-KSP-Incremental-processing)


docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,93 @@
...
-* In an isolating Java annotation processor, all the outputs are isolating in KSP.
-* In an aggregating Java annotation processor, some outputs can be isolating and some can be aggregating in KSP.
+    **Why?** An output is made out of a set of inputs. Processors may need all the inputs to regenerate the output.
 
-### How it is implemented
+* If an unchanged input file isn't associated with any aggregating outputs, it won't be reprocessed.
 
-The dependencies are calculated by the association of input and output files, instead of annotations.
-This is a many-to-many relation.
+    **Why?** If an output file is categorized as isolating, the processor is sure that the information comes from a
```
jaschdoc: It seems a bit unclear to me how the 'why' part answers the statement. Maybe it's clearer to say something along the lines "Since it's not associated with an output and it's unchanged (meaning not new nor modified), it can not contribute to any output change."
I see that the reason is borrowed from the original document and the new version is not incorrect. Maybe you could also write it as "_If an output file is categorized as not aggregating, it must be isolating, so the processor_ is sure that information comes from a ..."

Additionally, it could be a bit more precise to update the statement to "If an unchanged input file ..., it won't be reprocessed _unless one of the rules above apply_".

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,93 @@
...
-A processor generates `outputForA` after reading class `A` in `A.kt` and class `B` in `B.kt`, where `A` extends `B`.
-The processor got `A` by `Resolver.getSymbolsWithAnnotation` and then got `B` by `KSClassDeclaration.superTypes` from `A`.
-Because the inclusion of `B` is due to `A`, `B.kt` doesn't need to be specified in `dependencies` for `outputForA`.
-You can still specify `B.kt` in this case, but it is unnecessary.
+A processor generates `outputForA` after reading class A in `A.kt` and class B in `B.kt`, where A extends B.
+
+The processor gets A by `Resolver.getSymbolsWithAnnotation` and B by `KSClassDeclaration.superTypes` from A.
+
+Because the inclusion of B is due to A, `B.kt` doesn't need to be specified in dependencies for `outputForA`. You can 
+still specify `B.kt` in this case, but it is unnecessary.
```
jaschdoc: Maybe this can be clarified a bit by saying something like "... since `B` is obtained from `A`, KSP associates `B` as a dependency of `A` via resolution tracing."

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,93 @@
...
 
 ## Example 1
 
-A processor generates `outputForA` after reading class `A` in `A.kt` and class `B` in `B.kt`, where `A` extends `B`.
-The processor got `A` by `Resolver.getSymbolsWithAnnotation` and then got `B` by `KSClassDeclaration.superTypes` from `A`.
-Because the inclusion of `B` is due to `A`, `B.kt` doesn't need to be specified in `dependencies` for `outputForA`.
-You can still specify `B.kt` in this case, but it is unnecessary.
+A processor generates `outputForA` after reading class A in `A.kt` and class B in `B.kt`, where A extends B.
+
+The processor gets A by `Resolver.getSymbolsWithAnnotation` and B by `KSClassDeclaration.superTypes` from A.
```
jaschdoc: This is just a suggestion (for both examples): The files could be listed in a `tree` view, so readers will recognize the file tree view. Additionally, the code that the processors execute could be formatted in an ordered list so it's clear what happens when. E.g.,

```
.
├── A.kt
└── B.kt
```
The processor generates `outputForA` by:

1. Obtains `A` by calling `Resolver.getSymbolsWithAnnotation`
2. Obtains `B` by calling `KSClassDeclaration.superTypes` on `A`.

jaschdoc: For the second example, the order in which files are read is a bit unclear so again an ordered list there might also help.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -102,44 +117,42 @@ class Example1Processor : SymbolProcessor {
...
-Consider that a processor generates `outputA` after reading `sourceA` and `outputB` after reading `sourceB`.
+A processor generates `outputA` after reading `sourceA` and `outputB` after reading `sourceB`.
 
 When `sourceA` is changed:
-* If `outputB` is aggregating, both `sourceA` and `sourceB` are reprocessed.
-* If `outputB` is isolating, only `sourceA` is reprocessed.
+
+* If outputB is aggregating, both sourceA and sourceB are reprocessed.
+
+* If outputB is isolating, only sourceA is reprocessed.
```
jaschdoc: Nit: should `sourceA`, `sourceB`, and `outputB` be wrapped in backticks?

budindepunk: Yes, of course. Missed it!

docs/topics/ksp/ksp-incremental.md
```diff
@@ -115,22 +111,43 @@ class Example1Processor : SymbolProcessor {
...
+.
+├── src
+│   ├── sourceA.kt
+│   └── sourceB.kt
+└── generated
+    ├── outputA
+    └── outputB
+```
+
 A processor generates `outputA` after reading `sourceA` and `outputB` after reading `sourceB`.
```
jaschdoc: I still find this sentence confusing. Could we make an ordered list here that shows exactly what happens when? E.g.,

1. Reads sourceB
2. Generates outputB
3. Reads sourceA
4. Generates outputA

docs/topics/ksp/ksp-incremental.md
```diff
@@ -115,22 +111,43 @@ class Example1Processor : SymbolProcessor {
...
 ## Example 2
 
+```none
+.
+├── src
+│   ├── sourceA.kt
+│   └── sourceB.kt
+└── generated
+    ├── outputA
+    └── outputB
```
jaschdoc: nice :)

docs/topics/ksp/ksp-incremental.md
```diff
@@ -115,22 +111,43 @@ class Example1Processor : SymbolProcessor {
...
+
 A processor generates `outputA` after reading `sourceA` and `outputB` after reading `sourceB`.
 
-When `sourceA` is changed:
+If `sourceA` changes:
+
+* If `outputB` is aggregating, KSP reprocesses both `sourceA` and `sourceB`.
 
-* If outputB is aggregating, both sourceA and sourceB are reprocessed.
+* If `outputB` is isolating, KSP reprocesses `sourceA`only.
```
jaschdoc: Missing space at the end :)

jaschdoc (commented):
One nit. Looks good otherwise! :)

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
-Incremental processing is currently enabled by default. To disable it, set the Gradle property `ksp.incremental=false`.
-To enable logs that dump the dirty set according to dependencies and outputs, use `ksp.incremental.log=true`.
-You can find these log files in the `build` output directory with a `.log` file extension.
+```
+ksp.incremental=false
+```
+
+> For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+>
+{style=”tip”}
```
zamulla: How useful is this link here? It is general information indeed, but I'm afraid it might be too general :)

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
 [//]: # (title: Incremental processing)
 
-Incremental processing is a processing technique that avoids re-processing of sources as much as possible.
-The primary goal of incremental processing is to reduce the turn-around time of a typical change-compile-test cycle.
-For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+KSP supports incremental processing. KSP reprocesses a file only when one or more of its dependencies change. This
```
zamulla:
```suggestion
KSP supports incremental processing: KSP reprocesses a file only when one or more of its dependencies change. This 
```

Since these sentences are basically the same statement, I have the urge to connect them somehow.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
 [//]: # (title: Incremental processing)
 
-Incremental processing is a processing technique that avoids re-processing of sources as much as possible.
-The primary goal of incremental processing is to reduce the turn-around time of a typical change-compile-test cycle.
-For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+KSP supports incremental processing. KSP reprocesses a file only when one or more of its dependencies change. This 
+reduces compilation time by avoiding unnecessary reprocessing.
```
zamulla: "This avoids unnecessary reprocessing and therefore reduces compilation time" sounds a bit clearer to me; verbs are better than gerunds and especially gerund chains.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
-To determine which sources are _dirty_ (those that need to be reprocessed), KSP needs processors' help to identify
-which input sources correspond to which generated outputs. To help with this often cumbersome and error-prone process,
-KSP is designed to require only a minimal set of _root sources_ that processors use as starting points to navigate the
-code structure. In other words, a processor needs to associate an output with the sources of the corresponding `KSNode`
-if the `KSNode` is obtained from any of the following:
-* `Resolver.getAllFiles`
-* `Resolver.getSymbolsWithAnnotation`
-* `Resolver.getClassDeclarationByName`
-* `Resolver.getDeclarationsFromPackage`
+Incremental processing is enabled by default. To disable it, add this line to your `gradle.properties` file:
```
zamulla: Is there a common reason to disable it? I'd add it in parentheses here, to anchor the use case.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+ksp.incremental=false
+```
+
+> For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+>
+{style=”tip”}
+
+## Dirty files
+
+A source that requires reprocessing is considered dirty. To determine which sources are dirty, KSP relies on processors
```
zamulla:
```suggestion
A source that requires reprocessing is considered _dirty_. To determine which sources are dirty, KSP relies on processors 
```

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+```
+
+> For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+>
+{style=”tip”}
+
+## Dirty files
+
+A source that requires reprocessing is considered dirty. To determine which sources are dirty, KSP relies on processors 
+to associate generated outputs with their corresponding input sources. KSP uses these associations to identify the
```
zamulla: "To do X, KSP relies on processors to do Y" can be confusing — would it be correct to say "...KSP relies on processors, which associate..."?

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+
+> For general information, see Wikipedia's article on [incremental computing](https://en.wikipedia.org/wiki/Incremental_computing).
+>
+{style=”tip”}
+
+## Dirty files
+
+A source that requires reprocessing is considered dirty. To determine which sources are dirty, KSP relies on processors 
+to associate generated outputs with their corresponding input sources. KSP uses these associations to identify the 
+sources that must be reprocessed when changes occur.
```
zamulla:
```suggestion
sources that must be reprocessed when a change occurs.
```

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+
+KSP requires only a minimal set of root sources. Processors use these sources as entry points to navigate the code structure.
```
zamulla, line 22: The concept of root sources and the mention of `KSNode` below doesn't paint a picture for me, although I'm sure it should. The general idea of "processors match outputs and sources → KSP uses this matching to find sources to be reprocessed" is clear; what role do root sources play here? What is `KSNode` and how is it related to root sources?

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+
+When a processor generates an output, it must associate that output with the sources of the corresponding `KSNode`. This 
+association is required if the `KSNode` is obtained from any of the following methods:
```
zamulla: I feel that these are conflicting statements: either a processor must always associate an output with sources, or the association is only required if a condition is met. Am I missing something?

docs/topics/ksp/ksp-incremental.md
```diff
@@ -100,46 +111,73 @@ class Example1Processor : SymbolProcessor {
...
+
+If either `sourceA` or `sourceB` is removed, KSP doesn't need to reprocess any files.
 
-Note that both of them are transitive and the second forms equivalence classes.
 
 ## Reporting bugs
 
-To report a bug, please set Gradle properties `ksp.incremental=true` and `ksp.incremental.log=true`, and perform a clean build.
-This build produces two log files:
+If you encounter any error that only occurs in incremental mode, report it by creating an issue in GitHub. Follow the
```
zamulla: There should probably be a link to the repo where the issue should be created.

zamulla:
```suggestion
If you encounter any error that only occurs in incremental mode, report it by creating an issue in GitHub.
```

docs/topics/ksp/ksp-incremental.md
```diff
@@ -100,46 +111,73 @@ class Example1Processor : SymbolProcessor {
...
+If either `sourceA` or `sourceB` is removed, KSP doesn't need to reprocess any files.
 
-Note that both of them are transitive and the second forms equivalence classes.
 
 ## Reporting bugs
 
-To report a bug, please set Gradle properties `ksp.incremental=true` and `ksp.incremental.log=true`, and perform a clean build.
-This build produces two log files:
+If you encounter any error that only occurs in incremental mode, report it by creating an issue in GitHub. Follow the 
+following steps to retrieve the relevant logs:
```
zamulla:
```suggestion
Attach the relevant logs to the issue:
```

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+|-------------------|---------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------|
+| Aggregating       | Can be affected by changes in any source file, except for removals that do not impact other files | Any input change triggers a rebuild of all aggregating outputs and reprocessing of all corresponding registered, new, or modified source files | An output that collects all symbols with a particular annotation  |
+| Isolating         | Depends only on its specified sources                                                             | Changes to other sources do not affect the output. Multiple source files can be associated with a single output                                | A generated class that is dedicated to an interface it implements |
```
zamulla: In the current kotlinlang.org frontend, the table is almost unusable — maybe the rows should become columns, or columns should have fixed width. That's one of the reasons we generally use `<table>`, not Markdown tables.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+
+> Unlike Gradle annotation processing, KSP applies the categorization to individual outputs
+> rather than entire processors.
+>
+{style=”note”}
```
zamulla: Maybe this note can be after the first paragraph here? The assumption that the categorization works the same should probably be dealt with as soon as it appears.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
-Otherwise, the output is isolating.
+## Dirtiness propagation
 
-Here's a summary for readers familiar with Java annotation processing:
-* In an isolating Java annotation processor, all the outputs are isolating in KSP.
-* In an aggregating Java annotation processor, some outputs can be isolating and some can be aggregating in KSP.
+A file is considered dirty if it is either directly modified by the user or indirectly affected by changes in other dirty files.
 
-### How it is implemented
+KSP propagates dirtiness in two steps:
```
zamulla: If these are sequential, the following list should probably be numbered.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -100,46 +111,73 @@ class Example1Processor : SymbolProcessor {
...
-When `sourceC` is added:
-* If `outputB` is aggregating, both `sourceC` and `sourceB` are reprocessed.
-* If `outputB` is isolating, only `sourceC` is reprocessed.
+3. reads `sourceA`.
 
-When `sourceA` is removed, nothing needs to be reprocessed.
+4. generates `outputA`.
 
-When `sourceB` is removed, nothing needs to be reprocessed.
+If `sourceA` changes:
```
zamulla:
```suggestion
When `sourceA` changes:
```

docs/topics/ksp/ksp-incremental.md
```diff
@@ -100,46 +111,73 @@ class Example1Processor : SymbolProcessor {
...
+│   ├── sourceA.kt
+│   └── sourceB.kt
+└── generated
+    ├── outputA
+    └── outputB
+```
+
+A processor:
+
+1. reads `sourceB`.
```
zamulla: Is there a meaning behind reading B before A? If not, doing the default order is going to be less prone to questions like that )

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,89 @@
...
+shared outputs.
 
-Reasons are:
-1. If an input is changed, new information can be introduced and therefore processors need to run again with the input.
-2. An output is made out of a set of inputs. Processors may need all the inputs to regenerate the output.
-3. `aggregating=true` means that an output may potentially depend on new information, which can come from either new
-   files, or changed, existing files.
-   `aggregating=false` means that processor is sure that the information only comes from certain input files and never
-   from other or new files.
+Note that both of these steps will be repeated until there are no dirty files.
```
zamulla: I'd add this to the intro paragraph: "KSP propagates dirtiness in two steps, which are repeated until there are no dirty files."

zamulla (changes requested):
I left some suggestions for clarity, and I suggest reviewing the section structure: I think we can group and reorder some sections to better outline the page.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,140 @@
...
-To enable logs that dump the dirty set according to dependencies and outputs, use `ksp.incremental.log=true`.
-You can find these log files in the `build` output directory with a `.log` file extension.
+```
+ksp.incremental=false
+```
+
+## Dirty files
+
+A file is considered _dirty_ if it is either directly modified by the user or indirectly affected by changes in other
+dirty files. A dirty source requires reprocessing.
```
zamulla: I think "requires reprocessing" is a part of definition of a dirty source rather than a consequence? This feels closer, but feel free to correct me: "A file is considered _dirty_ (needing to be reprocessed) if..."

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,140 @@
...
+
+When a processor generates an output, it must associate that output with the sources of the corresponding `KSNode`. This 
+association is required if the `KSNode` is obtained from any of the following methods:
+
+* `Resolver.getAllFiles()`
+* `Resolver.getSymbolsWithAnnotation()`
+* `Resolver.getClassDeclarationByName()`
+* `Resolver.getDeclarationsFromPackage()`
+
+> To learn more about the `KSNode` API, see the diagram in [How KSP models Kotlin code](https://kotlinlang.org/docs/ksp-additional-details.html).
```
zamulla: I would introduce this link before KSNode: those who read it already would at least understand where we are, those who haven't can, like me, panic at the sight of `KSNode` and start googling before the paragraph ends.

Also this link should not be absolute, right? If it's a topic in the same repo.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,140 @@
...
+</def>
+</deflist>
 
-## Aggregating vs Isolating
+### Dirtiness propagation
 
-Similar to the concepts in [Gradle annotation processing](https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing),
-KSP supports both _aggregating_ and _isolating_ modes. Note that unlike Gradle annotation processing, KSP categorizes
-each output as either aggregating or isolating, rather than the entire processor.
+KSP propagates dirtiness in two steps, which are repeated until there are no dirty files:
```
zamulla: Suddenly I'm thinking that "propagating dirtiness until there are no dirty files" doesn't make sense. Does it? Dirty files are going to be reprocessed, that's how we get rid of dirty files; until then we just need to know which ones are dirty, and for that we monitor changes + the propagation mechanism?

zamulla (changes requested):
I have a couple more questions, but in general page is in satisfying shape, I think!

docs/topics/ksp/ksp-incremental.md
budindepunk: In the end, the dev and I rewrote this section a bit for clarity and removed the mention of KSNode altogether.

docs/topics/ksp/ksp-incremental.md
budindepunk: Yes, you are right. The two things can trigger each other, though, and there can be some back and forth before it's over. That's where the idea of repetition "until there are no new dirty files" came from. More accurate would be to say "until no new files are marked as dirty". I'll rephrase.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,153 @@
...
+or modified source files.
+
+For example, an output that collects all symbols with a particular annotation is aggregating.
+
+</def>
+<def title="Isolating">
+
+Isolating outputs depend only on their specified sources.
+
+Changes to other sources do not affect the output. Multiple source files can be associated with a single output
```
hfmehmed: missing full stop

docs/topics/ksp/ksp-incremental.md
```diff
@@ -84,55 +159,44 @@ open class B
...
+   `gradle.properties`:
 
-## Reporting bugs
+  ```
+  ksp.incremental.log=true
+   ```
 
-To report a bug, please set Gradle properties `ksp.incremental=true` and `ksp.incremental.log=true`, and perform a clean build.
-This build produces two log files:
+2. Perform a clean build. This build produces two log files in the build output directory with a `.log` file extension:
```
hfmehmed: if the user is reporting a bug related to incrementality - performing a clean build will actually fix the issue.

hfmehmed: Maybe we can say 

2. Perform a clean build to initialize the caches. 
3. Make a change to a source file that triggers the issue, and run the build again. 
4. Collect the log files produced in the build output directory

jaschdoc: Also, I'm not sure the last two logs are generated. There is no code path at least that reaches these names:
`build/kspCaches/<source set>/logs/kspDirtySetByDeps.log`
`build/kspCaches/<source set>/logs/kspDirtySetByOutputs.log`.
The `kspDirtySet.log` and `kspSourceToOutputs.log` are generated though.
For error reporting, they should have the logs from the successful build as well as the failed build so we can check what the diff is. So the idea is that they should copy the generated logs from the successful build somewhere else first, then perform the change and run the incremental build which regenerates the logs which can then be copied somewhere else. That should in total yield 4 files.
Just wanted to point it out and clear up the mismatch between the code and the documentation :)

budindepunk: Could you tell me what the step by step would be then? Something like this?

1. Perform a clean build (successful)
2. save the logs somewhere
3. modify a source file, triggering the issue
4. run the build again (failed)
5. collect the logs

budindepunk: And this part should not be there at all:

```
You can then run successive incremental builds, which will generate two additional log files:

* `build/kspCaches/<source set>/logs/kspDirtySetByDeps.log`
* `build/kspCaches/<source set>/logs/kspDirtySetByOutputs.log`
```

Correct?

docs/topics/ksp/ksp-incremental.md
```diff
@@ -65,16 +73,21 @@ For example, a generated class that is dedicated to an interface it implements i
 
 ### Dirtiness propagation
 
-KSP propagates dirtiness in two steps, which are repeated until there are no dirty files:
+KSP propagates dirtiness in two ways:
```
zamulla: I'm allergic to hardcoding the number of things in a list, so I would just write "in the following ways".

docs/topics/ksp/ksp-incremental.md
```diff
@@ -65,16 +73,21 @@ For example, a generated class that is dedicated to an interface it implements i
...
 
-* Propagation by **resolution tracing**: Type resolution is the only way to traverse from one file to another. When a
-  processor resolves a type reference (explicitly or implicitly), KSP considers dependencies between the file containing
-  the reference and any file that defines a symbol affecting that resolution. As a result, a change in a resolved symbol
-  may mark the referencing file as dirty.
+1. by **resolution tracing**: Type resolution is the only way to traverse from one file to another. When a
```
zamulla: Lower case looks strange here, let's capitalize the beginning of the sentence.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -84,16 +97,16 @@ This is how KSP determines which files need to be reprocessed:
 
 * If an input file is changed, it will always be reprocessed.
 
-    **Why?** If an input is changed, new information can be introduced. Processors need to run again with the input.
+**Why?** If an input is changed, new information can be introduced. Processors need to run again with the input.
```
zamulla: Not sure why the indent was removed — now a comment to a list item is on a higher level than the thing it's commenting on, and visually it's hard to track their relationship.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -168,29 +181,32 @@ To generate `outputForA`, the processor:
...
+If you encounter any error that occurs only when incremental processing is enabled, report it by creating an issue in
 [the GitHub repository](https://github.com/google/ksp/issues). Attach the relevant logs to the issue:
 
-1. Enable logs that dump the dirty set according to dependencies and outputs. To do this, add this line to 
-`gradle.properties`: 
+1. Enable incremental processing logs by adding the following line to `gradle.properties`:
+
+  ```
+  ksp.incremental.log=true
+   ```
```
zamulla, line 195: The indentation is a little mismatched here.

zamulla (approved):
LGTM! A couple of superficial fixes suggested.

docs/topics/ksp/ksp-incremental.md
jaschdoc: > And this part should not be there at all:
> 
> ```
> You can then run successive incremental builds, which will generate two additional log files:
> 
> * `build/kspCaches/<source set>/logs/kspDirtySetByDeps.log`
> * `build/kspCaches/<source set>/logs/kspDirtySetByOutputs.log`
> ```
> 
> Correct?

Yes, exactly.

jaschdoc: > Could you tell me what the step by step would be then? Something like this?
> 
> 1. Perform a clean build (successful)
> 2. save the logs somewhere
> 3. modify a source file, triggering the issue
> 4. run the build again (failed)
> 5. collect the logs

Yes that's correct :)

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,153 @@
...
-You can find these log files in the `build` output directory with a `.log` file extension.
+```
+ksp.incremental=false
+```
+
+You might want to disable incremental processing when troubleshooting or to force a full rebuild.
+
+## Dirty files
+
+A file is considered _dirty_ (needing to be reprocessed) if it is either directly modified by the user or indirectly
```
koshachy: A "user" here is the developer who reads this doc? 
In this case I'd like to make it more obvious.

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,153 @@
...
+Incremental processing is enabled by default. To disable it, add this line to your `gradle.properties` file:
 
-Incremental processing is currently enabled by default. To disable it, set the Gradle property `ksp.incremental=false`.
-To enable logs that dump the dirty set according to dependencies and outputs, use `ksp.incremental.log=true`.
-You can find these log files in the `build` output directory with a `.log` file extension.
+```
+ksp.incremental=false
+```
+
+You might want to disable incremental processing when troubleshooting or to force a full rebuild.
```
koshachy: I believe that this line should be somewhere (mb even rephrased) before the instruction, how to disable the IC

docs/topics/ksp/ksp-incremental.md
koshachy: BTW, we avoid saying "User" when we mean "A developer"

koshachy (changes requested):

docs/topics/ksp/ksp-incremental.md
```diff
@@ -1,78 +1,152 @@
...
-which input sources correspond to which generated outputs. To help with this often cumbersome and error-prone process,
-KSP is designed to require only a minimal set of _root sources_ that processors use as starting points to navigate the
-code structure. In other words, a processor needs to associate an output with the sources of the corresponding `KSNode`
-if the `KSNode` is obtained from any of the following:
-* `Resolver.getAllFiles`
-* `Resolver.getSymbolsWithAnnotation`
-* `Resolver.getClassDeclarationByName`
-* `Resolver.getDeclarationsFromPackage`
+Incremental processing is enabled by default. Disable it when troubleshooting or forcing a full rebuild. To disable it, 
+add this line to your `gradle.properties` file:
```
koshachy: What do you think about the following option, does it work?

> Incremental processing is enabled by default. You can disable it when troubleshooting or when you need to force a full rebuild. To disable it, add the following line to your gradle.properties file:

koshachy (approved):

docs/topics/ksp/ksp-incremental.md
budindepunk: Yes, better. On it!

PR #5584, opened 2026-06-01
update: ksp-multiplatform.md

budindepunk:
[KEDOC-134](https://youtrack.jetbrains.com/issue/KEDOC-134/Update-KSP-for-Multiplatform)


jaschdoc (approved):
@hfmehmed, any comments on this? :)

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,236 @@
...
+    // KSP configurations are per-compilation target. Shared intermediate 
+    // source sets like 'iosMain' do NOT have their own KSP configuration.
+    // You must specify them for each iOS target individually.
+    add("kspIosX64", project(":test-processor"))
+    add("kspIosArm64", project(":test-processor"))
+    add("kspIosSimulatorArm64", project(":test-processor"))
+    
+    // TIP: If you have many iOS targets, you can avoid repetition by looping:
+    kotlin.targets.filter { it.name.startsWith("ios") }.forEach { target ->
+        add("ksp${target.name.capitalize()}", project(":test-processor"))
```
hfmehmed: if my memory serves well capitalize if deprecated api in kotlin language since  few years. use `replaceFirstChar { it.upperCase() }` instead. Check if what i suggested is actually correct as i didnt test it

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,236 @@
...
+This document explains how to use Kotlin Symbol Processing (KSP) in a Kotlin Multiplatform project. To get started with 
+Kotlin Multiplatform, see the [Kotlin Multiplatform overview](https://kotlinlang.org/docs/multiplatform/kmp-overview.html).
 
-Starting from KSP 1.0.1, applying KSP on a multiplatform project is similar to that on a single platform, JVM project.
-The main difference is that, instead of writing the `ksp(...)` configuration in dependencies, `add(ksp<Target>)` or `add(ksp<SourceSet>)`
-is used to specify which compilation targets need symbol processing, before compilation.
+To use KSP-based processors in a multiplatform project, assign a processor to each target that needs symbol processing:
+
+```
+add(ksp<Target>, <processor>)
```
hfmehmed: can you put in quotes like this 

`"ksp<Target>"` for people who copy paste this into their build file and replace manually. The quotes are needed in the build file

hfmehmed (commented):
Apart from the 2 comments i left, all good. 🟢

docs/topics/ksp/ksp-multiplatform.md
budindepunk: Thank you! Could you let me know if this snippet is correct?

```
kotlin.targets.filter { it.name.startsWith("ios") }.forEach { target ->
    add(
        "ksp${target.name.replaceFirstChar { it.uppercaseChar() }}",
        project(":test-processor")
    )
}
```

hfmehmed: yes, thank you - that looks correct to me

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,239 @@
...
+To use KSP-based processors in a multiplatform project, assign a processor to each target that needs symbol processing:
+
+```
+add("ksp<Target>", <processor>)
+```
+
+`<processor>` is a Gradle project path. It can be:
+
+* the specific folder in your project that contains the logic for your symbol processor.
+* an external processor such as Room or Dagger.
```
zsmb13: Dagger isn't multiplatform, maybe mention just Room

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,239 @@
...
+For example:
+
+<tabs group="build-script">
+<tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
+// build.gradle.kts
+
+// This example demonstrates various ways to declare KSP dependencies
+// in a Kotlin project, covering multiplatform setups.
```
zsmb13: I feel like the comments in this snippet are a bit too heavy, do we usually do this? These two lines for example don't add much extra information, and it feels like just body that could be outside of the snippet (if anywhere).

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,239 @@
...
-        binaries {
-            executable()
-        }
+dependencies {
+    // ===============================================================
+    // 1. MAIN COMPILATIONS (Production Code)
+    // ===============================================================
+
+    // In KMP, the global 'ksp' configuration is deprecated to avoid running 
+    // processors on targets unnecessarily. Use target-specific configurations instead.
```
zsmb13: These comments for sections and their explanations could also be just body text in the doc in my opinion

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,239 @@
...
+```
+
+Look for configurations matching your target names to find the right one.
+
+Since KSP 2 the catch-all `ksp(...)` configuration has been deprecated. Although this may lead to some duplication in 
+the build file, specifying each platform individually avoids the cost of applying target-specific processors to targets 
+in which they do nothing.
+
+> To use the deprecated ksp(...) configuration, set the flag `ksp.allow.all.target.configuration=true`
+> in gradle.properties.
```
zsmb13: I'm wondering if this is only for migration purposes, if it will be removed in the future, etc. Could use a little more context if we have it.

budindepunk: Yes. After talking it over with the devs, we agreed that this information is not very relevant to new users. The previous paragraph gives some context for people who were using KSP1 before, so it's probably worh keeping.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
 [//]: # (title: KSP with Kotlin Multiplatform)
+[//]: # (description: Add KSP to a Kotlin multiplatform project)
 
-For a quick start, see a [sample Kotlin Multiplatform project](https://github.com/google/ksp/tree/main/examples/multiplatform) 
-defining a KSP processor.
+This document explains how to use Kotlin Symbol Processing (KSP) in a Kotlin Multiplatform project. To get started with
```
zamulla: "This document" sounds a little formal to me, and we don't really define "document". Maybe something like "Here you can learn ..."?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
+
+    // iOS Targets
+    add('kspIosX64', project(':test-processor'))
+    add('kspIosArm64', project(':test-processor'))
+    add('kspIosSimulatorArm64', project(':test-processor'))
+
+    // TIP: If you have many iOS targets, you can avoid repetition by looping:
+    kotlin.targets.findAll { it.name.startsWith("ios") }.each { target ->
+        add(
+                "ksp${target.name.replaceFirstChar { it.toUpperCase() }}",
```
zamulla:
```suggestion
            "ksp${target.name.replaceFirstChar { it.toUpperCase() }}",
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
+                project(":test-processor")
```
zamulla:
```suggestion
            project(":test-processor")
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
 
-Before KSP 1.0.1, there is only one, unified `ksp(...)` configuration available. Therefore, processors either applies to all
-compilation targets, or nothing at all. Note that the `ksp(...)` configuration not only applies to the main source set, but also
-the test source set if it exists, even on traditional, non-multiplatform projects. This brought unnecessary overheads to build time.
+* `kspJvm` and `kspJvmTest`
+* `kspJs` and `kspJsTest`
+* `kspAndroidNativeX64` and `kspAndroidNativeX64Test`
+* `kspAndroidNativeArm64` and `kspAndroidNativeArm64Test`
+* `kspLinuxX64` (test is omitted/commented out)
+* `kspMingwX64` (test is omitted/commented out)
```
zamulla: Slash is hard to interpret here. I see that other test configurations are included, why not these two?
The fact that they are commented out and not simply left out obviously has meaning, but not sure which one.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
+* Android Native Arm64
+* Mingw X64
 
-For example, in the above `build.gradle.kts`, there are 4 compilation tasks: common/metadata, JVM main, Linux x64 main, Linux x64 test,
-and 3 symbol processing tasks: common/metadata, JVM main, Linux x64 test.
+Each of these targets includes a `main` and a `test` compilation. As a result, the project creates at least 12 Kotlin 
+compilation tasks.
 
-## Avoid the ksp(...) configuration on KSP 1.0.1+
+In the example's `workload/build.gradle.kts`, KSP dependencies are declared for the following configurations:
```
zamulla:
```suggestion
In the example's `workload/build.gradle.kts` file, KSP dependencies are declared for the following configurations:
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
+* `kspAndroidNativeX64` and `kspAndroidNativeX64Test`
+* `kspAndroidNativeArm64` and `kspAndroidNativeArm64Test`
+* `kspLinuxX64` (test is omitted/commented out)
+* `kspMingwX64` (test is omitted/commented out)
 
-Starting from KSP 1.0.1, per-target configurations are provided as shown in the above example. In the future:
-1. For multiplatform projects, the `ksp(...)` configuration will be deprecated and removed.
-2. For single platform projects, the `ksp(...)` configuration will only apply to the main, default compilation. 
-   Other targets like `test` will need to specify `kspTest(...)` in order to apply processors.
+KSP creates one symbol processing task for each configuration where a KSP dependency is declared. In this example, the
```
zamulla:
```suggestion
KSP creates a symbol processing task for each configuration where a KSP dependency is declared. In this example, the 
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
+`test`. For every eligible Kotlin compilation task, KSP creates a corresponding symbol processing task.
+
+The [example project](https://github.com/google/ksp/tree/main/examples/multiplatform) defines six targets:
+
+* JVM
+* JS
+* Linux X64
+* Android Native X64
+* Android Native Arm64
+* Mingw X64
```
zamulla: It's a little strange to see that only some of them actually have a corresponding source set. Adding a `jvmMain` would allow to quickly run the project on any machine and see the output.

I guess my problem is that it's not clear how to see that (or how) the example works: maybe we should explicitly say what the processor is supposed to do? It's not that hard to read the code and figure it out, but could be easier.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
-is used to specify which compilation targets need symbol processing, before compilation.
+To use KSP-based processors in a multiplatform project, assign a processor to each target that needs symbol processing:
+
+```
+add("ksp<Target>", <processor>)
+```
+
+`<processor>` is a Gradle project path. It can be:
+
+* the specific folder in your project that contains the logic for your symbol processor.
```
zamulla:
```suggestion
* a specific directory in your project that contains the logic for your symbol processor.
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
 ```kotlin
+// build.gradle.kts
+
 plugins {
-    kotlin("multiplatform")
-    id("com.google.devtools.ksp")
+    // Apply the KSP plugin
+    id("com.google.devtools.ksp") version "%kspVersion%"     
+    // Apply other plugins as needed (e.g., kotlin("jvm"), 
+    // kotlin("multiplatform"), etc.)
```
zamulla: The KMP plugin is probably mandatory in this example?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,183 @@
...
-            executable()
-        }
+dependencies {
+
+    // Targets with multiple processors
+    add("kspJvm", project(":local-processor"))
+    add("kspJvm", "com.example:external-processor:1.0")
+
+    add("kspAndroid", project(":test-processor"))
+    add("kspAndroid", "androidx.room:room-compiler:2.6.1")
```
zamulla: The Room processor also needs a library dependency and a Gradle plugin, right? Should we mention this, or do people aiming to use Room as a processor obviously have it in their project already?

zamulla (commented):
Thanks! There's a couple of questions I had that may need answering in the doc.

docs/topics/ksp/ksp-multiplatform.md
hfmehmed: +1

docs/topics/ksp/ksp-multiplatform.md
hfmehmed: That's right. Althought i think this is just an example of how to add an external (maven) depdency as a processor to ksp

docs/topics/ksp/ksp-multiplatform.md
budindepunk: @hfmehmed What do you think about this? This is referring to the example in the [ksp repo](https://github.com/google/ksp/tree/main/examples/multiplatform). Is it relevant to keep these two things in the list?

hfmehmed: I think we can drop them from this doc. They were probably added in the ksp repo project just to serve as further example of other configurations but they are not really used in that project.

docs/topics/ksp/ksp-multiplatform.md
hfmehmed: That's a fair point. I am not sure why we dont have jvmMain/jvmTest. But regardless of the existance of the source set, jvm target is enabled and will generate an output artifact (a jar). It will inherit everything from commonMain but it will not add anything on top. (apart from the ksp processor generated classes)

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,200 @@
 [//]: # (title: KSP with Kotlin Multiplatform)
+[//]: # (description: Add KSP to a Kotlin multiplatform project)
 
-For a quick start, see a [sample Kotlin Multiplatform project](https://github.com/google/ksp/tree/main/examples/multiplatform) 
-defining a KSP processor.
+Here you can learn how to use Kotlin Symbol Processing (KSP) in a Kotlin Multiplatform project. To get started with 
+Kotlin Multiplatform, see the [Kotlin Multiplatform overview](https://kotlinlang.org/docs/multiplatform/kmp-overview.html).
```
zamulla: I've thought about it, and especially if we add the link to the sample below, this link doesn't make a lot of sense to me. People who arrive on this page surely have encountered KMP already, to introduce KSP into a project.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,200 @@
...
+* an external processor such as Room:
+
+    ```
+    add("kspJvm", "androidx.room:room-compiler:2.6.1")
+    ```
+
+> For a full list of targets, see [Multiplatform Gradle DSL reference](https://kotlinlang.org/docs/multiplatform/multiplatform-dsl-reference.html#targets) 
+> and [Kotlin/Native supported targets](https://kotlinlang.org/docs/native-target-support.html).
+>
+{style=”tip”}
```
zamulla: Didn't register the first time, but I think we're losing consistency here: <processor> has its options written out in the body of the text, while <Target> is confined to an admonition. I'd present them in the same way, if not in the same list.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,200 @@
...
+> processors where they are not needed.
+> 
+{style="note"}
+
 ## Compilation and processing
 
-In a multiplatform project, Kotlin compilation may happen multiple times (`main`, `test`, or other build flavors) for each platform.
-So is symbol processing. A symbol processing task is created whenever there is a Kotlin compilation task and a
-corresponding `ksp<Target>` or `ksp<SourceSet>` configuration is specified.
+In a multiplatform project, Kotlin creates a separate compilation for each target and source set, such as `main` and
```
zamulla:
```suggestion
In a multiplatform project, Kotlin creates a separate [compilation](https://kotlinlang.org/docs/multiplatform/multiplatform-advanced-project-structure.html#compilations) for each target and source set, such as `main` and 
```

zamulla (commented):
Generally LGTM, added a couple of polishing suggestions.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+## Add KSP to a multiplatform project
+
+In the client module's `build.gradle.kts` file, assign a processor to each target that needs symbol processing in the 
+`dependencies {}` block:
+
+```
+add("ksp<Target>", <processor>)
+```
+
+`ksp<Target>` is one of the targets used in your multiplatform project. For a full list of targets, see
```
zamulla:
```suggestion
`<Target>` is one of the targets used in your multiplatform project. For a full list of targets, see 
```

`ksp<Target>` is not a target, right, it's a KSP configuration?

zamulla (approved):
Awesome, thanks! One final nit but it's done 👌

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
 [//]: # (title: KSP with Kotlin Multiplatform)
+[//]: # (description: Add KSP to a Kotlin multiplatform project)
 
-For a quick start, see a [sample Kotlin Multiplatform project](https://github.com/google/ksp/tree/main/examples/multiplatform) 
-defining a KSP processor.
+Here you can learn how to use Kotlin Symbol Processing (KSP) in a Kotlin Multiplatform project. For a quick start, see
```
koshachy:
```suggestion
Here you'll learn how to use Kotlin Symbol Processing (KSP) in a Kotlin Multiplatform project. For a quick start, see 
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+an example of a multiplatform project with several targets using KSP in the 
+[source repository](https://github.com/google/ksp/tree/main/examples/multiplatform). The processor in this example
+generates a `Foo` class used by the project.
 
-Starting from KSP 1.0.1, applying KSP on a multiplatform project is similar to that on a single platform, JVM project.
-The main difference is that, instead of writing the `ksp(...)` configuration in dependencies, `add(ksp<Target>)` or `add(ksp<SourceSet>)`
-is used to specify which compilation targets need symbol processing, before compilation.
+## Add KSP to a multiplatform project
+
+In the client module's `build.gradle.kts` file, assign a processor to each target that needs symbol processing in the
```
koshachy: What is a **client** module? 🤔

koshachy: I'm not sure that we use this term anywhere, maybe I'm wrong?

koshachy: Why do we use "assign" as a word here? Isn't it just adding a dependency?
Have we ever used "assign" to explain something in the KSP?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+## Add KSP to a multiplatform project
+
+In the client module's `build.gradle.kts` file, assign a processor to each target that needs symbol processing in the 
+`dependencies {}` block:
+
+```
+add("ksp<Target>", <processor>)
+```
+
+`<Target>` is one of the targets used in your multiplatform project. For a full list of targets, see
```
koshachy: For a full list of target — could be a tip, no?
It is not that important rn since we expect that person knows about targets already

koshachy: You can place both Target and Processor in a list, so the info about them would be structured.

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+    add("kspJvm", project(":local-processor"))
+    ```
+
+* an external processor such as Room:
+
+    ```
+    add("kspJvm", "androidx.room:room-compiler:2.6.1")
+    ```
+
+A single target can have multiple processors:
```
koshachy: Don't you want to put these different build set ups into separte subsections?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+add('kspAndroidHostTest', project(':test-processor'))
+add('kspAndroidDeviceTest', project(':test-processor'))
+```
+
+</tab>
+</tabs>
+
+## Find KSP configuration names
+
+KSP configuration names depend on the source set names generated by the Kotlin plugin. To view the complete
```
koshachy: What is the Kotlin plugin? Is it KGP (Kotlin Gradle Plugin)? Or Kotlin IntelliJ Plugin? Or Kotlin Multiplatfrom Plugin?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+KSP configuration names depend on the source set names generated by the Kotlin plugin. To view the complete 
+list of KSP configurations for a module, run:
+
+```Bash
+./gradlew :<your-module-name>:dependencies | grep ksp
+```
+
+Look for the configuration names that correspond to your target source sets.
+
+> Since KSP 2, the catch-all `ksp(...)` configuration is deprecated. Configure each target explicitly to avoid running
```
koshachy: WDYT about making this one a warning instead of note? Will it work?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+> 
+{style="note"}
+
 ## Compilation and processing
 
-In a multiplatform project, Kotlin compilation may happen multiple times (`main`, `test`, or other build flavors) for each platform.
-So is symbol processing. A symbol processing task is created whenever there is a Kotlin compilation task and a
-corresponding `ksp<Target>` or `ksp<SourceSet>` configuration is specified.
+In a multiplatform project, Kotlin creates a separate [compilation](https://kotlinlang.org/docs/multiplatform/multiplatform-advanced-project-structure.html#compilations) 
+for each target and source set, such as `main` and`test`. For every eligible Kotlin compilation task, KSP creates a
```
koshachy: What is **eligible** Kotlin compilation task?

koshachy: Can we just omit it?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
+for each target and source set, such as `main` and`test`. For every eligible Kotlin compilation task, KSP creates a 
+corresponding symbol processing task.
+
+The [example project](https://github.com/google/ksp/tree/main/examples/multiplatform) defines six targets:
 
-For example, in the above `build.gradle.kts`, there are 4 compilation tasks: common/metadata, JVM main, Linux x64 main, Linux x64 test,
-and 3 symbol processing tasks: common/metadata, JVM main, Linux x64 test.
+* JVM
+* JS
+* Linux X64
```
koshachy: Shouldn't we write the target names as one word?

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,198 @@
...
-corresponding `ksp<Target>` or `ksp<SourceSet>` configuration is specified.
+In a multiplatform project, Kotlin creates a separate [compilation](https://kotlinlang.org/docs/multiplatform/multiplatform-advanced-project-structure.html#compilations) 
+for each target and source set, such as `main` and`test`. For every eligible Kotlin compilation task, KSP creates a 
+corresponding symbol processing task.
+
+The [example project](https://github.com/google/ksp/tree/main/examples/multiplatform) defines six targets:
 
-For example, in the above `build.gradle.kts`, there are 4 compilation tasks: common/metadata, JVM main, Linux x64 main, Linux x64 test,
-and 3 symbol processing tasks: common/metadata, JVM main, Linux x64 test.
+* JVM
```
koshachy: What do you think to move the info about main and test before this list, maybe rephrase, and in the list mention it in the way:
* For the JVM target `jvmMain` and `jvmTest`
* ...

koshachy (changes requested):

docs/topics/ksp/ksp-multiplatform.md
budindepunk: we use "client module" also in the KSP quickstart, it is the module that uses the processor. Maybe here it can be explained

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,214 @@
...
 
-Starting from KSP 1.0.1, applying KSP on a multiplatform project is similar to that on a single platform, JVM project.
-The main difference is that, instead of writing the `ksp(...)` configuration in dependencies, `add(ksp<Target>)` or `add(ksp<SourceSet>)`
-is used to specify which compilation targets need symbol processing, before compilation.
+## Add KSP to a multiplatform project
+
+In the `build.gradle.kts` file of the client module (the module that uses the processor), add the appropriate KSP 
+processor dependency for each target that requires symbol processing.
+
+`dependencies {}` block:
```
koshachy: Why it is not a part of the prevoius sentence?

koshachy: We also could provide the whole block:

```kotlin
dependencies {
    .....
}
```

docs/topics/ksp/ksp-multiplatform.md
```diff
@@ -1,56 +1,214 @@
...
+
+```
+add("ksp<Target>", <processor>)
+```
+
+Where:
```
koshachy: No need to put this word. 
Just proceed with the list of things after the codeblock.

koshachy (approved):
LGTM!
Please, fix some small things and we are good to go!

PR #5585, opened 2026-06-01
feat: runtime library features for Power-assert

danil-pavlov:


docs/topics/compiler-plugins/power-assert.md
```diff
@@ -372,6 +376,48 @@ powerAssert {
...
+           Mascot(name=Unknown)
+    require(subject is Mascot)
+    check(subject.name == "Kodee")
+                  |    |
+                  |    false
+                  "Unknown"
+}
+```
+
+No additional build configuration is needed for functions annotated with `@PowerAssert`.
```
bnorm: This seems like a repeat of the first paragraph in this section.

docs/topics/compiler-plugins/power-assert.md
bnorm, line 19: This output will be different in 2.4. Should update it with the new version. Should look something similar to this:

```text
Incorrect length
assert(hello.length == world.substring(1, 4).length) { "Incorrect length" }
       |     |      |  |     |               |
       |     5      |  |     "orl"           3
       "Hello"      |  "world!"
                    false
```

bnorm, line 435: Again, this output will be different in 2.4.

bnorm, line 464: This output will be different 2.4.

bnorm: Basically, you should double-check all the example outputs, as they will likely be different in 2.4.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+## Support the Power-assert plugin
+
+If you are a library author, you can add out-of-the-box Power-assert support to your library using the `@PowerAssert`
+annotation and the `CallExplanation` class from the Power-assert runtime library.
+
+### The @PowerAssert annotation
+
+The [`@PowerAssert` annotation](https://github.com/JetBrains/kotlin/blob/master/plugins/power-assert/power-assert-runtime/src/commonMain/kotlin/kotlin/powerassert/PowerAssert.kt)
+marks a function as Power-assert capable. If users of your library have the Power-assert compiler plugin in their projects
+and make calls to your annotated functions they calls automatically transformed without additional build configuration.
```
bnorm: > your annotated functions they calls automatically transformed

"they" doesn't make sense. I'm not sure if this is a mistype, a missing word, or what was meant to be said here.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+
+To add support for Power-assert to your library:
+
+1. Add the Power-assert runtime library as a dependency:
```
bnorm: This is only required for maven. The dependency is automatically added with the Gradle plugin for power-assert.

This should also be step 2, and step 1 should be adding the compiler plugin to the build. The compiler plugin is required on both the project which defines the `@PowerAssert` function and the project which uses the function.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+   import kotlin.powerassert.toDefaultMessage
+   import kotlin.contracts.ExperimentalContracts
+   import kotlin.contracts.contract
+   
+   @OptIn(ExperimentalContracts::class)
+   @PowerAssert
+   fun powerAssert(condition: Boolean, @PowerAssert.Ignore message: String? = null) {
+       contract { returns() implies condition }
+       if (!condition) {
+           val explanation = PowerAssert.explanation
```
bnorm, line 764: May want to use an explicit type here to help with clarity.
```suggestion
           val explanation: CallExplanation? = PowerAssert.explanation
```

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+           val failureMessage = buildString {
+               if (message?.isNotBlank() == true) appendLine(message)
+               if (explanation != null) append(explanation.toDefaultMessage())
+           }
+           throw AssertionError(failureMessage)
+       }
+   }
+   ```
+
+   * The `PowerAssert.explanation` property provides access to the `Explanation` object containing the call site information.
```
bnorm:
```suggestion
   * The `PowerAssert.explanation` property provides access to the `CallExplanation` object containing the call site information.
```

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+{style="tip"}
+
+### The CallExplanation class
+
+The [`CallExplanation`](https://github.com/JetBrains/kotlin/blob/master/plugins/power-assert/power-assert-runtime/src/commonMain/kotlin/kotlin/powerassert/CallExplanation.kt)
+data structure provides detailed information about the call site, including intermediate expression values.
+This enables dynamic diagram rendering for assertion failures and better integration with external tools.
+
+The `CallExplanation` parameter must be the last parameter of the function and must have a default value of `null`.
+The compiler plugin automatically fills in this parameter at the call site with the relevant call information.
```
bnorm: This doesn't make sense. The transformation is automatically performed by the compiler plugin, so the library author doesn't need to do anything here if the compiler plugin is applied to the project and the function is annotated with `@PowerAssert`.

But it may be worth mentioning that the `PowerAssert.explanation` property can return null if the function is called from Java, a project where power-assert is not applied, or via reflection.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -703,7 +749,169 @@ assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.ag
...
+            for ((msg, _) in failures) {
+                if (msg != null) appendLine(" * $msg")
+            }
+            appendLine()
+            append(combined.toDefaultMessage())
+        }
+        throw AssertionError(message)
+    }
+}
+```
```
bnorm, line 847: Without the `require` or `check` extension function on `AssertScope`, this example feels incomplete.

https://github.com/bnorm/power-assert-examples/blob/main/fluent-assert/src/commonMain/kotlin/kotlinx/test/fluent/general.kt

bnorm (changes requested):

bnorm (approved):

docs/topics/compiler-plugins/power-assert.md
koshachy: So the output will be in quotes?
I tried the code with 2.4.0-RC2 and I still have the old output, so it will be like that only with 2.4.0?

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -30,6 +27,10 @@ The Power-assert plugin key features:
 * **Support for multiple functions**: By default, it transforms `assert()` function calls but can also transform other functions,
    such as `require()`, `check()`, and `assertTrue()`.
 
+The plugin also includes a runtime library that provides the `@PowerAssert` annotation and the `CallExplanation` class.
```
koshachy: Isn't it a key feature of the plugin? 🤔 
If yes, it should be mention in the list of key features before.

If not, I'd like to remove "also" here.

koshachy: It looks like it is, since we put it on the first place (before **Assert function** section).
So I think it also should be prioritized and on the top of key features list.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -372,6 +373,41 @@ powerAssert {
 </tab>
 </tabs>
 
+### `@PowerAssert`-annotated functions
+
+If a library function is annotated with `@PowerAssert`, the Power-assert plugin transforms calls to it automatically.
```
koshachy: What if we change it to "If you annotate a library function with ..."?

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -372,6 +373,41 @@ powerAssert {
...
+        // the plugin transforms this call automatically
+        assertThat(subject) {
+            require(subject is Mascot)
+            check(subject.name == "Kodee")
+        }
+    }
+}
+```
+
+The plugin provides detailed failure messages with intermediate expression values, just like with `assert()`:
```
koshachy: Well, the section with the `assert()` function is below this one. Not sure that we could reference to it here.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
 Dave has an invalid age: 150
 assert(employee.age < 100) { "${employee.name} has an invalid age: ${employee.age}" }
        |        |   |
-       |        |   false
-       |        150
+       |        150 false
        Employee(name=Dave, age=150, salary=70000)
 ```
 
+## Support the Power-assert plugin
```
koshachy: I find the title a bit awkward, like "The plugin needs your support, here what you can do".

Could we make it more straightforward? Such as "Add the plugin support to your library?"

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+   <dependencies>
+       <dependency>
+           <groupId>org.jetbrains.kotlin</groupId>
+           <artifactId>kotlin-power-assert-runtime</artifactId>
+           <version>%kotlinVersion%</version>
+       </dependency>
+   </dependencies>
+   ```
+
+   For Gradle, this dependency is automatically added together with the Power-assert compiler plugin.
```
koshachy:
```suggestion
   For Gradle, this dependency is automatically added with the Power-assert compiler plugin.
```

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+         }
+   
+         val failureMessage = buildString {
+            if (message?.isNotBlank() == true) appendLine(message)
+            append(explanation.toDefaultMessage())
+         }
+   
+         fail(failureMessage, equalityErrors)
+      }
+   }
```
koshachy, line 782: We use 4 spaces for indentation, not 3.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+   ```
+
+   * The `PowerAssert.explanation` property provides access to the `CallExplanation` object containing the call site information.
+   * The `toDefaultMessage()` function renders the standard Power-assert diagram.
```
koshachy: To be honest, I didn't undestand what the diagram is.
Then I realized what is it but there was no mention of this term before.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+   * The `@PowerAssert.Ignore` annotation on the `message` parameter excludes it from the diagram.
+
+The compiler plugin detects the `@PowerAssert` annotation and transforms calls to the function at compile time.
+
+> For a full example, see the [`kotlin-test-power-assert`](https://github.com/bnorm/power-assert-examples/tree/main/kotlin-test-power-assert) project.
```
koshachy:
```suggestion
> For a complete example, see the [`kotlin-test-power-assert`](https://github.com/bnorm/power-assert-examples/tree/main/kotlin-test-power-assert) project.
```

koshachy: Nitpicking

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+>
+{style="tip"}
+
+### The `CallExplanation` class
+
+The [`CallExplanation`](https://github.com/JetBrains/kotlin/blob/master/plugins/power-assert/power-assert-runtime/src/commonMain/kotlin/kotlin/powerassert/CallExplanation.kt)
+data structure provides detailed information about the call site, including intermediate expression values.
```
koshachy: Just curious, why do we need to use "data structure" instead of "class". Does it change something?

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+This enables dynamic diagram rendering for assertion failures and better integration with external tools.
+
+When a function is annotated with `@PowerAssert` and the compiler plugin is applied, the transformation
```
koshachy: is it about function **in your library**? 
or on developer's call-site?

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
 
+## Support the Power-assert plugin
+
+If you are a library author, you can add out-of-the-box Power-assert support to your library using the `@PowerAssert`
+annotation and the `CallExplanation` class from the Power-assert runtime library.
+
+### The `@PowerAssert` annotation
+
+The [`@PowerAssert` annotation](https://github.com/JetBrains/kotlin/blob/master/plugins/power-assert/power-assert-runtime/src/commonMain/kotlin/kotlin/powerassert/PowerAssert.kt)
+marks a function as Power-assert capable. If users of your library have the Power-assert compiler plugin in their projects
```
koshachy, line 730: We usually don't use term "user" if we mean a "developer".

If developers that use your library have...

It makes the sentence more complicated but I think it sounds better.

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +706,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+The [`CallExplanation`](https://github.com/JetBrains/kotlin/blob/master/plugins/power-assert/power-assert-runtime/src/commonMain/kotlin/kotlin/powerassert/CallExplanation.kt)
+data structure provides detailed information about the call site, including intermediate expression values.
+This enables dynamic diagram rendering for assertion failures and better integration with external tools.
+
+When a function is annotated with `@PowerAssert` and the compiler plugin is applied, the transformation
+is performed automatically. The `PowerAssert.explanation`
+property provides access to the `CallExplanation` object inside the function body.
+
+> The `PowerAssert.explanation` property can return `null` if the annotated function is called from Java,
+> from a project where the Power-assert plugin is not applied, or via reflection.
```
koshachy: WDYT about having a link to the [Reflection](https://kotlinlang.org/docs/reflection.html) article here?

koshachy (changes requested):
Nice work.
Please go through my comments before merging.

Thanks!

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: well, there's also an option to use an external library, added both

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: I disagree here because it creates more confusion between library author/library developer
Our guide on designing libraries uses the term 'user' for people who use libraries

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: Thank you! replaced all mentions

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: Added an explanation below

docs/topics/compiler-plugins/power-assert.md
koshachy: Fair enough!

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -692,18 +709,155 @@ In the output, all the `assert()` function error messages will be printed one af
...
+        fail(message, explanation)
+    }
+}
+```
+
+In this example, the `check()` function collects failures for later reporting, while the `require()` function
+fails immediately. Both functions use `CallExplanation` to extract the source code of the failed condition
+and include it in the failure message.
+
+> For a full example, see the [`fluent-assert`](https://github.com/bnorm/power-assert-examples/tree/main/fluent-assert) project.
```
koshachy:
```suggestion
> For a complete example, see the [`fluent-assert`](https://github.com/bnorm/power-assert-examples/tree/main/fluent-assert) project.
```

koshachy (approved):
Update the fix and we are ready to go!
Good job!

PR #5609, opened 2026-06-11
Add Workday case study

EmilFlach:


EmilFlach:
I have confirmation from Workday that this addition is approved.

nikpachoo:
@berezinant, could you please review the change? 🥺

berezinant (approved):
Tested, all good 👍 
Just unified case id name to match the other cases

PR #5615, opened 2026-06-16
upd: this expressions

aepanchi:
Fixes [KT-48913](https://youtrack.jetbrains.com/issue/KT-48913).

### DIFF
--- a/docs/topics/this-expressions.md
+++ b/docs/topics/this-expressions.md
@@ -67,7 +67,7 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
-The output for this example:
+The output:
 
 ```kotlin
 Local function
--- a/docs/topics/this-expressions.md
+++ b/docs/topics/this-expressions.md
@@ -56,20 +56,16 @@ fun main() {
             fun printLine() {
                 println("Local function")
             }
- 
-            printLine() // Calls the local function
-            this.printLine() // Calls the member function
+         
+            printLine()
+            // Local function
+         
+            this.printLine()
+            // Member function
         }
     }
 
     A().invokePrintLine()
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-
-The output:
-
-```kotlin
-Local function
-Member function
-```
+{kotlin-runnable="true"}
--- a/docs/topics/this-expressions.md
+++ b/docs/topics/this-expressions.md
@@ -49,22 +49,27 @@ use the `this.` qualifier:
 fun main() {
     class A {
         fun printLine() {
-            println("Member function") 
+            println("Member function")
         }
 
-        fun invokePrintLine(omitThis: Boolean = false) {
+        fun invokePrintLine() {
             fun printLine() {
                 println("Local function")
             }
-            if (omitThis) printLine()
-            else this.printLine()
-        } 
+ 
+            printLine() // Calls the local function
+            this.printLine() // Calls the member function
+        }
     }
 
-    A().invokePrintLine() 
-    // Member function
-    A().invokePrintLine(omitThis = true) 
-    // Local function
+    A().invokePrintLine()
 }
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
+The output for this example:
+
+```kotlin
+Local function
+Member function
+```
--- a/docs/topics/this-expressions.md
+++ b/docs/topics/this-expressions.md
@@ -40,8 +40,10 @@ class A { // implicit label @A
 
 ## Implicit this
 
-When you call a member function on `this`, you can skip the `this.` part.
-If you have a non-member function with the same name, use this with caution because in some cases it can be called instead:
+When you call a member function on `this`, you can omit the `this.` qualifier.
+However, if another callable with the same name is available in a closer lexical scope, Kotlin resolves
+the unqualified call to that callable instead of the member function. To explicitly call the member function,
+use the `this.` qualifier:
 
 ```kotlin
 fun main() {
--- a/docs/topics/this-expressions.md
+++ b/docs/topics/this-expressions.md
@@ -47,19 +47,24 @@ use the `this.` qualifier:
 
 ```kotlin
 fun main() {
-    fun printLine() { println("Local function") }
-    
     class A {
-        fun printLine() { println("Member function") }
+        fun printLine() {
+            println("Member function") 
+        }
 
         fun invokePrintLine(omitThis: Boolean = false) {
+            fun printLine() {
+                println("Local function")
+            }
             if (omitThis) printLine()
             else this.printLine()
-        }
+        } 
     }
-    
-    A().invokePrintLine() // Member function
-    A().invokePrintLine(omitThis = true) // Local function
+
+    A().invokePrintLine() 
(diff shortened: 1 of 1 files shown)


docs/topics/this-expressions.md
```diff
@@ -40,8 +40,10 @@ class A { // implicit label @A
 
 ## Implicit this
 
-When you call a member function on `this`, you can skip the `this.` part.
-If you have a non-member function with the same name, use this with caution because in some cases it can be called instead:
+When you call a member function on `this`, you can omit the `this.` qualifier.
+However, if another callable with the same name is available in a closer lexical scope, Kotlin resolves
```
e2e4b6b7, line 44: "closer lexical scope" sounds good in the abstract most common scenario, but does not match the provided example, where member is actually lexically closer. Either the example should be simplified by making local function local to `invokePrintLine` or using a more vague terms in the description (f.e. "if another callable with the same name is available in the scope, Kotlin might resolve ...")

e2e4b6b7 (changes requested):

docs/topics/this-expressions.md
aepanchi: Good point! I have updated the example

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,31 @@ class A { // implicit label @A
...
+        } 
     }
-    
-    A().invokePrintLine() // Member function
-    A().invokePrintLine(omitThis = true) // Local function
+
+    A().invokePrintLine() 
+    // Member function
+    A().invokePrintLine(omitThis = true) 
+    // Local function
```
e2e4b6b7: Now it is hard to understand which comment relates to which line. They should be either on the same line as the call, or there should be a separation line in between.

e2e4b6b7: nit: Also I do not see a reason why it should be two external calls with boolean for dispatch. Isn't something like the following easier to understand:

```
        fun invokePrintLine(omitThis: Boolean = false) {
            fun printLine() {
                println("Local function")
            }
            printLine() // Local function
            this.printLine() // Member function
        } 
```

e2e4b6b7 (changes requested):

docs/topics/this-expressions.md
aepanchi: What do you think about dropping omitThis at all? Maybe smth like: 
```
fun main() {
    class A {
        fun printLine() {
            println("Member function")
        }

        fun invokePrintLine() {
            fun printLine() {
                println("Local function")
            }

            printLine() // Calls the local function
            this.printLine() // Calls the member function

        }

    }

    A().invokePrintLine()
    
    // Local function
    // Member function
}
```

e2e4b6b7: Looks good, but the meaning of two last comment is not that straightforward. Either remove them or explain like:
```
    // Prints:
    // Local function
    // Member function
```

Or it might be fine if the same is used in lots of other examples

aepanchi: agree, I have added output separately to avoid confusion

e2e4b6b7 (approved):

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
...
-    
-    A().invokePrintLine() // Member function
-    A().invokePrintLine(omitThis = true) // Local function
+
+    A().invokePrintLine()
 }
 ```
+
+The output:
```
daniCsorbaJB: I suggest to put this below the relevant lines instead and removing it from here.

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
...
         }
     }
-    
-    A().invokePrintLine() // Member function
-    A().invokePrintLine(omitThis = true) // Local function
+
+    A().invokePrintLine()
 }
 ```
```
daniCsorbaJB:
```suggestion
{kotlin-runnable="true"}
```

daniCsorbaJB: 1.3 was so long ago I say we can safely remove this

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
...
 
-        fun invokePrintLine(omitThis: Boolean = false) {
-            if (omitThis) printLine()
-            else this.printLine()
+        fun invokePrintLine() {
+            fun printLine() {
+                println("Local function")
+            }
+ 
+            printLine() // Calls the local function
```
daniCsorbaJB: Let's put these comments above the relevant lines

daniCsorbaJB:
```suggestion
            printLine()
```

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
...
+        }
 
-        fun invokePrintLine(omitThis: Boolean = false) {
-            if (omitThis) printLine()
-            else this.printLine()
+        fun invokePrintLine() {
+            fun printLine() {
+                println("Local function")
+            }
+
```
daniCsorbaJB:
```suggestion
     // Calls the local function
```

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
...
-        fun invokePrintLine(omitThis: Boolean = false) {
-            if (omitThis) printLine()
-            else this.printLine()
+        fun invokePrintLine() {
+            fun printLine() {
+                println("Local function")
+            }
+ 
+            printLine() // Calls the local function
+            this.printLine() // Calls the member function
```
daniCsorbaJB: Same suggestion here, and separate these two lines with an extra line just for ease of reading.

docs/topics/this-expressions.md
```diff
@@ -40,24 +40,36 @@ class A { // implicit label @A
 
 ## Implicit this
 
-When you call a member function on `this`, you can skip the `this.` part.
-If you have a non-member function with the same name, use this with caution because in some cases it can be called instead:
+When you call a member function on `this`, you can omit the `this.` qualifier.
+However, if another callable with the same name is available in a closer lexical scope, Kotlin resolves
```
daniCsorbaJB, line 44:
```suggestion
However, if another callable with the same name is available in a closer lexical scope, such as a local function, Kotlin resolves
```

daniCsorbaJB: Just to align with the example a bit more closely, WDYT? 🤔

daniCsorbaJB (changes requested):
Really nice improvement, great work! 💯 
I added a couple of suggestions. Let me know what you think.

docs/topics/this-expressions.md
aepanchi: It does align the text with the example, but I think it can confuse users, making them think that it is applicable to local functions only. However, the rule isn’t limited to local functions - other declarations can also shadow members. I would prefer to use a general wording in that case

daniCsorbaJB: sure thing 👍 works as it is, I suggested with "such as ..." to just give an example, but I think we can keep it generic.

daniCsorbaJB (approved):
LGTM! 🚀

PR #5620, opened 2026-06-17
Improve description of type parameter inference in generics

Junio00r:
Clarify the explanation about inferring class type parameters in Kotlin generics differentiating:

1. parameters of the class; 
2. parameters of the constructor by itself;

### DIFF
--- a/docs/topics/generics.md
+++ b/docs/topics/generics.md
@@ -14,7 +14,7 @@ To create an instance of such a class, simply provide the type arguments:
 val box: Box<Int> = Box<Int>(1)
 ```
 
-But if the class type parameters can be inferred by the constructor type parameters with the generic signature, for example, from the constructor arguments,
+If the compiler can infer the type arguments, for example, from the constructor arguments, you don’t need to specify them explicitly:
 you can omit the type arguments:
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


docs/topics/generics.md
```diff
@@ -14,7 +14,7 @@ To create an instance of such a class, simply provide the type arguments:
 val box: Box<Int> = Box<Int>(1)
 ```
 
-But if the parameters can be inferred, for example, from the constructor arguments,
+But if the class type parameters can be inferred by the constructor type parameters with the generic signature, for example, from the constructor arguments,
```
daniCsorbaJB:
```suggestion
If the compiler can infer the type arguments, for example, from the constructor arguments, you don’t need to specify them explicitly:
```

docs/topics/generics.md
```diff
@@ -14,7 +14,7 @@ To create an instance of such a class, simply provide the type arguments:
...
 you can omit the type arguments:
```
daniCsorbaJB:
```suggestion
```

daniCsorbaJB (changes requested):
Hey @Junio00r, thank you for your feedback! 

I agree that the current wording is a bit unclear, but I don’t think your proposed wording would fix the core issue. 🤔 
I've sent back an alternative to align the description with the example that follows and to clean up the text. 

Can you update your PR? Please let me know if you need any assistance or if something is unclear.

daniCsorbaJB (approved):
Thank you for the changes! LGTM 🚀

PR #5629, opened 2026-06-22
Improve Packages and imports

ElviraMustafina:


docs/topics/packages.md
```diff
@@ -1,73 +1,132 @@
...
 ```
 
-All the contents, such as classes and functions, of the source file are included in this package.
-So, in the example above, the full name of `printMessage()` is `org.example.printMessage`,
-and the full name of `Message` is `org.example.Message`. 
+All contents of the source file (such as classes and functions) belong to this package.
+In the example above:
 
-If the package is not specified, the contents of such a file belong to the _default_ package with no name.
+* The fully qualified name of `printMessage()` is `org.example.printMessage`.
```
serras, line 23: It would be nice to prefix this with a sentence like "The combination of the package name and the name of the declaration is called its _fully qualified name_"

docs/topics/packages.md
```diff
@@ -1,73 +1,132 @@
...
-- [kotlin.comparisons.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.comparisons/index.html)
-- [kotlin.io.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/index.html)
-- [kotlin.ranges.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/index.html)
-- [kotlin.sequences.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/index.html)
-- [kotlin.text.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/index.html)
+To use an entity from a file in a different package, use an `import` directive.
+Apart from the default imports, each file may declare its own imports.
 
-Additional packages are imported depending on the target platform:
+### Import a single name
```
serras:
```suggestion
### Import a single declaration
```

docs/topics/packages.md
```diff
@@ -1,73 +1,132 @@
...
+    val message = Message("Hello")
+    println(message.text)
+}
+```
 
-Apart from the default imports, each file may contain its own `import` directives.
+### Import the contents of a scope
 
-You can import either a single name:
+You can import everything accessible within a scope — a package, class, or object:
```
serras: This is not correct, you cannot use `*` with an object

serras (changes requested):

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+    import org.example.Color.GREEN
+    ```
+* Nested classes:
+    ```kotlin
+    import org.example.Outer.Nested
+    ```
+
+## Default imports
+
+Kotlin imports several packages into every file by default:
```
serras: Nitpick: if you say you import a package, you should just write `kotlin.annotations`. Otherwise it's better to say "Kotlin includes the following imports by default" or something like that

serras (approved):

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
-// ...
+class Message(val text: String) { /*...*/ }
 ```
 
-All the contents, such as classes and functions, of the source file are included in this package.
-So, in the example above, the full name of `printMessage()` is `org.example.printMessage`,
-and the full name of `Message` is `org.example.Message`. 
+All contents of the source file (such as classes and functions) belong to this package.
+A declaration's fully qualified name combines the package name with the name of the declaration.
+In the example above:
```
daniCsorbaJB: (suggestion)
WDYT about simply: In this example? 🤔 — we are in one paragraph so should be just as understandable and one fewer word (tbh, both works so whichever you prefer)

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
 [//]: # (title: Packages and imports)
 
+In a Kotlin project, code is organized using packages and imports:
+
+* A **package** is a container for one or more Kotlin files. Files are linked to a package using a `package` header.
+* An **import** is a directive that makes entities from other packages available in the current file.
```
daniCsorbaJB, line 6: Would this part work with declarations instead of entities here? 🤔 

* An **import** is a directive that makes declarations from other packages available in the current file.

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
-class Message { /*...*/ }
-
-// ...
+class Message(val text: String) { /*...*/ }
 ```
 
-All the contents, such as classes and functions, of the source file are included in this package.
-So, in the example above, the full name of `printMessage()` is `org.example.printMessage`,
-and the full name of `Message` is `org.example.Message`. 
+All contents of the source file (such as classes and functions) belong to this package.
```
daniCsorbaJB:
```suggestion
All contents of the source file, such as classes and functions, belong to this package.
```

daniCsorbaJB: (suggestion)
It's mostly preference, but I think you made the sentence read smoothly, so I think it parses well even with commas instead of parentheses.

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+A declaration's fully qualified name combines the package name with the name of the declaration.
+In the example above:
 
-If the package is not specified, the contents of such a file belong to the _default_ package with no name.
+* The fully qualified name of `printMessage()` is `org.example.printMessage`.
+* The fully qualified name of `Message` is `org.example.Message`.
 
-## Default imports
+If no package is specified, the file's contents belong to the default root package.
```
daniCsorbaJB:
```suggestion
If a file has no `package` header, its declarations belong to the default root package.
```

daniCsorbaJB: Just as a nod to the terminology we used at the beginning of the doc, and to avoid the passive structure we could make file the subject.

I'm not sure about the declaration vs contents switch. I'd assume comments would also belong to the root package, but perhaps the focus should be on the declarations, so that's why I suggested it 🤔 
I'm not sure, to me both feel correct, so whichever you think is right 👍

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
 
-- [kotlin.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/index.html)
-- [kotlin.annotation.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.annotation/index.html)
-- [kotlin.collections.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/index.html)
-- [kotlin.comparisons.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.comparisons/index.html)
-- [kotlin.io.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/index.html)
-- [kotlin.ranges.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/index.html)
-- [kotlin.sequences.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/index.html)
-- [kotlin.text.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/index.html)
+To use an entity from a file in a different package, use an `import` directive.
```
daniCsorbaJB, line 30: same here as above, entity vs declaration? 🤔

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+Apart from the default imports, each file may declare its own imports.
```
daniCsorbaJB:
```suggestion
In addition to the default imports, each file may declare its own imports.
```

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
 
-- JVM:
-  - java.lang.*
-  - [kotlin.jvm.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/index.html)
+Import one specific declaration so it can be used without qualification:
 
-- JS:    
-  - [kotlin.js.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/index.html)
+```kotlin
+import org.example.Message // Message is accessible without qualification
```
daniCsorbaJB: In general it may be better to have the comments above (it can avoid scrolling especially for the later ones 🤔 — here I think is fine, but then it might become inconsistent.

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+To use an entity from a file in a different package, use an `import` directive.
+Apart from the default imports, each file may declare its own imports.
 
-Additional packages are imported depending on the target platform:
+### Import a single declaration
 
-- JVM:
-  - java.lang.*
-  - [kotlin.jvm.*](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/index.html)
+Import one specific declaration so it can be used without qualification:
```
daniCsorbaJB:
```suggestion
Import a specific declaration so you can use it without qualification:
```

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+    val message = Message("Hello")
+    println(message.text)
+}
+```
 
-Apart from the default imports, each file may contain its own `import` directives.
+### Import the contents of a scope
 
-You can import either a single name:
+Star-imports ending in an asterisk (`*`) import all named entities inside the corresponding scope:
```
daniCsorbaJB:
```suggestion
Star imports ending in an asterisk (`*`) import all named declarations inside the corresponding scope:
```

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
 
 ```kotlin
-import org.example.Message // Message is now accessible without qualification
+import org.example.* // Everything in 'org.example' is accessible
```
daniCsorbaJB:
```suggestion
import org.example.* // Everything in org.example is accessible
```

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+
+fun main() {
+    printMessage()
+    val message = Message("Hi")
+}
 ```
 
-or all the accessible contents of a scope: package, class, object, and so on:
+If a name is imported via both a star import and an explicit single import,
```
daniCsorbaJB:
```suggestion
If you import the same name with both a star import and an explicit single import,
```

daniCsorbaJB: just to remove latin + passive 🤔 WDYT?

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
-```
+### Resolve name clashes with aliases
 
-If there is a name clash, you can disambiguate by using `as` keyword to locally rename the clashing entity:
+If two imported names clash, use the `as` keyword to locally rename one of them:
 
 ```kotlin
-import org.example.Message // Message is accessible
-import org.test.Message as TestMessage // TestMessage stands for 'org.test.Message'
+import org.example.Message             // Message refers to 'org.example.Message'
```
daniCsorbaJB: Here, I suggest placing the code comments above to eliminate scrolling and adding a line between them for clarity

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
+* [kotlin.annotation.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.annotation/index.html)
+* [kotlin.collections.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/index.html)
+* [kotlin.comparisons.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.comparisons/index.html)
+* [kotlin.io.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/index.html)
+* [kotlin.ranges.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/index.html)
+* [kotlin.sequences.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/index.html)
+* [kotlin.text.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/index.html)
+* [kotlin.math.*](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.math/index.html)
+
+Additional packages are imported based on the target platform:
```
daniCsorbaJB:
```suggestion
Kotlin imports additional packages by default depending on the target platform:
```

daniCsorbaJB: (suggestion)
again, it's understandable with the passive, but wanted to leave it here as an option

docs/topics/packages.md
```diff
@@ -1,73 +1,136 @@
...
-  * top-level functions and properties
-  * functions and properties declared in [object declarations](object-declarations.md#object-declarations-overview)
-  * [enum constants](enum-classes.md)
+## Visibility and imports
 
-## Visibility of top-level declarations
+The ability to import an entity depends on its [visibility modifiers](visibility-modifiers.md):
 
-If a top-level declaration is marked `private`, it is private to the file it's declared in (see [Visibility modifiers](visibility-modifiers.md)).
+* `public` entities can be imported anywhere.
```
daniCsorbaJB, line 137: same question here about declarations vs entities

daniCsorbaJB (commented):
Really nice update thank you! 💯 I left a couple of comments please let me know what you think

docs/topics/packages.md
ElviraMustafina: So, I double-checked the Kotlin language specification:
The spec uses "entity" to describe what you are importing and interacting with, while "declaration" is mostly used to describe the syntactic construct or scope.

docs/topics/packages.md
ElviraMustafina: There’s also a switch from “header” to “declaration”! The spec consistently uses "package header"

docs/topics/packages.md
```diff
@@ -45,27 +46,31 @@ fun main() {
...
-If two imported names clash, use the `as` keyword to locally rename one of them:
+If two imported entities have the same name, use the `as` keyword to locally rename one of them:
 
 ```kotlin
-import org.example.Message             // Message refers to 'org.example.Message'
-import org.test.Message as TestMessage // TestMessage refers to 'org.test.Message'
+// Message refers to `org.example.Message`
+import org.example.Message
+
+// TestMessage refers to `org.test.Message`
```
daniCsorbaJB: Just as a last question as a general comment: 
Do you think we need the backticks for these fully qualified names like in: `org.test.Message`? 🤔 I'm asking because for TestMessage we don't add it but for the fully qualified names we do 🤔

daniCsorbaJB (approved):
Thank you for the updates! LGTM! 🚀 💯 — I only left one additional question.

docs/topics/packages.md
ElviraMustafina: Thanks for pointing that out! According to our guidelines, there should be no backticks, so I'll get this fixed right away

PR #5642, opened 2026-06-26
[upd] Document non-denotable types

aepanchi:


docs/topics/jvm/java-interop.md
```diff
@@ -148,8 +148,8 @@ foo.`is`(bar)
 ## Null-safety and platform types
 
 Any reference in Java may be `null`, which makes Kotlin's requirements of strict null-safety impractical for objects coming from Java.
-Types of Java declarations are treated in Kotlin in a specific manner and called *platform types*. Null-checks are relaxed
-for such types, so that safety guarantees for them are the same as in Java (see more [below](#mapped-types)).
+Types of Java declarations are treated in Kotlin as non-denotable and called [*platform types*](https://kotlinlang.org/spec/type-system.html#platform-types).
```
InversionSpaces, line 151: Nitpicking: not sure that "non-denotableness" is a good first thing to tell about Java types.

docs/topics/jvm/java-interop.md
```diff
@@ -168,9 +168,9 @@ prevent nulls from propagating:
 item.substring(1) // allowed, throws an exception if item == null
 ```
 
-Platform types are *non-denotable*, meaning that you can't write them down explicitly in the language.
-When a platform value is assigned to a Kotlin variable, you can rely on the type inference (the variable will have an inferred
-platform type then, as `item` has in the example above), or you can choose the type you expect (both nullable and non-nullable types are allowed):
+You can assign a platform type to both nullable and non-nullable Kotlin types. However, if you assign a platform type to
```
InversionSpaces: You do not assign types, you assign values of some type to variables of some other type.

docs/topics/generics.md
```diff
@@ -267,6 +267,32 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
 
+### Captured types
+
+When you project a generic type with `out` or `in`, the compiler internally captures the unknown concrete type
+behind the projection into a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). For example,
+when you write `Array<out Any>`, the compiler doesn't know the actual element type. It captures that unknown type to
+enforce the variance rules:
+
+* You can call `get()` to return the captured type.
+* You can't call `set()`, because it needs to know the exact type to write safely.
```
InversionSpaces: Formally not true, it is enough to know lower bound for the captured type to write safely.

docs/topics/generics.md
```diff
@@ -267,6 +267,32 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+
+Since intersection types are non-denotable, you can't write them in Kotlin code. However, if you violate a variance
```
InversionSpaces: I don't get it, how intersections types are related to information in this section?

docs/topics/generics.md
```diff
@@ -267,6 +267,32 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+constraint on a projected type, the compiler may mention them in error messages, displayed as `CapturedType(out X)`
+or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`, where `L` is the
+lower bound (non-nullable) and `U` is the upper bound (nullable). For example:
+
+```kotlin
+fun example(list: MutableList) {
```
InversionSpaces: I believe argument to `MutableList` type is forgotten here.

docs/topics/generics.md
```diff
@@ -267,6 +267,32 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+    list.add(42)
+    // Error: out-projected type 'MutableList' prohibits
+    // the use of 'fun add(element: E): Boolean'
+}
+```
+
+If you see `CapturedType` in a compiler diagnostic, the underlying issue is a variance mismatch on a projected type. This
+means, you try to use the type in a position that the projection doesn't allow.
```
InversionSpaces: Technically, it is not necessary true. You can see `CapturedType` in a TYPE_MISMATCH diagnostic, for example:
```kotlin
val l: Array<out CharSequence> = arrayOf("str")
val i: Int = l.get(0)
```
It does not mean that you violated variance of the projection.

docs/topics/typecasts.md
```diff
@@ -289,6 +292,33 @@ fun signalCheck(signalStatus: Any) {
...
+    if (animal is Bird && animal is Fish) {
+        // Accesses fly() and swim() without additional checks or casts
+        animal.fly()
+        animal.swim()
+    }
+}
+```
+
+Since intersection types are non-denotable, you can't write them in Kotlin code. However, you may encounter them in
+compiler error messages and IDE tooltips, typically displayed as `A & B`.
```
InversionSpaces: Possibly worth to mention the exception of `T & Any`:
```kotlin
fun <T> method(): T & Any = TODO()
```

docs/topics/types-overview.md
```diff
@@ -18,6 +18,19 @@ To learn about other Kotlin types, such as `Nothing`, `Any`, and `Unit`, look th
...
+compiler uses them internally, for example, for interoperability with other languages. Kotlin creates non-denotable
+types to represent type information that is more precise than what Kotlin source syntax allows.
+
+Even though you can't declare non-denotable types yourself, you may encounter them in compiler diagnostics, IDE
+tooltips, or inferred type displays. Learn more about non-denotable types in:
+
+* [Platform types](java-interop.md#null-safety-and-platform-types)
+* [](typecasts.md#intersection-types)
+* [](numbers.md#integer-literal-types)
+* [](generics.md#captured-types)
```
InversionSpaces, line 31: Display text for links is missing?

docs/topics/types-overview.md
```diff
@@ -18,6 +18,19 @@ To learn about other Kotlin types, such as `Nothing`, `Any`, and `Unit`, look th
 * [`Nothing`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-nothing.html)
 * [`Unit`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-unit/)
 
+Kotlin also support non-denotable types. They are types that you can't write directly in the Kotlin code. Instead, the
```
InversionSpaces: Nitpicking: I am not sure "supports" is the right word here. I would phrase it like "Some types in Kotlin are non-denotable. The compiler uses them internally..."

InversionSpaces (changes requested):

docs/topics/types-overview.md
aepanchi: No, it will be just fine on production:) It uses the name of the topic/section that it links to. It is rendered the following way: 
image

docs/topics/jvm/java-interop.md
aepanchi: fixed

docs/topics/generics.md
aepanchi: nice catch! fixed

docs/topics/generics.md
aepanchi: Would be better to phrase it like that?:
For example, when you write `Array<out Any>`, the compiler doesn't
know the actual element type. It captures that unknown type to enforce the variance rules: 
* You can call `get()`, since it only returns the captured type. 
* You can't call `set()`, because the captured type from an`out`-projection has no known lower bound. Therefore,  there's no type the compiler can guarantee is safe to write.

docs/topics/generics.md
aepanchi: typo! to many types :D thanks!

docs/topics/generics.md
aepanchi: Nice point! What do you think about:

You may see `CapturedType` mentioned in various compiler diagnostics. It means that the compiler reports back the type it inferres for the projected value. The specific problem depends on the context the message appears in. For example, a type mismatch can also reference a captured type if you try to use it where an incompatible type is expected:

```
val list: List = listOf("str")
val i: Int = list.get(0)
// Error: Initializer type mismatch: expected 'Int', actual 'CharSequence'
```

docs/topics/typecasts.md
aepanchi: done

docs/topics/jvm/java-interop.md
```diff
@@ -168,9 +168,9 @@ prevent nulls from propagating:
 item.substring(1) // allowed, throws an exception if item == null
 ```
 
-Platform types are *non-denotable*, meaning that you can't write them down explicitly in the language.
-When a platform value is assigned to a Kotlin variable, you can rely on the type inference (the variable will have an inferred
-platform type then, as `item` has in the example above), or you can choose the type you expect (both nullable and non-nullable types are allowed):
+You can assign a value of a platform type to a variable both nullable and non-nullable Kotlin types. However,
```
InversionSpaces:
```suggestion
You can assign a value of a platform type to a variable of both nullable and non-nullable Kotlin types. However,
```

docs/topics/generics.md
InversionSpaces: Well, I guess this is okay. A more generally comment from me is that it may be hard to grasp the idea of a captured type from this section, I think. A captured type is an unknown type within some range, i.e. we know lower and upper bounds for it. Usually one of the bounds is trivial (`Nothing` as lower or `Any` as upper). When type appears in a covariant position, we approximate to upper bound, when in a contravariant postion, we approximate to lower bound (or outright prohibit such usage if lower bound is `Nothing`). So I think it would be a better example if you take `Array<out X>` where `X` is **not** `Any`. Then `get()` returns technically a captured type, yes, but if user writes `val result = array.get(0)`, then type of `result` would be inferred to `X`, not `CapturedType<out X>`.

docs/topics/generics.md
InversionSpaces: Text is okay, but in the example `List` misses a type argument and also `List` is a bad example, because it has declaration-site `out` variance, so `val list: List<out CharSequence>` is not a thing, IDE will suggest you to remove `out`.

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
...
+The one exception is `T & Any`, used to declare a [definitely non-nullable type](generics.md#definitely-non-nullable-types).
+The syntax is reserved specifically to combine a type parameter with `Any`. You still can't write other
+intersection types explicitly.
+
+```kotlin
+fun  method(): T & Any = TODO()
```
InversionSpaces: It would be great if examples had some meaning
```suggestion
fun  <T> T.assertNotNull(): T & Any = this ?: throw IllegalStateException("null value")
```

InversionSpaces (changes requested):

InversionSpaces (approved):

docs/topics/jvm/java-interop.md
```diff
@@ -168,9 +168,9 @@ prevent nulls from propagating:
 item.substring(1) // allowed, throws an exception if item == null
 ```
 
-Platform types are *non-denotable*, meaning that you can't write them down explicitly in the language.
-When a platform value is assigned to a Kotlin variable, you can rely on the type inference (the variable will have an inferred
```
danil-pavlov, line 172: Thank you! Nice addition to the interop docs, now this part is much clearer. One thing, do you think we can keep this explanation of type inference?

danil-pavlov (approved):

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
 >
 {style="note"}
 
+### Captured types
+
+When you use a type projection like `out T` or `in T`, the compiler doesn't know the exact type argument behind it. In this case,
```
daniCsorbaJB: Do you think we could start with the new term here? 🤔 This description has lots of anthropomorphism

Something like: 
When you use a type projection, such as `out T` or `in T`, Kotlin/the compiler/the Kotlin compiler represents the type argument in the projection as a *captured type*.

The following sentence defines it: A captured type is an unknown type within a known range.

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+the compiler internally represents the unknown concrete type as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown
```
daniCsorbaJB: I don't think we need the link to the language spec unless there is something we didn't define in the docs. 🤔 (but imho your description is quite good already so we can drop it)

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+type within a known range. The compiler uses this range to decide which operations are safe:
```
daniCsorbaJB: by safe, do we mean type safe? 🤔

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+
+* When you read a value, the compiler uses the captured type’s upper bound.
```
daniCsorbaJB:
```suggestion
* When you read a value, the compiler uses the captured type's upper bound.
```

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+array.set(0, "New value")
+// Receiver type 'Array<out CharSequence>' contains out projection
+// which prohibits the use of 'fun set(index: Int, value: T): Unit'
+```
+
+The `.set()` function uses the element type in a write position. In this case, the compiler needs a lower bound for the
+captured type. Since for `Array<out CharSequence>`, the lower bound is `Nothing`, there is no ordinary value that the
+compiler can safely accept. This prevents unsafe writes. For example, the actual array might be `Array<StringBuilder>`.
+
+Since captured types are non-denotable, you can’t write them directly in Kotlin code. However, you might see them in the compiler
```
daniCsorbaJB: I suggest we start with this when we define what a captured type is. Now, I feel like it comes way too late (it could cause some frustration if someone reads until this point and then realizes they can't/won't even use these)

docs/topics/generics.md
daniCsorbaJB: please check for the other occurences for `’` vs `'`

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+### Captured types
+
+When you use a type projection like `out T` or `in T`, the compiler doesn't know the exact type argument behind it. In this case,
+the compiler internally represents the unknown concrete type as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown
+type within a known range. The compiler uses this range to decide which operations are safe:
+
+* When you read a value, the compiler uses the captured type’s upper bound.
+* When you write a value, the compiler requires a valid lower bound.
+
+Usually, one of these bounds is trivial. For example, an `out T` projection gives the captured type an upper bound (`T`),
```
daniCsorbaJB: hmm 🤔 I find this a bit confusing. Does the user read or write, or the compiler? 🤔 the UL makes it seem like the user is writing/reading - but in the description below, which expands on the information, it makes it seem like the user is using a type projection and the compiler is reading and writing 🤔

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+As a result, the compiler treats any value read from the array as `CharSequence`:
+
+```kotlin
+val item = array.get(0)
+```
+
+Here, the `.get()` function technically returns a captured type. However, the captured type only appears in a covariant
+(read) position. Therefore, the compiler approximates it to its upper bound when inferring the type for `item`.
+
+However, it's not safe to write to this array:
```
daniCsorbaJB: type safe? 🤔

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
 >
 {style="note"}
 
+### Intersection types
+
+When the compiler smart-casts an object through multiple `&&` checks, it infers an [*intersection type*](https://kotlinlang.org/spec/type-system.html#intersection-types).
+It is an internal type that simultaneously satisfies all the checked constraints:
```
sarahhaggarty:
```suggestion
This is an internal type that simultaneously satisfies all the checked constraints:
```
Nit: Sounds more natural to a native speaker.

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
...
+    // Infers the Bird and Fish types
+    if (animal is Bird && animal is Fish) {
+        // Accesses fly() and swim() without additional checks or casts
+        animal.fly()
+        animal.swim()
+    }
+}
+```
+
+Since intersection types are non-denotable, you can't write them in Kotlin code. However, you may encounter them in
```
sarahhaggarty:
```suggestion
Intersection types are non-denotable. They only exist in the compiler's internal type system to preserve type information during type checking. You can't write them directly in Kotlin code. You may encounter intersection types in
```
It's not clear what "non-denotable" means just from this text and I think it's also a fairly advanced English word. What do you think about expanding your explanation here a little? I made a suggestion for you to consider :)

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
...
+compiler error messages and IDE tooltips, typically displayed as `A & B`.
+The one exception is `T & Any`, used to declare a [definitely non-nullable type](generics.md#definitely-non-nullable-types).
```
sarahhaggarty: Can you change the passive "is used" to something more active? Maybe "which declares"?

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
...
+The syntax is reserved specifically to combine a type parameter with `Any`. You still can't write other
```
sarahhaggarty:
```suggestion
This syntax is reserved specifically to combine a type parameter with `Any`. You still can't write other
```

docs/topics/typecasts.md
```diff
@@ -289,6 +292,40 @@ fun signalCheck(signalStatus: Any) {
...
+intersection types explicitly.
```
sarahhaggarty: Do we really need the last sentence since the paragraph starts by explaining that you can't write them in code?

sarahhaggarty (changes requested):
A good start! It's only request changes because I'd like you to expand your explanation of what non-denotable means a little more :)

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+type can appear in a type mismatch diagnostic:
+
+```kotlin
+val array: Array<out CharSequence> = arrayOf("str")
+val item: Int = array.get(0)
+// Initializer type mismatch: expected 'Int', actual 'CapturedType(out CharSequence)'
+```
+
+This error doesn’t mean that the code violates the variance rules of the projection. The problem is that the compiler
+knows only that the result is some unknown subtype of `CharSequence`, and such a value can’t be assigned to `Int`.
```
sarahhaggarty:
```suggestion
knows only that the result is some unknown subtype of `CharSequence`, and such a value can't be assigned to `Int`.
```

docs/topics/generics.md
sarahhaggarty:
```suggestion
Since captured types are non-denotable, you can't write them directly in Kotlin code. However, you might see them in the compiler
```

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+diagnostics, displayed as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, a captured
+type can appear in a type mismatch diagnostic:
+
+```kotlin
+val array: Array<out CharSequence> = arrayOf("str")
+val item: Int = array.get(0)
+// Initializer type mismatch: expected 'Int', actual 'CapturedType(out CharSequence)'
+```
+
+This error doesn’t mean that the code violates the variance rules of the projection. The problem is that the compiler
```
sarahhaggarty:
```suggestion
This error doesn't mean that the code violates the variance rules of the projection. The problem is that the compiler
```

docs/topics/generics.md
sarahhaggarty:
```suggestion
* When you read a value, the compiler uses the captured type's upper bound.
```

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+the compiler internally represents the unknown concrete type as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown
+type within a known range. The compiler uses this range to decide which operations are safe:
+
+* When you read a value, the compiler uses the captured type’s upper bound.
+* When you write a value, the compiler requires a valid lower bound.
+
+Usually, one of these bounds is trivial. For example, an `out T` projection gives the captured type an upper bound (`T`),
+but its lower bound is `Nothing`. This means that the compiler can read values as `T`, but can't write ordinary values.
+
+For example, in the following code, the actual element type is unknown. It can be `String`, `StringBuilder`, or another
```
daniCsorbaJB: I'm a bit confused by this example so please correct me if I'm wrong:

🤔 Wouldn't the exact element type be a String here, though? 
`val array: Array<out CharSequence>` so here it's unknown, but then we initialize it with  `= arrayOf("Kotlin")` — Wouldn't that make the element type a String? 🤔 
Or do we mean that the code that uses the variable array only sees the projected type `Array<out CharSequence>` so from their perspective it is unknown?

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+Therefore, the compiler captures this unknown type as a type with `CharSequence` as an upper bound.
+
+Even though the compiler doesn’t know the exact element type, it knows that every possible element type is a subtype of `CharSequence`.
+As a result, the compiler treats any value read from the array as `CharSequence`:
+
+```kotlin
+val item = array.get(0)
+```
+
+Here, the `.get()` function technically returns a captured type. However, the captured type only appears in a covariant
```
daniCsorbaJB:
```suggestion
Here, the `get()` function technically returns a captured type. However, the captured type only appears in a covariant
```

docs/topics/jvm/java-interop.md
aepanchi: Thanks, added :)

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+
+However, it's not safe to write to this array:
+
+```kotlin
+array.set(0, "New value")
+// Receiver type 'Array<out CharSequence>' contains out projection
+// which prohibits the use of 'fun set(index: Int, value: T): Unit'
+```
+
+The `.set()` function uses the element type in a write position. In this case, the compiler needs a lower bound for the
```
daniCsorbaJB:
```suggestion
The `set()` function uses the element type in a write position. In this case, the compiler needs a lower bound for the
```

daniCsorbaJB: 🤔 The docs don't use the terms: read position vs write position. I'd suggest rephrasing these for clarity, here it could be something like: The `set()` function takes a value of the array's element type as an argument. or something similarly specific that avoids the need to call it write position.

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+captured type. Since for `Array<out CharSequence>`, the lower bound is `Nothing`, there is no ordinary value that the
```
daniCsorbaJB: I suggest rewriting this part as well to avoid anthropomorphism, for example

You can't use the `set()` function here, because ...

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+For example, in the following code, the actual element type is unknown. It can be `String`, `StringBuilder`, or another
+subtype of `CharSequence`:
+
+```kotlin
+val array: Array<out CharSequence> = arrayOf("Kotlin")
+```
+The projection `out CharSequence` guarantees that the actual element type is `CharSequence` or one of its subtypes.
+Therefore, the compiler captures this unknown type as a type with `CharSequence` as an upper bound.
+
+Even though the compiler doesn’t know the exact element type, it knows that every possible element type is a subtype of `CharSequence`.
```
daniCsorbaJB: This sentence feels anthropomorphic again, could you please rephrase?

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+diagnostics, displayed as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, a captured
+type can appear in a type mismatch diagnostic:
+
+```kotlin
+val array: Array<out CharSequence> = arrayOf("str")
+val item: Int = array.get(0)
+// Initializer type mismatch: expected 'Int', actual 'CapturedType(out CharSequence)'
+```
+
+This error doesn't mean that the code violates the variance rules of the projection. The problem is that the compiler
```
daniCsorbaJB: I feel like we already have enough examples, and this description feels exactly the same as a previous one. Still, I believe the example itself is good, since this should be the most common way the user will ever encounter captured types, but I'd suggest removing the redundancy in the description at least.

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+array.set(0, "New value")
+// Receiver type 'Array<out CharSequence>' contains out projection
+// which prohibits the use of 'fun set(index: Int, value: T): Unit'
+```
+
+The `.set()` function uses the element type in a write position. In this case, the compiler needs a lower bound for the
+captured type. Since for `Array<out CharSequence>`, the lower bound is `Nothing`, there is no ordinary value that the
+compiler can safely accept. This prevents unsafe writes. For example, the actual array might be `Array<StringBuilder>`.
+
+Since captured types are non-denotable, you can't write them directly in Kotlin code. However, you might see them in the compiler
```
daniCsorbaJB: I think we should start with this after describing what captured types are. I think logically this is the most important part someone must read to be able to decide if they want to read further or not 🤔 Reading until this part and just learning that these are non-denotable types, meaning that they'll most likely never use them  (only encounter them) might cause some frustration.

daniCsorbaJB:
```suggestion
Captured types are non-denotable, so you can't write them directly in Kotlin code. You can see them in the compiler
```

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+diagnostics, displayed as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, a captured
```
daniCsorbaJB:
```suggestion
diagnostics, as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, a captured
```

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+type can appear in a type mismatch diagnostic:
```
daniCsorbaJB: (suggestion)
"can appear" can be a bit fuzzy - can appear when? ) WDYT about: Here's an example where a type mismatch diagnostic contains a captured type:

docs/topics/generics.md
daniCsorbaJB: Also I'm a bit uncertain what we mean by trivial in this context. 🤔

docs/topics/generics.md
```diff
@@ -267,6 +267,62 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+
+When you use a type projection like `out T` or `in T`, the compiler doesn't know the exact type argument behind it. In this case,
+the compiler internally represents the unknown concrete type as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown
+type within a known range. The compiler uses this range to decide which operations are safe:
+
+* When you read a value, the compiler uses the captured type's upper bound.
+* When you write a value, the compiler requires a valid lower bound.
+
+Usually, one of these bounds is trivial. For example, an `out T` projection gives the captured type an upper bound (`T`),
+but its lower bound is `Nothing`. This means that the compiler can read values as `T`, but can't write ordinary values.
```
daniCsorbaJB: What do we mean by ordinary? - that the user can't pass `T` values to it? 🤔

daniCsorbaJB (changes requested):
Thank you for adding this, I added some comments. In general, I think we could make this slightly more concise. 🤔

docs/topics/generics.md
aepanchi: Agree about anthropomorphism, but at the same time I think this description makes it easier for users to grasp the idea. Updated the wording though:)

docs/topics/generics.md
aepanchi: I believe it can establish a good knowledge graph for AI agents though.

docs/topics/generics.md
aepanchi: agree, the wording may be confusing. Updated

sarahhaggarty (approved):

docs/topics/generics.md
```diff
@@ -267,6 +267,57 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
 >
 {style="note"}
 
+### Captured types
+
+When you use a type projection, such as `out T` or `in T`, the compiler internally represents the unknown concrete type
+as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown type within a known range.
+
+Captured types are non-denotable, so you can't write them directly in Kotlin code. Instead, you can see captured types
```
daniCsorbaJB:
```suggestion
Captured types are non-denotable, so you can't write them directly in Kotlin code. Instead, you can most frequently see captured types
```

docs/topics/generics.md
```diff
@@ -267,6 +267,57 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+in the compiler diagnostics as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, the
```
daniCsorbaJB:
```suggestion
in compiler diagnostics, for example as `CapturedType(out X)` or in [flexible type](https://kotlinlang.org/spec/type-system.html#flexible-types) notation as `(L..U)`. For example, the
```

daniCsorbaJB: is flexible type relevant here? I feel like it weakens this explanation. 🤔 Could we remove it from here or am I missing something?

docs/topics/generics.md
```diff
@@ -267,6 +267,57 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+
+```kotlin
+val array: Array<out CharSequence> = arrayOf("str")
+val item: Int = array.get(0)
+// Initializer type mismatch: expected 'Int', actual 'CapturedType(out CharSequence)'
+```
+
+The compiler uses the range of a captured type to decide which operations are type-safe:
+
+* When you read a value, the compiler uses the captured type's upper bound.
```
daniCsorbaJB: 🤔 imho, this part still parses a bit strangely.
Perhaps it could be easier to write this part without an UL and merging the examples to make it more concise, WDYT? 

I'm thinking of something like:

> The compiler uses the captured type's upper bound for read operations and its lower bound to determine which values are type-safe for write operations: 

We could then combine the three examples into one, following this description, making this part a lot more concise without losing clarity:

```
// The projected type is Array<out CharSequence>
val array: Array<out CharSequence> = arrayOf("Kotlin")

// The get() read operation uses the captured type's upper bound, CharSequence
val item: CharSequence = array.get(0)

//The set() write operation uses the captured type's lower bound, Nothing, which results in an error
array.set(0, "New value")
// Receiver type 'Array<out CharSequence>' contains out projection
// which prohibits the use of 'fun set(index: Int, value: T): Unit'
```
We could then further explain this mechanism based on the example afterward.
like:

>In this example, the projected type is `Array<out CharSequence>`.  The compiler represents its projected type argument, `out CharSequence`, as a captured type with `CharSequence` as the upper bound and `Nothing` as the lower bound. 
>
>The captured type has `CharSequence` as its upper bound, so the compiler infers `CharSequence` as the type of `array.get(0)`, even though the actual element type may be `CharSequence` itself or one of its subtypes.
>
>Since the lower bound is `Nothing` and `Nothing` has no instances, writing to the projected type with the `set()` function isn't type-safe and results in an error.

docs/topics/generics.md
```diff
@@ -267,6 +267,57 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
 >
 {style="note"}
 
+### Captured types
+
+When you use a type projection, such as `out T` or `in T`, the compiler internally represents the unknown concrete type
+as a [*captured type*](https://kotlinlang.org/spec/type-system.html#type-capturing). A captured type is an unknown type within a known range.
```
daniCsorbaJB: with known bounds? 🤔 WDYT? - since further down we use the terms - upper bound lower bound etc.
We could even say: A captured type is an unknown type with known upper and lower bounds.

daniCsorbaJB (changes requested):
Nice improvements, I think it's getting there - I left a couple of additional minor suggestions and structural comments

docs/topics/generics.md
aepanchi: Yes, I had the same thoughts just wanted to highlight the connection. But I totally agree that it may no be reasonable here. Removed.

docs/topics/generics.md
aepanchi: Updated, but I think range is also fine here, since it has bounds by defenition. But I agree that new wording streamlines it more

docs/topics/generics.md
aepanchi: Updated

docs/topics/generics.md
```diff
@@ -267,6 +267,49 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+// The set() write operation uses the captured type's lower bound, Nothing,
+// which results in an error
+array.set(0, "New value")
+// Receiver type 'Array<out CharSequence>' contains out projection
+// which prohibits the use of 'fun set(index: Int, value: T): Unit'
+```
+
+In this example:
+
+* Array has the projected type `Array<out CharSequence>`. The compiler represents the projected type
```
daniCsorbaJB: does "Array" it refer to the variable `array`? 

I'd suggest either saying something like: 
The variable `array` has the projected type `...` 

But I think this is both more concise and more understandable, unless we're afraid people might not 
understand it immediately and need the extra clarification:

The projected type is `Array<out CharSequence>`.

docs/topics/generics.md
```diff
@@ -267,6 +267,49 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+argument `out CharSequence` as a captured type with `CharSequence` as its upper bound and `Nothing` as its lower bound.
+* For the `get()` operation, the compiler approximates the captured type to its upper bound and infers `CharSequence` as
+the type of `item`. The exact element type behind the projection can be `CharSequence` or one of its subtypes.
```
daniCsorbaJB: > The exact element type behind the projection can be `CharSequence` or one of its subtypes.

I think this is sounds complicated or perhaps I'm not seeing the connection this way: "element type behind the projection" 

WDYT would this work? 
This upper bound includes `CharSequence` and its subtypes.

docs/topics/generics.md
```diff
@@ -267,6 +267,49 @@ For example, if the type is declared as `interface Function<in T, out U>` you co
...
+* For the `set()` operation, the captured type has `Nothing` as its lower bound. Because `Nothing` has no instances, no
+value can satisfy this bound. Therefore, writing to the projected type with the `set()` function isn't type-safe and
```
daniCsorbaJB: I think we can make this part more concise:
Since `Nothing` has no instances, writing to the projected type isn't type-safe and results in an error.

daniCsorbaJB (changes requested):
Thank you, only a couple more suggestions 👍

docs/topics/generics.md
aepanchi: Yes, I must have forgotten to format it as a code. Rewrote to clarify

docs/topics/generics.md
aepanchi: I would't put it this way. The issue is that an upper bound does not really include types like a set or collection. CharSequence is the upper bound, and the captured type is constrained to be a subtype of it.
But your comment made me think that maybe we actually don't need to mention the possible subtypes here. The example already establishes why the exact type is unknown, and bringing subtypes back in makes the explanation more complex again. Rewrote.

daniCsorbaJB (approved):
Thank you! LGTM! 🚀

docs/topics/generics.md
daniCsorbaJB: Fair enough — I was also of the mindset that the easiest would be to remove this, as it makes the point more difficult to get across. I like the new version! 👍

PR #5645, opened 2026-06-30
update: add information about underscore syntax in unused exceptions

daniCsorbaJB:
This PR adds information about the underscore `_` syntax in unused exceptions

### DIFF
--- a/docs/topics/exceptions.md
+++ b/docs/topics/exceptions.md
@@ -234,6 +234,31 @@ fun count(): Int {
 ```
 {kotlin-runnable="true"}
 
+You can handle an exception without using the exception instance.
+This is useful when the `catch` block can handle the exception without it, for example by providing a fallback value or a generic error message.
+
+Use an underscore (`_`) instead of the exception parameter name to indicate that the exception instance is intentionally ignored:
+
+```kotlin
+import java.io.File
+import java.io.IOException
+
+//sampleStart
+fun main() {
+    val userSettings = try {
+        File("user-settings.json").readText()
+    
+    // Catches IOException without using the exception instance
+    } catch (_: IOException) {
+        // Uses a fallback value if loading the file fails
+        "{}"
+    }
+
+    println(userSettings)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
 
 You can use multiple `catch` handlers for the same `try` block.
 You can add as many `catch` blocks as needed to handle different exceptions distinctively.
--- a/docs/topics/exceptions.md
+++ b/docs/topics/exceptions.md
@@ -235,7 +235,7 @@ fun count(): Int {
 {kotlin-runnable="true"}
 
 You can handle an exception without using the exception instance.
-This is useful when the `catch` block can handle the exception without it, for example by providing a fallback value or a generic error message.
+For example, you can provide a fallback value or a generic error message in the `catch` block.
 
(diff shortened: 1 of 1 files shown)


sandwwraith (approved):

docs/topics/exceptions.md
```diff
@@ -234,6 +234,31 @@ fun count(): Int {
 ```
 {kotlin-runnable="true"}
 
+You can handle an exception without using the exception instance.
```
zamulla, line 237:
```suggestion
You can catch an exception without using an exception instance.
```

docs/topics/exceptions.md
```diff
@@ -234,6 +234,31 @@ fun count(): Int {
...
+This is useful when the `catch` block can handle the exception without it, for example by providing a fallback value or a generic error message.
```
zamulla: To avoid anthropomorphizing here, maybe something more like "when you don't need an exception instance to handle it, for example..."?

It's also maybe harder to figure out what "it" refers to here, so I'd repeat the phrase.

docs/topics/exceptions.md
```diff
@@ -234,6 +234,31 @@ fun count(): Int {
...
+    } catch (_: IOException) {
+        // Uses a fallback value if loading the file fails
+        "{}"
+    }
+
+    println(userSettings)
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
zamulla, line 261:
```suggestion
{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
```

Not sure why exactly, but it seems that 1.3 was a watershed version, and even simple "Hello, world" stuff is not working on Playground with the earlier version.

Not that anyone but us cares!

zamulla (approved):
LGTM, a couple of nitpicks there.

docs/topics/exceptions.md
daniCsorbaJB: tbh I don't like adding these unless it's for something recent (2 major releases prior) 😓 
The reason being that some functionalities like this one doesn't work on the playground for 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, etc (even though it should be available since 1.4 because of java API permission issues on the playground — it only works there since 2.3 so I find adding it misleading either way. If someone is testing it on the playground, according to the version, they might run into issues.)

docs/topics/exceptions.md
daniCsorbaJB: I agree that “catch” would work well outside this specific context. I chose “handle” because this section introduces the broader concept of handling exceptions, so it reinforces the existing terminology. Since the exception is both caught and handled here, “handle” felt more aligned with the section.

docs/topics/exceptions.md
daniCsorbaJB: Good point 🤔 I was thinking of the same thing. I chose the `catch` block wording to make it clear that the exception is already being handled in that block, but I agree that it makes the sentence less natural. 🤔  I also agree that it would help repeating “the exception".
My other issue was that it repeats the first sentence... So we could in theory merge the two like so:

> You can handle an exception without using the exception instance, for example by providing a fallback value or a generic error message in the `catch` block.

This is way more concise, which I used while drafting this update, but it felt a bit less explanatory for new users 🤔 But looking at it I think it should be better after all! 🙏

PR #5649, opened 2026-07-02
update: Gradle workers API

danil-pavlov:
* [KT-87071](https://youtrack.jetbrains.com/issue/KT-87071)


Tapchicoma (approved):

docs/topics/compiler-plugins/kapt.md
```diff
@@ -381,7 +365,22 @@ tasks.withType(org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask.class
 </tab>
 </tabs>
 
-### Caching for annotation processors' classloaders
+### Use Gradle build cache safely
+
+The kapt annotation processing tasks are [cached in Gradle](https://guides.gradle.org/using-build-cache/) by default.
```
daniCsorbaJB: This link redirects to https://docs.gradle.org/current/userguide/build_cache_use_cases.html — should we change it? 🤔

daniCsorbaJB: Could we flip this sentence around to avoid passive? 
Gradle caches kapt annotation processing tasks by default.

docs/topics/compiler-plugins/kapt.md
```diff
@@ -381,7 +365,22 @@ tasks.withType(org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask.class
...
+However, annotation processors can run arbitrary code, which may not reliably transform task inputs into outputs,
+or may access and modify files that Gradle doesn't track.
```
daniCsorbaJB: This sentence feels a bit on the longer side, do you think if we could break it up 🤔

docs/topics/compiler-plugins/kapt.md
```diff
@@ -381,7 +365,22 @@ tasks.withType(org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask.class
...
+If the annotation processors used in the build cannot be properly cached,
```
daniCsorbaJB:
```suggestion
If the annotation processors used in the build can't be properly cached,
```

daniCsorbaJB: since we used contracted doesn't above.

daniCsorbaJB: What do you think about flipping this, so the instruction is followed by the code example. 
I'm thinking of something like: 

Annotation processors that don’t support safe caching may cause false-positive cache hits for kapt tasks. To prevent this, set the `useBuildCache` property to `false` in your build script:

docs/topics/compiler-plugins/kapt.md
```diff
@@ -407,12 +406,28 @@ If you run into any problems with caching for annotation processors, disable cac
...
+
+```none
+kapt.incremental.apt=false
+```
+
+Note that incremental annotation processing requires [incremental compilation](gradle-compilation-and-caches.md#incremental-compilation)
+to be enabled as well.
+
+## Analyze performance
+
```
daniCsorbaJB, line 430: It would be nice to have a short intro - although for these two it might not be necessary, up to you, but I thought I'd mention 🤔

docs/topics/compiler-plugins/kapt.md
```diff
@@ -407,12 +406,28 @@ If you run into any problems with caching for annotation processors, disable cac
...
+kapt supports incremental annotation processing by default.
+Currently, annotation processing can be incremental only if all annotation processors being used are incremental.
+
+To disable incremental annotation processing, add this line to your `gradle.properties` file:
+
+```none
+kapt.incremental.apt=false
+```
+
+Note that incremental annotation processing requires [incremental compilation](gradle-compilation-and-caches.md#incremental-compilation)
```
daniCsorbaJB: I'd suggest putting this at the beginning of the 
something like: 

You must also enable [incremental compilation](gradle-compilation-and-caches.md#incremental-compilation) to use incremental annotation processing.

docs/topics/compiler-plugins/kapt.md
```diff
@@ -407,12 +406,28 @@ If you run into any problems with caching for annotation processors, disable cac
...
 > we would appreciate your feedback in [YouTrack](https://youtrack.jetbrains.com/issue/KT-28901).
-> 
+>
 {style="note"}
 
-### Measure performance of annotation processors
+### Use incremental annotation processing
+
+kapt supports incremental annotation processing by default.
+Currently, annotation processing can be incremental only if all annotation processors being used are incremental.
```
daniCsorbaJB:
```suggestion
Currently, annotation processing can be incremental only if all annotation processors used in the build are incremental.
```

daniCsorbaJB: or perhaps something like this, would this work to avoid the passive 🤔 : 

Currently, kapt uses incremental annotation processing only when all annotation processors in the build are incremental.

daniCsorbaJB (changes requested):
Nice update, I only left a couple of comments, let me know what you think! 🎉

docs/topics/compiler-plugins/kapt.md
danil-pavlov: Made a list of conditions instead

daniCsorbaJB: that's an even better solution — I like it 👍

daniCsorbaJB (approved):
LGTM! 🚀

PR #5657, opened 2026-07-06
[upd] this expressions

aepanchi:


docs/topics/this-expressions.md
```diff
@@ -1,42 +1,218 @@
 [//]: # (title: This expressions)
 
-To denote the current _receiver_, you use `this` expressions:
+The `this` expression refers to the current object or receiver.
```
serras: To be honest, I'm not a fan of saying "current object or receiver". `this` always refers to the current receiver, and it could be an instance of a class or interface, or an object.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,218 @@
...
+
+How you use `this` depends on the context:
 
 * In a member of a [class](classes.md#inheritance), `this` refers to the current object of that class.
-* In an [extension function](extensions.md) or a [function literal with receiver](lambdas.md#function-literals-with-receiver)
-`this` denotes the _receiver_ parameter that is passed on the left-hand side of a dot.
 
-If `this` has no qualifiers, it refers to the _innermost enclosing scope_. To refer to `this` in other scopes, _label qualifiers_ are used:
```
serras, line 9: This is actually very important: `this` doesn't only refer to the current receiver. It actually can refer to any of the nested receivers. For example, if I have

```kotlin
class Foo {
  fun foo() { }
}

class Bar {
  fun bar() { }

  fun Foo.moo() {
    this.bar()  // 'this' refers to the enclosing 'Bar' instance
    this.foo()  // 'this' refers to the extension receiver
  }
}
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,218 @@
...
+
+  fun main() {
+      println("Kotlin".lastCharacter())
+      // n
+  }
+  //sampleEnd
+  ```
+  {kotlin-runnable="true"}
+
+> In simple cases, you don't need to write `this` explicitly. Kotlin infers it from the current scope.
```
serras: The use of the word _infers_ is a bit overloaded in Kotlin already, and mostly refers to type inference.

serras (changes requested):

docs/topics/this-expressions.md
aepanchi: Added to the Qualified this section. Do you think it also makes sense to include the example like the one you've provided?

serras: I still think this is incorrect. You write "If `this` has no qualifier, it refers to the receiver of the _innermost enclosing scope_", but my example above shows that `this` may refer to different enclosing scopes, even if it appears unqualified.

aepanchi: Got it now, thank you! What do you think about smth like this:

Kotlin code can have several receivers available at the same time. This happens when receiver scopes are nested. For example, an extension function declared inside a class can access members of both the extension receiver and the enclosing class object.
Kotlin can use these receivers implicitly when you access their members. Unqualified this refers to the receiver with the highest priority in the current scope. To refer to another receiver explicitly, use a qualified this.

serras: The problem with your comment is in the "Unqualified this refers to the receiver with the highest priority in the current scope." part, unqualified `this` may refer to any receiver in scope, the closest one is just given more priority. So if, for example, several receivers have the same type, using the qualified `this` is the only possibility to access the outer one.

aepanchi: Yeah in that case my wording is too restrictive ig, so is the following would be correct: 

Your code can have several receivers available at the same time when receiver scopes are nested. Kotlin can use any available receiver to access its members implicitly, but receivers from inner scopes have higher priority. To explicitly refer to a particular receiver, use a qualified this. This is especially useful when multiple receivers have members with the same name and you need to access a member of an outer receiver.

serras: That's much better, thanks!

serras (approved):

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
 [//]: # (title: This expressions)
 
-To denote the current _receiver_, you use `this` expressions:
+The `this` expression refers to the current receiver. How you use `this` depends on the context:
```
zamulla: It's a little off-topic, but to someone new to Kotlin it's not actually clear what a receiver is, whether we mark it as a term or not. Since it's a relatively general OOP term, we can probably link somewhere to clarify, but I bet as is a lot of people either stumble on this or skip forward.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+  fun main() {
+      val language = Language("Kotlin")
+      language.printName()
+      // Kotlin
+  }
+  //sampleEnd
+  ```
+  {kotlin-runnable="true"}
+
+* In an [extension function](extensions.md) or a [function literal with receiver](lambdas.md#function-literals-with-receiver),
```
zamulla, line 25: Links forward (in the context of the ToC at least) are always risky, and actually I wonder whether `this` should be under "Functions" at all. It feels like it belongs both to "Classes" and "Functions" which usually means it belongs to neither.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+  fun main() {
+      println("Kotlin".lastCharacter())
+      // n
+  }
+  //sampleEnd
+  ```
+  {kotlin-runnable="true"}
+
+> In simple cases, you don't need to write `this` explicitly. Kotlin interprets it from the current scope.
+> Learn more about [](#implicit-this).
```
zamulla, line 46: Should we simply put the "Implicit `this`" section next instead of linking to the end of the same page? It's small enough that it shouldn't affect the flow in the same way as going to the bottom and then returning back up.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+> 
+{style="tip"}
+
+## Qualified this
```
zamulla:
```suggestion
## Qualified `this`
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+```
+{kotlin-runnable="true"}
+
+> If you create an anonymous object inside another class, you can use qualified `this` to access the outer class object.
+> Use it to access an outer receiver, not the anonymous object itself.
+> 
+{style="tip"}
+
 
 ## Implicit this
```
zamulla:
```suggestion
## Implicit `this`
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+    user.printUser()
+    // Name: JANE DOE
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+Here:
+
+* `this` refers to the extension receiver (the `"Jane Doe"` string).
```
zamulla:
```suggestion
* `this.uppercase()` refers to the extension function of `String`.
```

zamulla: Simply `this` is hard to attribute.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+
+    fun printUser() {
+        println("Jane Doe".formatName())
+    }
+}
+
+fun main() {
+    val user = User()
+    user.printUser()
+    // Name: JANE DOE
```
zamulla, line 125: Nit: a `User` class that doesn't have even a hardcoded name property reads very synthetic to me :) it's an extra line, but still.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+
+Here:
+
+* `this` refers to the extension receiver (the `"Jane Doe"` string).
+* `this@User` refers to the current `User` object.
+* `this@User.prefix` accesses the prefix property of the current `User` object.
+
+### Access from a lambda
+
+Unlike a regular lambda, a [lambda with receiver](lambdas.md#function-literals-with-receiver) has its own receiver. Therefore, `this` inside it
```
zamulla: "Lambda with receiver has a receiver" is funny :)

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+refers to that receiver. If a lambda with receiver is nested inside another receiver scope, add a label to the lambda
```
zamulla: I think the key to the funny is "that receiver" which is not immediately opposed by something else. "To that receiver... instead of an outside scope".

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+
+* `stringLabel@` labels the lambda.
+* `this@stringLabel` refers to the string receiver of the lambda.
+* `this@User` refers to the current `User` object.
+
+The label doesn't call anything or change how the lambda works. It only helps you refer to the lambda's receiver.
+
+### Access an anonymous object
+
+Since anonymous objects don't have class names, you can't use a name like `this@YourClassName` to refer
```
zamulla:
```suggestion
Since [anonymous objects](object-declarations.md#object-expressions) don't have class names, you can't use a name like `this@YourClassName` to refer
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+to it. Inside an anonymous object, use plain `this` to refer to the object:
```
zamulla:
```suggestion
to it. Inside an anonymous object, use unqualified `this` to refer to the object:
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+      
+        override fun print() {
+            // Refers to the name property of the anonymous object
+            println(this.name)
+        }
+    }
+    printUser(printer)
+    // Jane Doe
+}
+//sampleEnd
```
zamulla: Since `// sampleStart` and `// sampleEnd` coincide with the code block, can we get rid of them? The rest of the samples are the same, maybe I'm missing something.

docs/topics/this-expressions.md
zamulla: There is a term conflict here: above the very same link has text "function literal with receiver", here it's "lambda with receiver". We should the same term to refer to the same thing; probably "lambda" since it's the shorter expression (and that's why we use in the rest of the docs)?

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
 [//]: # (title: This expressions)
 
-To denote the current _receiver_, you use `this` expressions:
+The `this` expression refers to the current receiver. How you use `this` depends on the context:
 
 * In a member of a [class](classes.md#inheritance), `this` refers to the current object of that class.
```
zamulla:
```suggestion
* In a member of a [class](classes.md), `this` refers to the current object of that class.
```

docs/topics/this-expressions.md
zamulla, line 255: Since we're here, let's make this runnable too? :)

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+    }
+    printUser(printer)
+    // Jane Doe
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> If you create an anonymous object inside another class, you can use qualified `this` to access the outer class object.
+> Use it to access an outer receiver, not the anonymous object itself.
```
zamulla: I think I'd appreciate a sample for this, too, as it's hardly a side note. And the section may be called "Access an anonymous object or its outer class", since we don't have that covered anywhere else on the page — even though it's arguably intuitive from the basic rules of `this`.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+      // Kotlin
+  }
+  //sampleEnd
+  ```
+  {kotlin-runnable="true"}
+
+* In an [extension function](extensions.md) or a [function literal with receiver](lambdas.md#function-literals-with-receiver),
+`this` refers to the receiver. 
+
+  In the following example, we call `lastCharacter()`on the string
```
zamulla:
```suggestion
  In the following example, `lastCharacter()` is called on the string
```

To avoid first person where it's not necessary.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+Here:
+
+* `this` refers to the extension receiver (the `"Jane Doe"` string).
+* `this@User` refers to the current `User` object.
+* `this@User.prefix` accesses the prefix property of the current `User` object.
```
zamulla:
```suggestion
* `this@User.prefix` accesses the `prefix` property of the current `User` object.
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+
+  fun main() {
+      println("Kotlin".lastCharacter())
+      // n
+  }
+  //sampleEnd
+  ```
+  {kotlin-runnable="true"}
+
+> In simple cases, you don't need to write `this` explicitly. Kotlin interprets it from the current scope.
```
zamulla:
```suggestion
> In simple cases, you don't need to write `this` explicitly. Kotlin resolves it from the current scope.
```

Nitnit.

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+
+Unlike a regular lambda, a [lambda with receiver](lambdas.md#function-literals-with-receiver) has its own receiver. Therefore, `this` inside it
+refers to that receiver. If a lambda with receiver is nested inside another receiver scope, add a label to the lambda
+and use qualified `this` to specify which receiver you need:
+
+```kotlin
+//sampleStart
+class User(val name: String) {
+    fun printWithPrefix() {
+        val print: String.() -> Unit = stringLabel@ {
```
zamulla: Talking about name resolution, maybe let's not call custom stuff after stdlib functions :)
```suggestion
        val printString: String.() -> Unit = stringLabel@ {
```

docs/topics/this-expressions.md
```diff
@@ -1,42 +1,217 @@
...
+            println("${this@stringLabel}: ${this@User.name}")
         }
+      
+        print("User")
```
zamulla:
```suggestion
        printString("User")
```
See above.

docs/topics/this-expressions.md
zamulla: "`this` inside it refers to that receiver" is not illustrated in the sample, and I think it would be beneficial to see the examples side-by-side like in other sections.

zamulla (changes requested):
An article about basics is a great time to nitpick. Most of the comments are superficial, but there are a couple that I think can affect clarity seriously.

docs/topics/this-expressions.md
aepanchi: I wanted to do so, but I'm not sure there is a good ebough place to link to tbh. The wiki page is okay, but I don't like that it's called command pattern

docs/topics/this-expressions.md
aepanchi: Agree, I tried different places but I don't want to be the one who suggests to change the toc AGAIN :D

docs/topics/this-expressions.md
aepanchi: It would make sense, but it describes qualified vs unqualified calls that we explain later

docs/topics/this-expressions.md
aepanchi: Agree, but `uppercase()` isn’t an extension function of String in the sense your wording explains. More importantly, the point is that `this` is the extension receiver, not that `uppercase()` is an extension. I have rewritten that bullets, let me know what of think of them!

docs/topics/this-expressions.md
aepanchi: Hm, but the page to whichj we link says: lambda expressions can be used as function literals with receiver when the receiver type can be inferred from the context. So I wouldn't treat these terms as identical.

docs/topics/this-expressions.md
aepanchi: But it's already runnable?

docs/topics/this-expressions.md
aepanchi: done

docs/topics/this-expressions.md
aepanchi: It's funny cause I used first person to avoid passive :D

docs/topics/this-expressions.md
aepanchi: Oh, good point! Thanks

docs/topics/this-expressions.md
zamulla: We can just define it right here, "receiver (object in which context the code is executed)" or something. The wiki page is highly abstract and linking there I don't think will add a lot of value.

docs/topics/this-expressions.md
zamulla: GitHub doesn't want to show me the last line apparently. Probably because we need an empty line before EOF.

docs/topics/this-expressions.md
zamulla: Actually I think I misunderstood here, it's fine as is :)

docs/topics/this-expressions.md
```diff
@@ -1,44 +1,226 @@
...
+interface UserPrinter {
+    fun print()
+}
+
+fun main() {
+    val printer = object : UserPrinter {
+        val prefix = "User"
+        
+        override fun print() {
+            // this refers to the anonymous object, so this.prefix accesses its `prefix` property
```
zamulla:
```suggestion
            // `this` refers to the anonymous object, so `this.prefix` accesses its `prefix` property
```
With the human-readable syntax it's hard to see immediately whether the lower-case letter is a typo or a keyword (it's all the more confusing because `prefix` is marked). I don't think we have a guide line for that, but it just makes sense™

docs/topics/this-expressions.md
```diff
@@ -1,44 +1,226 @@
...
+
+interface UserPrinter {
+    fun print()
+}
+
+class User(val name: String) {
+    fun createPrinter(): UserPrinter {
+        return object : UserPrinter {
+            override fun print() {
+                // this@User refers to the enclosing `User` object, so this@User.name accesses its name property
```
zamulla: I can just see the warning about the code line being too long for the web page, let's add a newline after the comma?

zamulla (approved):
Thanks! I left a couple of final comments, but they are hardly consequential 👍

PR #5665, opened 2026-07-09
Fix error message in functions.md example

SaSteMi:
Hi I noticed a small mistake at [Language Guide/Functions/Functions/Function Usage/Parameters with default values](https://kotlinlang.org/docs/functions.html#parameters-with-default-values) in the second code example


sarahhaggarty (approved):

sarahhaggarty:
Good catch! Thanks for your contribution @SaSteMi!

PR #5667, opened 2026-07-12
feat: add memory allocation details for SAM conversions

sakethpathike:
This PR adds a new section explaining the memory allocation behavior when using SAM conversions. 

I am not exactly sure if ~~this should have been~~ I should have placed this in the lambdas page or here, but since I was using SAM and got to know about the allocation from Baeldung, there might be common cases where this explicit explanation might make sense alongside the usual anonymous object's allocation.

### DIFF
--- a/docs/topics/fun-interfaces.md
+++ b/docs/topics/fun-interfaces.md
@@ -62,6 +62,25 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.4"}
 
+### Memory allocation
+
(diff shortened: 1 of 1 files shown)


sarahhaggarty:
Hi @sakethpathike!
Thanks for your suggestion and for thoroughly researching the behavior of SAM conversions. I checked with the team, and although what you've written is correct, we can't guarantee that this behavior will remain the same in future releases. For that reason, I'm not going to add it to the documentation.

Thanks again for your time, and I hope you'll consider contributing again in the future!

PR #5669, opened 2026-07-13
Clarify API and language version behavior

sarahhaggarty:
This PR adds more explanation for the behavior of API and language versions in the compiler reference and updates the related pages.


3flex:
Suggestion: adding a note about the metadata version also being controlled by setting the language version. If targeting a platform with embedded Kotlin (such as a Gradle plugin written in Kotlin) the metadata version needs to align to the metadata versions supported by the embedded Kotlin version.

If it doesn't belong with this doc update I can raise a KT issue.

docs/topics/gradle/gradle-compiler-options.md
```diff
@@ -597,14 +597,14 @@ Here is a complete list of options for the Gradle compiler:
...
+| `suppressWarnings`    | Don't generate warnings                                                                                                                                                              |                                                         | false         |
+| `verbose`             | Enable verbose logging output. Works only when the [Gradle debug log level enabled](https://docs.gradle.org/current/userguide/logging.html)                                          |                                                         | false         |
+| `freeCompilerArgs`    | A list of additional compiler arguments. You can use experimental `-X` arguments here too. See an [example](#example-of-additional-arguments-usage-via-freecompilerargs)             |                                                         | []            |
+| `apiVersion`          | Control which Kotlin APIs your code can use. For more information, see [`-api-version`](compiler-reference.md#api-version-version).                                                   | "2.0", "2.1", "2.2", "2.3", "2.4", "2.5" (EXPERIMENTAL) |               |
```
kunyavskiy, line 606: The list of valid arguments changes from version to version, not sure if it's a good way to specify it.

docs/topics/kotlin-evolution-principles.md
```diff
@@ -206,20 +206,24 @@ and even when it's over and the change ships in a stable version, there's still
...
-We provide compatiblity options that make a new Kotlin version emulate the behavior of an old one for compatibility purposes:
+We provide compatibility options that make a new Kotlin version emulate the behavior of an older one for compatibility purposes:
 
-* `-language-version X.Y` - compatibility mode for Kotlin language version X.Y, reports errors for all language features
-  that came out later.
-* `-api-version X.Y` - compatibility mode for Kotlin API version X.Y, reports errors for all code using newer APIs from
-  the Kotlin Standard Library (including the code generated by the compiler).
+* `-language-version X.Y` – compatibility mode for Kotlin language version X.Y. Reports errors when your code uses language
+  features introduced in later versions.
+* `-api-version X.Y` – compatibility mode for Kotlin API version X.Y. Reports errors when your code uses Kotlin standard
```
kunyavskiy: That's not completely true. You wouldn't get errors as if "you can't call this declaration, it's too new". The compiler would pretend this declaration doesn't exist at all, this can lead to unresolved reference, but also to just choosing diffrent overload, if it is availiable.

docs/topics/kotlin-evolution-principles.md
```diff
@@ -206,20 +206,24 @@ and even when it's over and the change ships in a stable version, there's still
...
+  library APIs introduced in later versions, including APIs referenced by compiler-generated code.
 
-To give you more time for migration, we support the development for at least three previous language and API versions
-in addition to the latest stable one.
+To give you more time to migrate, we support at least three previous language and API versions in addition to the
```
kunyavskiy: That's only true for jvm.

kunyavskiy: To be more precice, you can use older LV with other backends, but it wouldn't produce more compatible binary, just would make you unable to use new features, so it's not very useful.

kunyavskiy (changes requested):

docs/topics/gradle/gradle-compiler-options.md
sarahhaggarty: We have an action item with every release to manually update this line so we won't miss it.

docs/topics/kotlin-evolution-principles.md
sarahhaggarty: I updated the text. Is this correct now?

docs/topics/kotlin-evolution-principles.md
sarahhaggarty: I updated the text. Is this correct now?

docs/topics/kotlin-evolution-principles.md
```diff
@@ -208,13 +208,15 @@ and even when it's over and the change ships in a stable version, there's still
 
 We provide compatibility options that make a new Kotlin version emulate the behavior of an older one for compatibility purposes:
 
-* `-language-version X.Y` – compatibility mode for Kotlin language version X.Y. Reports errors when your code uses language
+* `-language-version X.Y` – compatibility mode for Kotlin language version X.Y. The compiler reports errors when your code uses language
   features introduced in later versions.
-* `-api-version X.Y` – compatibility mode for Kotlin API version X.Y. Reports errors when your code uses Kotlin standard
+* `-api-version X.Y` – compatibility mode for Kotlin API version X.Y. The compiler ignores declarations in your code that use Kotlin standard
```
kunyavskiy: Maybe something like? 

The compiler would ignore declarations from dependencies and standard library, marked to be introduced later that the mentioned version by (SinceKotlin)[https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-since-kotlin/] annotation. This would lead to unresolved reference or resolution to alternative existing eariler if available on attempt to use them. That way compiled code would be compatible to run with corresponding old standard library.

docs/topics/kotlin-evolution-principles.md
```diff
@@ -206,20 +206,27 @@ and even when it's over and the change ships in a stable version, there's still
...
-We provide compatiblity options that make a new Kotlin version emulate the behavior of an old one for compatibility purposes:
+We provide compatibility options that make a new Kotlin version emulate the behavior of an older one:
 
-* `-language-version X.Y` - compatibility mode for Kotlin language version X.Y, reports errors for all language features
-  that came out later.
-* `-api-version X.Y` - compatibility mode for Kotlin API version X.Y, reports errors for all code using newer APIs from
-  the Kotlin Standard Library (including the code generated by the compiler).
+* `-language-version X.Y` – compatibility mode for Kotlin language version X.Y. The compiler reports errors when your code uses language
+  features introduced in later versions.
+* `-api-version X.Y` – compatibility mode for Kotlin API version X.Y. The compiler ignores declarations in your code that use Kotlin standard
```
kunyavskiy: not in your code, but in libraries you use.

docs/topics/kotlin-evolution-principles.md
```diff
@@ -206,20 +206,27 @@ and even when it's over and the change ships in a stable version, there's still
...
+  library APIs introduced in later versions, including APIs referenced by compiler-generated code.
 
-To give you more time for migration, we support the development for at least three previous language and API versions
-in addition to the latest stable one.
+To give you more time to migrate, on the JVM we support at least three previous language and API versions in addition to the 
+latest stable version. On other platforms, you can use older language and API versions, but then you can't use newer
```
kunyavskiy: I didn't get it. you can't use it on either platform. The problem on other platforms is "you wouldn't be able to use, but it wouldn't help your consumers to update faster, as they would still need newest Kotlin to use your library".

kunyavskiy (changes requested):

docs/topics/kotlin-evolution-principles.md
sarahhaggarty: Would it be OK if I just removed "in your code"?

docs/topics/kotlin-evolution-principles.md
sarahhaggarty: You're right. The second sentence isn't making the right point. I adjusted the text to focus the difference on being about whether consumers can use older compiler versions or not.

kunyavskiy (approved):

sarahhaggarty:
> Suggestion: adding a note about the metadata version also being controlled by setting the language version. If targeting a platform with embedded Kotlin (such as a Gradle plugin written in Kotlin) the metadata version needs to align to the metadata versions supported by the embedded Kotlin version.
> 
> If it doesn't belong with this doc update I can raise a KT issue.

Hi @3flex!
Sorry for the slow reply. It's summer vacation time so it took longer than I expected to discuss this within the team.
Firstly, thanks for the suggestion! The point you make about metadata version is correct, but after discussing it with the team, we don’t think we should introduce metadata version as part of these updates.

The metadata version follows the language version used for compilation, so the guidance here is still to choose a language version that is compatible with the Kotlin version in the target environment. Introducing metadata version separately would mean explaining an additional, fairly advanced concept without changing the actual recommendation.

There's already information about metadata compatibility in the `kotlinx-metadata-jvm` [documentation](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm#detailed-explanation) for cases where users need to work with metadata directly.

So for now, we'll keep these updates focused on language-version compatibility. Thanks for raising it though!

docs/topics/compiler/compiler-reference.md
```diff
@@ -52,7 +52,16 @@ The following options are common for all Kotlin compilers.
...
 ### -api-version _version_
 
-Allow using declarations only from the specified version of Kotlin bundled libraries.
+Set the API version to control which Kotlin APIs your code can use at runtime. For example, if you use Kotlin compiler version 2.4.0
+with `-api-version=2.1`, your code remains compatible with Kotlin standard library 2.1.0.
+
+The `-api-version` value can't be higher than the `-language-version` value.
+
+In most cases, the API version and [language version](#language-version-version) should be the same. One exception is
+when you develop a library for consumers who must run an older version of the Kotlin standard library. In that
```
koshachy, line 61:
```suggestion
when you develop a library for consumers who must run an older version of the Kotlin Standard Library. In that
```

docs/topics/compiler/compiler-reference.md
```diff
@@ -52,7 +52,16 @@ The following options are common for all Kotlin compilers.
 
 ### -api-version _version_
 
-Allow using declarations only from the specified version of Kotlin bundled libraries.
+Set the API version to control which Kotlin APIs your code can use at runtime. For example, if you use Kotlin compiler version 2.4.0
+with `-api-version=2.1`, your code remains compatible with Kotlin standard library 2.1.0.
```
koshachy, line 56:
```suggestion
with `-api-version=2.1`, your code remains compatible with Kotlin Standard Library 2.1.0.
```

docs/topics/compiler/compiler-reference.md
```diff
@@ -52,7 +52,16 @@ The following options are common for all Kotlin compilers.
 
 ### -api-version _version_
 
-Allow using declarations only from the specified version of Kotlin bundled libraries.
+Set the API version to control which Kotlin APIs your code can use at runtime. For example, if you use Kotlin compiler version 2.4.0
```
koshachy, line 55: "can use" or just "use"? 🤔

docs/topics/compiler/compiler-reference.md
```diff
@@ -52,7 +52,16 @@ The following options are common for all Kotlin compilers.
...
+with `-api-version=2.1`, your code remains compatible with Kotlin standard library 2.1.0.
+
+The `-api-version` value can't be higher than the `-language-version` value.
```
koshachy: To discuss: You can't set the ... value higher then the.... 

Will it be a bit more direct? Or we can lose the informaiton about the restriction? 🤔

docs/topics/compiler/compiler-reference.md
```diff
@@ -52,7 +52,16 @@ The following options are common for all Kotlin compilers.
...
+
+In most cases, the API version and [language version](#language-version-version) should be the same. One exception is
+when you develop a library for consumers who must run an older version of the Kotlin standard library. In that
+case, set an older API version to help ensure that you don't accidentally use APIs that aren't available to those consumers.
```
koshachy: `to help ensure that you don't accidentally use` -> `to avoid accidentally using`

WIll it works? 🤔

docs/topics/compiler/compiler-reference.md
```diff
@@ -65,9 +74,23 @@ Specify a custom path to the Kotlin compiler used for the discovery of runtime l
...
 ### -language-version _version_
 
-This option sets the supported syntax and semantics according to the specified language version. For example, using
-Kotlin compiler version 2.4.0 with `-language-version=2.2` lets you use only the language features and standard library APIs from
-version 2.2 or earlier. This can help with a gradual migration to newer Kotlin versions.
+Set the language version to control which Kotlin language features are available during compilation. 
+
+For example, if
+you want to benefit from new compilation performance improvements without changing compiler behavior,
+you can use a new compiler version with an older language version. If you use an older language version, you can't
```
koshachy: With an older language version, you can't use newer language features

To exclude use+use

docs/topics/compiler/compiler-reference.md
```diff
@@ -65,9 +74,23 @@ Specify a custom path to the Kotlin compiler used for the discovery of runtime l
 
 ### -language-version _version_
 
-This option sets the supported syntax and semantics according to the specified language version. For example, using
-Kotlin compiler version 2.4.0 with `-language-version=2.2` lets you use only the language features and standard library APIs from
-version 2.2 or earlier. This can help with a gradual migration to newer Kotlin versions.
+Set the language version to control which Kotlin language features are available during compilation. 
+
+For example, if
+you want to benefit from new compilation performance improvements without changing compiler behavior,
```
koshachy, line 80: The first a bit complicated but still fine.
WDYT if we swap the parts, will it be a bit more easier to read? 🤔 

For example, you can use a newer compiler version with an older language version to benefit from compilation performance improvements without changing compiler behavior.

docs/topics/compiler/compiler-reference.md
```diff
@@ -65,9 +74,23 @@ Specify a custom path to the Kotlin compiler used for the discovery of runtime l
...
+For more information, see [Backward compatibility guidelines for library authors](api-guidelines-backward-compatibility.md#choose-compatible-language-and-api-versions).
+
+You can configure one of the latest three stable versions of Kotlin as the language version. For example, Kotlin 2.5.0
+supports language versions as old as 2.2.
+
+If you use an older language version, you also need to use an older API version.
+For more information, see [](#api-version-version).
+
+Technically, you can configure a newer language version to try pre-stable language features before they are stabilized.
+However, we recommend enabling individual features by following their dedicated instructions.
```
koshachy: I guess it is about upcoming laguage versions? Such as 2.5?

It is not clear from the text that it is about it 🤔

koshachy: Also we could change it to a tip?

koshachy: It is not critical, but we also could get rid of "we" here (but I think it is not a problem in this context):

However, it is better/recommended to enable individual features by following their dedicated instructions.

koshachy (changes requested):
Thank you!
See my comments. Feel free to start a dicsussion.
Cheers!

docs/topics/compiler/compiler-reference.md
sarahhaggarty: I'd say "can use" because the compiler will only allow use of APIs with the correct version :)

docs/topics/compiler/compiler-reference.md
sarahhaggarty: Yes! Nice one.

docs/topics/compiler/compiler-reference.md
sarahhaggarty: WDYT about "However, as a best practice, enable individual features by following their dedicated instructions?" Or the closest to your suggestion: "However, it's best to enable individual features by following their dedicated instructions." 🤔

sarahhaggarty: And yes, it is about upcoming language features so I swapped "pre-stable" for "upcoming" to try and make that clearer :)

docs/topics/compiler/compiler-reference.md
sarahhaggarty: I agree it would read more easily if it was swapped but I wrote it in this order to highlight the most common use case early on in the sentence. I worry that if we swap it, readers might skip over the sentence 🤔

docs/topics/compiler/compiler-reference.md
sarahhaggarty: I've always used lowercase since the standard library isn't a product name but I see that it's written in uppercase in evolution principles so it should be consistent. We've got 60 cases of lowercase and 7 cases of uppercase. Is it alright if I change the 7 to lowercase?

koshachy: Let's discuss it in another place.
Our documentation is not the source of truth, the stdlib reference is.
API references, Kotlin sources, our marketing materials uses Standard Library a lot.

docs/topics/compiler/compiler-reference.md
koshachy: So you decided not to use the suggestion?

docs/topics/compiler/compiler-reference.md
koshachy: I'm OK with it, i just wanted to highlight that the sentence is hard to read.

docs/topics/compiler/compiler-reference.md
sarahhaggarty: Sorry, missed pushing the change. Fixed now.

koshachy (approved):
LGTM, thank you!

PR #5671, opened 2026-07-13
upd: document adapted function references

aepanchi:


aepanchi:
Close, since such documentation is not needed at this point.

PR #5672, opened 2026-07-14
ktl-4696 feat: update education page

krutilov:


berezinant (approved):
we might want to rename it to backend later but it's ok for now

PR #5676, opened 2026-07-14
[upd] type aliases

aepanchi:


docs/topics/type-aliases.md
```diff
@@ -1,100 +1,141 @@
...
+    }
+}
+
+fun synchronizeUsers(users: UserRepository.UserIndex) {
+  // ...
 }
 ```
 
-To fix this issue, declare the type parameter directly in the type alias:
+The containing declaration establishes the alias's scope, but visibility can restrict access further. Therefore,
```
e2e4b6b7: There are two places with the description of the fact that type aliases could have visibility modifier. Probably it is better to just mention it in some shared section (first one or "Declare type aliases"). Also I do not think that such a detailed explanation of how visibility works and what kinds of visibilities are there is needed. It works the same way as for any other entity.

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,141 @@
...
+the effective accessibility of a nested alias depends on both the alias and its containing declaration. For example, you can't access
+a `public` alias inside an `internal` class from outside the module because the containing class isn't accessible there.
+
+A type alias also can't expose an underlying type whose visibility is more restrictive than the alias. For example, a `public`
```
e2e4b6b7: This applies not only to nested type aliases, but to both of them. Probably it should be in any of the common sections

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,141 @@
...
+
+// user-service.kt
+package org.example.services
+
+import org.example.users.UserId
 }
 ```
 
-Capturing means that the type alias refers to a type parameter defined in the outer class:
+The package determines whether you can refer to the alias by its unqualified name or need to import it or use its fully
```
e2e4b6b7: I do not understand what is the meaning of this first sentence. What new idea does it introduces compared to the text before the snippet?

docs/topics/type-aliases.md
aepanchi: Fair enough, removed description of visibility modifiers

docs/topics/type-aliases.md
aepanchi: removed

docs/topics/type-aliases.md
aepanchi: Moved visability info to the top section and proveded relevant link to avoid additional explanation

docs/topics/type-aliases.md
```diff
@@ -52,6 +52,11 @@ You can declare a type alias:
 
 You can't declare a type alias in a local scope, such as inside a function or lambda expression.
 
+The declaration location determines the scope of a type alias, while its visibility determines which code can access it.
+By default, a type alias is `public`. The effective accessibility of a type alias also depends on the accessibility of
+its containing declarations and visibility of the underlying type. A type alias can't expose a type whose visibility is
```
e2e4b6b7: > The effective accessibility of a type alias also depends on the accessibility of its containing declarations and visibility of the underlying type

Effective accessibility of a type alias does not depends on the visibility of the underlying type. The reason for that is in your next sentence: "A type alias can't expose a type whose visibility is more restrictive than the alias". So, the correctness of the visibility of underlying type is checked in regards to typealias' visibility, but it does not affect the visibility furthermore

e2e4b6b7 (approved):

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+typealias DatabaseTransaction = Database.Transaction
 ```
 
-Type aliases do not introduce new types. 
-They are equivalent to the corresponding underlying types.
-When you add `typealias Predicate<T>` and use `Predicate<Int>` in your code, the Kotlin compiler always expands it to `(Int) -> Boolean`. 
-Thus you can pass a variable of your type whenever a general function type is required and vice versa:
+A type alias doesn't create a new type. It introduces an alternative name for an existing type. The alias and its underlying
+type are interchangeable. For example, when you add `typealias Predicate<T>` and use `Predicate<Int>`, the compiler
+expands it to `(Int) -> Boolean`. You can use a value of the aliased type wherever the underlying type is expected, and
```
daniCsorbaJB:
```suggestion
expands it to `(Int) -> Boolean`. You can use a value declared with the alias wherever the underlying type is expected, and
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+vice versa:
```
daniCsorbaJB:
```suggestion
the other way around:
```

daniCsorbaJB: just to replace the latin with English 👍

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
 }
 ```
 {kotlin-runnable="true"}
 
-## Nested type aliases
+## Declare type aliases
+
+You can declare a type alias:
+
+* At the top level of a Kotlin file (top-level type aliases).
```
daniCsorbaJB: I don't think the UL adds much here. 🤔  

WDYT about something more concise? 🤔 
You can declare a type alias at the top level of a Kotlin file (a top-level type alias) or inside a class, interface, or object (a nested type alias).

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+* Inside a class, interface, or object (nested type aliases).
 
-In Kotlin, you can define type aliases inside other declarations, as long as they
-don't capture type parameters from their outer class:
+You can't declare a type alias in a local scope, such as inside a function or lambda expression.
```
daniCsorbaJB:
```suggestion
You can't declare a type alias in a local scope, such as inside a function or [lambda expression](lambdas.md#lambda-expressions-and-anonymous-functions).
```

daniCsorbaJB: Since this a bit more advanced concept, let's link to it

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+
+The declaration location determines the scope of a type alias, while its visibility determines which code can access it.
+By default, a type alias is `public`. The effective accessibility of a type alias depends on its visibility and the accessibility
+of its containing declarations. For example, a `public` alias inside an `internal` class isn't accessible from outside
+the module because the containing class isn't accessible there. When you declare a type alias, its underlying type must
```
daniCsorbaJB:
```suggestion
the module because the containing class isn't accessible there. 

When you declare a type alias, its underlying type must
```

daniCsorbaJB: must be at least as visible sounds a bit mechancial 🤔
WDYT about having a simpler sentence — the example you use below demonstrates this beautifully we could even cut the last sentence:

A type alias can't expose a type with more restrictive [visibility](visibility-modifiers.md) than its own. 
For example, ...

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+be at least as visible as the alias. A type alias can't expose a type whose visibility is more restrictive than the alias.
+For example, a `public` type alias can't refer to a `private` class. Learn how visibility affects access in [](visibility-modifiers.md).
```
daniCsorbaJB:
```suggestion
For example, a `public` type alias can't refer to a `private` class.
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+
+* At the top level of a Kotlin file (top-level type aliases).
+* Inside a class, interface, or object (nested type aliases).
 
-In Kotlin, you can define type aliases inside other declarations, as long as they
-don't capture type parameters from their outer class:
+You can't declare a type alias in a local scope, such as inside a function or lambda expression.
+
+The declaration location determines the scope of a type alias, while its visibility determines which code can access it.
+By default, a type alias is `public`. The effective accessibility of a type alias depends on its visibility and the accessibility
```
daniCsorbaJB:
```suggestion
By default, a type alias is `public`. A nested type alias is accessible only where its containing class, interface, or object is accessible.
```

daniCsorbaJB: If I understand correctly, this is just for nested aliases. For top-level ones, only their own visibility applies, no? 🤔

docs/topics/type-aliases.md
daniCsorbaJB:
```suggestion
the module. When you declare a type alias, its underlying type must
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+
+### Top-level type aliases
+
+A top-level type alias is a package-level declaration. Within the same package, you can refer to an alias by its unqualified name.
+To use the alias from another package, import the alias or refer to it by its qualified name:
 
 ```kotlin
-class Dijkstra {
-    typealias VisitedNodes = Set<Node>
+// user-id.kt
```
daniCsorbaJB:
```suggestion
// UserId.kt
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+package org.example.users
+
+typealias UserId = Long
 
-    private fun step(visited: VisitedNodes, ...) = ...
+// user-service.kt
```
daniCsorbaJB:
```suggestion
// UserService.kt
```

docs/topics/type-aliases.md
daniCsorbaJB: Just to follow [Kotlin naming conventions](https://kotlinlang.org/docs/coding-conventions.html#source-file-names)

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+package org.example.services
+
+import org.example.users.UserId
```
daniCsorbaJB, line 81: could we add an unqualified name here? 🙏 

something like (but up to you )

fun findUser(id: UserId) { // ... 
}

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
 }
 ```
 
-Capturing means that the type alias refers to a type parameter defined in the outer class:
+### Nested type aliases
+
+Declare a type alias inside a class, interface, or object when the alternative name is relevant only in the context of
+that declaration. This keeps the alias close to the code that uses it and avoids adding another name to the package scope.
+Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
```
daniCsorbaJB, line 96:
```suggestion

Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
```

daniCsorbaJB: hmm 🤔 
I think here we should focus on the rules first WDYT?

We could put this sentence about their use case to the top of the section: 
> Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter, and simplifying internal implementations.

Then we could keep this separately here.

Nested type aliases follow the same scope and name resolution rules as [nested classes](link).

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+    }
+}
+
+fun synchronizeUsers(users: UserRepository.UserIndex) {
+  // ...
 }
 ```
 
-To fix this issue, declare the type parameter directly in the type alias:
+> Nested type aliases are not supported in Kotlin Multiplatform [`expect/actual` declarations](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html).
```
daniCsorbaJB:
```suggestion
> Nested type aliases aren't supported in Kotlin Multiplatform [`expect/actual` declarations](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html).
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+Within the containing declaration, you can refer to the alias by its unqualified name. Outside the declaration, qualify
+the alias with the name of its containing declaration:
 
 ```kotlin
-class Graph<Node> {
-    // Incorrect because captures Node
-    typealias Path = List<Node>
+class UserRepository {
+    typealias UserIndex = Map<UserId, User>
+
```
daniCsorbaJB, line 108:
```suggestion

    // Refers to the alias by its unqualified name inside UserRepository
```

daniCsorbaJB: (suggestion)
perhaps overkill, but I thought it could help up to you

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+    fun saveAll(users: UserIndex) {
+        // ...
+    }
+}
+
```
daniCsorbaJB, line 114:
```suggestion

// Refers to the alias by its qualified name outside UserRepository
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
-    typealias Path<Node> = List<Node>
+    // Error: Path captures Graph's Node type parameter
+    typealias Path = List<Node>
 }
 ```
 
-Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
-and simplifying internal implementations.
+In this example, `Path` can't use `Node` because a nested type alias doesn't capture type parameters from its containing class.
+To make the alias generic, declare its own type parameter:
```
daniCsorbaJB:
```suggestion

To make the alias generic, declare its own type parameter:
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
+#### Type parameters
+
+A nested type alias can't capture type parameters from its containing declaration. Capturing occurs when the alias refers
+to a type parameter declared by its outer class or interface:
 
 ```kotlin
 class Graph<Node> {
-    // Correct because Node is a type alias parameter
-    typealias Path<Node> = List<Node>
+    // Error: Path captures Graph's Node type parameter
```
daniCsorbaJB: this is a bit misleading — it seems like this is the error compiler would give I'd suggest:

This results in an error, because Path captures Graph's Node type parameter

Then add below: 
// Unresolved reference 'Node'.

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,131 @@
...
 ```
 
-To fix this issue, declare the type parameter directly in the type alias:
+> Nested type aliases are not supported in Kotlin Multiplatform [`expect/actual` declarations](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html).
+>
+{style="note"}
+
+#### Type parameters
+
+A nested type alias can't capture type parameters from its containing declaration. Capturing occurs when the alias refers
```
daniCsorbaJB: As a general comment. It feels strange to lead with what the user can't do 🤔 

What do you think about flipping this section a bit? 

To use type parameters in a nested type alias, add them to the alias declaration:

[correct example]

In this example, `Path` declares `T`.  In `Graph.Path<String>`, `String` provides the type argument for `T` and doesn't depend on the `Node` type parameter declared by `Graph`. (a bit wordy, but I couldn't come up with a smoother version. 😢 )

If you refer to a type parameter declared by the class or interface that contains the alias, the compiler reports an error:

[incorrect example]

Personally, I also would suggest dropping the part about "capturing" — I see that the original had it, but I think we can make this more understandable without it.

daniCsorbaJB (changes requested):
nice improvement to the page 🔥 — I added my comments, please let me know what you think.

docs/topics/type-aliases.md
aepanchi: I think helps users clearly see where the subsections come from.

docs/topics/type-aliases.md
aepanchi: No, it applies to both

docs/topics/type-aliases.md
aepanchi: No, I like it! Thanks :)

docs/topics/type-aliases.md
daniCsorbaJB: hmm 🤔 if that's the intention, I think it works, but then we should put the focus on that (right now it's in parentheses without links to the relevant sections, so I don't think many people would make this connection)

docs/topics/type-aliases.md
daniCsorbaJB: hmm 🤔 Then I'm confused by this sentence: 

> The effective accessibility of a type alias depends on its visibility and the accessibility of its containing declarations.

What containing declaration do we mean in relation to a top-level declaration?

docs/topics/type-aliases.md
aepanchi: Yes, it's a good point, updated!

docs/topics/type-aliases.md
aepanchi: Sorry, I missunderstood your point! I was talking about underlying type visibility. You are correct, I have updated that part

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,147 @@
...
 {kotlin-runnable="true"}
 
-## Nested type aliases
+## Declare type aliases
 
-In Kotlin, you can define type aliases inside other declarations, as long as they
-don't capture type parameters from their outer class:
+You can declare a type alias:
+
+* At the top level of a Kotlin file with [top-level type aliases](#top-level-type-aliases).
```
daniCsorbaJB:
```suggestion
* At the top level of a Kotlin file, as a [top-level type alias](#top-level-type-aliases).
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,147 @@
...
+* Inside a class, interface, or object with [nested type aliases](#nested-type-aliases).
```
daniCsorbaJB:
```suggestion
* Inside a class, interface, or object, as a [nested type alias](#nested-type-aliases).
```

daniCsorbaJB: as is a bit cleaner than with in this case imho. 🤔 with suggests to me that you use a type alias to declare something else.

docs/topics/type-aliases.md
daniCsorbaJB: nice that's a great improvement 🎉 — my only concern is the "effective accessibility" part now. The docs don't use this term, because it can be a bit vague - so I still prefer a more exact wording, like:

> A nested type alias is accessible only where its containing class, interface, or object is accessible. 

this sentence sidesteps the issue with the abstract noun phrase/technical term

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,147 @@
...
+> Nested type aliases aren't supported in Kotlin Multiplatform [`expect/actual` declarations](https://kotlinlang.org/docs/multiplatform/multiplatform-expect-actual.html).
+>
+{style="note"}
 
-```kotlin
+#### Type parameters
+
+To use type parameters in a nested type alias, add them to the alias declaration:
+
+```
```
daniCsorbaJB:
```suggestion
```kotlin
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,147 @@
...
-Nested type aliases allow for cleaner, more maintainable code by improving encapsulation, reducing package-level clutter,
-and simplifying internal implementations.
+In this example, `Path` declares its type parameter `T`. In `Graph.Path<String>`, `String` is the type argument for `T`
+and is independent of the `Node` type parameter declared by `Graph`.
 
-### Rules for nested type aliases
+If you refer to a type parameter declared by its containing class or interface, the compiler reports an error:
 
-Nested type aliases follow specific rules to ensure clear and consistent behavior:
+```
```
daniCsorbaJB:
```suggestion
```kotlin
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,147 @@
...
+To use the alias from another package, import the alias or refer to it by its qualified name:
 
 ```kotlin
-class Dijkstra {
-    typealias VisitedNodes = Set<Node>
+// UserId.kt
+package org.example.users
+
+typealias UserId = Long
+
```
daniCsorbaJB, line 72: some nits here:
Technically it would be more correct imho to:
- place the comment ABOVE the line (fun createUser(id: UserId) { and similar here)) instead of inside
- let's use third person comments and not passive

Something like
// Uses the unqualified name within the same package / or // Refers to the alias within the same package by its unqualified name
// Uses the imported alias by its unqualified name / // ...
// Uses the fully qualified name / // ...

daniCsorbaJB (commented):
Some last nitpicks here and there and final cleanup otherwise it's great!

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,149 @@
...
+    // Refers to the alias by its unqualified name inside UserRepository
+    fun saveAll(users: UserIndex) {
+        // ...
+    }
+}
+
+// Refers to the alias by its qualified name outside UserRepository
+fun synchronizeUsers(users: UserRepository.UserIndex) {
+  // ...
```
daniCsorbaJB:
```suggestion
    // ...
```

docs/topics/type-aliases.md
```diff
@@ -1,100 +1,149 @@
...
+    // ...
+}
+
+// UserService.kt
+package org.example.services
 
-    private fun step(visited: VisitedNodes, ...) = ...
+import org.example.users.UserId
+
+// Uses the imported alias by its unqualified name
```
daniCsorbaJB:
```suggestion
// Uses the imported alias by its unqualified name
```

daniCsorbaJB (approved):
Nice! Just two final clean-up suggestions, otherwise LGTM! 🚀

PR #5679, opened 2026-07-16
update: dependency graph

budindepunk:
New feature. See [issue in KSP repository](https://github.com/google/ksp/issues/3015)


docs/topics/ksp/ksp-incremental.md
```diff
@@ -183,14 +183,34 @@ To generate `outputForA`, the processor:
...
+```
+
+If you're using KSP from the command line, add the following options:
+
+```bash
+-incremental-log=true -incremental-log.graph.origin=<fully-qualified-name>
+```
+
+The DOT file is generated in the `logs` directory.
+
```
jaschdoc: I think this should come after 'Reporting bugs', so people know that it's related to KSP bug reporting. It's there to make bug reporting easier so that if they expect to see dependency `A -> B` but that edge is missing, then it's easier to narrow down the bug.

Also here's a small example to help users understand what the fully qualified name is:
```kotlin
package com.example.app

class MyEntrypoint
```
Then the fully qualified name is `com.example.app.MyEntrypoint`, i.e., `<package name>.<class name>`.

I'm not sure if it's relevant to include, since the users of this is most likely process authors and they are very likely to know what a fully qualified name is, but regardless it may help clear up confusion.

jaschdoc (commented):
Thanks for updating the documentation!

docs/topics/ksp/ksp-incremental.md
budindepunk: I moved it down and added a link to somewhere where fully qualified names are explained. I felt explaining it here a little out of place. Does that work?

jaschdoc (approved):
Looks good, thank you :)

docs/topics/ksp/ksp-incremental.md
```diff
@@ -205,3 +206,23 @@ If you encounter any error that occurs only when incremental processing is enabl
 
 5. Attach the log files from both the successful build and the build that reproduces the issue to the GitHub issue.
 
+### Visualizing the symbol dependency graph
+
+To help debug incremental processing, KSP can generate a Graphviz DOT file that visualizes the symbol dependency graph
+explored from a specified symbol.
```
daniCsorbaJB:
```suggestion
starting from a specified symbol.
```

daniCsorbaJB: If I understand correctly, this serves as the starting point for the dependency graph, so this feels slightly clearer to me. WDYT would it work?

docs/topics/ksp/ksp-incremental.md
```diff
@@ -205,3 +206,23 @@ If you encounter any error that occurs only when incremental processing is enabl
...
+
+Enable incremental logging and specify the [fully qualified name](https://kotlinlang.org/docs/packages.html#package-headers) 
+of the symbol to visualize:
```
daniCsorbaJB:
```suggestion
of the symbol to use as the starting point for visualizing the graph:
```

daniCsorbaJB: same clarification here

daniCsorbaJB (commented):
Nice addition, thank you! 🚀 — I have two small clarity-related comments, but aside from that, all looks great!

daniCsorbaJB (approved):
nice, LGTM 🚀

PR #5682, opened 2026-07-16
chore: updating coroutines cancellation links

daniCsorbaJB:
This PR updates the links for the coroutines cancellation page.

### DIFF
--- a/redirects/pages.yml
+++ b/redirects/pages.yml
@@ -494,7 +494,7 @@
   to: /docs/exception-handling.html
 
 - from: /docs/reference/coroutines/cancellation-and-timeouts.html
-  to: /docs/cancellation-and-timeouts.html
+  to: /docs/coroutines-cancellation.html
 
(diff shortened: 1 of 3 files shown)


redirects/pages.yml
```diff
@@ -494,7 +494,7 @@
   to: /docs/exception-handling.html
 
 - from: /docs/reference/coroutines/cancellation-and-timeouts.html
-  to: /docs/cancellation-and-timeouts.html
```
danil-pavlov, line 497: It's better to make a list of "from" and include cancellation-and-timeouts.html there instead of just replacing it

danil-pavlov (changes requested):

redirects/pages.yml
daniCsorbaJB: good idea 👍

danil-pavlov (approved):

PR #5691, opened 2026-07-20
Fix typo in method name from isMoreThat3 to isMoreThan3

berakoc:


aepanchi:
Nice catch! Thanks for contribution @berakoc :)

aepanchi (approved):

PR #5697, opened 2026-07-21
feat: JPMS support for Maven projects

danil-pavlov:
* [KT-86759](https://youtrack.jetbrains.com/issue/KT-86759) Maven: how to configure JPMS

### DIFF
--- a/docs/topics/maven/maven-configure-project.md
+++ b/docs/topics/maven/maven-configure-project.md
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
 >
 {style="note"}
 
+## Configure Java modules (JPMS)
+
+The Kotlin Maven plugin supports the [Java Platform Module System (JPMS)](https://dev.java/learn/modules/), so you can
+compile Kotlin code alongside a `module-info.java` descriptor and consume the resulting module like other Java modules.
+
+On Maven's side, no extra JPMS-specific options are required. The configuration is set by the `module-info.java`
+descriptor and correct plugin ordering: Kotlin sources are compiled before the Maven compiler processes `module-info.java`.
+
+To configure a Java module, create the `module-info.java` file with a module descriptor in the `src/main/java` directory.
+List the Kotlin modules that your module requires and exports your packages. For example:
+
+```java
+module org.example.myapp {
+    requires transitive kotlin.stdlib;
+    requires kotlin.stdlib.jdk8;
+
+    exports org.example.myapp;
+}
+```
+
+* For a module, the package name in Kotlin files must be equal to the package name from `module-info.java` to avoid a
+  "package is empty or does not exist" build failure.
+* The `pom.xml` build file should be configured so that [Kotlin is compiled before Java](#compile-kotlin-and-java-sources).
+  If you use [automatic project configuration](#automatic-configuration), the `<extensions>` option already ensures that. 
+
 ## What's next?
 
 [Set dependencies in your Kotlin Maven project](maven-set-dependencies.md)
--- a/docs/topics/maven/maven-configure-project.md
+++ b/docs/topics/maven/maven-configure-project.md
@@ -365,12 +365,12 @@ and has no impact on other plugins in the build.
 The Kotlin Maven plugin supports the [Java Platform Module System (JPMS)](https://dev.java/learn/modules/), so you can
 compile Kotlin code alongside a `module-info.java` descriptor and consume the resulting module like other Java modules.
 
-In the build file, no extra JPMS-specific options are required beyond the correct plugin ordering: the Kotlin compiler
-should be configured before Maven.
+You don't need any extra JPMS-specific options in the build file. 
+Just configure the Kotlin compiler before Maven.
 
 When a `module-info.java` descriptor is present, the Kotlin compiler uses it as a source file and compiles against
-its module path instead of the classpath. The Kotlin compiler reads the descriptor to resolve the module graph, and then
-the Maven compiler compiles it into the `module-info.class` file.
+its module path instead of the classpath. The Kotlin compiler reads the descriptor to resolve the module graph, and
+the Maven compiler then compiles it into a `module-info.class` file.
 
 To configure a Java module, create the `module-info.java` file in the `${project.basedir}/src/main/java` directory.
 In the module descriptor, declare all the dependencies that your module requires and the packages it exports. For example:
@@ -386,13 +386,13 @@ module org.example.myapp {
 
 Keep in mind that:
 
-* Your Java module can only use what you declared. Since compilation goes through the module path instead of the classpath,
+* Your Java module can only use what you declare. Since compilation uses the module path instead of the classpath,
   the descriptor should include all dependencies that your Kotlin code uses: the standard library, JDK modules (except for `java.base`),
-  and other libraries to avoid `Unresolved reference` errors.
-* For a module, the package name in Kotlin files must be equal to the package name from `module-info.java` to avoid the
+  and other libraries. Otherwise, you may get `Unresolved reference` errors.
+* For a module, the package name in Kotlin files must match the package name from `module-info.java` to avoid the
   `Package is empty or does not exist` build failures.
 * The `pom.xml` build file should be configured so that [Kotlin is compiled before Java](#compile-kotlin-and-java-sources).
-  If you use [automatic project configuration](#automatic-configuration), the `<extensions>` option already ensures that. 
+  If you use [automatic project configuration](#automatic-configuration), the `<extensions>` option already ensures this. 
 
 ## What's next?
 
--- a/docs/topics/maven/maven-configure-project.md
+++ b/docs/topics/maven/maven-configure-project.md
@@ -191,9 +191,9 @@ To apply the Kotlin Maven plugin, update your `pom.xml` build file as follows:
                     </goals>
                     <configuration>
                         <sourceDirs>
-                            <sourceDir>${project.basedir}/main/kotlin</sourceDir>
+                            <sourceDir>${project.basedir}/src/main/kotlin</sourceDir>
                             <!-- Ensure Kotlin code can reference Java code -->
-                            <sourceDir>${project.basedir}/main/java</sourceDir>
+                            <sourceDir>${project.basedir}/src/main/java</sourceDir>
                         </sourceDirs>
                     </configuration>
                 </execution>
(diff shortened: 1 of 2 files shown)


docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
 >
 {style="note"}
 
+## Configure Java modules (JPMS)
+
+The Kotlin Maven plugin supports the [Java Platform Module System (JPMS)](https://dev.java/learn/modules/), so you can
+compile Kotlin code alongside a `module-info.java` descriptor and consume the resulting module like other Java modules.
+
+On Maven's side, no extra JPMS-specific options are required. The configuration is set by the `module-info.java`
```
shinkai-tester: The section says Maven needs no extra options, which is true, but it would be nice to mention what the descriptor changes for the Kotlin compiler. With a `module-info.java`, the plugin compiles against the module path instead of the classpath, so every dependency whose API the Kotlin code uses must be declared in the descriptor: the standard library, other dependencies, and JDK modules other than `java.base`. Missing clauses show up as plain `Unresolved reference` errors that don't mention modules at all (the standard library is the exception - it has a dedicated message).

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
...
+descriptor and correct plugin ordering: Kotlin sources are compiled before the Maven compiler processes `module-info.java`.
```
shinkai-tester: This reads as if the Kotlin compiler knows nothing about the descriptor. It does: the plugin passes
`module-info.java` to the Kotlin compiler as a source file, and the compiler uses it to resolve the module graph - which is why its errors point at `module-info.java`, for example "The Kotlin standard library is not found in the module graph". Only javac compiles the descriptor into `module-info.class`.
Suggested rewording: "the Kotlin compiler reads the descriptor to resolve the module graph, and the
Maven compiler compiles it into `module-info.class`."

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
...
+
+To configure a Java module, create the `module-info.java` file with a module descriptor in the `src/main/java` directory.
+List the Kotlin modules that your module requires and exports your packages. For example:
```
shinkai-tester: Suggestion: "Declare the modules that your module requires and export your packages."
Why: every dependency must be declared, not only Kotlin ones.

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
...
+
+```java
+module org.example.myapp {
+    requires transitive kotlin.stdlib;
+    requires kotlin.stdlib.jdk8;
+
+    exports org.example.myapp;
+}
+```
```
shinkai-tester, line 385: This example doesn't compile:
[ERROR] Module kotlin.stdlib.jdk8 cannot be found in the module graph
Please drop this line: `requires kotlin.stdlib.jdk8`.

I am not sure, but maybe worth extending the example with a JDK module, for example `requires java.net.http`. Inside a module only `java.base` is readable, so Kotlin code that uses `java.sql`, `java.net.http`, or `java.desktop`  stops compiling — and the diagnostic mentions no modules at all:                                                                                                                                                               
[ERROR] Greeter.kt:6:28 Unresolved reference 'http'                                                                                                                                                                        
That looks like a missing dependency, so readers will go looking in `pom.xml` instead of the descriptor. One extra line in the example makes the point for free:
```
     module org.example.myapp {                                                                                                                                                                                                     
         requires transitive kotlin.stdlib;                                                                                                                                                                                         
         requires java.net.http;                                                                                                                                                                                                    
         exports org.example.myapp;                                                                                                                                                                                                 
     }  
```

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,31 @@ and has no impact on other plugins in the build.
...
+
+* For a module, the package name in Kotlin files must be equal to the package name from `module-info.java` to avoid a
+  "package is empty or does not exist" build failure.
+* The `pom.xml` build file should be configured so that [Kotlin is compiled before Java](#compile-kotlin-and-java-sources).
```
shinkai-tester, line 394: This link leads to a snippet with `<sourceDir>src/main/java</sourceDir>`, and in a modular project
that exact line breaks the build :(  :
    [ERROR] .../src/main/java/module-info.java Too many source module declarations found
    [ERROR] java.lang.AssertionError: Modules computed by ClasspathRootsResolver cannot have
            more than one source module: [org.example.myapp, org.example.myapp, ...]
Only the relative form fails, `${project.basedir}/src/main/java` builds fine. Since we send readers
to that snippet from here, it is better to add some warning, please see freshly created ticket:
https://youtrack.jetbrains.com/issue/KT-88152/Maven-Too-many-source-module-declarations-found-and-AssertionError-when-sourceDir-uses-a-relative-path-to-a-directory-containing.

shinkai-tester (changes requested):

shinkai-tester:
@Tapchicoma I’ve added a few comments - please take a look when you have a chance. I tried to recall everything I know about Maven and JPMS in the Kotlin context, but I’m not entirely sure all of my suggestions are correct, as I haven’t seen many examples of configurations with JPMS support.

docs/topics/maven/maven-configure-project.md
```diff
@@ -205,8 +205,8 @@ To apply the Kotlin Maven plugin, update your `pom.xml` build file as follows:
                     </goals>
                     <configuration>
                         <sourceDirs>
-                            <sourceDir>src/test/kotlin</sourceDir>
-                            <sourceDir>src/test/java</sourceDir>
+                            <sourceDir>${project.basedir}/test/kotlin</sourceDir>
+                            <sourceDir>${project.basedir}/test/java</sourceDir>
```
shinkai-tester: ${project.basedir}/src/test/kotlin
${project.basedir}/src/test/java

docs/topics/maven/maven-configure-project.md
```diff
@@ -191,9 +191,9 @@ To apply the Kotlin Maven plugin, update your `pom.xml` build file as follows:
                     </goals>
                     <configuration>
                         <sourceDirs>
-                            <sourceDir>src/main/kotlin</sourceDir>
+                            <sourceDir>${project.basedir}/main/kotlin</sourceDir>
                             <!-- Ensure Kotlin code can reference Java code -->
-                            <sourceDir>src/main/java</sourceDir>
+                            <sourceDir>${project.basedir}/main/java</sourceDir>
```
shinkai-tester: src is missing:
${project.basedir}/src/main/kotlin
${project.basedir}/src/main/java

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,39 @@ and has no impact on other plugins in the build.
...
+    requires transitive kotlin.stdlib;
+    requires java.net.http;
+    
+    exports org.example.myapp;
+}
+```
+
+Keep in mind that:
+
+* Since your Java module is used instead of the classpath during compilation, it should include all dependencies that your Kotlin code uses:
```
shinkai-tester: I would reformulate it like this (just a suggestion):

> Your Java module can only use what it declares: compilation goes through the module path instead of the classpath, so the descriptor should declare the standard library, JDK modules (except for `java.base`), and other libraries whose API your Kotlin code uses. Otherwise, you get `Unresolved reference` errors.

shinkai-tester (changes requested):

shinkai-tester (approved):

Tapchicoma (approved):

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
 >
 {style="note"}
 
+## Configure Java modules (JPMS)
+
+The Kotlin Maven plugin supports the [Java Platform Module System (JPMS)](https://dev.java/learn/modules/), so you can
+compile Kotlin code alongside a `module-info.java` descriptor and consume the resulting module like other Java modules.
+
+In the build file, no extra JPMS-specific options are required beyond the correct plugin ordering: the Kotlin compiler
+should be configured before Maven.
```
sarahhaggarty:
```suggestion
You don't need any extra JPMS-specific options in the build file. 
Just configure the Kotlin compiler before Maven.
```
What do you think about this as an alternative?

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
...
+
+When a `module-info.java` descriptor is present, the Kotlin compiler uses it as a source file and compiles against
+its module path instead of the classpath. The Kotlin compiler reads the descriptor to resolve the module graph, and then
+the Maven compiler compiles it into the `module-info.class` file.
```
sarahhaggarty:
```suggestion
its module path instead of the classpath. The Kotlin compiler reads the descriptor to resolve the module graph, and
the Maven compiler then compiles it into a `module-info.class` file.
```

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
...
+    requires transitive kotlin.stdlib;
+    requires java.net.http;
+    
+    exports org.example.myapp;
+}
+```
+
+Keep in mind that:
+
+* Your Java module can only use what you declared. Since compilation goes through the module path instead of the classpath,
```
sarahhaggarty:
```suggestion
* Your Java module can only use what you declare. Since compilation uses the module path instead of the classpath,
```

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
...
+  the descriptor should include all dependencies that your Kotlin code uses: the standard library, JDK modules (except for `java.base`),
+  and other libraries to avoid `Unresolved reference` errors.
```
sarahhaggarty:
```suggestion
  and other libraries. Otherwise, you may get `Unresolved reference` errors.
```
What do you think about breaking the sentence here?

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
...
+* For a module, the package name in Kotlin files must be equal to the package name from `module-info.java` to avoid the
```
sarahhaggarty:
```suggestion
* For a module, the package name in Kotlin files must match the package name from `module-info.java` to avoid the
```
A bit more concise?

docs/topics/maven/maven-configure-project.md
```diff
@@ -360,6 +360,40 @@ and has no impact on other plugins in the build.
...
+  `Package is empty or does not exist` build failures.
+* The `pom.xml` build file should be configured so that [Kotlin is compiled before Java](#compile-kotlin-and-java-sources).
+  If you use [automatic project configuration](#automatic-configuration), the `<extensions>` option already ensures that.
```
sarahhaggarty:
```suggestion
  If you use [automatic project configuration](#automatic-configuration), the `<extensions>` option already ensures this. 
```

sarahhaggarty (approved):
It's fine as it is but I made some suggestions for you to consider!

PR #5698, opened 2026-07-21
Clarify anonymous context receiver semantics

lowasser:
"The parameter's value is available for resolution" actively confused some of my team members, so I thought some additional examples might help.


docs/topics/context-parameters.md
netdpb: What about "available for resolution for context parameters of called functions" here, just to clarify what "resolution" refers to.

aepanchi: Agree, I would rephrase it to: 
You can use `_` as a context parameter name when you don't need to refer to the parameter directly. An anonymous context parameter is still available for resolving the context parameters of called functions. However it isn't available by name and doesn't act as an implicit receiver. To access its value explicitly, use `contextOf<T>()`.

aepanchi (commented):
Thank you for the update! The changes look very nice and bring the value to the docs. I have left some minor comments. Let me know if you would like me to commit them myself :)

aepanchi:
Thanks for your contribution @lowasser! After checking with developers in our team, I modified your PR a bit. I will be merging it, but if you have any additional thoughts, feel free to comment this PR or open a new one. Thanks once again for the update :)

aepanchi (approved):

lowasser:
(Thank you!  I was on vacation for about a week, so I appreciate that these got merged while I was gone.)

ccolin1841-jpg (commented):
.

PR #5700, opened 2026-07-23
update: outdated tasks and APIs

danil-pavlov:
* [KT-56009](https://youtrack.jetbrains.com/issue/KT-56009)

### DIFF
--- a/docs/topics/js/js-project-setup.md
+++ b/docs/topics/js/js-project-setup.md
@@ -58,12 +58,12 @@ Inside the `kotlin {}` block, you can manage the following aspects:
 Kotlin/JS projects can target two different execution environments: 
 
 * Browser for client-side scripting in browsers
-* [Node.js](https://nodejs.org/) for running JavaScript code outside of a browser, for example, for server-side scripting.
+* [Node.js](https://nodejs.org/) for running JavaScript code outside a browser, for example, for server-side scripting.
 
 To define the target execution environment for a Kotlin/JS project, add the `js {}` block with `browser {}` or `nodejs {}`
 inside:
 
-```groovy
+```kotlin
 kotlin {
     js {
         browser {
@@ -75,7 +75,7 @@ kotlin {
 
 The instruction `binaries.executable()` explicitly instructs the Kotlin compiler to emit executable `.js` files.
 Omitting `binaries.executable()` will cause the compiler to only generate 
-Kotlin-internal library files, which can be used from other projects, but not run on their own.
+Kotlin-internal library files, which can be used from other projects but not run on their own.
 
 > This is typically faster than creating executable files,
 > and can be a possible optimization when dealing with non-leaf modules of your project.
@@ -84,8 +84,8 @@ Kotlin-internal library files, which can be used from other projects, but not ru
 
 The Kotlin Multiplatform plugin automatically configures its tasks for working with the selected environment.
 This includes downloading and installing the required environment and dependencies for running and testing the application.
-This allows developers to build, run, and test simple projects without additional configuration. For projects targeting
-Node.js, there is also an option to use an existing Node.js installation. Learn how to [use pre-installed Node.js](#use-pre-installed-node-js).
+This allows developers to build, run, and test simple projects without additional configuration. There is also an option
+to use an existing installation. Learn how to [use pre-installed Node.js](#use-pre-installed-node-js).
 
 ## Support for ES2015 features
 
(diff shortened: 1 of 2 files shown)


docs/topics/js/js-project-setup.md
```diff
@@ -166,40 +166,15 @@ un-webpacked JavaScript code.
...
-    implementation 'org.example.myproject:1.1.0'
-}
-```
-
-</tab>
-</tabs>
-
-The Kotlin Multiplatform Gradle plugin also supports dependency declarations for particular source sets in the `kotlin {}` block 
-of the build script:
+To declare a dependency, use the `dependencies {}` block in the `jsMain` source set in your `build.gradle(.kts)` file:
```
JSMonk, line 169: or `webMain`?

docs/topics/js/js-project-setup.md
```diff
@@ -370,10 +361,10 @@ The plugin also provides useful testing features, for example:
 * Test run results in the console
 
 For running browser tests, the plugin uses [Headless Chrome](https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md)
-by default. You can also choose another browser to run tests in, by adding the corresponding entries inside the `useKarma {}`
+by default. You can also choose another browser to run tests in by adding the corresponding entries inside the `useKarma {}`
```
JSMonk, line 364: I believe we also need to mention the experimental support of playwright, which came in 2.4.20: https://kotlinlang.org/docs/whatsnew-eap.html#new-dsl-for-browser-testing

JSMonk (changes requested):

docs/topics/js/js-project-setup.md
danil-pavlov: I agree, but that will be a separate task for the final 2.4.20 release

docs/topics/js/js-project-setup.md
danil-pavlov: I'm not sure TBH, both Kotlin Multiplatform IDE plugin and [kmp-js-wizard](https://github.com/Kotlin/kmp-js-wizard/tree/main/src) have projects with `jsMain`, should we change that?

JSMonk (approved):

docs/topics/js/js-project-setup.md
daniCsorbaJB: nit: here, we use `allProjects` but below we use `allprojects`

docs/topics/js/js-project-setup.md
```diff
@@ -654,17 +640,17 @@ You can also use an existing Node.js instance if you have it.
...
+For example, the default Node.js version is currently 24.16.0, but you can use a different version for a specific subproject.
+Add the following lines to your subproject's `build.gradle(.kts)` file:
 
 <tabs group="build-script">
 <tab title="Kotlin" group-key="kotlin">
 
 ```kotlin
 project.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsPlugin> {
-    project.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsEnvSpec>().version = "your Node.js version"
+    project.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsEnvSpec>().version = "26.2.0"
```
daniCsorbaJB, line 652: is there a reason to change to this version specifically? 🤔 it feels arbitrary without any description. WDYT about adding the default intstead if we want to have a specific version (24.16.0)

docs/topics/js/js-project-setup.md
```diff
@@ -517,25 +496,32 @@ capabilities are well described in its [documentation](https://webpack.js.org/co
...
+ be available in `build/kotlin-webpack` unless [specified otherwise](#distribution-target-directory).
 
 ```bash
-./gradlew browserProductionWebpack
+./gradlew jsBrowserProductionWebpack
 ```
 
 Note that these tasks will only be available if your target is configured to generate executable files (via `binaries.executable()`).
 
+To generate the distribution placed into `build/dist`, run the `jsBrowserDistribution` task instead:
```
daniCsorbaJB:
```suggestion
To generate the distribution in the `build/dist/<targetName>/<binaryName>` directory, run the `jsBrowserDistribution` task instead:
```

docs/topics/js/js-project-setup.md
```diff
@@ -517,25 +496,32 @@ capabilities are well described in its [documentation](https://webpack.js.org/co
...
+
+```bash
+./gradlew jsBrowserDistribution
+```
+
+This task produces the ready-to-use distribution together with the project resources.
```
daniCsorbaJB:
```suggestion
This task generates a ready-to-use distribution that includes the project resources.
```

daniCsorbaJB: just to keep the terms consistent with the above used one (generate - generate)

docs/topics/js/js-project-setup.md
daniCsorbaJB: just to align with what we have further down.

daniCsorbaJB (approved):
(I ignored the language part of the old content as discussed.) I only have a couple of minor suggestions/questions, otherwise LGTM!

docs/topics/js/js-project-setup.md
danil-pavlov: There was a lot of feedback regarding changing the version, so I think it makes sense to show it (since the newest versions are far from default), but let's highligh that it's just an example

daniCsorbaJB: ah I see 👍 nice, LGTM

PR #5710, opened 2026-07-30
fix: remove redirect in pages.yml and fix link in ksp-overview.md

budindepunk:
Remove redirection and add link to ksp-command-line.md from ksp-overview.md

Part of [KEDOC-236](https://youtrack.jetbrains.com/issue/KEDOC-236/Create-new-KSP-CLI-article)


zamulla (approved):
👍

PR #5711, opened 2026-07-30
update: BOM file description

danil-pavlov:


docs/topics/maven/maven-set-dependencies.md
```diff
@@ -156,6 +163,48 @@ add a dependency on [`kotlin-bom`](https://mvnrepository.com/artifact/org.jetbra
 </dependencyManagement>
 ```
 
+After importing the BOM, you can declare Kotlin dependencies in the `<dependencies>` section without specifying their versions;
```
e5l, line 166: this section is about autoring BOM, not using it. I would refer external official documentation on how to author BOM instead

e5l (approved):

docs/topics/maven/maven-set-dependencies.md
```diff
@@ -137,10 +137,17 @@ To add a dependency on the [`kotlinx-datetime`](https://kotlinlang.org/api/kotli
 </dependencies>
 ```
 
-### Use BOM dependency mechanism
+## Manage dependencies with BOM
```
daniCsorbaJB:
```suggestion
## Manage dependencies with a BOM
```

docs/topics/maven/maven-set-dependencies.md
```diff
@@ -137,10 +137,17 @@ To add a dependency on the [`kotlinx-datetime`](https://kotlinlang.org/api/kotli
...
 
-To use a Kotlin [Bill of Materials (BOM)](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms),
-add a dependency on [`kotlin-bom`](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-bom):
+A [Bill of Materials (BOM)](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms)
+is a special POM file that manages dependency versions in your project. This keeps related artifacts aligned and avoids version conflicts.
+
+Kotlin publishes the [`kotlin-bom`](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-bom) artifact, which
+sets the versions of Kotlin libraries (such as `kotlin-stdlib`, `kotlin-reflect`, and `kotlin-test`) that belong to a
```
daniCsorbaJB:
```suggestion
specifies the versions of Kotlin libraries, such as `kotlin-stdlib`, `kotlin-reflect`, and `kotlin-test`, that corresponds to a
```

daniCsorbaJB: does it set it or specify it? 🤔

docs/topics/maven/maven-set-dependencies.md
```diff
@@ -137,10 +137,17 @@ To add a dependency on the [`kotlinx-datetime`](https://kotlinlang.org/api/kotli
...
+single Kotlin release. This is useful when other libraries in your project have transitive dependencies on Kotlin
+artifacts, ensuring that all dependencies resolve to the same Kotlin version.
```
daniCsorbaJB: ensuring is a bit of a dangling modifier here 🤔 
WDYT about moving it earlier to connect it more closely: 

This is useful when other libraries in your project have transitive dependencies on Kotlin
artifacts because it ensures that all Kotlin dependencies resolve to the same version.

docs/topics/maven/maven-set-dependencies.md
```diff
@@ -156,6 +163,17 @@ add a dependency on [`kotlin-bom`](https://mvnrepository.com/artifact/org.jetbra
 </dependencyManagement>
 ```
 
+After importing the BOM, you can declare Kotlin dependencies in the `<dependencies>` section without specifying their versions;
+they come from the BOM file automatically.
+
+* Importing a BOM doesn't add any dependencies to your project on its own. It only controls the versions of dependencies
+  that you declare explicitly or that come in transitively.
```
daniCsorbaJB:
```suggestion
  that you declare explicitly or that are introduced as transitive dependencies.
```

daniCsorbaJB: come in transitively is a bit awkward. Here passive could work because the actor is unclear 🤔 

Perhaps, if we want to avoid passive we could say it like this to have only one actor: 
It only controls the versions of explicitly declared and transitive dependencies.

docs/topics/maven/maven-set-dependencies.md
```diff
@@ -156,6 +163,17 @@ add a dependency on [`kotlin-bom`](https://mvnrepository.com/artifact/org.jetbra
...
+* If you still specify a `<version>` for a dependency, that value overrides the version from the BOM.
+
+If your project publishes several libraries that are released together, you can provide your own BOM so that users
+can align their versions the same way the Kotlin BOM does. To learn how to author your own BOM, see the
```
daniCsorbaJB:
```suggestion
can align the versions of those libraries in the same way. To learn how to author your own BOM, see the
```

daniCsorbaJB: or perhaps: 
...that users can keep the versions of your project's libraries consistent.

daniCsorbaJB (changes requested):
Nice addition 💯 — I left a couple of comments

daniCsorbaJB (approved):
nice, thank you for the changes! LGTM! 🚀

PR #5722, opened 2026-08-04
fix: @JvmStatic description for objects

danil-pavlov:
* [KT-54370](https://youtrack.jetbrains.com/issue/KT-54370) Clarify that @JvmStatic on elements of object declaration replaces instance methods

### DIFF
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -205,11 +205,11 @@ int version = C.VERSION;
 ## Static methods
 
 Kotlin represents package-level functions as static methods.
-It also generates static methods for functions defined in named objects or companion objects if you annotate such
-functions with [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/).
+You can also generate static methods for functions defined in named objects or companion objects if you annotate such
+functions with the [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/) annotation.
 
 If you use `@JvmStatic` on a function in a companion object, the compiler generates both a static method in the enclosing
-class and an instance method in the companion object itself:
+class and an instance method in the companion object:
 
 ```kotlin
 // Kotlin
@@ -221,18 +221,19 @@ class C {
 }
 ```
 
-In Java, the `callStatic()` method is static while `callNonStatic()` is not:
+In Java, you can call `callStatic()` on both the enclosing class and the companion object, while `callNonStatic()` is only
+available through the companion object:
 
 ```java
 // Java
-C.callStatic(); // Success
-C.callNonStatic(); // Error: not a static method
-C.Companion.callStatic(); // Instance method remains
+C.callStatic();              // Success
+C.callNonStatic();           // Error: not a static method
+C.Companion.callStatic();    // Instance method remains
 C.Companion.callNonStatic(); // Success
 ```
 
-For a named object (a singleton), `@JvmStatic` turns the function into a static method of the object's class;
-no separate instance method is generated:
+For a named object (a singleton), `@JvmStatic` turns the function into a static method of the object's class but
+doesn't generate a separate instance method:
 
 ```kotlin
 // Kotlin
@@ -242,16 +243,17 @@ object Obj {
 }
 ```
 
-In Java, the `callStatic()` method is static, while `callNonStatic()` is only available through the singleton instance:
+In Java, you can call the `callStatic()` method on the named object, while `callNonStatic()` is only available through
+the singleton instance:
 
 ```java
 // Java
-Obj.callStatic(); // Success
-Obj.callNonStatic(); // Error: not a static method
-Obj.INSTANCE.callNonStatic(); // Success, a call through the singleton instance
+Obj.callStatic();             // Success
+Obj.callNonStatic();          // Error: not a static method
+Obj.INSTANCE.callNonStatic(); // Success, the call passed through the singleton instance
 ```
 
-`@JvmStatic` also applies to functions defined in companion objects of interfaces.
+You can also annotate a function in an interface's companion object with `@JvmStatic`.
 Such functions compile to static methods in interfaces:
 
 ```kotlin
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -208,8 +208,8 @@ Kotlin represents package-level functions as static methods.
 It also generates static methods for functions defined in named objects or companion objects if you annotate such
 functions with [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/).
 
-For a function in a companion object, the compiler generates both a static method in the enclosing class and
-an instance method in the companion object itself:
+If you use `@JvmStatic` on a function in a companion object, the compiler generates both a static method in the enclosing
+class and an instance method in the companion object itself:
 
 ```kotlin
 // Kotlin
@@ -221,7 +221,7 @@ class C {
 }
 ```
 
-Now, the `callStatic()` method is static in Java while `callNonStatic()` is not:
+In Java, the `callStatic()` method is static while `callNonStatic()` is not:
 
 ```java
 // Java
@@ -242,7 +242,7 @@ object Obj {
 }
 ```
 
-In Java:
+In Java, the `callStatic()` method is static, while `callNonStatic()` is only available through the singleton instance:
 
 ```java
 // Java
--- a/docs/topics/jvm/java-to-kotlin-interop.md
+++ b/docs/topics/jvm/java-to-kotlin-interop.md
@@ -122,7 +122,6 @@ class User(id: String) {
 ```
 
 ```java
-
 // Java
 class JavaClient {
     public String getID(User user) {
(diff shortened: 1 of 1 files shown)


docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -199,21 +197,22 @@ const val MAX = 239
...
-Kotlin can also generate static methods for functions defined in named objects or companion objects if you annotate those
-functions as [`@JvmStatic`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-static/index.html).
-If you use this annotation, the compiler generates both a static method in the enclosing class of the object and
-an instance method in the object itself. For example:
+Kotlin represents package-level functions as static methods.
+It also generates static methods for functions defined in named objects or companion objects if you annotate such
+functions with [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/).
+
+For a function in a companion object, the compiler generates both a static method in the enclosing class and
+an instance method in the companion object itself:
```
udalov: Maybe this sentence should mention somehow that this only happens when `@JvmStatic` is applied. Of course, it can be deduced from above and below, but this sentence is in a separate paragraph, so it's not obvious that it applies to annotated functions only...

udalov: In other words, I kinda prefer the previous version of these several lines, as it was more explicit.

udalov (changes requested):

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: Right, let's add more details here

udalov (approved):

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -199,21 +197,22 @@ const val MAX = 239
...
 
 ## Static methods
 
-As mentioned above, Kotlin represents package-level functions as static methods.
-Kotlin can also generate static methods for functions defined in named objects or companion objects if you annotate those
-functions as [`@JvmStatic`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-static/index.html).
-If you use this annotation, the compiler generates both a static method in the enclosing class of the object and
-an instance method in the object itself. For example:
+Kotlin represents package-level functions as static methods.
+It also generates static methods for functions defined in named objects or companion objects if you annotate such
```
daniCsorbaJB: WDYT would this work with something a bit more direct? 🤔 
I'm thinking of: 

> You can generate static methods for functions in named objects or companion objects with the [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/) annotation.

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -199,21 +197,22 @@ const val MAX = 239
...
+functions with [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/).
+
+If you use `@JvmStatic` on a function in a companion object, the compiler generates both a static method in the enclosing
+class and an instance method in the companion object itself:
```
daniCsorbaJB:
```suggestion
class and an instance method in the companion object:
```

daniCsorbaJB: could we remove this? 🤔

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -222,38 +221,38 @@ class C {
 }
 ```
 
-Now, `callStatic()` is static in Java while `callNonStatic()` is not:
+In Java, the `callStatic()` method is static while `callNonStatic()` is not:
```
daniCsorbaJB: hmm 🤔 
I find this a bit difficult to understand — would this be the same? 

In Java, you can call `callStatic()` on either the enclosing class or the companion object.
You can call `callNonStatic()` only on the companion object:

or as one sentence perhaps:

In Java, you can call `callStatic()` on either the enclosing class or the companion object, but you can call `callNonStatic()` only on the companion object:

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -222,38 +221,38 @@ class C {
...
+// Java
+C.callStatic(); // Success
+C.callNonStatic(); // Error: not a static method
+C.Companion.callStatic(); // Instance method remains
+C.Companion.callNonStatic(); // Success
 ```
 
-Similarly, for named objects:
+For a named object (a singleton), `@JvmStatic` turns the function into a static method of the object's class;
+no separate instance method is generated:
```
daniCsorbaJB: to avoid passive would something like this work, WDYT: 

For a named object (a singleton), the compiler generates static methods for functions annotated with `@JvmStatic`, but it doesn't generate a separate instance method:

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -222,38 +221,38 @@ class C {
...
 ```java
-
-Obj.callStatic(); // works fine
-Obj.callNonStatic(); // error
-Obj.INSTANCE.callNonStatic(); // works, a call through the singleton instance
-Obj.INSTANCE.callStatic(); // works too
+// Java
+Obj.callStatic(); // Success
+Obj.callNonStatic(); // Error: not a static method
+Obj.INSTANCE.callNonStatic(); // Success, a call through the singleton instance
```
daniCsorbaJB: this comment reads a bit strange to me. — perhaps we could add this part above the line?

// Calls the method through the singleton instance
Obj.INSTANCE.callNonStatic(); // Success

docs/topics/jvm/java-to-kotlin-interop.md
```diff
@@ -222,38 +221,38 @@ class C {
...
 ```
 
-Starting from Kotlin 1.3, `@JvmStatic` applies to functions defined in companion objects of interfaces as well.
-Such functions compile to static methods in interfaces. Note that static method in interfaces were introduced in Java 1.8,
-so be sure to use the corresponding targets.  
+`@JvmStatic` also applies to functions defined in companion objects of interfaces.
```
daniCsorbaJB:
```suggestion
The `@JvmStatic` annotation also applies to functions defined in companion objects of interfaces.
```

daniCsorbaJB: hmm 🤔 or perhaps if this works might be better, WDYT: 

When you annotate a function in an interface's companion object with `@JvmStatic`, the compiler generates a static method in the interface:

daniCsorbaJB (changes requested):
Nice changes - thank you! 🔥 I added a few comments

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: I'd prefer to keep it as detailed as possible to avoid confusion here

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: I agree, rephrased similar to the sentence below

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: I don't want to change the subject here because the meaning changes a bit, but let's get rid off the passive

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: I think the explanation is covered by the into phrase before the code. I'd rather keep all comment inline for consistency

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: Again, I don't want to change the subject to 'compiler' here, but changed the beginning to avoid 'applies'

docs/topics/jvm/java-to-kotlin-interop.md
daniCsorbaJB: looks good! 💯

docs/topics/jvm/java-to-kotlin-interop.md
daniCsorbaJB: hmm 🤔  the issue I see with it is that we change the structure for the second part of the sentence, so it gives me some pause while reading to understand what we mean:

> In Java, you can call `callStatic()` on both the enclosing class and the companion object, while `callNonStatic()` is only available through the companion object:

You can call vs is available 🤔 so could we make it parallel? 

> In Java, you can call `callStatic()` through both the enclosing class and the companion object, while you can call `callNonStatic()` only through the companion object.

daniCsorbaJB (approved):
nice, I left additional comments in two of the previous conversations, otherwise LGTM 🚀

docs/topics/jvm/java-to-kotlin-interop.md
daniCsorbaJB: hmm 🤔 perhaps with a gerund (even though usually it's not recommended, for me it reads smoother) 

You can also generate static methods for functions defined in named objects or companion objects by annotating them with the [`@JvmStatic`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.jvm/-jvm-static/) annotation.

danil-pavlov: I still see a possible confusion here because it reads like you're annotating objects, not functions

docs/topics/jvm/java-to-kotlin-interop.md
danil-pavlov: I see your point, but the exact repetition here seems a bit weird

PR #5723, opened 2026-08-05
update: adding information about wasmtime in KGP

daniCsorbaJB:
This PR adds information for adding support for Wasmtime in KGP
Related issue is here: [KT-86633](https://youtrack.jetbrains.com/issue/KT-86633)


ilgonmic (approved):

docs/topics/wasm/wasm-wasi.md
danil-pavlov, line 32: Will the template be updated to include Wasmtime? Without it, the task list doesn't match

another option is to show the Gradle build file, like in [#756](https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/756)

docs/topics/wasm/wasm-wasi.md
```diff
@@ -61,19 +62,25 @@ in the terminal from the ` kotlin-wasm-wasi-template` root directory:
 * To run the application in Node.js:
```
danil-pavlov, line 62: There are several "how to do it with Node.js - how to do the same with wasmtime" instructions on this page. What about using tabs?

docs/topics/wasm/wasm-wasi.md
daniCsorbaJB: Yes, the template will be updated for Kotlin 2.4.20 afaik.

> another option is to show the Gradle build file, like in #756

Unfortunately, that way it would require more changes (I did them when testing but you have to remove the custom tests as well then, so it would feel pretty strange without an update)

danil-pavlov (approved):

docs/topics/wasm/wasm-wasi.md
daniCsorbaJB: (as discussed separately, we'll keep the list)

PR #5725, opened 2026-08-05
update: add link to Open VSX registry for Kotlin Language Server

koshachy:


docs/topics/kotlin-lsp.md
```diff
@@ -20,7 +20,9 @@ Kotlin Language Server provides official Kotlin language support for [Visual Stu
 
 If you use Visual Studio Code for Kotlin development,
 install the official [Kotlin by JetBrains](https://marketplace.visualstudio.com/items?itemName=JetBrains.kotlin-server)
-extension from the Visual Studio Marketplace.
+extension from the Visual Studio Marketplace.  
+For VS Code-compatible editors that use Open VSX, 
+install the extension from the [Open VSX Registry](https://open-vsx.org/extension/JetBrains/kotlin-server).
```
fedochet, line 26: I am not sure if there is such a thing as `Open VSX` (without the "Registry")

We can probably write something like this:

```
For other VS Code compatible editors like Cursor or VSCodium, install the extension from Open VSX Registry.
```

docs/topics/kotlin-lsp.md
```diff
@@ -20,7 +20,9 @@ Kotlin Language Server provides official Kotlin language support for [Visual Stu
 
 If you use Visual Studio Code for Kotlin development,
 install the official [Kotlin by JetBrains](https://marketplace.visualstudio.com/items?itemName=JetBrains.kotlin-server)
-extension from the Visual Studio Marketplace.
+extension from the Visual Studio Marketplace.  
+For VS Code-compatible editors that use Open VSX,
```
fedochet: I feel like we should have a newline here, so that paragraphs about VSCode and OpenVSX are clearly separated

fedochet (changes requested):

docs/topics/kotlin-lsp.md
koshachy: Makes sense, added.

docs/topics/kotlin-lsp.md
koshachy: Yes, thank you.
Rewrote it.

fedochet (approved):
Thank you very much!

PR #5726, opened 2026-08-05
fix: correct comment in functions.md

davidc-swl:
Thanks for the great docs.

This example comment incorrectly says items are added to a list `twice`.  

Output of running the example:
```
Adding: first
Adding: second
Items = [first, second]
```


sarahhaggarty (approved):

sarahhaggarty:
Thanks for your contribution @davidc-swl !

PR #5727, opened 2026-08-06
Docs: Add performance tip to map() section

hyfhaozero:
The associate() section has a note about performance, but map() doesn't mention anything about operation order. Added a small tip suggesting filter() before map() to reduce unnecessary work, and asSequence() for large collections.

The asSequence link points to the absolute API URL like other references in the doc.

Let me know if anything needs adjusting.Thank you.

### DIFF
--- a/docs/topics/collection-transformations.md
+++ b/docs/topics/collection-transformations.md
@@ -58,6 +58,19 @@ fun main() {
 ```
 {kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
+**Performance tip**: To improve performance when chaining `map()` with operations like `filter()` or `sorted()`, place `filter()` before `map()` to reduce the number of elements processed. For large collections or complex chains, consider using [`asSequence()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-sequence.html) to avoid creating intermediate collections.
+
+```kotlin
+fun main() {
+//sampleStart
+    val numbers = listOf(1, 2, 3, 4, 5, 6)
+    val result = numbers.filter { it % 2 == 0 }.map { it * 2 }
+    println(result) // [4, 8, 12]
+//sampleEnd
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+
 ## Zip
 
 _Zipping_ transformation is building pairs from elements with the same positions in both collections.


docs/topics/collection-transformations.md
```diff
@@ -58,6 +58,19 @@ fun main() {
...
 
+**Performance tip**: To improve performance when chaining `map()` with operations like `filter()` or `sorted()`, place `filter()` before `map()` to reduce the number of elements processed. For large collections or complex chains, consider using [`asSequence()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-sequence.html) to avoid creating intermediate collections.
+
+```kotlin
+fun main() {
+//sampleStart
+    val numbers = listOf(1, 2, 3, 4, 5, 6)
+    val result = numbers.filter { it % 2 == 0 }.map { it * 2 }
+    println(result) // [4, 8, 12]
+//sampleEnd
```
sarahhaggarty, line 69: It's not always possible to filter before transforming elements because the predicate needs to cover all the original collection elements. But we can show an example of when it is possible to make this swap. Can you update the example to show a before and after with these samples?

Before:
```kotlin
data class Product(val name: String, val price: Double)

fun main() {
//sampleStart
    val products = listOf(
        Product("Laptop", 999.99),
        Product("Mouse", 29.99),
        Product("Keyboard", 79.99),
        Product("Monitor", 249.99)
    )

    val affordableProducts = products
        .map { Pair(it.name, it.price) }
        .filter { (_, price) -> price < 100.0 }

    println(affordableProducts)
    // [(Mouse, 29.99), (Keyboard, 79.99)]
//sampleEnd
}
```

After:

```kotlin
data class Product(val name: String, val price: Double)

fun main() {
//sampleStart
    val products = listOf(
        Product("Laptop", 999.99),
        Product("Mouse", 29.99),
        Product("Keyboard", 79.99),
        Product("Monitor", 249.99)
    )

    val affordableProducts = products
        .filter { it.price < 100.0 }
        .map { Pair(it.name, it.price) }

    println(affordableProducts)
    // [(Mouse, 29.99), (Keyboard, 79.99)]
//sampleEnd
}
```
You can see that filtering uses the same predicate and both versions have the same result.

docs/topics/collection-transformations.md
```diff
@@ -58,6 +58,19 @@ fun main() {
 ```
 
+**Performance tip**: To improve performance when chaining `map()` with operations like `filter()` or `sorted()`, place `filter()` before `map()` to reduce the number of elements processed. For large collections or complex chains, consider using [`asSequence()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-sequence.html) to avoid creating intermediate collections.
```
sarahhaggarty, line 61:
```suggestion
To improve performance when chaining `map()` with operations like `filter()` or `sorted()`, place `filter()` before `map()` to reduce the number of elements processed.
```
You're right about suggesting using sequences instead but this is actually more universal than just for maps so to mention it here specifically isn't the best place. Please can you remove the sentence?

sarahhaggarty (changes requested):
Thanks for the suggestion @hyfhaozero !
I reviewed it with our developer team who came back with an example that more easily demonstrates what you are suggesting. Please take a look at my review comments and if you have any questions, don't hesitate to ask!

docs/topics/collection-transformations.md
hyfhaozero: Oh, I see. Thank you very much for your reminder, explanation, and guidance. I now understand this mistake. I appreciate your reply.And I'll take your advice

sarahhaggarty:
Hi @hyfhaozero !
Do you need some help updating the pull request based on my feedback? Please let me know!

PR #5729, opened 2026-08-06
feat: new compilation filter for Power-assert

danil-pavlov:
This PR adds the info on the new `compilationFilter` option ([KT-84811](https://youtrack.jetbrains.com/issue/KT-84811))

### DIFF
--- a/docs/topics/compiler-plugins/power-assert.md
+++ b/docs/topics/compiler-plugins/power-assert.md
@@ -63,13 +63,16 @@ plugins {
 
 The Power-assert plugin provides several options to customize its behavior:
 
-* **`functions`** lists fully qualified paths to functions whose calls the Power-assert plugin transforms. If not specified, only `kotlin.assert()` calls are transformed.
-* **`compilationFilter`** controls which Kotlin compilations the Power-assert plugin is applied to. You can create your own custom filter or use the predefined options:
+* **`functions`** lists the fully qualified paths to functions that the Power-assert plugin transforms when they're called.
+  If not specified, the plugin transforms only `kotlin.assert()` calls.
+* **`compilationFilter`** controls which Kotlin compilations the Power-assert plugin is applied to. You can create your
+  own custom filter or use the predefined options:
   * `PowerAssertCompilationFilter.TESTS` applies to all test source sets (default).
   * `PowerAssertCompilationFilter.ALL` applies to all source sets.
 
-  > The `compilationFilter` option has replaced the deprecated `includedSourceSets`, which listed Gradle source sets that the Power-assert plugin transforms.
-  > These two options are mutually exclusive: if `includedSourceSets` is specified, `compilationFilter` is ignored.
+  > The `compilationFilter` option has replaced the deprecated `includedSourceSets`, which listed Gradle source sets that
+  > the Power-assert plugin transforms. These two options are mutually exclusive: if `includedSourceSets` is specified,
+  > `compilationFilter` is ignored.
   >
   {style="note"}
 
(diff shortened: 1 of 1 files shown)


docs/topics/compiler-plugins/power-assert.md
```diff
@@ -63,8 +63,9 @@ plugins {
 
 The Power-assert plugin provides several options to customize its behavior:
 
-* **`functions`**: A list of fully-qualified function paths. The Power-assert plugin will transform the calls to these functions. If not specified, only `kotlin.assert()` calls will be transformed by default.
-* **`includedSourceSets`**: A list of Gradle source sets that the Power-assert plugin will transform. If not specified, all _test source sets_ will be transformed by default.
+* **`functions`** lists fully qualified paths to functions whose calls the Power-assert plugin transforms. If not specified, only `kotlin.assert()` calls are transformed.
+* **`includedSourceSets`** lists Gradle source sets that the Power-assert plugin transforms. If not specified, all _test source sets_ are transformed.
+* **`compilationFilter`** controls if the Power-assert plugin is applied to all Kotlin compilations or only to test compilations. If not specified, the plugin is applied to all compilations.
```
bnorm: These 2 options are mutually exclusive. If `includedSourceSets` is specified, the `compilationFilter` is ignored. `includedSourceSets` is now deprecated, so it should not be recommended. You can achieve the same behavior with `compilationFilter` like this:
```kotlin
powerAssert {
  compilationFilter = PowerAssertCompilationFilter {
    it.name in setOf("commonMain", "jvmMain", "jsMain", "nativeMain")
  }
}
```

bnorm: I would say that it controls to which Kotlin compilations Power-assert plugin is applied. It's not limited to all or tests, there's a lot more control since you can create your own custom filter.

And the default for `compilationFilter` is `PowerAssertCompilationFilter.TESTS`, so all _test source sets_, just like it is for `includedSourceSets`.

bnorm (changes requested):

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: Thank you for the clarification! added a note about the deprecation

bnorm (approved):

docs/topics/compiler-plugins/power-assert.md
```diff
@@ -63,8 +63,15 @@ plugins {
 
 The Power-assert plugin provides several options to customize its behavior:
 
-* **`functions`**: A list of fully-qualified function paths. The Power-assert plugin will transform the calls to these functions. If not specified, only `kotlin.assert()` calls will be transformed by default.
-* **`includedSourceSets`**: A list of Gradle source sets that the Power-assert plugin will transform. If not specified, all _test source sets_ will be transformed by default.
+* **`functions`** lists fully qualified paths to functions whose calls the Power-assert plugin transforms. If not specified, only `kotlin.assert()` calls are transformed.
```
sarahhaggarty:
```suggestion
* **`functions`** lists the fully qualified paths to functions that the Power-assert plugin transforms when called. If not specified, the plugin transforms only `kotlin.assert()` calls.
```
I find "whose calls" a bit hard to read, though I can't put a finger on exactly why. It's probably the nesting. I suggested an alternative that avoids that. I also suggested making the second sentence active.

sarahhaggarty (approved):
Looks good! Only two suggestions for you to consider.

docs/topics/compiler-plugins/power-assert.md
danil-pavlov: I like it, but I think 'when called' may sound like 'when the plugin is called', so I changed it a bit

PR #5732, opened 2026-08-07
feat: new page for enum classes

danil-pavlov:


docs/topics/enum-classes.md
```diff
@@ -1,123 +1,192 @@
...
+    // `Direction.NORTH` is an enum constant of type `Direction`.
+    val direction: Direction = Direction.NORTH
+
+    println(direction) // NORTH
 }
 ```
+{kotlin-runnable="true" id="create-enum-class-kotlin"}
 
-## Anonymous classes
+Every enum class in Kotlin inherits from the [`Enum<T>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-enum/) constructor,
```
InversionSpaces: In internal kotlin compiler terminology, `Enum<T>` is a **type constructor**, but I am a bit confused here, from the user POV, in what sense `Enum<T>` is a "constructor"?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,192 @@
 [//]: # (title: Enum classes)
 
-The most basic use case for enum classes is the implementation of type-safe enums:
+Enum classes represent a fixed set of possible values. Use an enum class when a value can only be one of several
+predefined options, such as available states or modes.
+
+Each value in an enum class is called an _enum constant_. Enum constants are [singleton objects](object-declarations.md)
```
InversionSpaces: I think it is important to mention that enum entries are similar to singleton objects, but not quite the same.
One important user-facing distinction, I believe, is that you can't access enum entry declarations:

```kotlin
enum class Test {
    A {
        val prop = 42
    };
}

fun main() {
    println(Test.A.prop) // UNRESOLVED_REFERENCE
}
```

A name of singleton object is also a valid type, while name of a enum entry is not:

```kotlin
object B

fun test(b: B) {
    val b = b is B
}

fun invalid(a: Test.A) {} // ENUM_ENTRY_AS_TYPE
```

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,192 @@
...
+of the enum class type, so they can have properties, functions, and custom behavior.
+
+Use enum classes when all possible values are known in advance and have the same structure. Use [sealed classes or interfaces](sealed-classes.md)
+when you need to hold different data or have a different structure for each case.
+
+## Create enum classes
```
InversionSpaces: nit-picking: is not "declare" a more appropriate instead of "create"  here?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,192 @@
...
+Since enum constants are values, you can assign them to variables, print them, pass them to functions, compare them,
+and use them in `when` expressions.
 
-    TALKING {
-        override fun signal() = WAITING
-    };
+### Declare enum constants
 
-    abstract fun signal(): ProtocolState
+In Kotlin, enum constants don't use assignment syntax like `RED = "#FF0000"` or `LOW = 0`.
```
InversionSpaces: nit-picking: I believe it is very annoying when a paragraph is intended to tell you how to do something, but the first thing it tells you is "how NOT to do it". I suppose this is for people coming from languages like C++ where enums are defined like this, but for general reader it feels off.

docs/topics/enum-classes.md
```diff
@@ -126,27 +195,387 @@ fun main() {
...
+fun main() {
+    val color = Color.entries.getOrNull(0)
+
+    println(color) // RED
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="find-enum-getornull-kotlin"}
+
+Enum positions start from `0`. In this example, `RED` has position 0, `GREEN` has position 1, and `BLUE` has position 2.
```
InversionSpaces: nit-picking: I think number constants are quoted inconsistently in this text.

docs/topics/enum-classes.md
```diff
@@ -126,27 +195,387 @@ fun main() {
...
+Companion object helper functions are useful when you want safe lookups that return null instead of throwing an exception.
+
+The lookup APIs used above, such as `entries` and `valueOf()`, are examples of _synthetic_ members.
+In this context, synthetic means that Kotlin provides these members automatically, even though you don't declare them
+yourself. This is why every enum class can list its constants with `entries` and get a constant by name with `valueOf()`
+without you writing extra code.
+
+Some generic enum helper functions use reified type parameters. A reified type parameter keeps the actual enum type
+available inside an inline generic function. This allows functions such as [`enumEntries<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.enums/enum-entries.html) and [`enumValueOf<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/enum-value-of.html) to
+work with the enum type `T` directly. You can access the constants in an enum class using the following functions:
```
InversionSpaces: nit-picking: this paragraph goes from "how these functions work" to "what they can be used for". I think for the user it is more convenient to have other way around.

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,192 @@
...
+
+    println(direction) // NORTH
 }
 ```
+{kotlin-runnable="true" id="create-enum-class-kotlin"}
 
-## Anonymous classes
+Every enum class in Kotlin inherits from the [`Enum<T>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-enum/) constructor,
+where `T` is the enum class itself. For example, the `Direction` enum class inherits from `Enum<Direction>`.
+This is why enum constants have built-in properties, such as [`name`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-enum/name.html) and [`ordinal`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-enum/ordinal.html).
```
InversionSpaces: nit-picking: a little bit confusing here where __implementation__ of those properties comes from. It is described better later, but here might be a little bit not that clear.

docs/topics/enum-classes.md
danil-pavlov: Right, let's use the same term as in the API docs

docs/topics/enum-classes.md
danil-pavlov: Thank you for the clarification! I fixed the wording and added these two points to the corresponding sections

docs/topics/enum-classes.md
danil-pavlov: Yeah, it was one of the requests from confused users with knowledge of other languages
I suggest keeping this point, but let's rephrase

docs/topics/enum-classes.md
danil-pavlov: True, let's add a direct link to the detailed explanation below

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,197 @@
...
-    override fun applyAsInt(t: Int, u: Int) = apply(t, u)
+fun main() {
+    println(Priority.MEDIUM.level) // 1
+}
+```
+{kotlin-runnable="true" id="declare-enum-level-kotlin"}
+
+### Access enum constants and their properties
+
+You can access an enum constant through the enum class name:
```
InversionSpaces: nit: Repetition is okay, but this info seems a little bit out of order, because you already accessed enum constants in snippets above.

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,197 @@
...
+//sampleEnd
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.9" id="rgb-enums-kotlin"}
+{kotlin-runnable="true" id="pass-enum-to-function-kotlin"}
+
+Here, the `printColor()` function accepts a value of type `Color`, so you can pass any `Color` enum constant to it.
 
-The `valueOf()` method throws an `IllegalArgumentException` if the specified name does
-not match any of the enum constants defined in the class.
+> Although enum constants behave like singleton objects, they aren't the same. Only the enum class can be used as a type.
```
InversionSpaces: nit: an example would be good, I think it can be confusing for a user what does this mean without an example

InversionSpaces (approved):
Feel free to ignore my last comments

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+    println(color) // RED
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="find-enum-first-kotlin"}
+
+The `first()` function throws a `NoSuchElementException` if no matching constant is found. To get `null` instead, use
+[`firstOrNull()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/first-or-null.html).
+
+To get the number of enum constants, use the [`size`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-list/size.html) function. For example:
```
sarahhaggarty:
```suggestion
To get the number of enum constants, use the [`size`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-list/size.html) property. For example:
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+
+        fun fromRgb(rgb: Int): Color? =
+            entries.find { it.rgb == rgb }
+    }
+}
+
+fun main() {
+    println(Color.fromName("RED"))   // RED
+    println(Color.fromPosition(1))   // GREEN
+    println(Color.fromRgb(0x0000FF)) // BLUE
```
sarahhaggarty:
```suggestion
    println(Color.fromRgb(0x0000FF)) // BLUE
```

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
 
-The most basic use case for enum classes is the implementation of type-safe enums:
+Enum classes represent a fixed set of possible values. Use an enum class when a value can only be one of several
+predefined options, such as available states or modes.
+
+Each value in an enum class is called an _enum constant_. Enum constants behave like [singleton objects](object-declarations.md)
+of the enum class type, so they can have properties, functions, and custom behavior.
+
+Use enum classes when all possible values are known in advance and have the same structure. Use [sealed classes or interfaces](sealed-classes.md)
+when you need to hold different data or have a different structure for each case.
```
sarahhaggarty:
```suggestion
Enum classes are the best fit when all possible values are known in advance and have the same structure. If you need to hold different data or have a different structure for each case, use [sealed classes or interfaces](sealed-classes.md).
```
I got a bit confused with the duplicate "use an enum class/use enum classes". I think the main point is to contrast with sealed classes so what do you think about adjusting the text like this?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
+
+## Declare enum classes
+
+To create an enum class, use the `enum` keyword and list the enum constants inside the class body, separated by
+commas:
```
sarahhaggarty:
```suggestion
To create an enum class, use the `enum` keyword and follow the usual class syntax with a body enclosed in curly braces. Inside the class body, list the enum constants separated by commas:
```
You don't mention the curly brace syntax. What do you think about adding it like this?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
+and use them in `when` expressions.
 
-    TALKING {
-        override fun signal() = WAITING
-    };
+### Declare enum constants
 
-    abstract fun signal(): ProtocolState
+To declare enum constants, first define properties in the enum class constructor and then pass values to each enum
+constant in parentheses. Don't use assignment syntax, for example `RED = "#FF0000"`, like in C#.
```
sarahhaggarty: We don't usually make comparisons to other programming languages. Is this for a specific feedback ticket?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
-{kotlin-runnable="true" kotlin-min-compiler-version="1.9"}
+{kotlin-runnable="true" id="access-enum-properties-kotlin"}
 
-All enum classes implement the [Comparable](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-comparable/index.html)
-interface by default. Constants in the enum class are defined in the natural order. For more information, see [Ordering](collection-ordering.md).
+Here, `Color.RED` is an enum constant of the `Color` type. The color variable stores this enum constant.
 
-## Working with enum constants
+When you print an enum constant directly, Kotlin prints its name. To access a property associated with an enum constant,
+use dot notation, such as `color.hex` or `Color.GREEN.hex`.
```
sarahhaggarty: Can you add the syntax explanation before the example? It seems late afterwards.

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
 
-Enum classes in Kotlin have synthetic properties and methods for listing the defined enum constants and getting an enum constant by 
-its name. The signatures of these methods are as follows (assuming the name of the enum class is `EnumClass`):
+Besides any properties you define, every enum constant also has the built-in `name` and `ordinal` properties for
+obtaining its name and position (starting from 0) in the enum class declaration:
```
sarahhaggarty:
```suggestion
obtaining its name and position (starting from zero) in the enum class declaration:
```

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
+    println(RGB.RED.name)    // RED
+    println(RGB.RED.ordinal) // 0
+}
 ```
+{kotlin-runnable="true" id="rgb-enums-properties-kotlin"}
 
-Below is an example of them in action:
+### Pass enum constants to functions
+
+Because enum constants are values, you can pass them to functions:
```
sarahhaggarty: It would be nice to explain why this is useful.

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
-{kotlin-runnable="true" kotlin-min-compiler-version="1.9" id="rgb-enums-kotlin"}
+{kotlin-runnable="true" id="pass-enum-to-function-kotlin"}
 
-The `valueOf()` method throws an `IllegalArgumentException` if the specified name does
-not match any of the enum constants defined in the class.
+Here, the `printColor()` function accepts a value of type `Color`, so you can pass any `Color` enum constant to it.
 
-Prior to the introduction of `entries` in Kotlin 1.9.0, the `values()` function was used to retrieve an array of enum constants.
+Although enum constants behave like singleton objects, they aren't the same. You can only use the enum class name as a type.
+On the contrary, enum constants cannot be valid types, only values:
```
sarahhaggarty:
```suggestion
Although enum constants behave like singleton objects, the compiler treats them as values of the enum class type. You can use the enum class name itself as a type, but you can't use enum constants as enum types:
```
I struggled with this. I think because the wording is a little formal and "they aren't the same" felt a bit vague. What do you think about this version?

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
+
+fun printRed(color: Color.RED) {
+    println(color)
+    // Error: enum entry cannot be used as a type
+}
+```
+
+### Use enum constants in `when` expressions
+
+Enum classes are commonly used with when expressions when you want to handle each constant separately:
```
sarahhaggarty:
```suggestion
Enum classes work best with `when` expressions when you want to handle each constant separately:
```
To avoid the passiveness of "are commonly used".

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
+    // Red is a warm color
 }
+//sampleEnd
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="rgb-enums-properties-kotlin"}
+{kotlin-runnable="true" id="enum-when-expression-kotlin"}
+
+When you use all enum constants in a when expression, you don't need an `else` branch.
 
 > To reduce repetition when working with enum entries, try out context-sensitive resolution (currently in preview).
```
sarahhaggarty:
```suggestion
> To reduce repetition when working with enum entries, try context-sensitive resolution (currently in preview).
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+
+    println(color) // RED
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="find-enum-getornull-kotlin"}
+
+Enum positions start from `0`. In this example, `RED` has the position `0`, `GREEN` is at `1`, and `BLUE` is at `2`.
+
+This is useful when you need to look up an enum constant by index. Kotlin doesn't support directly casting an `Int` to
```
sarahhaggarty: It's not clear to me how Kotlin not supporting casting to `Int` relates to accessing an enum constant via index. Is the "integer" an enum entry?

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+fun main() {
+    println(Color.fromName("RED"))   // RED
+    println(Color.fromPosition(1))   // GREEN
+    println(Color.fromRgb(0x0000FF)) // BLUE 
+    println(Color.fromRgb(0xABCDEF)) // null
+}
+```
+{kotlin-runnable="true" id="find-enum-companion-object-kotlin"}
+
+Companion object helper functions are useful when you want safe lookups that return null instead of throwing an exception.
```
sarahhaggarty:
```suggestion
Companion object helper functions are useful when you want safe lookups that return `null` instead of throwing an exception.
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+
+The lookup APIs used above, such as `entries` and `valueOf()`, are examples of _synthetic_ members.
+In this context, synthetic means that Kotlin provides these members automatically, even though you don't declare them
+yourself. This is why every enum class can list its constants with `entries` and get a constant by name with `valueOf()`
```
sarahhaggarty:
```suggestion
yourself. This is why every enum class can list its constants with the `entries` property and get a constant by name with the `valueOf()` function
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+without you writing extra code.
```
sarahhaggarty:
```suggestion
without you needing to write extra code.
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+
+You can access the constants in an enum class using generic helper functions such as [`enumEntries<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.enums/enum-entries.html) and [`enumValueOf<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/enum-value-of.html).
+These functions use reified type parameters. Such parameters keep the actual enum type available inside a generic inline
```
sarahhaggarty: What do you think about linking to https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters?

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+    HIGH, LOW, MEDIUM
+}
+
+fun main() {
+    println(Priority.entries.sorted()) // [HIGH, LOW, MEDIUM]
+}
+```
+{kotlin-runnable="true" id="sort-enum-declaration-order-kotlin"}
+
+The `Enum<T>` class provides the `compareTo()`, `equals()`, and `hashCode()` functions, and you can't override them to
```
sarahhaggarty, line 441: Just checking, are these extension functions?

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+}
+
+fun main() {
+    println(Direction.NORTH.isVertical()) // true
+    println(Direction.EAST.isVertical())  // false
+}
+```
+{kotlin-runnable="true" id="enum-shared-function-kotlin"}
+
+Every enum constant can call the shared function. Inside the function, `this` refers to the enum constant it was called on.
```
sarahhaggarty:
```suggestion
Every enum constant can call the shared function. Inside the function, `this` refers to the enum constant it's called on.
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+
+fun main() {
+    println(!Direction.NORTH) // SOUTH
+}
+```
+{kotlin-runnable="true" id="enum-operator-function-kotlin"}
+
+## Use anonymous classes
+
+Enum constants can declare their own anonymous classes with their corresponding methods, as well as with overriding base
```
sarahhaggarty: Can you use "functions" instead of "methods"? Methods is very reminiscent of Java.

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+methods. With anonymous classes, you write the class body directly after the enum constant
+name, and Kotlin infers the enum class as the supertype.
```
sarahhaggarty:
```suggestion
name and Kotlin infers the enum class as the supertype.
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+    state = state.signal()
+    println(state) // TALKING
+}
+```
+{kotlin-runnable="true" id="enum-abstract-function-kotlin"}
+
+Here, each constant implements the abstract `signal()` function differently, so calling `signal()` returns a different
+next state depending on the constant.
+
+Although enum constants behave like singleton objects, they aren't the same. You can't access members declared inside
```
sarahhaggarty: Please see my suggestion above to try and make "they aren't the same" clearer.

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+the body anonymous class:
```
sarahhaggarty:
```suggestion
the body of an anonymous class:
```

docs/topics/enum-classes.md
```diff
@@ -126,27 +203,413 @@ fun main() {
...
+```
+
+To expose data or behavior for every constant, declare it in the enum class body, using an abstract member when each
+constant needs its own implementation.
+
+## Implement interfaces in enum classes
+
+An enum class can implement an interface (but it cannot derive from a class), providing either a common implementation of
+interface members for all the entries or separate implementations for each entry within its anonymous class.
+This is done by adding the interfaces you want to implement to the enum class declaration as follows:
```
sarahhaggarty:
```suggestion
An enum class can implement an interface, but it cannot inherit from a class. You can provide a common implementation of 
the interface members for all enum constants or let each constant provide its own implementation in an anonymous class.

To implement an interface, add it to the enum class declaration:
```
The first sentence is so long 😿 
I suggest to change "derive" to "inherit" because it is more commonly used in our inheritance docs. I also suggested a more direct alternative to the passive "this is done by adding".

docs/topics/enum-classes.md
```diff
@@ -1,123 +1,200 @@
...
-    BLUE(0x0000FF)
+enum class Direction {
+    NORTH, SOUTH, WEST, EAST
+}
+
+fun main() {
+    // `Direction.NORTH` is an enum constant of type `Direction`.
+    val direction: Direction = Direction.NORTH
+
+    println(direction) // NORTH
```
sarahhaggarty: Sometimes your output code comments are on the same line and sometimes they are below. Can you make them consistent in the doc? The guidelines say to add them below but if you have a big code block demonstrating different outputs sequentially then I think it's OK to add them on the same line.

sarahhaggarty (changes requested):
This is sooooooo much better! I do have some questions and suggestions though 🙃 Let me know if anything I've said is unclear.

docs/topics/enum-classes.md
danil-pavlov: Yeah, it was, I changed the wording a bit

docs/topics/enum-classes.md
danil-pavlov: Thank you for noticing, I changed it to be consistent with other mentions of positions

docs/topics/enum-classes.md
danil-pavlov: added more details

docs/topics/enum-classes.md
danil-pavlov: No, they are not extensions

docs/topics/enum-classes.md
danil-pavlov: Yeah, it was the old text

docs/topics/enum-classes.md
danil-pavlov: I moved all the comments to the next line

docs/topics/enum-classes.md
```diff
@@ -1,152 +1,641 @@
...
+```
+{kotlin-runnable="true" id="rgb-enums-properties-kotlin"}
 
-## Implementing interfaces in enum classes
+### Pass enum constants to functions
 
-An enum class can implement an interface (but it cannot derive from a class), providing either a common implementation of
-interface members for all the entries, or separate implementations for each entry within its anonymous class.
-This is done by adding the interfaces you want to implement to the enum class declaration as follows:
+Because enum constants are values, you can pass them to functions. This way a function accepts only the fixed set of
```
sarahhaggarty:
```suggestion
Because enum constants are values, you can pass them to functions. This way, a function accepts only the fixed set of
```

docs/topics/enum-classes.md
```diff
@@ -1,152 +1,641 @@
...
+    // RED
 }
+//sampleEnd
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.9" id="rgb-enums-kotlin"}
+{kotlin-runnable="true" id="find-enum-getornull-kotlin"}
+
+Enum positions start from `0`. In this example, `RED` has the position `0`, `GREEN` is at `1`, and `BLUE` is at `2`.
+
+This is useful when you have an integer, for example read from a file or user input, and need the matching enum constant.
```
sarahhaggarty:
```suggestion
This is useful when you have an integer, for example from a file or user input, that represents an enum constant's position.
```

docs/topics/enum-classes.md
```diff
@@ -1,152 +1,641 @@
...
+Unlike some other languages, Kotlin doesn't let you cast an `Int` directly to an enum constant. Instead, use the integer
+as a position index and look up the constant with `entries.getOrNull(index)` or define a stable numeric property and
+search for the constant with the matching value.
```
sarahhaggarty:
```suggestion
as an index and look up the constant with `entries.getOrNull(index)`. If the integer represents
a value that should remain stable even if you reorder the enum constants, define a numeric property on the enum class and search for the constant with the matching value. For example, `rgb` or `code`.
```
What do you think about combining the note into the text above?

docs/topics/enum-classes.md
```diff
@@ -1,152 +1,641 @@
...
-you call `enumEntries<T>()` the same list is returned each time, which is far more efficient.
+Companion object helper functions are useful when you want safe lookups that return `null` instead of throwing an exception.
+
+The lookup APIs used above, such as `entries` and `valueOf()`, are examples of _synthetic_ members.
+In this context, synthetic means that Kotlin provides these members automatically, even though you don't declare them
+yourself. This is why every enum class can list its constants with the `entries` property and get a constant by name with
+the `valueOf()` function without the need to write extra code.
+
+You can access the constants in an enum class using generic helper functions such as [`enumEntries<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.enums/enum-entries.html) and [`enumValueOf<T>()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/enum-value-of.html).
+These functions use [reified type parameters](https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters).
```
sarahhaggarty:
```suggestion
These functions use [reified type parameters](inline-functions.md#reified-type-parameters).
```

docs/topics/enum-classes.md
```diff
@@ -1,152 +1,641 @@
...
+    println(state)
+    // TALKING
+}
+```
+{kotlin-runnable="true" id="enum-abstract-function-kotlin"}
+
+Here, each constant implements the abstract `signal()` function differently, so calling `signal()` returns a different
+next state depending on the constant.
+
+Although enum constants behave like singleton objects, the type of enum constant is the enum class itself,
```
sarahhaggarty:
```suggestion
Although enum constants behave like singleton objects, the type of an enum constant is the enum class itself,
```

sarahhaggarty (approved):
Nice! Only a few more minor suggestions from me.

PR #5737, opened 2026-08-11
update: new exercise for objects

danil-pavlov:


docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,48 +293,68 @@ fun main() {
 
 ### Exercise 3 {initial-collapse-state="collapsed" collapsible="true" id="objects-exercise-3"}
 
-You have an app where you want to record temperatures. The class itself stores the information in Celsius, but 
-you want to provide an easy way to create an instance in Fahrenheit as well. Complete the data class so that
+You are building a user registration module for an app. When users are creating accounts, you want to
+validate email addresses without instantiating a `User` object first. Complete the data class so that
```
sarahhaggarty:
```suggestion
You are building a user registration module for an app. You want to keep email validation associated with the `User` class 
and make it available without a `User` instance. Complete the data class so that
```

docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,48 +293,68 @@ fun main() {
...
 the following code in the `main()` function runs successfully:
```
sarahhaggarty, line 298:
```suggestion

For this exercise, consider an email address valid if it contains both `@` and `.`.

```

docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,48 +293,68 @@ fun main() {
...
 
 <deflist collapsible="true">
     <def title="Hint">
-        Use a companion object.
+        Use a companion object with a utility function.
```
sarahhaggarty:
```suggestion
        Add an email validation function in a companion object for the `User` class so that you can call the function directly on `User`.
```

sarahhaggarty (commented):
I like the exercise! I suggested some changes to give a little more guidance and explanation why the exercise is set up the way it is.

docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,48 +293,70 @@ fun main() {
 
 ### Exercise 3 {initial-collapse-state="collapsed" collapsible="true" id="objects-exercise-3"}
 
-You have an app where you want to record temperatures. The class itself stores the information in Celsius, but 
-you want to provide an easy way to create an instance in Fahrenheit as well. Complete the data class so that
+You are building a user registration module for an app. You want to keep email validation associated with the `User` class 
+and make it available without a `User` instance. Complete the data class so that
```
serras: It may be useful to elaborate _why_ we need to make it available without a `User` instance.

serras (approved):
I like the change, I think it should a nicer example of when you definitely cannot use an instance. However, in doing so we lose an example of another different pattern: using companion object methods to _construct_ an instance. For me, both usages of companions are equally interesting, and it would be great if the tour could highlight both.

docs/topics/tour/kotlin-tour-intermediate-objects.md
danil-pavlov: updated the description

danil-pavlov:
> I like the change, I think it should a nicer example of when you definitely cannot use an instance. However, in doing so we lose an example of another different pattern: using companion object methods to _construct_ an instance. For me, both usages of companions are equally interesting, and it would be great if the tour could highlight both.

The problem is that we don't have a theory in the tour for this pattern, but let's add a note that leads to the theory and to examples of factory functions (as an extension of the last exercise for the curious).

docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,52 +293,79 @@ fun main() {
...
-        Use a companion object.
+        Add an email validation function in a companion object for the `User` class so that you can call the function directly on `User`.
     </def>
 </deflist>
 
 |---|---|
 ```kotlin
-data class Temperature(val celsius: Double) {
-    val fahrenheit: Double = celsius * 9 / 5 + 32
+data class User(val name: String, val email: String) {
```
sarahhaggarty:
```suggestion
```

docs/topics/tour/kotlin-tour-intermediate-objects.md
```diff
@@ -293,52 +293,79 @@ fun main() {
...
+            println("Error: '${email}' is not valid. The email should contain '@' and '.'")
+            // Error: 'bob2example-com' is not valid. The email should contain '@' and '.'
+        }
+    }
 }
 ```
 {initial-collapse-state="collapsed" collapsible="true" collapsed-title="Example solution" id="kotlin-tour-objects-solution-3"}
 
+> As an extension of this exercise, try using class-level functions in companion objects as factory methods to construct
+> instances of a class. For an example of this pattern and more information, see [](object-declarations.md#companion-objects).
```
sarahhaggarty:
```suggestion
> As an extension of this exercise, try using functions in companion objects as factory methods to construct
> instances of a class. For an example and more information about this pattern, see [](object-declarations.md#companion-objects).
```

sarahhaggarty (approved):
Looks great! Just a couple more minor suggestions for you to consider.

PR #5745, opened 2026-08-12
update: new syntax for tours and buttons

danil-pavlov:


docs/topics/native/mapping-function-pointers-from-c.md
```diff
@@ -170,7 +161,14 @@ or use the console command in your terminal, in this example:
...
+<list columns="2" id="tour-nav">
+  <li>
+    <a as="button" href="mapping-struct-union-types-from-c.md" mode="outline" icon="arrow-left" icon-position="left">Previous step</a>
+  </li>
+  <li>
+    <a as="button" href="mapping-strings-from-c.md" mode="classic" icon="arrow-right" icon-position="right">Next step</a>
+  </li>
+</list>
 
 ### See also
```
sarahhaggarty, line 173:
```suggestion
## See also
```

sarahhaggarty (approved):
I noticed one pre-existing nesting error! 🙃

docs/topics/native/mapping-function-pointers-from-c.md
danil-pavlov: I checked, it is consistent with other pages, "See also" under "Next step"

sarahhaggarty: Hmm. But on the staging it shows the first two pages as "See also" on the same level as "Next step" 🤔

danil-pavlov: Oh, thank you for pointing that out! there was a syntax error

docs/topics/jvm/jvm-get-started-spring-boot.md
```diff
@@ -7,17 +7,15 @@ application with Spring Boot and adding a database to store the information.
...
+   <img src="icon-3.svg" width="20" alt="Third step"/> <a href="jvm-spring-boot-add-db-support.md">Add database support for the Spring Boot project</a><br/>
+   <img src="icon-4.svg" width="20" alt="Fourth step"/> <a href="jvm-spring-boot-using-crudrepository.md">Use Spring Data CrudRepository for database access</a><br/>
+</p>
 
 ## Next step
 
-Start by [creating a Spring Boot project](jvm-create-project-with-spring-boot.md) with Kotlin using IntelliJ IDEA.
+<a href="jvm-create-project-with-spring-boot.md" as="button" mode="classic" icon="arrow-right" icon-position="right">Start</a>
 
 ### See also
```
koshachy, line 20: Could you please check (and remove) join the community section if it is not performing?
If it is still there, you can press Start button without scrolling and the first step wouldn't be marked as Completed.

koshachy (approved):
Approved!
Please, take a look at the Spring Boot Join the community section.

docs/topics/jvm/jvm-get-started-spring-boot.md
danil-pavlov: True, there aren't many clicks on those links. Should we move "Community section" to the bottom of the Backend overview instead?

koshachy: I think it makes sense

PR #5748, opened 2026-08-14
chore: update core api reference docs (2.4.20)

ilya-g:
Please ensure that the default Kotlin version on play.kotlinlang.org is updated to 2.4.20-... before merging this, otherwise runnable samples for the functions added in 2.4.20 would produce compilation errors, for example one in `kotlin-stdlib/kotlin.collections/all-distinct.html`


berezinant:
We are going to release this after 2.4.20 on Sept 1

berezinant (approved):

PR #5753, opened 2026-08-18
Remove-why-ksp

budindepunk:
[KEDOC-323](https://youtrack.jetbrains.com/issue/KEDOC-323/Delete-Why-KSP-from-KSP-docs)

### DIFF
--- a/docs/kr.tree
+++ b/docs/kr.tree
@@ -356,7 +356,7 @@
             <toc-element toc-title="Compiler options" topic="compose-compiler-options.md"/>
         </toc-element>
         <toc-element toc-title="Kotlin Symbol Processing API">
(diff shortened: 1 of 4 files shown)


docs/kr.tree
```diff
@@ -358,7 +358,6 @@
         <toc-element toc-title="Kotlin Symbol Processing API">
             <toc-element toc-title="KSP overview" topic="ksp-overview.md"/>
```
koshachy: In this case, it is better to use Writerside redirect.
So there is no need to update `pages.yml`

```suggestion
            <toc-element accepts-web-file-names="ksp-why-ksp.html" toc-title="KSP overview" topic="ksp-overview.md"/>
```

koshachy (changes requested):

koshachy (approved):
LGTM!

PR #5754, opened 2026-08-18
KTL-4761 Clean up of the old files

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-4761/Simplify-the-local-setup-of-kotlinlang.org-for-external-contributors


package.json
berezinant: look like webpack-dev-server is not used after `"start": "webpack serve"` is not needed

berezinant (commented):
in api reference testing helper script (e.g. [up.sh](https://github.com/user-attachments/files/31307937/up.sh) for [ApiReferencesTemplateTest.kt](https://github.com/JetBrains/kotlin-web-site/blob/41b12c92afc8fbd7afcd0c8f6083df0e6d7abccf/.teamcity/tests/buildTypes/ApiReferencesTemplateTest.kt)) we rely on dockerfiles (e.g. docker build -t "$APP_CONTAINER_NAME" -f ./dockerfiles/nginx-server/Dockerfile .)

migth worth checking it

berezinant (approved):

PR #5755, opened 2026-08-18
update: document the reflection library

aepanchi:
Document kotlin reflection. Changes:

- reflection.md describes kotlin-reflect; mobed to the Library guides section
- Callable references are documented on the corresponding pages:
  - classes.md: Constructor references
  - fun-interfaces.md: Pass callable references to SAM parameters
  - lambdas.md: Callable reference
  - properties.md: Property references
  - sealed-classes.md: Inspect sealed subclasses with reflection

### DIFF
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -444,6 +444,31 @@ fun main() {
 ```
 {kotlin-runnable="true" id="class-delegation-sequence"}
 
+## Constructor references
+
+Use a constructor reference when an API needs a factory, but construction already provides the required behavior. In this case,
+you don't need to repeat it in a [lambda](lambdas.md).
+
+To create a reference to a constructor, use `::ClassName`:
+
+```kotlin
+data class User(val name: String, val age: Int)
+
+fun createUser(factory: (String, Int) -> User): User =
+    factory("Jane Doe", 22)
+
+fun main() {
+    val user = createUser(::User)
+    println(user)
+    // User(name=Jane Doe, age=22)
+}
+```
+{kotlin-runnable="true"}
+
+Callable references to constructors are typed as one of the [`KFunction<out R>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-function/) subtypes depending on the parameter count.
+However, the compiler resolves and checks the constructor references at compile time and doesn't require the
+[kotlin-reflect](reflection.md) library.
+
 ### Classes without constructors
 
 Classes that don't declare any constructors (primary or secondary) have an implicit primary constructor
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -444,31 +444,6 @@ fun main() {
 ```
 {kotlin-runnable="true" id="class-delegation-sequence"}
 
-### Constructor references
-
-Use a constructor reference when an API needs a factory, but construction already provides the required behavior. In this case,
-you don't need to repeat it in a [lambda](lambdas.md).
-
-To create a reference to a constructor, use `::ClassName`:
-
-```kotlin
-data class User(val name: String, val age: Int)
-
-fun createUser(factory: (String, Int) -> User): User =
-    factory("Jane Doe", 22)
-
-fun main() {
-    val user = createUser(::User)
-    println(user)
-    // User(name=Jane Doe, age=22)
-}
-```
-{kotlin-runnable="true"}
-
-Callable references to constructors are typed as one of the [`KFunction<out R>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-function/) subtypes depending on the parameter count.
-However, the compiler resolves and checks the constructor references at compile time and doesn't require the
-[kotlin-reflect](reflection.md) library.
-
 ### Classes without constructors
 
 Classes that don't declare any constructors (primary or secondary) have an implicit primary constructor
--- a/docs/topics/classes.md
+++ b/docs/topics/classes.md
@@ -444,7 +444,7 @@ fun main() {
 ```
 {kotlin-runnable="true" id="class-delegation-sequence"}
 
-## Constructor references
+### Constructor references
 
 Use a constructor reference when an API needs a factory, but construction already provides the required behavior. In this case,
 you don't need to repeat it in a [lambda](lambdas.md).
--- a/docs/topics/fun-interfaces.md
+++ b/docs/topics/fun-interfaces.md
@@ -66,7 +66,7 @@ You can also use [SAM conversions for Java interfaces](java-interop.md#sam-conve
 
 ### Pass callable references to SAM parameters
 
-If the required behavior already exists as a named function, you can pass a callable reference instead of repeating the call
+If the required behavior already exists as a named function, you can pass a [callable reference](lambdas.md#callable-reference) instead of repeating the call
 in a lambda. Kotlin adapts a compatible callable reference to the expected functional interface.
 
 In the following example, `::start` refers to the existing `start()` function. Because `execute()` expects `Action`,
--- a/docs/topics/fun-interfaces.md
+++ b/docs/topics/fun-interfaces.md
@@ -64,9 +64,43 @@ fun main() {
 
 You can also use [SAM conversions for Java interfaces](java-interop.md#sam-conversions).
 
+### Pass callable references to SAM parameters
+
+If the required behavior already exists as a named function, you can pass a callable reference instead of repeating the call
+in a lambda. Kotlin adapts a compatible callable reference to the expected functional interface.
+
+In the following example, `::start` refers to the existing `start()` function. Because `execute()` expects `Action`,
+the Kotlin compiler adapts the reference to an `Action` instance, which `run()` method calls `start()`:
+
+```kotlin
+fun interface Action {
+    fun run()
+}
+
+fun execute(action: Action) {
+    action.run()
+}
+
+fun start() {
+    println("Run from a callable reference")
+}
+
+fun main() {
+    execute(::start)
+    // Run from a callable reference
+}
+```
+{kotlin-runnable="true"}
+
+The lambda and the callable reference produce compatible `Action` instances, but they begin as different expressions.
+The first is a function literal and uses SAM conversion. The second refers to an existing declaration and uses SAM adaptation.
+
+SAM adaptation happens at compile time and doesn't perform runtime [reflection](reflection.md) or require the `kotlin-reflect` library.
+The compiler checks that the referenced function is compatible with the interface's single abstract method.
+
 ## Migration from an interface with constructor function to a functional interface
 
-Starting from 1.6.20, Kotlin supports [callable references](reflection.md#callable-references) to functional interface constructors, which
+Starting from 1.6.20, Kotlin supports callable references to functional interface constructors, which
 adds a source-compatible way to migrate from an interface with a constructor function to a functional interface.
 Consider the following code:
 
--- a/docs/topics/fun-interfaces.md
+++ b/docs/topics/fun-interfaces.md
@@ -66,11 +66,8 @@ You can also use [SAM conversions for Java interfaces](java-interop.md#sam-conve
 
 ### Pass callable references to SAM parameters
 
-If the required behavior already exists as a named function, you can pass a [callable reference](lambdas.md#callable-reference) instead of repeating the call
-in a lambda. Kotlin adapts a compatible callable reference to the expected functional interface.
-
-In the following example, `::start` refers to the existing `start()` function. Because `execute()` expects `Action`,
-the Kotlin compiler adapts the reference to an `Action` instance, which `run()` method calls `start()`:
+If the required behavior already exists as a named function, you can pass a [callable reference](lambdas.md#callable-reference).
+Kotlin adapts a compatible callable reference to the expected functional interface:
 
 ```kotlin
 fun interface Action {
@@ -86,17 +83,27 @@ fun start() {
 }
 
 fun main() {
+    // Kotlin converts the lambda to an Action instance
+    execute { start() }
+    // Run from a callable reference
+
+    // Kotlin adapts the reference to an Action instance
     execute(::start)
     // Run from a callable reference
 }
 ```
 {kotlin-runnable="true"}
 
-The lambda and the callable reference produce compatible `Action` instances, but they begin as different expressions.
-The first is a function literal and uses SAM conversion. The second refers to an existing declaration and uses SAM adaptation.
+The lambda in `execute { start ()}` and the callable reference in `execute(::start)` produce compatible `Action` instances,
+but they begin as different expressions. The first is a function literal and uses SAM conversion.
+The second refers to an existing declaration and uses SAM adaptation.
 
 SAM adaptation happens at compile time and doesn't perform runtime [reflection](reflection.md) or require the `kotlin-reflect` library.
 The compiler checks that the referenced function is compatible with the interface's single abstract method.
+
+Don't confuse a callable reference adapted to a SAM parameter with a callable reference to a functional interface constructor.
+In `execute(::start)`, `::start` supplies the implementation of `Action.run()`. A reference, such as `::Action`, refers
+to the implicit constructor that creates an `Action` from compatible behavior.
 
 ## Migration from an interface with constructor function to a functional interface
 
--- a/docs/topics/lambdas.md
+++ b/docs/topics/lambdas.md
@@ -114,11 +114,11 @@ There are several ways to obtain an instance of a function type:
   [Function literals with receiver](#function-literals-with-receiver) can be used as values of function types with receiver.
 
 * Use a callable reference to an existing declaration:
-    * a top-level, local, member, or extension [function](reflection.md#function-references): `::isOdd`, `String::toInt`,
-    * a top-level, member, or extension [property](reflection.md#property-references): `List<Int>::size`,
-    * a [constructor](reflection.md#constructor-references): `::Regex`
+    * a top-level, local, member, or extension [function](#function-references): `::isOdd`, `String::toInt`,
+    * a top-level, member, or extension [property](properties.md#property-references): `List<Int>::size`,
+    * a [constructor](classes.md#constructor-references): `::Regex`
 
-  These include [bound callable references](reflection.md#bound-function-and-property-references) that point to a member of a particular instance: `foo::toString`.
+  These include [bound callable references](#bound-and-unbound-references) that point to a member of a particular instance: `foo::toString`.
 
 * Use instances of a custom class that implements a function type as an interface:
 
@@ -193,6 +193,101 @@ fun main() {
 ### Inline functions
 
 Sometimes it is beneficial to use [inline functions](inline-functions.md), which provide flexible control flow, for higher-order functions.
+
+## Callable reference
+
+Callable references let you refer to a function, property, or constructor without calling or accessing it immediately.
+You can store the reference in a variable, pass it to another function, or invoke it later.
+
+To create a callable reference, use the `::` operator:
+
+```kotlin
+fun isOdd(number: Int) = number % 2 != 0
+
+fun main() {
+val numbers = listOf(1, 2, 3, 4, 5)
+
+    // Pass a reference to isOdd()
+    println(numbers.filter(::isOdd))
+    // [1, 3, 5]
+}
+```
+{kotlin-runnable="true"}
+
+Callable-reference objects implement [reflection](reflection.md) interfaces, such as `KFunction` and `KProperty`.
+
+### Function references
+
+Use a function reference when an API expects behavior as a value, and you already have a function that implements it.
+To refer to a top-level or local named function, use `::functionName`:
+
+```kotlin
+fun calculateLength(text: String) = text.length
+
+fun main() {
+val length: (String) -> Int = ::calculateLength
+
+    println(length("Kotlin"))
+    // 6
+    println(::calculateLength.name)
+    // calculateLength
+}
+```
+{kotlin-runnable="true"}
+
+Function references belong to one of the [`KFunction<out R>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-function/) subtypes, depending on the parameter count.
+For instance, `KFunction3<T1, T2, T3, R>`.
+
+If several functions have the same name, the compiler uses the expected function type to select the matching overload:
+
+```kotlin
+fun parse(value: String) = "String: $value"
+fun parse(value: Int) = "Int: $value"
+
+fun main() {
+val parseText: (String) -> String = ::parse
+val parseNumber: (Int) -> String = ::parse
+
+    println(parseText("Kotlin"))
+    // String: Kotlin
+    println(parseNumber(22))
+    // Int: 22
+}
+```
+{kotlin-runnable="true"}
+
+### Bound and unbound references
+
+A reference to a member can be _bound_ and _unbound_:
+
+* Bound reference captures a particular instance and requires only the function's declared arguments. Use it when every
+  invocation should use the same object.
+* Unbound reference isn't attached to an instance, so its function type includes the receiver. Use it when the caller
+  should provide the object.
+
+```kotlin
+class User(val name: String) {
+fun hasName(prefix: String) = name.startsWith(prefix)
+}
+
+fun main() {
+    // The first parameter of an unbound reference is the User receiver
+    val unbound: (User, String) -> Boolean = User::hasName
+    println(unbound(User("Jane"), "Ja"))
+    // true
+
+    val jane = User("Jane")
+
+    // Reference captures jane as its receiver
+    val bound: (String) -> Boolean = jane::hasName
+    println(bound("Doe"))
+    // false
+}
+```
+{kotlin-runnable="true"}
+
+Extension functions follow the same rule. For example, `String::isBlank` is unbound and needs a `String` receiver.
+`"Kotlin"::isBlank` is bound to one string.
 
 ## Lambda expressions and anonymous functions
 
--- a/docs/topics/lambdas.md
+++ b/docs/topics/lambdas.md
@@ -115,8 +115,8 @@ There are several ways to obtain an instance of a function type:
 
 * Use a callable reference to an existing declaration:
     * a top-level, local, member, or extension [function](#function-references): `::isOdd`, `String::toInt`,
-    * a top-level, member, or extension [property](properties.md#property-references): `List<Int>::size`,
-    * a [constructor](classes.md#constructor-references): `::Regex`
+    * a top-level, member, or extension [property](#property-references): `List<Int>::size`,
+    * a [constructor](#constructor-references): `::Regex`
 
   These include [bound callable references](#bound-and-unbound-references) that point to a member of a particular instance: `foo::toString`.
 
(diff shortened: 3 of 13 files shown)


docs/topics/classes.md
```diff
@@ -444,6 +444,31 @@ fun main() {
 ```
 {kotlin-runnable="true" id="class-delegation-sequence"}
 
+### Constructor references
+
+Use a constructor reference when an API needs a factory, but construction already provides the required behavior. In this case,
```
sandwwraith: 'instead of repeating it in a lambda' sounds extremely confusing when you don't have the example with lambda at hand

docs/topics/fun-interfaces.md
```diff
@@ -64,9 +64,43 @@ fun main() {
 
 You can also use [SAM conversions for Java interfaces](java-interop.md#sam-conversions).
 
+### Pass callable references to SAM parameters
+
+If the required behavior already exists as a named function, you can pass a [callable reference](lambdas.md#callable-reference) instead of repeating the call
```
sandwwraith: same here

docs/topics/fun-interfaces.md
```diff
@@ -64,9 +64,43 @@ fun main() {
...
+}
+
+fun main() {
+    execute(::start)
+    // Run from a callable reference
+}
+```
+{kotlin-runnable="true"}
+
+The lambda and the callable reference produce compatible `Action` instances, but they begin as different expressions.
```
sandwwraith: If you want to make such comparison, add an example with `execute { start() }`, otherwise readers will have no idea what is happening

docs/topics/lambdas.md
```diff
@@ -194,6 +194,101 @@ fun main() {
...
+    // Pass a reference to isOdd()
+    println(numbers.filter(::isOdd))
+    // [1, 3, 5]
+}
+```
+{kotlin-runnable="true"}
+
+Callable-reference objects implement [reflection](reflection.md) interfaces, such as `KFunction` and `KProperty`.
+
+### Function references
```
sandwwraith, line 219: I think properties/constructor references should also be placed here. Or into separate `callable-references.md` doc. I don't think it's better to place them together with the rest of the properties/constructors

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
+that consists of the following packages:
 
-* In Gradle:
+* [`kotlin.reflect`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/) with core reflection types and functions.
+* [`kotlin.reflect.full`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/) with extensions for inspecting Kotlin declarations.
+* [`kotlin.reflect.jvm`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.jvm/) with JVM-specific extensions that connect Kotlin and Java reflections.
 
-    <tabs group="build-script">
-    <tab title="Kotlin" group-key="kotlin">
+Reflection can only inspect declarations that are present in the compiled program. It doesn't parse the original source code or
```
sandwwraith: one can use `Class.forName(...).kotlin` tho. reflect can access any class in the runtime classpath

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
-On the JVM platform, the Kotlin compiler distribution includes the runtime component required for using the reflection features as a separate
-artifact, `kotlin-reflect.jar`. This is done to reduce the required size of the runtime
-library for applications that do not use reflection features.
+The Kotlin compiler records metadata about declarations in the compiled output. The [reflection API](https://kotlinlang.org/api/core/kotlin-reflect/) reads this metadata and
+presents it through Kotlin types such as `KClass` or `KFunction`. A reflection object describes this declaration, but it isn't
+the declaration's value. For example, you can use a `KProperty` to get a property's name and return its type without reading
+that property from an object.
 
-To use reflection in a Gradle or Maven project, add the dependency on `kotlin-reflect`:
+Kotlin provides some basic features, such as class literals or callable references, as part of the language and standard
```
sandwwraith, line 18: why did you move down instructions on how to add the library to dependencies? they usually go first on most of our libraries' pages

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
 
-If you don't use Gradle or Maven, make sure you have `kotlin-reflect.jar` in the classpath of your project.
-In other supported cases (IntelliJ IDEA projects that use the command-line compiler),
-it is added by default. In the command-line compiler, you can use the `-no-reflect` compiler option to exclude
-`kotlin-reflect.jar` from the classpath.
+To obtain a runtime class, choose processing based on the concrete object it received:
 
-## Class references
+* If you know the class from the sourse code, use `ClassName::class`.
+* If you need the actual class of a runtime value, use `object::class`.
```
sandwwraith: I'd suggest using `value::class` to avoid clash with `object` keyword and 'singleton' meaning

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
 
-To access properties as first-class objects in Kotlin, use the `::` operator:
+After obtaining a `KClass`, you can inspect its members. This way, your program learns which declarations exist before
+deciding whether to use any of them. For example, you can use:
+
+* The [`declaredMemberProperties`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/declared-member-properties.html) and [`declaredMemberFunctions`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/declared-member-functions.html) properties to return the declarations from the class itself.
+* The [`memberProperties`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/member-properties.html) and [`memberFunctions`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/member-functions.html) properties to return the declarations from the class and all of its superclasses.
+
+For example, the following code lists the properties declared directly in a class. Each returned `KProperty` describes a
+declaration without reading it from a `User` object. Therefore, you can inspect metadata without creating an instance:
```
sandwwraith: let's avoid using 'metadata' on a page with reflect. metadata is an underlying source of truth, but reflection provides much more powerful abstractions and mechanisms. so calling reflection usage 'inspecting metadata' is kinda wrong.

besides, we have `kotlin-metadata-jvm` for metadata inspections. as you can see, its abilities are very limited comparing to kotlin-reflect

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
+    User::class.declaredMemberProperties.forEach { property ->
+        println("${property.name}: ${property.returnType}")
+    }
 }
 ```
 
-The expression `::x` evaluates to a `KProperty0<Int>` type property object. You can read its
-value using `get()` or retrieve the property name using the `name` property. For more information, see
-the [docs on the `KProperty` class](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-property/index.html).
+Reflection can inspect classes to which you already have a reference. However, it doesn't scan the classpath or discover
```
sandwwraith: as I explained above, this is not true. you can get `KClass` object from whatever source you like (Class.forName, ServiceLoader, ClassLoader.loadClass, etc) by converting Java Class to KClass.

kotlin-reflect simply does not provide anything for scanning the classpath because java reflect already has all the tools for that

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
+every implementation of an open class or interface. For that, maintain an explicit registry or use a dedicated classpath-scanning
+library.
```
sandwwraith: without providing solid references, this sentence is simply confusing

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
+can use [`call()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-callable/call.html). Declare an unbound member function with its object instance first, then add its regular arguments:
 
 ```kotlin
-fun getKClass(o: Any): KClass<Any> = o.javaClass.kotlin
-```
+import kotlin.reflect.full.memberFunctions
 
-### Constructor references
+class Formatter {
+fun format(text: String, uppercase: Boolean): String =
```
sandwwraith: formatting: tabs missing

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
 
-Compare the types of the bound and the unbound references.
-The bound callable reference has its receiver "attached" to it, so the type of the receiver is no longer a parameter:
+## Add the JVM dependency
 
-```kotlin
-val isNumber: (CharSequence) -> Boolean = numberRegex::matches
+On the JVM platform, the Kotlin compiler distribution includes the runtime component required for using the reflection
+features as a separate artifact, `kotlin-reflect.jar`. This allows applications without reflection features to reduce the
+required size of the runtime library.
```
sandwwraith:
```suggestion
size of the runtime classpath.
```

docs/topics/reflection.md
```diff
@@ -1,369 +1,308 @@
...
-}
+If you don't use Gradle or Maven, make sure you have `kotlin-reflect.jar` in the classpath of your project.
+In other supported cases (IntelliJ IDEA projects that use the command-line compiler),
+it is added by default. In the command-line compiler, you can use the `-no-reflect` compiler option to exclude
+`kotlin-reflect.jar` from the classpath.
 
-val o = Outer()
-val boundInnerCtor = o::Inner
-```
+To convert the class representations themselves, use `.java` on a `KClass` and `.kotlin` on a Java Class.
```
sandwwraith: This seems very out of place

sandwwraith (changes requested):

docs/topics/sealed-classes.md
```diff
@@ -217,6 +217,41 @@ For more information, see [Guard conditions in when expressions](control-flow.md
...
+sealed interface Status
+data class Send(val message: String) : Status
+data object Start : Status
+data object Stop : Status
+
+fun main() {
+    // Get the direct subclasses known to the sealed hierarchy
+    val subclasses = Status::class.sealedSubclasses
+    println(subclasses.mapNotNull { it.simpleName })
+
```
fzhinkin, line 239: Should we mention what will be printed here, like for singletonStatus?

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+fun main() {
+    val user = createUser(::User)
+    println(user)
+    // User(name=Jane Doe, age=22)
+}
+```
+{kotlin-runnable="true"}
+
+Callable references to constructors are typed as one of the [`KFunction<out R>`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-function/) subtypes depending on the parameter count.
+However, the compiler resolves and checks the constructor references at compile time and doesn't require the
```
fzhinkin: Isn't it true for all reference types?

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+
+If several functions have the same name, the compiler uses the expected function type to select the matching overload:
+
+```kotlin
+fun parse(value: String) = "String: $value"
+fun parse(value: Int) = "Int: $value"
+
+fun main() {
+val parseString: (String) -> String = ::parse
+val parseInt: (Int) -> String = ::parse
```
fzhinkin:
```suggestion    
    val parseString: (String) -> String = ::parse
    val parseInt: (Int) -> String = ::parse
```

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+### Function references
+
+Use a function reference when an API expects behavior as a value, and you already have a function that implements it.
+To refer to a top-level or local named function, use `::functionName`:
+
+```kotlin
+fun calculateLength(text: String) = text.length
+
+fun main() {
+val length: (String) -> Int = ::calculateLength
```
fzhinkin:
```suggestion
    val length: (String) -> Int = ::calculateLength
```

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+Callable references let you refer to a function, property, or constructor without calling or accessing it immediately.
+You can store the reference in a variable, pass it to another function, or invoke it later.
+
+To create a callable reference, use the `::` operator:
+
+```kotlin
+fun isOdd(number: Int) = number % 2 != 0
+
+fun main() {
+val numbers = listOf(1, 2, 3, 4, 5)
```
fzhinkin:
```suggestion
    val numbers = listOf(1, 2, 3, 4, 5)
```

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+    // [1, 3, 5]
+}
+```
+{kotlin-runnable="true"}
+
+Callable-reference objects implement [reflection](reflection.md) interfaces, such as `KFunction` and `KProperty`.
+
+### Function references
+
+Use a function reference when an API expects behavior as a value, and you already have a function that implements it.
```
fzhinkin, line 221: > when an API expects behavior as a value

I don't really like this phrase, but I'm struggling to find any concise replacement.

Perhaps, `with higher order functions, when you have a function with the required signature`?

But neither cover full spectrum of scenarios where function references could be used.

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
-References to functions, properties, and constructors can
-also be called or used as instances of [function types](lambdas.md#function-types).
+If you don't use Gradle or Maven, make sure you have `kotlin-reflect.jar` in the classpath of your project.
+In other supported cases (IntelliJ IDEA projects that use the command-line compiler),
+it is added by default. In the command-line compiler, you can use the `-no-reflect` compiler option to exclude
+`kotlin-reflect.jar` from the classpath.
 
-The common supertype for all callable references is [`KCallable<out R>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-callable/index.html),
-where `R` is the return value type. It is the property type for properties, and the constructed type for constructors.
+To convert the class representations themselves, use `.java` on a `KClass` and `.kotlin` on a Java Class.
```
fzhinkin: This section is all about configuring project's dependencies and these two lines seems to be out of place.

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
+The obtained reference is a [`KClass`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/) type value.
 
-Function references belong to one of the [`KFunction<out R>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-function/index.html)
-subtypes, depending on the parameter count. For instance, `KFunction3<T1, T2, T3, R>`.
+### Inspect types
 
-`::` can be used with overloaded functions when the expected type is known from the context.
-For example:
+Even though [`KClass`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/) represents a class, it doesn't preserve type arguments.
+For example, `List<Int>` has the same `List::class` representaion as `List<String>`.
```
fzhinkin:
```suggestion
For example, `List<Int>` has the same `List::class` representation as `List<String>`.
```

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
 
-```kotlin
-fun isOdd(x: Int) = x % 2 != 0
-```
+* If you know the class from the source code, use `ClassName::class`.
+* If you need the actual class of a runtime value, use `value::class`.
 
-Alternatively, you can use the function as a function type value, that is, pass it
-to another function. To do so, use the `::` operator:
+The approach depends on the way an API accepts values: through a common superclass or through `Any`. The declared type tells
```
fzhinkin, line 75: I'd say it rather depends on whether you know a type statically and want to get some information about it, or when the exact type is not known and may change in runtime (because you have a reference only to a supertype, like `Any`) and you need to get this precise information.

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
-
-fun isOdd(x: Int) = x % 2 != 0
+import kotlin.reflect.full.cast
+import kotlin.reflect.full.safeCast
 
 fun main() {
-//sampleStart
-    fun length(s: String) = s.length
+val expectedClass = String::class
+val value: Any = "Kotlin"
```
fzhinkin:
```suggestion
    val expectedClass = String::class
    val value: Any = "Kotlin"
```

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
-fun main() {
-    println(String::lastChar.get("abc"))
-}
-```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
+data class User(val name: String, val age: Int)
 
-### Interoperability with Java reflection
+fun readProperty(instance: Any, propertyName: String): Any? {
+// Inspect the runtime class
```
fzhinkin: In almost all snippets the first lines in a function have incorrect indentation

fzhinkin:
@udalov, could you also take a look at the reflection.md, just in case?

docs/topics/lambdas.md
aepanchi: fair enough, remooved

docs/topics/lambdas.md
aepanchi: Maybe we should simplify it instead? 

> Use a function reference to pass or store an existing function as a value.

This should cover higher-order functions, callbacks, and storing references in variables or collections without introducing the abstract phrases. WDYT?

docs/topics/reflection.md
aepanchi: Good point. what if we cahnge this part and the bulleted list the following way: 
* Use `ClassName::class` to inspect a class known at compile time.
* Use `value::class` to inspect the actual class of a runtime value, which may differ from its declared type.

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+A reference to a member can be _bound_ and _unbound_:
+
+* Bound reference captures a particular instance and requires only the function's declared arguments. Use it when every
+  invocation should use the same object.
+* Unbound reference isn't attached to an instance, so its function type includes the receiver. Use it when the caller
+  should provide the object.
+
+```kotlin
+class User(val name: String) {
+fun hasName(prefix: String) = name.startsWith(prefix)
```
udalov:
```suggestion
    fun hasName(prefix: String) = name.startsWith(prefix)
```

docs/topics/lambdas.md
```diff
@@ -194,6 +194,176 @@ fun main() {
...
+    // Reference captures jane as its receiver
+    val bound: (String) -> Boolean = jane::hasName
+    println(bound("Doe"))
+    // false
+}
+```
+{kotlin-runnable="true"}
+
+Extension functions follow the same rule. For example, `String::isBlank` is unbound and needs a `String` receiver.
+`"Kotlin"::isBlank` is bound to one string.
```
udalov:
```suggestion
`"Kotlin"::isBlank` is bound and takes no parameters.
```

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
-
-```kotlin
-val isEmptyStringList: List<String>.() -> Boolean = List<String>::isEmpty
-```
+In this example, the `classifier` connects the type to its class or type parameter. The `arguments` collection contains
+its type arguments. Therefore, the code can inspect `String?` separately from `List`.
 
-#### Example: function composition
+> On the JVM, the created type has no annotations, even when you annotate the type in the source code. Support for type
+> annotations might be added in a future version.
```
udalov, line 125: Do we need to mention [KT-29919](https://youtrack.jetbrains.com/issue/KT-29919) here?

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
+>
+{style="note"}
 
-Consider the following function:
+### Check values
```
udalov, line 129: You can also mention `isInstance` (counterpart to `is`) here.

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
 ```
-
-### Property references
 
-To access properties as first-class objects in Kotlin, use the `::` operator:
-
-```kotlin
-val x = 1
+## Inspect a class
```
udalov, line 161: Probably, somewhere we should mention `members` which returns all members.

docs/topics/reflection.md
```diff
@@ -1,369 +1,300 @@
...
-On the JVM platform, the standard library contains extensions for reflection classes that provide a mapping to and from Java
-reflection objects (see package `kotlin.reflect.jvm`).
-For example, to find a backing field or a Java method that serves as a getter for a Kotlin property, you can write something like this:
+    // Find the property declaration
+    val property = runtimeClass.memberProperties
+        .firstOrNull { it.name == propertyName }
+        ?: error("Unknown property: $propertyName")
+
+    // Call the getter with the object as a receiver
+    return property.getter.call(instance)
```
udalov, line 210: Probably, `property.get(instance)` would be more idiomatic here.

udalov (changes requested):

docs/topics/reflection.md
aepanchi: We could, but the ticket itself doesn't offer any additional information or timeline, so I'm not sure

docs/topics/reflection.md
aepanchi: thanks, added!

docs/topics/reflection.md
aepanchi: added

docs/topics/reflection.md
aepanchi: It is more direct, but isn't the exact class is unknown to the compiler here? so the discovered property’s receiver type is also unknown, and `get()` requires that specific receiver type. Do I understand it incorrecty?

udalov: Oh, indeed, you have to cast to a specific property type such as `KProperty1` to invoke `get`.

I guess `property.getter.call(instance)` is fine then. However, consider that you can also call the property: `property.call(instance)`, and its equivalent to calling its getter. Personally I'd use `property.call`, but I'm not sure which one is clearer / more idiomatic.

docs/topics/reflection.md
```diff
@@ -160,6 +163,7 @@ fun main() {
 After obtaining a `KClass`, you can inspect its members. This way, your program learns which declarations exist before
 deciding whether to use any of them. For example, you can use:
 
+* The [`members`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/members.html) property to return all functions and properties accessible in this class, including inherited declarations.
 * The [`declaredMemberProperties`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/declared-member-properties.html) and [`declaredMemberFunctions`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/declared-member-functions.html) properties to return the declarations from the class itself.
 * The [`memberProperties`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/member-properties.html) and [`memberFunctions`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/member-functions.html) properties to return the declarations from the class and all of its superclasses.
```
udalov, line 168: To me it seems a bit strange that we're speaking about a way to get all members (which we need to, I think), and then just SOME of the properties that filter those members. There are [many more](https://github.com/JetBrains/kotlin/blob/ce0d62ea817843ea238af63b69e5b2a0d7c12b70/core/reflection.jvm/src/kotlin/reflect/full/KClasses.kt#L84):
```
declaredMembers
functions
staticFunctions
memberFunctions
memberExtensionFunctions
declaredFunctions
declaredMemberFunctions
declaredMemberExtensionFunctions
staticProperties
memberProperties
memberExtensionProperties
declaredMemberProperties
declaredMemberExtensionProperties
```

I don't suggest to list all of them in the same way as `declaredMemberProperties`/`memberProperties`, but maybe list them all _somehow_? :) And still, selecting just `declaredMemberProperties` and `memberProperties` seems a bit random...

udalov (changes requested):

udalov (approved):

docs/topics/lambdas.md
fzhinkin: Sounds good to me!

docs/topics/reflection.md
fzhinkin: > may differ from

I'd say it "may be more specific than".

Otherwise, looks good.

docs/topics/reflection.md
```diff
@@ -1,369 +1,312 @@
...
 
-Function references belong to one of the [`KFunction<out R>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-function/index.html)
-subtypes, depending on the parameter count. For instance, `KFunction3<T1, T2, T3, R>`.
+### Inspect types
 
-`::` can be used with overloaded functions when the expected type is known from the context.
-For example:
+Even though [`KClass`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/) represents a class, it doesn't preserve type arguments.
+For example, `List<Int>` has the same `List::class` representation as `List<String>`.
+To return a [`KType`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-type/) that includes type arguments and nullability, use the [`typeOf()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/type-of.html) function:
```
InversionSpaces, line 102: nit: I would highlight the trade-off between using `::class` and `typeOf`. The latter will reflect only statically known type, in contrast with the former which will represent the actual runtime type of the value (if you use `value::class`).
One might be tempted to write
```kotlin
inline fun <reified T> typeOfValue(value: T): KType = typeOf<T>()
```
and expect it to reflect the actual type of runtime value.

fzhinkin (approved):

docs/topics/reflection.md
```diff
@@ -1,369 +1,312 @@
...
+| Property                                                                                                                                            | Returns                                                                                                                                      |
+|-----------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------|
+| [`declaredMembers`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/declared-members.html)                                       | Functions and properties declared directly in the class, excluding inherited declarations                                                    |
+| [`functions`](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/functions.html)                                                    | All functions available from the class: non-static functions from the class and its superclasses, and static functions declared in the class |
```
InversionSpaces, line 178: nit: I understand that those lines are from the official docs, but Kotlin itself does not have a notion of a static method AFAIU, it is more of JVM/JS interopability thing. I am not strongly against this, but using this term without comments might create a little bit of confusion.

docs/topics/reflection.md
```diff
@@ -1,369 +1,312 @@
...
+    )
+
+    println(result)
+    // Hello, Kotlin!
 }
 ```
 
-You don't need to specify `this` as the receiver: `this::foo` and `::foo` are equivalent.
+In this example, `instanceParameter` identifies the member-function receiver, and `valueParameters` contains only parameters declared in the function signature.
```
InversionSpaces, line 306: nit: Worth also explaining in text that the previous example showcases that parameters with default values are handled by `call`?

docs/topics/sealed-classes.md
```diff
@@ -217,6 +217,42 @@ For more information, see [Guard conditions in when expressions](control-flow.md
 >
 {style="note"}
 
+## Inspect sealed subclasses with reflection
+
+You can use [Reflection](reflection.md) functionality to inspect sealed subclasses. 
+
+If you need to inspect a known direct subclass, use [sealedSubclasses](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/sealed-subclasses.html).
+The complier returns a [`KClass`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-class/) that can represent a regular class or an object declaration. For an object declaration,
```
InversionSpaces, line 225: I would say "the method returns", not the compiler

docs/topics/lambdas.md
```diff
@@ -194,6 +194,174 @@ fun main() {
...
+    // An unbound member property
+    // a User receiver is needed
+    val name = User::name
+    val user = User("Jane")
+
+    println(name.get(user))
+    // Jane
+    name.set(user, "Jane Doe")
+    println(user.name)
+    // Jane Doe
```
InversionSpaces, line 291: Bounded property reference is omitted deliberately?
```kotlin
val p = user::name
println(p.get())
```

docs/topics/lambdas.md
```diff
@@ -194,6 +194,174 @@ fun main() {
...
+
+```kotlin
+fun calculateLength(text: String) = text.length
+
+fun main() {
+    val length: (String) -> Int = ::calculateLength
+
+    println(length("Kotlin"))
+    // 6
+    println(::calculateLength.name)
```
InversionSpaces, line 232: nit: Worth explaining to the user why `.name` is available on `::calculateLength`, but it is not available on `length: (String) -> Int`?
it is also available on `val length = ::calculateLength` (without explicit type), even though IDE shows `(String) -> Int` as inferred type. 
This is confusing at least to me

docs/topics/fun-interfaces.md
```diff
@@ -64,9 +64,50 @@ fun main() {
...
+but they begin as different expressions. The first is a function literal and uses SAM conversion.
+The second refers to an existing declaration and uses SAM adaptation.
+
+SAM adaptation happens at compile time and doesn't perform runtime [reflection](reflection.md) or require the `kotlin-reflect` library.
+The compiler checks that the referenced function is compatible with the interface's single abstract method.
+
+Don't confuse a callable reference adapted to a SAM parameter with a callable reference to a functional interface constructor.
+In `execute(::start)`, `::start` supplies the implementation of `Action.run()`. A reference, such as `::Action`, refers
+to the implicit constructor that creates an `Action` from compatible behavior.
+
```
InversionSpaces, line 107: Sorry, maybe I am missing something, but I don't understand, why would I confuse `::start` and `::Action`?

InversionSpaces (changes requested):

docs/topics/reflection.md
aepanchi: Do you mean `callBy()`?

docs/topics/lambdas.md
aepanchi: good point, thanks!

docs/topics/lambdas.md
aepanchi: Good point, I agree that it should be documented, but I'm not sure if that's the right place tbh.

PR #5757, opened 2026-08-19
KTL-1215: Update verify samples GitHub workflow

sarahhaggarty:
This PR updates the verify samples GitHub workflow by adding a script that will check if it's necessary to run the full workflow.


.github/workflows/verify-samples.yml
nikpachoo, line 62: @zoobestik, is it expected that we still use the fork?

zoobestik (approved):

PR #5760, opened 2026-08-19
feat: release compiler caching

danil-pavlov:


docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,22 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
 >
 {style="tip"}
 
+### Enable caches for release binaries
+<primary-label ref="experimental-opt-in"/>
+
+By default, Kotlin/Native compiles release binaries in the link-time optimization (LTO) mode: all modules are compiled and
+optimized together. This makes release binaries faster but significantly increases compilation time.
+
+If you prefer a different trade-off, faster compilation time at the expense of some binary optimizations, you can enable
```
haitaka: Can we maybe add an explicit warning that the feature is in early stages of development, at this stage runtime performance might be unsatisfactory, there will be improvements in the upcoming releases

haitaka: binary optimizations -> compiler optimizations? WDYT?

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,22 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+caching in the release mode. To do so, add both following options to your `gradle.properties` file:
+
+```properties
+# Enables the compiler to use caches in the `opt` mode
```
haitaka: Maybe "the `opt` mode" -> "the `-opt` mode" or "the release mode"?

docs/topics/native/native-improving-compilation-time.md
danil-pavlov: Sure, let's add a warning. Plus, we already have an experimental label next to the header

haitaka (approved):

docs/topics/native/native-binary-options.md
```diff
@@ -70,6 +70,17 @@ kotlinc-native main.kt -Xbinary=enableSafepointSignposts=true
...
     </tr>
+    <tr>
+        <td><a href="native-improving-compilation-time.md#enable-caches-for-release-binaries"><code>enableReleaseBinaryCache</code></a></td>
+        <td>
+            <list>
+                <li><code>true</code></li>
+                <li><code>false</code> (default)</li>
+            </list>
+        </td>
+        <td>Enables the compiler to use caches in the release mode, making builds faster at the expense of some compiler optimizations.</td>
```
sarahhaggarty:
```suggestion
        <td>Enables the compiler to use caches in release mode, making builds faster at the expense of some compiler optimizations.</td>
```

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
 >
 {style="tip"}
 
+### Enable caches for release binaries
+<primary-label ref="experimental-opt-in"/>
+
+By default, Kotlin/Native compiles release binaries in the link-time optimization (LTO) mode: all modules are compiled and
+optimized together. This makes release binaries faster but significantly increases compilation time.
```
sarahhaggarty: Faster in what way? 🤔

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+
+If you prefer a different trade-off, faster compilation time at the expense of some compiler optimizations, you can enable
```
sarahhaggarty: "If you prefer a different trade-off" sounds unusual to me. What do you think about removing it to something like: "To prioritize faster compilation over some compiler optimizations, you can enable caching in release mode."

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+caching in the release mode. To do so, add both of the following options to your `gradle.properties` file:
```
sarahhaggarty:
```suggestion
caching in the release mode. To enable caching, add the following options to your `gradle.properties` file:
```
To avoid "to do so" and make it more concise.

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+
+```properties
+# Enables the compiler to use caches in the release mode
```
sarahhaggarty:
```suggestion
# Enables the compiler to use caches in release mode
```

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+kotlin.native.binary.enableReleaseBinaryCache=true
+# Makes the Kotlin Gradle plugin invoke the compiler with `-Xauto-cache-from` and related options
+kotlin.internal.native.enableReleaseBinaryCache=true
+```
+
+> The feature is in active development, so runtime performance is still suboptimal.
+> Performance improvements are planned for the upcoming Kotlin releases.
```
sarahhaggarty:
```suggestion
> Performance improvements are planned in upcoming Kotlin releases.
```

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -97,6 +97,27 @@ Similarly, instead of running `assembleXCFramework`, you can run `assembleShared
...
+caching in the release mode. To do so, add both of the following options to your `gradle.properties` file:
+
+```properties
+# Enables the compiler to use caches in the release mode
+kotlin.native.binary.enableReleaseBinaryCache=true
+# Makes the Kotlin Gradle plugin invoke the compiler with `-Xauto-cache-from` and related options
+kotlin.internal.native.enableReleaseBinaryCache=true
+```
+
+> The feature is in active development, so runtime performance is still suboptimal.
```
sarahhaggarty:
```suggestion
> The feature is in active development, so runtime performance still has room for improvement.
```
"Suboptimal" sounds quite severe. What do you think about this instead?

docs/topics/native/native-improving-compilation-time.md
```diff
@@ -146,11 +162,10 @@ To use the Gradle configuration cache, add the `org.gradle.configuration-cache=t
 
 ### Enable previously disabled features
 
-There are Kotlin/Native options that disable the Gradle daemon and compiler caches:
+There are Kotlin/Native options that disable some of the useful features:
```
sarahhaggarty: I'm a bit confused with this change. It seems strange to disable useful features. Why do you want to change it? 🤔

sarahhaggarty (changes requested):
A few suggestions and clarification questions for you!

docs/topics/native/native-improving-compilation-time.md
danil-pavlov: The list will get longer, so we need more general phrasing. And the reason users might have added these options is below, they could have issues with these features, disabled them, and then forgot to enable again (by removing these options)

sarahhaggarty: I see. How about we expand the intro a bit more then? Maybe something like this:

```
### Enable previously disabled features

You might have disabled some Kotlin/Native features in the past to work around build issues. For example:

* `kotlin.native.disableCompilerDaemon=true` disables the [Gradle daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html).
* `disableNativeCache` disables the [compilation cache](https://kotlinlang.org/docs/multiplatform/multiplatform-dsl-reference.html#binaries).

If your `gradle.properties` file or Gradle build files contain options that disable these features, remove them and check whether the build completes successfully. The issues that originally required these workarounds might have already been fixed.
```

danil-pavlov: I like it, thank you! just rephrased the bottom part a little bit

sarahhaggarty (approved):

PR #5784, opened 2026-08-31
Add guidance for returning multiple values

sarahhaggarty:
This PR updates the Functions page to add guidance on how to emulate returning multiple values.

### DIFF
--- a/docs/topics/functions.md
+++ b/docs/topics/functions.md
@@ -291,14 +291,98 @@ mergeStrings(strings = arrayOf("a", "b", "c"))
 
 ### Return types
 
-When you declare a function with a block body (by putting instructions within curly braces `{}`),
-you must always specify a return type explicitly.
-The only exception is when they return `Unit`,
-[in which case specifying the return type is optional](#unit-returning-functions).
+When you declare a function with a block body (by putting instructions within curly braces `{}`), you must always specify
+a return type explicitly. The only exception is when the function returns `Unit`,[in which case specifying the return type is optional](#unit-returning-functions).
 
-Kotlin doesn't infer return types for functions with block bodies.
-Their control flow can be complex, which makes the return type unclear to the reader and sometimes even to the compiler.
-However, Kotlin can infer the return type for [single-expression functions](#single-expression-functions) if you don't specify it.
+Kotlin doesn't infer return types for functions with block bodies. Their control flow can be complex, which makes the
+return type unclear to the reader and sometimes even to the compiler. However, Kotlin can infer the return type for
+[single-expression functions](#single-expression-functions) if you don't specify it.
+
+Kotlin functions return a single value, but that value can contain multiple pieces of data. For ways to represent these
+values, see [Return multiple values](#return-multiple-values).
+
+#### Return multiple values
+
+When you need to return multiple related values with distinct meanings, declare a [data class](data-classes.md), even if you only use
+it with one function:
+
+```kotlin
+data class OrderSummary(
+    val subtotal: Double,
+    val tax: Double,
+)
+
+fun calculateOrderSummary(prices: List<Double>): OrderSummary {
+    val subtotal = prices.sum()
+    val tax = subtotal * 0.2
+    return OrderSummary(subtotal, tax)
+}
+
+fun main() {
+    val summary = calculateOrderSummary(listOf(12.50, 8.00, 4.50))
+
+    println(summary.subtotal)
+    // 25.0
+    println(summary.tax)
+    // 5.0
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="return-multiple-values-data-class"}
+
+A data class works well when you need to return multiple values with distinct meanings. If the returned values
+are of the same kind and you want to handle them as a group, consider returning a collection instead:
+
+```kotlin
+data class Person(val name: String)
+
+val friendGroups = listOf(
+    listOf(Person("Alice"), Person("Bob")),
+    listOf(Person("Charlie"), Person("Diana"), Person("Eve")),
+    listOf(Person("Frank"))
+)
+
+fun findLargestGroupOfFriends(): List<Person> {
+    return friendGroups.maxByOrNull { it.size } ?: emptyList()
+}
+
+fun main() {
+    val largestGroup = findLargestGroupOfFriends()
+
+    println(largestGroup.map { it.name })
+    // [Charlie, Diana, Eve]
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.4" id="return-multiple-values-list"}
+
+If you need to return a fixed number of values, you can also use [`Pair`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-pair/) or [`Triple`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-triple/)
+data classes from the standard library. However, their properties have generic names such as `first`, `second`, and `third`,
+which can make the result difficult to understand.
+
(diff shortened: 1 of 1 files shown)


docs/topics/functions.md
```diff
@@ -294,6 +294,85 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+fun main() {
+    val largestGroup = findLargestGroupOfFriends()
+
+    println(largestGroup.map { it.name })
+    // [Charlie, Diana, Eve]
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.4" id="return-multiple-values-list"}
+
+For a small group of values whose meaning is clear from the context, you can also use [`Pair`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-pair/) or [`Triple`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-triple/).
```
e2e4b6b7: > clear from the context

It is very dangerous thing to say, as developer writing a function might consider a lot more things as obvious compared to themself 2 days later. 

It might be better to say something like ...

docs/topics/functions.md
```diff
@@ -294,6 +294,85 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+These types can contain values of different types. However, their properties have generic names such as `first`, `second`, and `third`,
+which don't describe what each value represents. Use a data class when descriptive property names would make the
+result easier to understand. For example:
```
e2e4b6b7: > Use a data class when descriptive property names would make the result easier to understand. For example:

inverted meaning to the example

docs/topics/functions.md
```diff
@@ -294,6 +294,85 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+    return OrderSummary(subtotal, tax)
+}
+
+fun main() {
+    val summary = calculateOrderSummary(listOf(12.50, 8.00, 4.50))
+
+    println(summary.subtotal)
+    // 25.0
+    println(summary.tax)
+    // 5.0
```
e2e4b6b7, line 327: Probably, a demonstration with a name-based destructuring might be useful

docs/topics/functions.md
e2e4b6b7: > meaning is clear from the context

It is very dangerous thing to say, as developer writing a function might consider a lot more things as obvious compared to themself 2 days later. 

> Use a data class when descriptive property names would make the result easier to understand. For example:

Description says to use data class, but demonstrates usage of `Pair`. It's a bit misleading.

It might be better to directly say that while it could be used for the shown example, it considered as anti-pattern. And after that say in which cases these structures might be useful

docs/topics/functions.md
sarahhaggarty: It would be nice but name-based destructuring is experimental. When it's stable, I'll feel more comfortable adding it here.

docs/topics/functions.md
sarahhaggarty: > It is very dangerous thing to say

That's fair. I removed this text.

> Description says to use data class, but demonstrates usage of Pair. It's a bit misleading.

That's a good point. I reworded it like you suggested. Does it look OK now?

docs/topics/functions.md
```diff
@@ -294,6 +294,88 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+  
+    // What does 'second' mean?
+    println(totals.second)
+    // 5.0
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="return-multiple-values-pair"}
+
+For results with distinct meanings, prefer a data class with descriptive property names, as in the `OrderSummary` data
+class example at the beginning.
```
e2e4b6b7: nit: Now it reads as if Pair and Triple are never useful. It might be better to add a single sentence about when it could be used. It might fit after or together with the paragraph about "collections". F.e.:

** Collection example

And if the collection is of known size, `Pair` or `Triple` might be used instead. While you can also use them to return multiple values with distinct meaning, their properties have generic names such as `first`, `second`, and `third`, which can make the result difficult to understand.

...

** Anti-pattern example

e2e4b6b7 (approved):

docs/topics/functions.md
sarahhaggarty: I see your point. I adjusted the text to be less negative about Pair and Triple:

> If you need to return a fixed number of values, you can also use a [`Pair`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-pair/) or [`Triple`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-triple/).
However, their properties have generic names such as `first`, `second`, and `third`, which can make the result difficult to understand.

docs/topics/functions.md
```diff
@@ -294,6 +294,88 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+  
+    // What does 'second' mean?
+    println(totals.second)
+    // 5.0
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3" id="return-multiple-values-pair"}
+
+For results with distinct meanings, prefer a data class with descriptive property names, as demonstrated in the `OrderSummary` data
+class example at the beginning of this section.
```
danil-pavlov: >at the beginning of this section.
Maybe put a link instead, to jump up to the beginning of the section?

docs/topics/functions.md
```diff
@@ -294,6 +294,88 @@ Kotlin doesn't infer return types for functions with block bodies.
...
+fun main() {
+    val largestGroup = findLargestGroupOfFriends()
+
+    println(largestGroup.map { it.name })
+    // [Charlie, Diana, Eve]
+}
+```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.4" id="return-multiple-values-list"}
+
+If you need to return a fixed number of values, you can also use a [`Pair`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-pair/) or [`Triple`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-triple/).
```
danil-pavlov:
```suggestion
If you need to return a fixed number of values, you can also use [`Pair`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-pair/) or [`Triple`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-triple/) data classes from the standard library.
```

docs/topics/functions.md
```diff
@@ -294,6 +294,88 @@ Kotlin doesn't infer return types for functions with block bodies.
 Their control flow can be complex, which makes the return type unclear to the reader and sometimes even to the compiler.
 However, Kotlin can infer the return type for [single-expression functions](#single-expression-functions) if you don't specify it.
 
+#### Return multiple values
```
danil-pavlov, line 304: I'd also add a mention of the possibility of returning multiple values to the intro of this section, because single-expression functions are mentioned, and multiple values are not

docs/topics/functions.md
danil-pavlov: Another idea is to change the structure of the whole section, to start with collections, then mention data classes from stdlib, and then move to custom data classes

danil-pavlov (changes requested):

docs/topics/functions.md
sarahhaggarty: Added!

docs/topics/functions.md
sarahhaggarty: Added a link instead. The recommendation from the language team is to use custom data classes primarily which is why the content is ordered as it is. Collections and data classes from stdlib are possible but they are comparatively less easy to understand by other developers because they provide less context. So I'd prefer to keep the current structure :)

danil-pavlov (approved):

PR #5810, opened 2026-09-10
chore: more consistent and readable code examples in calling the Derived class

zamulla:
We're already using the `val derived` in some samples. I think this is a more readable version so I made it consistent.


aepanchi (approved):
LGTM!

PR #5812, opened 2026-09-10
Update: document es2020 support

aepanchi:


docs/topics/js/js-project-setup.md
```diff
@@ -87,28 +87,30 @@ This includes downloading and installing the required environment and dependenci
...
-* Generators for compiling [suspend functions](https://kotlinlang.org/docs/composing-suspending-functions.html) that improve the final bundle size
-  and help with debugging.
-* [JavaScript code inlining](js-interop.md#inline-javascript).
+* `es5` provides compatibility with older JavaScript environments and is the default target.
+* `es2015` enables support for ES2015 features, including:
+    * Modules that simplify your codebase and improve maintainability.
+    * Classes that allow incorporating OOP principles, resulting in cleaner and more intuitive code.
+    * Generators for compiling [suspend functions](https://kotlinlang.org/docs/composing-suspending-functions.html) that improve the final bundle size and help with debugging.
+    * [JavaScript code inlining](js-interop.md#inline-javascript).
+* `es2020` includes all supported ES2015 features and enables newer JavaScript features like `BigInt`.
```
JSMonk: Could we re-word it a bit. Because it could be understanded as we allow `BigInt` inside of the `js` function, but basically what we do with es2020 is the compilation of `Long` type into `BigInt`

JSMonk (changes requested):

JSMonk (approved):

JSMonk (approved):

docs/topics/js/js-project-setup.md
```diff
@@ -87,28 +87,31 @@ This includes downloading and installing the required environment and dependenci
...
+    * Modules that simplify your codebase and improve maintainability.
+    * Classes that allow incorporating OOP principles, resulting in cleaner and more intuitive code.
+    * Generators for compiling [suspend functions](https://kotlinlang.org/docs/composing-suspending-functions.html) that improve the final bundle size and help with debugging.
+    * [JavaScript code inlining](js-interop.md#inline-javascript).
+* `es2020` includes all supported ES2015 features and allows the compiler to use ECMAScript 2020 features in the generated
+  JavaScript code. For example, it compiles Kotlin `Long` values to JavaScript `BigInt` values.
 
-You can enable all the supported ES2015 features at once by adding the `es2015` compilation target to your
-`build.gradle(.kts)` file:
+To set the ECMAScript target, configure the `target` property in the `build.gradle(.kts)` file. For example, to target ES2020:
```
zamulla: If there can be multiple ECMAScript targets per project and there is no other context, I'd use "an". Also "set" and "configure" sound right to me swapped.

```suggestion
To configure an ECMAScript target, set the `target` property in the `build.gradle(.kts)` file. For example, to target ES2020:
```

docs/topics/js/js-to-kotlin-interop.md
```diff
@@ -382,31 +382,31 @@ This feature is [Experimental](components-stability.md#stability-levels-explaine
...
+| `Unit`                                                           | Undefined                 | Exportable when used as return type, but not when used as parameter type.                                      |
+| `Any`                                                            | `Object`                  |                                                                                                                |
+| `Throwable`                                                      | `Error`                   |                                                                                                                |
+| `enum class Type`                                                | `Type`                    | Enum entries are exposed as static class properties (`Type.ENTRY`).                                            |
+| Nullable `Type?`                                                 | `Type | null | undefined` |                                                                                                                |
```
zamulla: Above there's "Undefined" that's not monospaced, I'd choose one variant; and I would write out "`Type` or `null` or `undefined`".

aepanchi: good point, thanks!

zamulla (approved):
👍

PR #5813, opened 2026-09-11
Add description of the Unit type

aepanchi:


docs/topics/functions.md
```diff
@@ -418,14 +418,17 @@ fun getDisplayNameOrDefault(userId: String?): String =
...
 
-If a function has a block body (instructions within curly braces `{}`) and does not return a useful value,
-the compiler assumes its return type  is `Unit`.
-`Unit` is a type that has only one value, also called `Unit`.
+If a function has a block body, doesn't produce a meaningful value, and has no explicit return type, the compiler infers
+its return type as `Unit`.
 
-You don't have to specify `Unit` as a return type, except for functional type parameters.
-You never have to return `Unit` explicitly.
+`Unit` is a type that has only one value, also called `Unit`. Unlike `void` in other programming languages, `Unit` is
```
ilya-g, line 424: It would make sense to give a reference to object types to expand on this concept of type with a single value: https://kotlinlang.org/docs/object-declarations.html

ilya-g (approved):

docs/topics/functions.md
```diff
@@ -418,14 +418,17 @@ fun getDisplayNameOrDefault(userId: String?): String =
 
 ### Unit-returning functions
 
-If a function has a block body (instructions within curly braces `{}`) and does not return a useful value,
-the compiler assumes its return type  is `Unit`.
-`Unit` is a type that has only one value, also called `Unit`.
+If a function has a block body, doesn't produce a meaningful value, and has no explicit return type, the compiler infers
```
zamulla, line 421: Not sure what a "meaningful value" is. Can we just drop this phrase? The return type is what the compiler cares about, no?

docs/topics/functions.md
```diff
@@ -418,14 +418,17 @@ fun getDisplayNameOrDefault(userId: String?): String =
...
+its return type as `Unit`.
 
-You don't have to specify `Unit` as a return type, except for functional type parameters.
-You never have to return `Unit` explicitly.
+`Unit` is a type that has only one value, also called `Unit`. This value is an [object](object-declarations.md), so only a single `Unit` instance exists.
+Unlike `void` in other programming languages, `Unit` is an actual type, and you can use its value if your code requires one.
+This gives functions that produce results and functions that only perform actions the same type model: both return a value
```
zamulla, line 426:
```suggestion
This way functions that produce results and functions that only perform actions fit the same type model: both return a value
```

"Giving something a model" doesn't sound right to me.

docs/topics/functions.md
zamulla, line 472: Here I would highlight the line with an explicit comment, too.

zamulla, line 465: I'd write a comment here, something like "() -> Unit isn't optional, but : Unit is.", to avoid playing find 10 differences.

docs/topics/functions.md
```diff
@@ -471,14 +475,14 @@ fun printHello(name: String?, action: () -> Unit): Unit {
 fun main() {
     printHello("Kodee") {
         println("This runs after the greeting.")
```
zamulla, line 477: Since there is no greeting apart from the one we're printing, I'm not sure how to interpret this. Should it be something like "This runs when a name is given"?

docs/topics/functions.md
```diff
@@ -471,14 +475,14 @@ fun printHello(name: String?, action: () -> Unit): Unit {
...
+        // Hello Kodee
+        // This runs after the greeting.
     }
-    // Hello Kodee
-    // This runs after the greeting.
-
     printHello(null) {
         println("No name provided, but action still runs.")
```
zamulla, line 482: And likewise, "This runs when no name is provided."

zamulla (commented):
Not a lot to comment on, but I found something to nitpick.

zamulla (approved):
🙌

PR #5817, opened 2026-09-15
update: ksp-additional-details.md

budindepunk:
[KEDOC-311](https://youtrack.jetbrains.com/issue/KEDOC-311/Rewrite-type-resolution-section)
[KEDOC-312](https://youtrack.jetbrains.com/issue/KEDOC-312/Rewrite-how-ksp-looks-at-source-files-section)


jaschdoc (approved):
Thanks!

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+      containingFile: String
+      typeParameters: KSTypeParameter
+      parentDeclaration: KSDeclaration
+      functionKind: FunctionKind
+      extensionReceiver: KSTypeReference?
+      returnType: KSTypeReference
+      parameters: List<KSValueParameter>
+      // contains local classes, local functions, local variables, etc.
+      declarations: List<KSDeclaration>
+    KSPropertyDeclaration // global variable
```
zamulla: Let's keep the comments according to the style guide – capitalized, on a new line, etc.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+
+![class diagram](ksp-class-diagram.svg){thumbnail="true" width="800" thumbnail-same-file="true"}
+
+> [See the full-sized diagram](https://kotlinlang.org/docs/images/ksp-class-diagram.svg).
+>
+{style="note"}
+
+You can find the complete API definition in the [KSP GitHub repository](https://github.com/google/ksp/tree/main/api/src/main/kotlin/com/google/devtools/ksp).
+
+## Type references and resolution
```
zamulla, line 70: Could we draw a clearer connection between this section and the overall theme of the KSP Kotlin model? From the first part I see that Kotlin model = hierarchy, and that's understood; this section is about specifics of type resolution within the model and we could make it clearer in the intro that there are specifics to be interested in.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+
+Type resolution is one of the most expensive operations in the KSP API. To avoid unnecessary work, processors resolve 
+most type references explicitly. Properties that refer to types, such as `KSFunctionDeclaration.returnType` and
```
zamulla, line 73: Maybe I did not get the point here: to avoid resolution work, processors resolve references; so which work is avoided exactly?

Having read the page, I understand that the point is that processors don't resolve anything unless I tell them to, and that's the decision I should learn how to make. That point should be highlighted, and also maybe that is exactly the connection between the intro and the resolution section: `KSType` is not in the basic model because it's likely you don't even need it. But how to decide? Read the section on resolution.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+}
+
+interface KSTypeReference : KSAnnotated, KSModifierListOwner {
+    val element: KSReferenceElement?
+    fun resolve(): KSType
+}
+```
+
+A `KSTypeReference` represents an unresolved type. It preserves the syntactic representation of the type as it appears 
+in the source code. Its `KSReferenceElement` models the corresponding type element in Kotlin's grammar, including its
```
zamulla, line 89: Maybe breaking into two sentences is not beneficial here: "...represents an unresolved type and preserves..." or "by preserving..."?

docs/kr.tree
```diff
@@ -366,7 +366,7 @@
             <toc-element topic="ksp-incremental.md"/>
             <toc-element topic="ksp-multi-round.md"/>
             <toc-element toc-title="Reference for Java annotation processor authors" topic="ksp-reference.md"/>
-            <toc-element topic="ksp-additional-details.md"/>
+            <toc-element topic="ksp-kotlin-model.md"/>
```
zamulla, line 369: Reminder to be safe: we also need a redirect here.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+in the source code. Its `KSReferenceElement` models the corresponding type element in Kotlin's grammar, including its 
+annotations and modifiers.
+
+A `KSReferenceElement` can be one of the following:
+
+* `KSClassifierReference`, which provides information such as `referencedName`.
+
+* `KSCallableReference`, which provides information such as `receiverType`, `functionArguments`, and `returnType`.
+
+You can inspect this information without resolving the reference.
```
zamulla: I feel like this should be before the list, as the purpose for `KSReferenceElement`.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+adds the generated source files to the compilation, and the Kotlin compiler later resolves and type-checks the type 
+references together with the rest of the source code.
+
+To access the type in Kotlin's type system, call `KSTypeReference.resolve()`. The returned `KSType` provides access to 
+the declaration that defines the type:
+
+```kotlin
+val ksType: KSType = ksTypeReference.resolve()
+val ksDeclaration: KSDeclaration = ksType.declaration
+```
```
zamulla, line 109: Should we declare `ksTypeReference` before using it? Usually it's not a big deal, but I'm not sure I can guess what a runnable sample would look like.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+
+Resolve a type reference only when you need information available from `KSType` or `KSDeclaration`. When possible, 
+inspect the `KSReferenceElement` first. For example, you can use `KSClassifierReference.referencedName` to filter 
+irrelevant references before resolving them.
```
zamulla, line 113: Yep, that's exactly the thing I'd like to see in the beginning. The model is great, but building it in full is expensive, so be judicious.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
+[//]: # (title: KSP Kotlin model)
+
+KSP represents source code as a hierarchy of symbols. Processors navigate this hierarchy to inspect declarations, types,
+annotations, and other elements of the source code.
+
+At the top level, a source file is represented by `KSFile`. A `KSFile` contains declarations such as classes, functions,
+and properties, which can contain additional declarations. The following simplified hierarchy shows some of the most
+common symbols and properties available through the KSP API:
+
+```none
```
zamulla, line 11: This whole thing is already present in ksp-overview.md. Does it need to be in both places? If so, an include block could be used, if not, then it should have been removed from the overview I guess, with a link here?

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+KSFile
+  packageName: KSName
+  fileName: String
+  annotations: List<KSAnnotation>  (File annotations)
+  declarations: List<KSDeclaration>
+    KSClassDeclaration // class, interface, object
+      simpleName: KSName
+      qualifiedName: KSName
+      containingFile: String
```
zamulla, line 21: I'm not sure what is the source of truth here, but [in the main branch right now](https://github.com/google/ksp/blob/main/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSDeclaration.kt) it says that `containingFile` is a `KSFile` rather than a `String`. Am I looking in the wrong direction, or is it worth rechecking? `annotations` is a `Sequence` rather than a `List` according to the same source.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+
+At the top level, a source file is represented by `KSFile`. A `KSFile` contains declarations such as classes, functions,
+and properties, which can contain additional declarations. The following simplified hierarchy shows some of the most
+common symbols and properties available through the KSP API:
+
+```none
+KSFile
+  packageName: KSName
+  fileName: String
+  annotations: List<KSAnnotation>  (File annotations)
```
zamulla: Should "(File annotations)" be a comment?

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+
+A `KSTypeReference` represents an unresolved type. It preserves the syntactic representation of the type as it appears 
+in the source code. Its `KSReferenceElement` models the corresponding type element in Kotlin's grammar, including its 
+annotations and modifiers.
+
+A `KSReferenceElement` can be one of the following:
+
+* `KSClassifierReference`, which provides information such as `referencedName`.
+
+* `KSCallableReference`, which provides information such as `receiverType`, `functionArguments`, and `returnType`.
```
zamulla:
```suggestion
* `KSCallableReference`, which provides information such as `receiverType`, `functionParameters`, and `returnType`.
```

Couldn't find `functionArguments` anywhere, kind of a typo?

docs/topics/ksp/ksp-kotlin-model.md
zamulla: To the same point: I understand that it's a simplified abstract notation, but losing the nullability markers (`KSTypeReference` vs `KSTypeReference?`) seems to hurt the consistency: the sample below reads as a bit of a contradiction to the model. I'd let the sleeping question marks lie.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+A `KSTypeReference` represents an unresolved type. It preserves the syntactic representation of the type as it appears 
+in the source code. Its `KSReferenceElement` models the corresponding type element in Kotlin's grammar, including its 
+annotations and modifiers.
+
+A `KSReferenceElement` can be one of the following:
+
+* `KSClassifierReference`, which provides information such as `referencedName`.
```
zamulla:
```suggestion
* `KSClassifierReference`, which provides information such as `referencedName()`.
```

If this is a function, as I'm reading in the source code.

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+To access the type in Kotlin's type system, call `KSTypeReference.resolve()`. The returned `KSType` provides access to 
+the declaration that defines the type:
+
+```kotlin
+val ksType: KSType = ksTypeReference.resolve()
+val ksDeclaration: KSDeclaration = ksType.declaration
+```
+
+Resolve a type reference only when you need information available from `KSType` or `KSDeclaration`. When possible, 
+inspect the `KSReferenceElement` first. For example, you can use `KSClassifierReference.referencedName` to filter
```
zamulla:
```suggestion
inspect the `KSReferenceElement` first. For example, you can use the `KSClassifierReference.referencedName()` function to filter 
```

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
+[//]: # (title: KSP Kotlin model)
```
zamulla: I like the previous title a little more. This one is shorter, but there's no _action_ in it :)

docs/topics/ksp/ksp-kotlin-model.md
zamulla: Separately, Kotlin has no global variables, I had a deja vu because we had the same conversation about Lincheck docs. In the end Anton explained the "[Kotlin equivalents](https://kotlinlang.org/docs/lincheck-model-checking.html#controlled-sources-of-non-determinism)" there.

But here it's probably simply "top-level property".

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+      qualifiedName: KSName
+      containingFile: String
+      typeParameters: KSTypeParameter
+      parentDeclaration: KSDeclaration
+      classKind: ClassKind
+      primaryConstructor: KSFunctionDeclaration
+      superTypes: List<KSTypeReference>
+      // contains inner classes, member functions, properties, etc.
+      declarations: List<KSDeclaration>
+    KSFunctionDeclaration // top level function
```
zamulla:
```suggestion
    KSFunctionDeclaration // top-level function
```

docs/topics/ksp/ksp-kotlin-model.md
```diff
@@ -0,0 +1,115 @@
...
+      setter: KSPropertySetter
+        parameter: KSValueParameter
+```
+
+This hierarchy shows some of the common declarations in a source file. The KSP API provides additional symbols and
+properties that aren't shown here.
+
+The following diagram illustrates the relationships between the main KSP API types:
+
+![class diagram](ksp-class-diagram.svg){thumbnail="true" width="800" thumbnail-same-file="true"}
```
zamulla:
```suggestion
![The full class diagram of the KSP 2 model](ksp-class-diagram.svg){thumbnail="true" width="800" thumbnail-same-file="true"}
```

Let's be nicer to people who read alt text :)

docs/topics/ksp/ksp-kotlin-model.md
zamulla: Also, let's add a meta description? This is not exactly a page the gist of which can be easily parsed by a crawler.

zamulla (changes requested):
That's a high-level one. The new page does a better job of introducing the model, but can be a little more engaging to lead the reader through these high-level concepts.

docs/topics/ksp/ksp-kotlin-model.md
budindepunk: How about "How KSP models Kotlin"? it's similar to the previous one but a bit more general.

docs/topics/ksp/ksp-kotlin-model.md
budindepunk: Yes, it would be obtained something like this:

```kotlin
val ksTypeReference = functionDeclaration.returnType ?: return
```

We can either use an example like that or something more abstract? Like so:

```kotlin
val ksTypeReference: KSTypeReference = (...)
```

What makes more sense to you?

docs/topics/ksp/ksp-kotlin-model.md
budindepunk: You mean, remove them all from this block?

PR #5829, opened 2026-09-21
KT-89437: Add "Migrate to the new Swift export DSL"

Egorand:


docs/kr.tree
```diff
@@ -219,6 +219,7 @@
                 <toc-element topic="apple-framework.md"/>
             </toc-element>
             <toc-element toc-title="Swift export" topic="native-swift-export.md"/>
+            <toc-element topic="native-swift-export-dsl-migration-guide.md"/>
```
Egorand, line 222: Wasn't sure what the best way to structure the content on Swift export is, with this new entry the sidebar looks like this:

Image

An alternative would be to move the existing and new articles under the `Swift export` section, e.g.:

```html
<toc-element toc-title="Swift export">
  <toc-element topic="native-swift-export.md"/>
  <toc-element topic="native-swift-export-dsl-migration-guide.md"/>
</toc-element>
```

Then, the sidebar would look like this:

Image

Let me which option you prefer.

